PR #3, opened 2019-03-17
Corrected command to get webapp id

eragonriddle:
- [ x] The purpose of this PR is to correct a command in the documentation file. The command to get the webapp id is incorrect.
- [ x] Proper license headers are included in each file.


joshgav:
Thanks @eragonriddle! Getting this merged now.

PR #5, opened 2019-03-19
Added PHP support

dorfire:


images/build/php/build.sh
```diff
@@ -0,0 +1,161 @@
...
+apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false $fetchDeps
+
+versionDevReqs='libssl-dev'
+if [ $PHP_MAJOR == '5' ]; then
+	versionDevReqs='libssl1.0-dev'
+fi
+
+savedAptMark="$(apt-mark showmanual)";
+apt-get update;
+apt-get install -y --no-install-recommends \
```
kichalla: These are being installed for each of the php versions. Can they be moved to the pre-req image step `php-build-prereqs` and this way its executed only once

images/build/php/install-prereqs.sh
```diff
@@ -0,0 +1,42 @@
...
+    echo 'Pin: release *';
+    echo 'Pin-Priority: -1';
+} > /etc/apt/preferences.d/no-debian-php
+
+# dependencies required for running "phpize"
+# (see persistent deps below)
+PHPIZE_DEPS="autoconf dpkg-dev file g++ gcc libc-dev make pkg-config re2c"
+
+# persistent / runtime deps
+apt-get update && apt-get install -y \
```
kichalla: Its worth checking if using `buildpack-deps:stable` can avoid installing these

images/runtime/php/Dockerfile.template
```diff
@@ -0,0 +1,114 @@
...
+FROM php:%PHP_BASE_IMAGE%
+SHELL ["/bin/bash", "-c"]
+ENV PHP_VERSION %PHP_VERSION%
+
+# Bake Application Insights key from pipeline variable into final image
+ARG AI_KEY
+ENV ORYX_AI_INSTRUMENTATION_KEY=${AI_KEY}
+
+# COPY images/runtime/python/install-dependencies.sh /tmp/scripts/install-dependencies.sh
+# RUN /tmp/scripts/install-dependencies.sh
```
kichalla: Can you see if rearranging the content here is feasible? For example, any trivial change in this file would bust the cache from this point on and all the dependencies in line 35 (`Install common PHP extensions`) would get installed again.

build/build-buildimages.sh
```diff
@@ -57,6 +57,11 @@ BuildAndTagStage python3.5-build
 BuildAndTagStage python3.6-build
 BuildAndTagStage python3.7-build
 BuildAndTagStage python
+BuildAndTagStage php-build-prereqs
+BuildAndTagStage php56-build
```
arroyc: php5.6-build, php7.0-build etc like we have for python

images/build/Dockerfile
```diff
@@ -264,6 +265,35 @@ RUN set -ex \
...
+FROM php-build-prereqs AS php72-build
+RUN set -eux && . /php/__php-versions.sh && PHP_VERSION=$PHP72_VERSION GPG_KEYS=$PHP72_KEYS PHP_SHA256=$PHP72_TAR_SHA256 /php/build.sh
+
+# Build PHP 7.3
+FROM php-build-prereqs AS php73-build
+RUN set -eux && . /php/__php-versions.sh && PHP_VERSION=$PHP73_VERSION GPG_KEYS=$PHP73_KEYS PHP_SHA256=$PHP73_TAR_SHA256 /php/build.sh
+# Install Composer so that it can be copied to the final image later
+RUN set -eux && . /php/__php-versions.sh && \
+    mkdir /opt/php-composer && \
+    export phpbin="/opt/php/$PHP73_VERSION/bin/php" && \
```
arroyc: IMO it will be cleaner to put this in a separate script

images/build/php/install-prereqs.sh
dorfire: even if they're already installed in `buildpack-deps:stable`, this command line will return right away (with something like "requirements already satisfied")

images/build/benv.sh
```diff
@@ -121,6 +130,23 @@ benv-resolve() {
     return 0
   fi
 
+  # Resolve PHP versions
+  if [ "$name" == "php" -o "${name::7}" == "php_" ] && [ "${value::1}" != "/" ]; then
```
gpcastro: why `name::7`? It should be ::4, right?

images/build/benv.sh
```diff
@@ -121,6 +130,23 @@ benv-resolve() {
...
+      echo >&2 benv: php version \'$value\' not found\; choose one of:
+      benv-versions >&2 /opt/php
+      return 1
+    fi
+    local DIR="/opt/php/$value/bin"
+    if [ "$name" == "php" ]; then
+      export PATH="$DIR:$PATH"
+      export php="$DIR/php"
+    else
+      eval export php_${name:7}=\"$DIR/php\"
```
gpcastro: same here

images/build/php/build.sh
```diff
@@ -0,0 +1,161 @@
...
+PHP_CFLAGS="-fstack-protector-strong -fpic -fpie -O2"
+PHP_CPPFLAGS="$PHP_CFLAGS"
+PHP_LDFLAGS="-Wl,-O1 -Wl,--hash-style=both -pie"
+
+PHP_URL="https://secure.php.net/get/php-$PHP_VERSION.tar.xz/from/this/mirror"
+PHP_ASC_URL="" # "https://secure.php.net/get/php-$PHP_VERSION.tar.xz.asc/from/this/mirror"
+GPG_KEYS=($GPG_KEYS) # Cast the string to an array
+PHP_MD5=""
+
+
```
gpcastro: nit: extra line

images/build/php/build.sh
```diff
@@ -0,0 +1,161 @@
...
+fi;
+
+if [ -n "$PHP_ASC_URL" ]; then
+	wget -O php.tar.xz.asc "$PHP_ASC_URL";
+	export GNUPGHOME="$(mktemp -d)";
+	for i in {1..5}; do
+		gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys "${GPG_KEYS[0]}" || \
+		gpg --batch --keyserver ha.pool.sks-keyservers.net --recv-keys "${GPG_KEYS[1]}";
+		if [ $? -eq 0 ]; then break; fi
+    done
```
gpcastro: indentation

build/build-buildimages.sh
dorfire: This whole section is going to change after we decide what to do with PR #12

images/build/benv.sh
dorfire: good catch!

images/build/php/build.sh
```diff
@@ -0,0 +1,162 @@
+#!/bin/bash
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+# --------------------------------------------------------------------------------------------
+
+set -eux
+
+START_TIME=`date +%s`
```
gpcastro: Did we take this script from somewhere? I see there's lots of references to GitHub issues. If yes, then we should reference the source

images/build/php/docker-php-entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
gpcastro: what is this file for? Found no reference to it....

images/build/php/docker-php-ext-install.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/sh
+set -e
```
gpcastro: copyright notice

dorfire: these scripts are actually copied from here:
https://github.com/docker-library/php/tree/master/7.3/stretch/apache

gpcastro: Let's add a readme.md to this folder and reference it

images/build/php/build.sh
dorfire: Will add a SOURCE.md file

images/build/php/docker-php-ext-install.sh
dorfire: Added a SOURCE.md

src/BuildScriptGenerator/Php/PhpBashBuildSnippet.sh.tpl
```diff
@@ -0,0 +1,13 @@
+﻿echo "PHP executable: $php"
+
+cd "$SOURCE_DIR"
```
gpcastro: nit: this is no longer needed, as the base script makes sure every snippet runs in this folder

src/BuildScriptGenerator/Php/PhpBashBuildSnippet.sh.tpl
```diff
@@ -0,0 +1,13 @@
...
+
+{{ if ComposerFileExists }}
+declare -r composer='/opt/php-composer/composer.phar'
+echo "Composer archive: $composer"
+echo "Running composer install..."
+echo
+$php $composer install
```
gpcastro: q: why do we need to use `$php` instead of just `php`? In pre/post build scripts, the user should be able to use just `php` and it must be in the right version

src/BuildScriptGenerator/Php/PhpBashBuildSnippet.sh.tpl
```diff
@@ -0,0 +1,13 @@
+﻿echo "PHP executable: $php"
+
+cd "$SOURCE_DIR"
+
+{{ if ComposerFileExists }}
```
gpcastro: if not use bash, `if [-f ]`?

src/Common/DefinitionListFormatter.cs
```diff
@@ -51,17 +53,27 @@ public override string ToString()
                     continue;
                 }
 
-                string[] lines = row.Item2.Split(Environment.NewLine);
+                string[] lines = row.Item2.Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None);
```
gpcastro: q: is "\r" needed? I'm wondering if it will split "\r\n" twice

src/startupscriptgenerator/php/main.go
```diff
@@ -0,0 +1,32 @@
...
+	"flag"
+	"startupscriptgenerator/common"
+)
+
+func main() {
+	common.PrintVersionInfo()
+
+	appPathPtr := flag.String("appPath", ".", "The path to the application folder, e.g. '/home/site/wwwroot/'.")
+	startupCmdPtr := flag.String("startupCommand", "apache2-foreground", "Command that will be executed to start the application server up.")
+	outputPathPtr := flag.String("output", "run.sh", "Path to the script to be generated.")
```
gpcastro: we should also accept a command from the user, if provided

src/BuildScriptGenerator/SourceRepo/SourceRepoFileHelpers.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Oryx.BuildScriptGenerator.SourceRepo
+{
+    internal static class SourceRepoFileHelpers
+    {
+        internal static dynamic ReadJsonObjectFromFile(ISourceRepo sourceRepo, string fileName)
```
gpcastro: sugg: we could make this an extension method, to make it easier to find?

src/Common/DefinitionListFormatter.cs
dorfire: It won't - it should work fine anyway. Buy I guess it really isn't needed as Mac OS (i.e. before Mac OS X) isn't used anymore

src/BuildScriptGenerator/SourceRepo/SourceRepoFileHelpers.cs
dorfire: good idea

src/BuildScriptGenerator/Php/PhpBashBuildSnippet.sh.tpl
dorfire: a style preference - why run extra commands in the build script if we know for certain they won't be needed?

src/startupscriptgenerator/php/main.go
dorfire: other than startupCmdPtr?

PR #8, opened 2019-03-19
Copy only production dependencies to destination directory when building Node apps

kichalla:


tests/Oryx.BuildImage.Tests/NodeJSSampleAppsTest.cs
```diff
@@ -32,7 +32,7 @@ public void GeneratesScript_AndBuilds()
             var appDir = volume.ContainerDir;
             var appOutputDir = "/tmp/webfrontend-output";
             var script = new ShellScriptBuilder()
-                .AddBuildCommand($"{appDir} -o {appOutputDir}")
+                .AddBuildCommand($"{appDir} -i /tmp/int -o {appOutputDir}")
```
kichalla: These changes were made because `rsync` and directories which are volume mounted do not behave well together.

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -1,44 +1,104 @@
...
+then
+	# Clean-up any earlier build artifacts (which contain only prod dependencies)
+	rm -rf node_modules
+	rm -f node_modules.zip
+
+	# Rename the folder which has all the node modules to reuse for later builds to improve perf
+	mv "$SOURCE_DIR"/$allNodeModulesDirName "$SOURCE_DIR"/node_modules
+fi
+
+# Directory where production dependencies are installed
```
kichalla: The logic here tries to install prod dependencies in a separate folder and then syncs up with the source folder where we run the full install. Finally we rename the node_modules folder having all dependencies to something else and copy the prod only dependencies from the separate folder back to source folder. This is done so that the base bash script code need not be touched and it does what it usually does.

However, I tried another approach which was simpler IMO but I hit a road block with it which I felt is risky. The approach is to restore prod dependencies in the source folder and instead use a separate folder for restoring all dependencies. The script looked less complex but the problem now was that the later commands like `npm run build` need to now know the custom node_modules folder having all the dependencies. This I fixed by using `NODE_PATH` which worked for `npm run build` but the same scenario did not work for `yarn run build`

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -1,44 +1,104 @@
...
+		# Rename folder having all node_modules as we want only prod dependencies
+		# to be synced with destination directory
+		mv node_modules __oryx_all_node_modules
+		
+		echo
+		echo "Copying production dependencies from '$prodDependencies' to '$SOURCE_DIR'..."
+		echo
+		cd "$prodDependencies"
+		{{ if ZipNodeModulesDir }}
+			echo
```
kichalla: I know this indentation shows up in the generated script but the script is not easily readable without it.

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -1,44 +1,104 @@
...
+	# Clean-up any earlier build artifacts (which contain only prod dependencies)
+	rm -rf node_modules
+	rm -f node_modules.zip
+
+	# Rename the folder which has all the node modules to reuse for later builds to improve perf
+	mv "$SOURCE_DIR"/$allNodeModulesDirName "$SOURCE_DIR"/node_modules
+fi
+
+# Directory where production dependencies are installed
+prodDependencies="/tmp/prod-dependencies"
```
gpcastro: I wouldn't hard code it inside tmp. Why not inside the `$SOURCE_DIR` itself?

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -1,44 +1,104 @@
...
 then
-	echo
-	echo Zipping 'node_modules' folder ...
-	rm -f "node_modules.zip"
-	tar -zcf node_modules.zip node_modules
+	if [ "$hasProductionOnlyDependencies" == "true" ]
+	then
+		# Rename folder having all node_modules as we want only prod dependencies
+		# to be synced with destination directory
+		mv node_modules __oryx_all_node_modules
```
gpcastro: `$allNodeModulesDirName`?

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -1,44 +1,104 @@
...
 
+allNodeModulesDirName=__oryx_all_node_modules
+if [ -d "$SOURCE_DIR"/$allNodeModulesDirName ]
+then
+	# Clean-up any earlier build artifacts (which contain only prod dependencies)
+	rm -rf node_modules
+	rm -f node_modules.zip
+
+	# Rename the folder which has all the node modules to reuse for later builds to improve perf
+	mv "$SOURCE_DIR"/$allNodeModulesDirName "$SOURCE_DIR"/node_modules
```
gpcastro: this is confusing... why would we have node_modules inside a node_modules folder?

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
gpcastro: why do we need this command?

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
kichalla: This is just renaming the folder

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -1,44 +1,104 @@
...
-    echo "Configuring Yarn cache folder ..."
+    echo "Configuring Yarn cache folder..."
     yarn config set cache-folder $YARN_CACHE_DIR
 fi
 
+allNodeModulesDirName=__oryx_all_node_modules
+if [ -d "$SOURCE_DIR"/$allNodeModulesDirName ]
+then
+	# Clean-up any earlier build artifacts (which contain only prod dependencies)
+	rm -rf node_modules
```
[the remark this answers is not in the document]
kichalla: You mean if it throws an error if it does not exist? No it does not throw, so we are good

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -1,44 +1,104 @@
...
 if [ -d $YARN_CACHE_DIR ]
 then
 	echo
-    echo "Configuring Yarn cache folder ..."
+    echo "Configuring Yarn cache folder..."
     yarn config set cache-folder $YARN_CACHE_DIR
 fi
 
+allNodeModulesDirName=__oryx_all_node_modules
+if [ -d "$SOURCE_DIR"/$allNodeModulesDirName ]
```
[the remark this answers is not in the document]
kichalla: Since base bash script copies the content from source directory to destination directory, I wanted the source directory to have the content that we actually want to be in the destination directory. Since we do not want all the modules to be copied to destination directory, we rename this folder and exclude it from rsync.

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
kichalla: Sure, will try that and see

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -1,44 +1,104 @@
...
+
+	# Rename the folder which has all the node modules to reuse for later builds to improve perf
+	mv "$SOURCE_DIR"/$allNodeModulesDirName "$SOURCE_DIR"/node_modules
+fi
+
+# Directory where production dependencies are installed
+prodDependencies="/tmp/prod-dependencies"
+
+hasProductionOnlyDependencies="{{ HasProductionOnlyDependencies }}"
+if [ "$SOURCE_DIR" != "$DESTINATION_DIR" ] && [ "$hasProductionOnlyDependencies" == "true" ]
```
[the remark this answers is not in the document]
kichalla: Ok

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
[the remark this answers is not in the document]
gpcastro: we already leave node_modules folder out when copying the files from source to target

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -6,39 +6,100 @@ YARN_CACHE_DIR=/usr/local/share/yarn-cache
...
 echo
-echo "Running '{{ NpmRunBuildAzureCommand }}' ..."
+echo "Running '{{ NpmRunBuildAzureCommand }}'..."
 echo
 {{ NpmRunBuildAzureCommand }}
 {{ end }}
 
+if [ "$ORYX_COPY_ONLY_PROD_MODULES_TO_OUTPUT" == "true" ]
+then
+	if [ "$SOURCE_DIR" != "$DESTINATION_DIR" ] && [ "$hasProductionOnlyDependencies" == "true" ]
```
gpcastro: the condition for those `mv` commands to execute should be equivalent to the lines 17 + 19, right? If we rename the folders there, we should rename them back here

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -6,39 +6,100 @@ YARN_CACHE_DIR=/usr/local/share/yarn-cache
...
+zippedModulesFileName=node_modules.zip
+allModulesDirName=__oryx_all_node_modules
+prodModulesDirName=__oryx_prod_node_modules
+
+if [ "$ORYX_COPY_ONLY_PROD_MODULES_TO_OUTPUT" == "true" ]
+then
+	if [ -d $allModulesDirName ]
+	then
+		# Rename existing node_modules back to prod modules
+		mv node_modules $prodModulesDirName
```
gpcastro: what if node_modules doesn't exist, like in the first build?

kichalla: Yeah, we could be more defensive here. I was depending on presence of `allModulesDirName` which would be present after one build.

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
kichalla: We are renaming the folders here but I think I am not handling a case like:

**Scenario-1**
1. first build with copyonlyprodmodules=true
2. first build with copyonlyprodmodules=false (the source folder would still have the folders with the reserved names)

Will update the PR.

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
kichalla: I made some changes in latest iteration which might look more complex but I think handle more scenarios correctly.

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -6,39 +6,114 @@ YARN_CACHE_DIR=/usr/local/share/yarn-cache
...
+	fi
+fi
+
+# if node modules exist separately for dev & prod (like from an earlier build),
+# rename the folders back appropriately for the current build
+if [ -d $allModulesDirName ]
+then
+	if [ -d node_modules ]
+	then
+		if [ "$copyOnlyProdModulesToOutput" == "true" ]
```
gpcastro: nit: combine those if's using `&&` ?

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -6,39 +6,114 @@ YARN_CACHE_DIR=/usr/local/share/yarn-cache
 if [ -d $YARN_CACHE_DIR ]
 then
 	echo
-    echo "Configuring Yarn cache folder ..."
+    echo "Configuring Yarn cache folder..."
     yarn config set cache-folder $YARN_CACHE_DIR
 fi
 
+zippedModulesFileName=node_modules.zip
```
gpcastro: I've renamed it to node_modules.tar.gz

PR #10, opened 2019-03-20
Arroyc/new yaml pipeline

arroyc:
Purpose of this PR is to have a sign build in our build pipeline and restructure existing build definitions accordingly.


images/build/Dockerfile
```diff
@@ -282,16 +282,6 @@ RUN if [ -z "$AGENTBUILD" ]; then dotnet publish -r linux-x64 -o /opt/buildscrip
 # This stage is only when building in devops agents
 FROM main AS copybuildscriptsymbol
 COPY symbols /opt/buildscriptgen/
-COPY src/BuildScriptGenerator /usr/oryx/src/BuildScriptGenerator
```
kichalla: Thanks for fixing this!

vsts/pipelines/_signbinary.yml
```diff
@@ -0,0 +1,63 @@
+steps:
+- powershell: 'Write-Host "Running Sign Oryx binaries"'
+  displayName: 'Show Build Phase'
+
+- powershell: |
+   Write-Host "##vso[task.setvariable variable=SignType]real"
+   Write-Host "##vso[task.setvariable variable=AGENTBUILD]true"
+  displayName: 'Set sign binary variable'
+  condition: and(in(variables['Build.Reason'], 'Schedule', 'Manual'), eq(variables['Build.DefinitionName'], 'Oryx-CI'))
```
gpcastro: we should avoid repeating these checks in every task. Add a new task that checks those things, and create a variable, e.g. "SignBinary" = "true" if they pass

vsts/pipelines/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+  displayName: Build and Publish Oryx Images
+  dependsOn: Job_2
+  condition: succeeded()
+  pool:
+    name: OryxLinux
+    demands: sh
+    steps:
+    - script: |
+        echo "##vso[task.setvariable variable=BuildBuildImages;]true"
+        echo "Hello World"
```
kichalla: remove

vsts/pipelines/_signbinary.yml
arroyc: yes I saw it … I already have a task where I'm setting signtype variable so from next task onward I will just use it

vsts/pipelines/_buildTemplate.yml
```diff
@@ -1,4 +1,12 @@
 steps:
+- task: DownloadPipelineArtifact@0
```
gpcastro: q: why do we need this task?

vsts/pipelines/_signbinary.yml
```diff
@@ -0,0 +1,63 @@
+steps:
+- powershell: 'Write-Host "Running Sign Oryx binaries"'
+  displayName: 'Show Build Phase'
+
+- powershell: |
+   Write-Host "##vso[task.setvariable variable=SignType]real"
+   Write-Host "##vso[task.setvariable variable=AGENTBUILD]true"
```
gpcastro: q: why do we also set AGENTBUILD here? Isn't signType enough?

vsts/pipelines/_securityChecks.yml
```diff
@@ -10,6 +10,12 @@ steps:
     projects: '$(Build.SourcesDirectory)\src\**\*.csproj'
     arguments: '-o $(Build.BinariesDirectory)'
     zipAfterPublish: false
+    modifyOutputPath: false
+
+- task: ms.vss-governance-buildtask.governance-build-task-component-detection.ComponentGovernanceComponentDetection@0
```
gpcastro: why are we duplicating this task here? It is already part of _buildTemplates.yaml.

vsts/pipelines/_signbinary.yml
arroyc: yes, the way we build the build definitions initially AGENTBUILD was required, now we have enough conditions to have just one variable which is signtype, I wanted to revisit all these cleaning tasks in a separate work-item.

vsts/pipelines/_securityChecks.yml
arroyc: I removed it, i think during rebase it came back, I will remove it again

vsts/pipelines/_buildTemplate.yml
arroyc: we need this when we do a sign build, we need to download the signed binaries and put them in right context/directory for doing a docker build for build image

PR #11, opened 2019-03-20
Print Oryx github repo info when doing a build

kichalla:


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -81,12 +81,16 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
             var buildOpId = logger.StartOperation(appName);
 
             var buildInfo = new DefinitionListFormatter();
-            buildInfo.AddDefinition("Build Operation ID", buildOpId);
             buildInfo.AddDefinition("Oryx Version", $"{Program.GetVersion()}, Commit: {Program.GetCommit()}");
+            buildInfo.AddDefinition("Oryx Repository", "https://github.com/microsoft/oryx");
```
kichalla: This is how looks like, FYI:
image

gpcastro: I'd use a more free form text, which also explains what Oryx is, e.g. `Build orchestrated by Microsoft Oryx, https://github.com/Microsoft/Oryx/. You can report issues at 'https://github.com/Microsoft/Oryx/Issues'`.
What do you think?

kichalla: cc @joshgav

kichalla: Updated. Latest looks like below:

image

PR #13, opened 2019-03-21
add manual build as a reason to do signed build

arroyc:
added one missing condition and formatted spaces to make sure yaml doesn't ignore tasks/steps in job_3


vsts/pipelines/ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
   pool:
     name: OryxLinux
     demands: sh
-    steps:
+  steps:
```
kichalla: Was this indentation the reason that the steps weren't getting executed?

PR #15, opened 2019-03-22
fix Oryx builds by fixing yml formatting and replacing "symbols" with "binaries"

arroyc:
Broke the build earlier, fixed it [here](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2521319)

### DIFF
--- a/vsts/pipelines/_buildTemplate.yml
+++ b/vsts/pipelines/_buildTemplate.yml
@@ -5,7 +5,7 @@ steps:
     pipelineId: 'Oryx-CI'
     artifactName: 'drop-$(Build.BuildNumber)'
     targetPath: '$(Build.SourcesDirectory)'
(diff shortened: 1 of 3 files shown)


vsts/pipelines/_buildTemplate.yml
```diff
@@ -5,7 +5,7 @@ steps:
     pipelineId: 'Oryx-CI'
     artifactName: 'drop-$(Build.BuildNumber)'
     targetPath: '$(Build.SourcesDirectory)'
-  condition: and(succeeded(), eq(variables['Build.DefinitionName'], 'Oryx-CI'), in(variables['Build.Reason'], 'Schedule', 'Manual'))
```
kichalla: Just FYI...make sure to have proper message on the PR as that is what is being put as the main commit message. For example, from earlier PR of yours:
image

PR #16, opened 2019-03-22
Adding automatic release notes from changelog file.

gpcastro:
Adding a changelog file and script to create a release notes file from it.

### DIFF
--- a/vsts/pipelines/_buildTemplate.yml
+++ b/vsts/pipelines/_buildTemplate.yml
@@ -167,6 +167,12 @@ steps:
     scriptPath: ./vsts/scripts/dockerCleanup.sh
   condition: or(eq(variables['TestBuildImages'], 'true'), eq(variables['TestRuntimeImages'], 'true'), eq(variables['TestIntegration'], 'true'))
 
+- task: ShellScript@2
+  displayName: 'Generate release notes'
+  inputs:
+    scriptPath: ./vsts/scripts/generate-release-notes.sh
+  condition: and(succeeded(), eq(variables['PushBuildImages'], 'true'), eq(variables['BuildBuildImages'], 'true'))
+
 - task: ArchiveFiles@2
   displayName: 'Archive docker files and scripts for Oryx build and runtime images'
   inputs:
--- a/vsts/scripts/generate-release-notes.sh
+++ b/vsts/scripts/generate-release-notes.sh
@@ -26,12 +26,12 @@ echo "Release notes will be placed in $OUTPUT_FILE"
 # First, we look for the latest tag that was pushed. Since our builds numbers are lexicographically ordered,
 # YYYYMMDD.P, we just take the latest value that starts with a `2` to avoid other tags that might be in the repo.
 # Optimistic note: yes, this script will break in year 3000, but we can fix it then.
-LAST_TAG=$(git tag --list 2* | sort -r | head -n 1)
+LAST_TAG=$(git tag --sort=committerdate | tail -n 1)
 
 if [ -z "$LAST_TAG" ]; then
     echo "Couldn't find a base tag, will output the entire file"
     # Ignore the lines starting with [//] which we're using as comments.
-    cat ../CHANGELOG.md | grep -v -e '^\[//\]' > $OUTPUT_FILE
+    cat $CHANGELOG_FILE | grep -v -e '^\[//\]' > $OUTPUT_FILE
 else
     echo "Getting the diff from latest tag, $LAST_TAG"
     # Get the diff for the changelog file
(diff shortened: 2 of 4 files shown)


vsts/pipelines/_buildTemplate.yml
```diff
@@ -167,6 +167,12 @@ steps:
     scriptPath: ./vsts/scripts/dockerCleanup.sh
   condition: or(eq(variables['TestBuildImages'], 'true'), eq(variables['TestRuntimeImages'], 'true'), eq(variables['TestIntegration'], 'true'))
 
+- task: ShellScript@2
+  displayName: 'Generate release notes'
+  inputs:
+    scriptPath: ./vsts/scripts/generate-release-notes.sh
+  condition: and(succeeded(), eq(variables['PushBuildImages'], 'true'), eq(variables['BuildBuildImages'], 'true'))
```
kichalla: I guess this condition was probably used for your testing? Ideally it should probably be only for our `Oryx-CI` builds?

vsts/scripts/generate-release-notes.sh
```diff
@@ -0,0 +1,42 @@
...
+    # Create a default file using DevOps' pipeline artifacts directory
+    OUTPUT_FILE=$BUILD_ARTIFACTSTAGINGDIRECTORY/Release-notes.md
+fi
+
+echo "Release notes will be placed in $OUTPUT_FILE"
+
+# First, we look for the latest tag that was pushed. Since our builds numbers are lexicographically ordered,
+# YYYYMMDD.P, we just take the latest value that starts with a `2` to avoid other tags that might be in the repo.
+# Optimistic note: yes, this script will break in year 3000, but we can fix it then.
+LAST_TAG=$(git tag --list 2* | sort -r | head -n 1)
```
kichalla: Seems like there are other ways of getting tags in a sorted order https://stackoverflow.com/a/36636526

vsts/scripts/generate-release-notes.sh
```diff
@@ -0,0 +1,42 @@
...
+
+if [ -z "$LAST_TAG" ]; then
+    echo "Couldn't find a base tag, will output the entire file"
+    # Ignore the lines starting with [//] which we're using as comments.
+    cat ../CHANGELOG.md | grep -v -e '^\[//\]' > $OUTPUT_FILE
```
kichalla: Can use `CHANGELOG_FILE` variable here

vsts/scripts/generate-release-notes.sh
```diff
@@ -0,0 +1,42 @@
...
+else
+    echo "Getting the diff from latest tag, $LAST_TAG"
+    # Get the diff for the changelog file
+    # The regex ^+[^+] is used to capture only the added lines, and the [^+], which means "exclude '+', removes the
```
kichalla: Is this typical...usually I have seen only commit messages being showed in a change log of a release? External users, including AppService, would feel it difficult to understand the contents of the diffed code IMO. If we maintain good commit messages those should be enough?

vsts/scripts/generate-release-notes.sh
kichalla: I was thinking of how this would work with branches...like if we have a 'patch' branch, then the CI build would be running using this patch branch(which is correct). We want to make sure that the tag list is only returns the tags which are part/visible to that branch.

vsts/scripts/generate-release-notes.sh
gpcastro: the idea is to have something more descriptive, e.g. how to use a new feature. Commit messages don't have that. Also, we have many commits that have zero impact in users (e.g. security, test automation, internal tools, etc), and the useful stuff will get lost in there

vsts/scripts/generate-release-notes.sh
gpcastro: That is fine, since patches are an exception. If things get messed up, we can always fix the release notes by hand. The idea is to have some minimal automation so that we don't have to worry about it for automated releases

vsts/scripts/generate-release-notes.sh
gpcastro: do you mean storing it as a variable in the script?

gpcastro: or accepting it as a parameter?

kichalla: I mean you have a variable defined in line 16 for it which you can use `cat $CHANGELOG_FILE`

vsts/pipelines/_buildTemplate.yml
gpcastro: It's fine to have it in other builds, so we can see how it would look like before the CI build happens. There's just going to be a text file with release notes in the build output folder

kichalla: Sure, sounds good

PR #17, opened 2019-03-25
Skip the failing integration test until the Node's run-script command…

kichalla:


tests/Oryx.Integration.Tests/LocalDockerTests/NodeEndToEndTests.cs
```diff
@@ -611,7 +611,7 @@ public async Task Node_CreateReactAppSample_singleImage()
                 });
         }
 
-        [Fact]
+        [Fact(Skip = "#824174: Sync the Node Go startup code with the C# 'run-script' code")]
```
arroyc: locally this test runs, I wonder why it fails in agent build

kichalla: Its because the generated startup file has execute permissions on your local windows machine but not on the Linux CI agent. Apart from the code that sets the execute permission on the generated file, there's some more code which also needs to be kept in sync, so skipping the test instead.

PR #19, opened 2019-03-25
fix Docker commands in README

joshgav:

Default to image stored in Docker Hub because some images are missing in MCR. Also fixed the commands a bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,29 +69,29 @@ run.sh, by default in the same folder as the compiled artifact.
 To build and run an app from a repo, follow these approximate steps. An example
 script follows.
 
-1. Mount the repo as a volume in Oryx's `mcr.microsoft.com/oryx/build` container.
+1. Mount the repo as a volume in Oryx's `docker.io/oryxprod/build` container.
 1. Run `oryx build ...` within the repo directory to build a runnable artifact.
 1. Mount the output directory from build in an appropriate Oryx "run"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -69,29 +69,29 @@ run.sh, by default in the same folder as the compiled artifact.
 To build and run an app from a repo, follow these approximate steps. An example
 script follows.
 
-1. Mount the repo as a volume in Oryx's `mcr.microsoft.com/oryx/build` container.
+1. Mount the repo as a volume in Oryx's `docker.io/oryxprod/build` container.
 1. Run `oryx build ...` within the repo directory to build a runnable artifact.
 1. Mount the output directory from build in an appropriate Oryx "run"
-   container, such as `mcr.microsoft.com/oryx/node-10.14`.
+   container, such as `docker.io/oryxprod/node-10.14`.
```
gpcastro: should we remove the `docker.io` from it?

joshgav: I've been putting the hostname in more frequently as a subtle way to encourage folks to be aware of the provenance of their images, and to make DockerHub URNs symmetrical to those for MCR and other registries. IMHO it would be a good practice, particularly in scripts, to include the full path so others can reason about it more quickly. I see where it might confuse a Docker-focused reader at first, but my vote is still to keep the full path and help these users understand it over time. Are you okay with that?

PR #20, opened 2019-03-26
Splitting yarn cache to a separate image

gpcastro:
Removing the yarn cache generation from the build image build to its own image, that will end up in MCR.
I've triggered a build to create a new image on MCR, and once that completes I'll update the PR with the changes to the build image Dockerfile pointing to that image.


build/__variables.sh
```diff
@@ -4,7 +4,7 @@ set -e
 declare -r __REPO_DIR=$( cd $( dirname "$0" ) && cd .. && pwd )
 
 # VSTS environment variables
-declare -r BUILD_NUMBER="$BUILD_BUILDNUMBER"
+declare -r BUILD_NUMBER="${BUILD_BUILDNUMBER:-latest}"
```
kichalla: I believe this change could have impact in other places of code(also `latest` thing feels docker specific), I would suggest to make a new variable(tag?) in `build-buildimage-bases.sh` which falls back to `latest` if there is no build number.

images/build/yarn-cache/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# This image contains a cache with common yarn packages.
+# It is never used directly, but rather as a base for our build image.
+# Yarn cache will be placed in /usr/local/share/yarn-cache, that is the folder
+# that must be copied to the build image.
+
+FROM node:lts
+
```
kichalla: nit: remove new line

build/build-buildimage-bases.sh
```diff
@@ -33,6 +33,12 @@ echo "$PYTHON_IMAGE_PREFIX-3.6:$BUILD_NUMBER" >> $BUILD_IMAGE_BASES_ARTIFACTS_FI
 docker build -f $__REPO_DIR/images/build/python/3.7/Dockerfile -t "$PYTHON_IMAGE_PREFIX-3.7:$BUILD_NUMBER" $__REPO_DIR
 echo "$PYTHON_IMAGE_PREFIX-3.7:$BUILD_NUMBER" >> $BUILD_IMAGE_BASES_ARTIFACTS_FILE
 
+# Build Yarn cache
+YARN_CACHE_IMAGE_BASE="$ACR_DEV_NAME/public/oryx/build-yarn-cache"
+YARN_CACHE_IMAGE_NAME=$YARN_CACHE_IMAGE_BASE:$BUILD_NUMBER
+
+docker build $__REPO_DIR/images/build/yarn-cache -t $YARN_CACHE_IMAGE_NAME
```
kichalla: You probably need to provide the docker build context here (`$__REPO_DIR`) explicitly like the above docker build commands...Or else I think we could just `cd` into the repo root and can simplify all the commands in this files a bit

dorfire: I think it uses `images/build/yarn-cache` as the build context - which totally works - but is different from all of our other Dockerfiles

gpcastro: @dorfire is correct. It is different because it is simpler, all we need is a bash script in there.

build/__variables.sh
gpcastro: Fair enough

build/build-buildimage-bases.sh
kichalla: I think the build context is the parameter that usually comes at the end: `Usage:  docker build [OPTIONS] PATH | URL | -`
Just to be clear, what i meant was something like below:
```
cd $__REPO_DIR

docker build -f images/build/python/3.7/Dockerfile -t "$PYTHON_IMAGE_PREFIX-3.7:$IMAGE_TAG" .

docker build -f images/build/yarn-cache/Dockerfile -t $YARN_CACHE_IMAGE_NAME .
```
Its the same consistent pattern

gpcastro: Trust me, the command works. You can locally if you want to be sure

gpcastro: There's no need to use -f here as the build context already has a dockerfile. It is the simplest possible docker build command.

kichalla: Sure, sounds good 👍

images/build/Dockerfile
```diff
@@ -284,8 +284,7 @@ RUN ln -s /opt/dotnet/lts/dotnet /usr/local/bin/dotnet
 # Copy NodeJs, NPM and Yarn related content
 COPY --from=node-install /opt /opt
 COPY --from=node-install /links/ /usr/local/bin
-COPY images/build/node-cache-packages.sh /tmp/
-RUN /tmp/node-cache-packages.sh && rm /tmp/node-cache-packages.sh
+COPY --from=mcr.microsoft.com/oryx/build-yarn-cache:20190326.8 /usr/local/share/yarn-cache /usr/local/share/yarn-cache
```
arroyc: should it be latest?

gpcastro: with latest it is hard to know precisely which build we took a dependency on

PR #23, opened 2019-03-26
Fixed DB integration tests

dorfire:


build/test-integration.sh
```diff
@@ -6,21 +6,12 @@
 
 set -e
 
-declare -r REPO_DIR=$( cd $( dirname "$0" ) && cd .. && pwd )
```
kichalla: We did it this way so that this script could be run from any directory. With this change, we are supposed to run from within `build` directory as `__variables.sh` would only be available there.

tests/Oryx.Integration.Tests/LocalDockerTests/DatabaseTestsBase.cs
```diff
@@ -8,51 +8,38 @@
...
 using Xunit.Abstractions;
 
 namespace Microsoft.Oryx.Integration.Tests.LocalDockerTests
 {
-    public class DatabaseTestsBase
+    public abstract class DatabaseTestsBase
     {
         protected readonly Fixtures.DbContainerFixtureBase _dbFixture;
         protected readonly ITestOutputHelper _output;
+        private readonly Random _rand = new Random();
```
kichalla: Should this be `static`? I say this because each test class inherits from this base and each of it will have its own instance and there could be collisions when `_rand.Next(100)` is called.

PR #27, opened 2019-03-27
Updated references of 'apt' to 'apt-get'

kichalla:


images/build/Dockerfile
```diff
@@ -264,9 +265,6 @@ RUN chmod a+x /opt/buildscriptgen/GenerateBuildScript
 FROM python AS final
 WORKDIR /
 
-# Install rsync
-RUN apt update && apt install rsync --yes
```
dorfire: :+1:

images/runtime/node/8.8/Dockerfile
```diff
@@ -81,8 +81,8 @@ CMD [ "node" ]
 
 FROM node-original-image as oryx
 
-RUN  if command -v apt; then \
-       apt update && apt upgrade --yes; \
+RUN  if command -v apt-get; then \
```
dorfire: Do we really need this `command` test? Isn't this image **always** debian-based?

images/runtime/node/8.2/Dockerfile
```diff
@@ -80,8 +80,8 @@ CMD [ "node" ]
 
 FROM node-original-image as oryx
 
-RUN  if command -v apt; then \
-       apt update && apt upgrade --yes; \
+RUN  if command -v apt-get; then \
```
dorfire: Same as my other comment

images/runtime/node/8.8/Dockerfile
kichalla: I do not know actually. Let's fix it in a separate PR if this we really care about it.

PR #30, opened 2019-03-28
DotNetCore app can be run when current directory is the output directory

kichalla:
This is AppService scenario where:
- the current directory is the output directory itself i.e it has the startup dll file
- the 'oryx' startup command is not supplied the output directory explicitly.

### DIFF
--- a/src/startupscriptgenerator/common/fsvalidation.go
+++ b/src/startupscriptgenerator/common/fsvalidation.go
@@ -19,6 +19,10 @@ func PathExists(path string) bool {
 
 func FileExists(path string) bool {
 	fi, err := os.Stat(path)
+	if os.IsNotExist(err) {
+		return false
+	}
+
 	if err != nil {
 		return false
 	}
--- a/src/startupscriptgenerator/dotnetcore/scriptgenerator.go
+++ b/src/startupscriptgenerator/dotnetcore/scriptgenerator.go
@@ -53,7 +53,9 @@ const OryxPublishOutputDirectory = "oryx_publish_output"
 const DefaultBindPort = "8080"
 
 var _retrievedProjectDetails = false
+var _gotStartupFileName = false
 var _projDetails projectDetails = projectDetails{}
+var _startupFileName = ""
 
 func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript() string {
 	logger := common.GetLogger("dotnetcore.scriptgenerator.GenerateEntrypointScript")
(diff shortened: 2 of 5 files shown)


src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -102,51 +104,43 @@ func (gen *DotnetCoreStartupScriptGenerator) getStartupCommand() (string, string
 	// as we want to generate a script which does a 'cd' to this directory and run the user startup command
 	publishOutputDir := gen.PublishedOutputPath
 	if publishOutputDir == "" {
-		projDetails := gen.getProjectDetailsAndCache()
-		if projDetails.FullPath == "" {
-			return "", ""
-		}
 
-		publishOutputDir = filepath.Join(projDetails.Directory, OryxPublishOutputDirectory)
+		// Check if current directory is indeed the output directory (Example, like in AppService's case)
```
kichalla: FYI: https://github.com/Azure-App-Service/dotnetcore/blob/master/2.1.1/Dockerfile#L30

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -102,51 +104,43 @@ func (gen *DotnetCoreStartupScriptGenerator) getStartupCommand() (string, string
...
-
-		if _, err := os.Stat(publishOutputDir); os.IsNotExist(err) {
-			logger.LogError(
-				"Could not find oryx publish output directory at '%s'. Error: %s",
-				publishOutputDir,
-				err.Error())
-			return "", ""
+		startupFileName := gen.getStartupDllFileName()
+		startupFileFullPath := filepath.Join(currentDir, startupFileName)
+		if _, err := os.Stat(startupFileFullPath); !os.IsNotExist(err) {
```
kichalla: So with this change we check if the current directory has the startup file and if not then fall back to looking for the reserved directory `oryx_publish_output` directory for the startup dll file.

src/startupscriptgenerator/common/fsvalidation.go
```diff
@@ -19,6 +19,10 @@ func PathExists(path string) bool {
 
 func FileExists(path string) bool {
 	fi, err := os.Stat(path)
+	if os.IsNotExist(err) {
+		return false
+	}
+
 	if err != nil {
 		return false
 	}
```
gpcastro: doesn't it already cover the `IsNotExist` case?

kichalla: Seems like not..at least this is what I read is a recommended pattern.

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -102,51 +104,43 @@ func (gen *DotnetCoreStartupScriptGenerator) getStartupCommand() (string, string
...
 	publishOutputDir := gen.PublishedOutputPath
 	if publishOutputDir == "" {
-		projDetails := gen.getProjectDetailsAndCache()
-		if projDetails.FullPath == "" {
-			return "", ""
-		}
 
-		publishOutputDir = filepath.Join(projDetails.Directory, OryxPublishOutputDirectory)
+		// Check if current directory is indeed the output directory (Example, like in AppService's case)
+		currentDir := common.GetValidatedFullPath(".")
```
gpcastro: >currentDir := common.GetValidatedFullPath(".") [](start = 2, length = 46)

what path is this supposed to return? The path from where the command was executed from?

kichalla: yes, right

PR #32, opened 2019-03-29
Add CLI flag for zipping Python virtualenv directory

dorfire:


src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -13,10 +13,9 @@
 
 namespace Microsoft.Oryx.BuildScriptGenerator.Node
 {
+    // TODO: is this used anywhere?
```
gpcastro: can you please remove this from your view? I'm working on this as part of my changes and want to avoid conflict :)

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -214,13 +218,16 @@ private void TryLogDependencies(string pythonVersion, ISourceRepo repo)
...
-            }
-
+            string virtualEnvName = string.Empty;
+            context.Properties?.TryGetValue(VirtualEnvironmentNamePropertyKey, out virtualEnvName);
             return virtualEnvName;
         }
+
+        private bool ShouldZipVenvDir(BuildScriptGeneratorContext context)
+        {
+            // Build property takes precedence over env var
```
gpcastro: you can just remove the env var, and rely only on the property

gpcastro: unless it starts with `ORYX_`, which case we can tackle this later

dorfire: yep, it's `ORYX_ZIP_VIRTUALENV_DIR`

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -14,12 +14,15 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Python
 {
     [BuildProperty(
         VirtualEnvironmentNamePropertyKey, "If provided, will create a virtual environment with the given name.")]
+    [BuildProperty(
+        ZipVenvDirPropertyKey, "If provided, the virtual environment folder will be zipped to the output folder.")]
```
kichalla: `If provided` => `if true`?

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -222,5 +218,12 @@ private string GetVirutalEnvironmentName(BuildScriptGeneratorContext context)
 
             return virtualEnvName;
         }
+
+        private bool ShouldZipVenvDir(BuildScriptGeneratorContext context)
+        {
+            // Build property takes precedence over env var
+            return context.Properties?.ContainsKey(VirtualEnvironmentNamePropertyKey) == true ||
```
kichalla: ZipVenvDirPropertyKey

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -14,12 +14,15 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Python
...
         VirtualEnvironmentNamePropertyKey, "If provided, will create a virtual environment with the given name.")]
+    [BuildProperty(
+        ZipVenvDirPropertyKey, "If provided, the virtual environment folder will be zipped to the output folder.")]
     [BuildProperty(
         TargetPackageDirectoryPropertyKey,
         "Directory to download the packages to, if no virtual environment is provided. Default: '" +
         DefaultTargetPackageDirectory + "'")]
     internal class PythonPlatform : IProgrammingPlatform
     {
+        internal const string ZipVenvDirPropertyKey = "zip_venv_dir";
```
kichalla: Add/change tests too

src/BuildScriptGenerator/Python/PythonPlatform.cs
dorfire: it doesn't need to be true, just to be present

src/BuildScriptGenerator/Python/PythonPlatform.cs
dorfire: good catch!

PR #39, opened 2019-04-01
Measure and print execution time of some commands

kichalla:


src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -364,7 +364,7 @@ private static bool ShouldPruneDevDependencies(BuildScriptGeneratorContext conte
                 else if (string.Equals(compressNodeModulesOption, ZipNodeModulesOption, StringComparison.InvariantCultureIgnoreCase))
                 {
                     compressedNodeModulesFileName = NodeConstants.NodeModulesZippedFileName;
-                    compressNodeModulesCommand = $"zip -r";
+                    compressNodeModulesCommand = $"zip -rq";
```
kichalla: ignore this

src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
```diff
@@ -31,7 +31,10 @@ then
 	then
 		echo
 		echo "Destination directory is not empty. Deleting its contents..."
+		START_TIME=$SECONDS
 		rm -rf "$DESTINATION_DIR"/*
+		ELAPSED_TIME=$(($SECONDS - $START_TIME))
```
kichalla: @gpcastro I was thinking if we can get these changes in since they could help in perf investigation by us and as well as Jenny. Its a short gap solution.

PR #40, opened 2019-04-02
Remove Python 3.5

gpcastro:


images/build/python/3.5/Dockerfile
```diff
@@ -1,5 +0,0 @@
-# Build Python 3.5
-FROM python-build-prereqs
```
kichalla: Remove the reference in `build-buildimage-bases.sh` too

PR #42, opened 2019-04-02
Improve CI build perf: Disable 'docker system prune' by default and a…

kichalla:


build/build-runtimeimages.sh
```diff
@@ -140,7 +140,7 @@ cat $ACR_RUNTIME_IMAGES_ARTIFACTS_FILE
 
 echo
 echo "Cleanup: Run 'docker system prune': $DOCKER_SYSTEM_PRUNE"
-if [ $DOCKER_SYSTEM_PRUNE = "true" ]
+if [ "$DOCKER_SYSTEM_PRUNE" == "true" ]
```
kichalla: Since this will be `false` by default now, we would need a job kind of thing to clean-up dangling layers.

build/build-buildimages.sh
```diff
@@ -114,7 +114,7 @@ cat $ACR_BUILD_IMAGES_ARTIFACTS_FILE
 
 echo
 echo "Cleanup: Run 'docker system prune': $DOCKER_SYSTEM_PRUNE"
-if [ $DOCKER_SYSTEM_PRUNE = "true" ]
+if [ "$DOCKER_SYSTEM_PRUNE" == "true" ]
```
kichalla: Since this will be `false` by default now, we would need a job kind of thing to clean-up dangling layers.

PR #45, opened 2019-04-02
Clean-up scripts related to building images etc.

kichalla:
This commits makes the following changes:
- When built locally, the images are always tagged with 'oryx' and 'latest' tag(examples: oryx/build:latest, oryx/node-10.14:latest)
- Tests would use this new tag always
- When building in a CI agent, we tag the locally (in this case built locally on the CI agent) built image with 4 tags (2 for DockerHub and 2 for ACR in the format 'repo-name:latest' and 'repo-name:buildnumber')


vsts/scripts/release-runtimeimages.sh
```diff
@@ -11,6 +11,6 @@ while read sourceImage; do
   docker pull "$sourceImage"
   newtag=$(echo "$sourceImage" | sed -r 's/oryxdevms/oryxprod/')
   echo "$newtag"
-  docker tag "$sourceImage" "$newtag"
-  docker push "$newtag"
-done <"$(System.DefaultWorkingDirectory)/_Oryx-Runtime/drop/images/runtime-images.txt"
+  docker tag "$sourceImage" "$newtag"``
```
kichalla: Unintended change. Will remove

vsts/scripts/dockerCleanup.sh
```diff
@@ -38,20 +38,29 @@ docker images
...
 UntagImages "oryxdevms/*:*.*"
-UntagImages "oryxtests/*:latest"
 UntagImages "oryxprod/*:latest"
 UntagImages "oryxprod/*:*.*"
+
+# Untag images with Dev ACR related tags
 UntagImages "oryxdevmcr.azurecr.io/public/oryx/*:latest"
 UntagImages "oryxdevmcr.azurecr.io/public/oryx/*:*.*"
+
+# Untag images with Staging ACR related tags
```
kichalla: This is required since "Release" related stuff pulls in images onto the build agent.

build/build-runtimeimages.sh
```diff
@@ -62,82 +67,68 @@ then
...
+            clearedOutput=true
+        fi
+
+        # add new content
+        echo
+        echo "Updating artifacts file with the built runtime image information..."
+    	echo "$dockerHubRuntimeImageTagNameRepo:latest" >> $DOCKERHUB_RUNTIME_IMAGES_ARTIFACTS_FILE
+    	echo "$dockerHubRuntimeImageTagNameRepo:$uniqueTag" >> $DOCKERHUB_RUNTIME_IMAGES_ARTIFACTS_FILE
+        echo "$acrRuntimeImageTagNameRepo:latest" >> $ACR_RUNTIME_IMAGES_ARTIFACTS_FILE
+        echo "$acrRuntimeImageTagNameRepo:$uniqueTag" >> $ACR_RUNTIME_IMAGES_ARTIFACTS_FILE
```
arroyc: nit: indentation

vsts/scripts/dockerCleanup.sh
```diff
@@ -38,20 +38,29 @@ docker images
...
 UntagImages "oryxmcr.azurecr.io/public/oryx/*:latest"
 UntagImages "oryxmcr.azurecr.io/public/oryx/*:*.*"
 
+# Untag older support images that are used to build Oryx build image
+# NOTE: Keep these around for a while so that they get untag across ALL the build agents
+UntagImages "mcr.microsoft.com/oryx/python-build-2.7:20190322.0"
```
kichalla: I put this here just as a template to make sure we tag older `base` build images.

arroyc:
this will break release flow .. so please make sure change the release pipeline according to changes made here regarding tag names.

vsts/pipelines/nightly.yml
```diff
@@ -20,8 +20,6 @@ jobs:
       echo "##vso[task.setvariable variable=BuildBuildImages;]true"
       echo "##vso[task.setvariable variable=TestBuildImages;]true"
       echo "##vso[task.setvariable variable=PushBuildImages;]true"
-      echo "##vso[task.setvariable variable=PushRuntimeImages;]false"
-      echo "##vso[task.setvariable variable=EmbedBuildContextInImages;]true"
```
kichalla: When build context is embedded, it causes cache invalidation (due to changes in git commit and build number) and hence causing the build to be slow. Since we `throw away` the images built in Nightly build, we can ignore it. We need to  however embed in CI builds.

vsts/pipelines/nightly.yml
```diff
@@ -51,12 +47,7 @@ jobs:
     name: OryxLinux
   steps:
   - script: |
-      echo "##vso[task.setvariable variable=BuildBuildImages;]false"
-      echo "##vso[task.setvariable variable=BuildRuntimeImages;]false"
       echo "##vso[task.setvariable variable=TestIntegration;]true"
-      echo "##vso[task.setvariable variable=PushBuildImages;]false"
```
kichalla: We need to be consistent. Either we specify all the variables with `false` and `true` or just have `true`. I saw a mix of these in one job and not the other. We could just agree that unless explicitly specified, its always `false`

arroyc: I agree with you, we should write explicitly IMO

vsts/scripts/release-runtimeimages.sh
arroyc: I checked in release we don't use this script, can you make sure we use this script to release runtime images https://devdiv.visualstudio.com/DevDiv/_apps/hub/ms.vss-releaseManagement-web.cd-workflow-hub?definitionId=1380&_a=definition-tasks&environmentId=6856

kichalla:
Closing this PR as I am ready to send out another one which does not have broader changes as in this PR.

PR #46, opened 2019-04-03
CI: Re-tag the pulled image with the correct repo name so that tests use that image

kichalla:


vsts/scripts/pullandretag.sh
```diff
@@ -12,8 +12,12 @@ while read sourceImage; do
     echo "Pulling the source image $sourceImage ..."
     docker pull "$sourceImage" | sed 's/^/     /'
         
-    # Trim the build number and append the '':latest' to end of it
+    # Trim the build number tag and append the '':latest' to end of it
     newtag="${sourceImage%:*}:latest"
+
+    # Replace the ACR registry repository name with a name that the tests know about
+    newtag=$(echo "$newtag" | sed 's,oryxdevmcr.azurecr.io/public/oryx,oryxdevms,g')
```
gpcastro: shouldn't we update the test framework instead? I'm thinking that since our mid term plan is to actually move to ACR/MCR only, using their naming makes sense

PR #49, opened 2019-04-03
stop test containers when tests fail

arroyc:
it creates problem in build agents when we have an immortal container .. weird but true docker kill and docker stop sometimes doesn't work .. so it's better to stop the container for exceptions

### DIFF
--- a/tests/Oryx.Tests.Common/EndToEndTestHelper.cs
+++ b/tests/Oryx.Tests.Common/EndToEndTestHelper.cs
@@ -172,7 +172,7 @@ namespace Microsoft.Oryx.Tests.Common
             }
             finally
             {
-                if (runResult != null && runResult.Exception == null)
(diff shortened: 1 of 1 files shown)


tests/Oryx.Tests.Common/EndToEndTestHelper.cs
```diff
@@ -172,11 +172,8 @@ public static class EndToEndTestHelper
             }
             finally
             {
-                if (runResult != null && runResult.Exception == null)
-                {
-                    // Stop the container so that shared resources (like ports) are disposed.
-                    dockerCli.StopContainer(runResult.ContainerName);
-                }
+                // Stop the container so that shared resources (like ports) are disposed.
+                dockerCli.StopContainer(runResult.ContainerName);
```
kichalla: Need a guard if `runResult == null`

PR #52, opened 2019-04-04
fix retagging in nightly builds with oryxtests images

arroyc:
We use both oryxdevms/build and oryxtests/build  in our tests, so we need to retag iamges  with oryxtests as well

### DIFF
--- a/Oryx.sln
+++ b/Oryx.sln
@@ -30,6 +30,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "build", "build", "{0FEBE7DD
 		build\build-constants.yaml = build\build-constants.yaml
 		build\build-runtimeimages.sh = build\build-runtimeimages.sh
 		build\build-solution.sh = build\build-solution.sh
+		build\build-testimage.sh = build\build-testimage.sh
 		build\test-buildimages.sh = build\test-buildimages.sh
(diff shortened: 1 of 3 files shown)


vsts/scripts/pullandretag.sh
```diff
@@ -14,13 +14,18 @@ while read sourceImage; do
         
     # Trim the build number tag and append the '':latest' to end of it
     newtag="${sourceImage%:*}:latest"
+	 
+    # Replace the ACR registry repository name with a oryxtests that the tests know about
+    oryxtesttag=$(echo "$newtag" | sed 's,oryxdevmcr.azurecr.io/public/oryx,oryxtests,g')
```
kichalla: Good catch about `oryxtests` but I think this change alone here is not sufficient as `oryxtests` is indeed a different image entirely with restricted access. We need to push & pull that image too now. Currently this image doesn't get written to the artifacts file, but we would need to now and need to make sure it does not get released to 'staging' repository during a release.

arroyc: I think we can do it without pushing it to dev acr as well, I applied necessary changes

kichalla: So we would be building the `oryxtests/build` 2 times? Once when building the build image and running its tests and once before running the integration tests? This feels more like a hack, but let's do it for now and file an issue to clean this up later.

PR #53, opened 2019-04-04
Restructure PHP dockerfile instructions to prevent the cache bust very early

kichalla:


images/runtime/php/Dockerfile.template
```diff
@@ -13,17 +13,6 @@ FROM php:%PHP_BASE_IMAGE%
 SHELL ["/bin/bash", "-c"]
 ENV PHP_VERSION %PHP_VERSION%
 
-# Bake Application Insights key from pipeline variable into final image
-ARG AI_KEY
-ENV ORYX_AI_INSTRUMENTATION_KEY=${AI_KEY}
-
-# COPY images/runtime/python/install-dependencies.sh /tmp/scripts/install-dependencies.sh
-# RUN /tmp/scripts/install-dependencies.sh
```
kichalla: @dorfire you ignored my PR comment about this :angry: https://github.com/Microsoft/Oryx/pull/5/files#r267138254 ...lol

dorfire: These lines are commented out though..

kichalla: Yeah, probably I should be more clear...the thing is the `COPY` command from the `Go` image artifacts is causing a cache bust because of which the later instructions in this docker stage are getting built again. So I had to re-arrange the instructions a bit...in general, the idea is to have the instructions which do not change quite often to be at the lowest layers of an image, this way we can maximize the cache usage.

PR #54, opened 2019-04-04
Parallel base image builds 

dorfire:


vsts/pipelines/validation.yml
```diff
@@ -4,15 +4,13 @@ resources:
 jobs:
 - job: Job_Security
   displayName: Security
-  condition: succeeded()
```
dorfire: (these `succeeded()` aren't needed since it's the default)

build/build-buildimage-bases.sh
```diff
@@ -7,59 +7,70 @@
...
 IMAGE_TAG="${BUILD_NUMBER:-latest}"
+BUILD_IMAGE_PREFIX="$__REPO_DIR/images/build"
 
 # Clean artifacts
 mkdir -p `dirname $BUILD_IMAGE_BASES_ARTIFACTS_FILE`
 > $BUILD_IMAGE_BASES_ARTIFACTS_FILE
 
-# Build Python
-docker build -f $__REPO_DIR/images/build/python/prereqs/Dockerfile -t "python-build-prereqs" $__REPO_DIR
+case $1 in
```
gpcastro: nit: put $1 in a local variable with a meaningful name, perhaps adding a validation if it is present when the script is run manually

dorfire: There's some form of validation in the last `case` clause - seems enough?

vsts/pipelines/_buildimage-bases.job-template.yml
```diff
@@ -0,0 +1,39 @@
...
+    condition: true
+
+  - task: Docker@1
+    displayName: Push built images to dev ACR
+    inputs:
+      command: push
+      azureSubscriptionEndpoint: $(ascName)
+      azureContainerRegistry: $(acrName)
+      pushMultipleImages: true
+      imageNamesPath: '$(Build.ArtifactStagingDirectory)/images/build-image-bases.txt'
```
kichalla: There could be a potential race condition here since the same file name is used by parallel build jobs (for example, a python build could overwrite this file which had content written by a php build). We should rather have unique file names which are based on the 'image directory' being built.

build/build-buildimage-bases.sh
```diff
@@ -7,59 +7,71 @@
...
 set -e
 
 source __variables.sh
 
+IMAGE_DIR_TO_BUILD=$1
 IMAGE_TAG="${BUILD_NUMBER:-latest}"
+BUILD_IMAGE_PREFIX="$__REPO_DIR/images/build"
 
 # Clean artifacts
 mkdir -p `dirname $BUILD_IMAGE_BASES_ARTIFACTS_FILE`
```
kichalla: Based on my other comment, we can make this file name dynamic which uses the `IMAGE_DIR_TO_BUILD` to construct it

vsts/pipelines/_buildimage-bases.job-template.yml
kichalla: Thinking more, looks like we are not going to hit the race condition issue here as `$(Build.ArtifactStagingDirectory)` resolves to a directory on the build agent itself and not on some common share. But there would race condition for the below task `Publish build artifacts` where parallel jobs would override the file with the same name.

vsts/pipelines/_buildimage-bases.job-template.yml
```diff
@@ -0,0 +1,39 @@
...
+
+  - task: PublishBuildArtifacts@1
```
kichalla: Let's add a docker cleanup task too here as we want to untag the image having the build number(as otherwise those images will stay on the agent consuming space). You could copy the task from our other yml files

arroyc:
oh one more thing, please add a security job in build-base-image yml as well, else this security job gets auto injected by azure pipeline and creates unnecessary alerts

PR #58, opened 2019-04-08
add cleanup for mounted directories

arroyc:
we have decomposed docker cleanup task in two different script now, one for untag and deleting images etc and another one for removing mounted directories from /tmp which gets generated every time integration tests ran. So ci builds should adopt that.


vsts/pipelines/_buildTemplate.yml
```diff
@@ -178,6 +178,12 @@ steps:
     scriptPath: ./vsts/scripts/dockerCleanup.sh
   condition: or(eq(variables['TestBuildImages'], 'true'), eq(variables['TestRuntimeImages'], 'true'), eq(variables['TestIntegration'], 'true'))
 
+- task: ShellScript@2
+  displayName: 'Clean up Docker artifacts from mounted directory'
```
kichalla: nit: `Clean up Docker mounted directories`?

PR #59, opened 2019-04-08
allowing manual builds to be signed only from master and patch branches

arroyc:
allowing manual builds to be signed only from they are master and patch branches and blocking signing when they are from other dev/random branches

### DIFF
--- a/vsts/pipelines/_signbinary.yml
+++ b/vsts/pipelines/_signbinary.yml
@@ -7,7 +7,7 @@ steps:
 - powershell: |
    Write-Host "##vso[task.setvariable variable=SignType]real"
   displayName: 'Set sign binary variable'
-  condition: and(in(variables['Build.Reason'], 'Schedule', 'Manual'), eq(variables['Build.DefinitionName'], 'Oryx-CI'))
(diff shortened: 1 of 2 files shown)


vsts/pipelines/_signbinary.yml
```diff
@@ -7,7 +7,7 @@ steps:
 - powershell: |
    Write-Host "##vso[task.setvariable variable=SignType]real"
   displayName: 'Set sign binary variable'
-  condition: and(in(variables['Build.Reason'], 'Schedule', 'Manual'), eq(variables['Build.DefinitionName'], 'Oryx-CI'))
+  condition: and(in(variables['Build.Reason'], 'Schedule', 'Manual'), eq(variables['Build.DefinitionName'], 'Oryx-CI'), or(startsWith(variables['Build.SourceBranch'], 'refs/heads/master'), startsWith(variables['Build.SourceBranch'],'refs/heads/patch/' )))
```
kichalla: Loooooooooong line 😄 ...Could you see if you can break it within the 120 limit? https://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines

arroyc: done !

PR #63, opened 2019-04-09
Enables feature to zip all output from build and running it in runtime container

kichalla:


build/build-runtimeimages.sh
```diff
@@ -49,11 +49,6 @@ echo
 echo "Generating Dockerfiles for Python runtime images..."
 $REPO_DIR/images/runtime/python/generateDockerfiles.sh
 
-echo
```
kichalla: Ignore this. Will revert.

src/BuildScriptGenerator/BaseBashBuildScriptProperties.cs
```diff
@@ -35,8 +35,6 @@ public class BaseBashBuildScriptProperties
         /// </summary>
         public string PostBuildScriptPath { get; set; }
 
-        public IEnumerable<string> DirectoriesToExcludeFromCopyToBuildOutputDir { get; set; }
```
kichalla: Removed this from common base bash script because each platform will handle it explicitly

src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
```diff
@@ -1,25 +1,56 @@
+tmpOutputDir="/tmp/puboutput"
+zippedOutputFileName=oryx_output.tar.gz
```
kichalla: We probably need to come up with a naming convention for any folders or files that Oryx creates

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -124,7 +129,13 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
                 buildProperties[NodeConstants.NodeModulesFileBuildProperty] = compressedNodeModulesFileName;
             }
 
+            var exlcudedDirs = GetDirectoriesToExcludeFromCopyToBuildOutputDir(context);
+
             bool pruneDevDependencies = ShouldPruneDevDependencies(context);
+            bool zipAllOutput = ShouldZipAllOutput(context);
+            // TODO: tightly coupled
```
kichalla: In runtime image the case of the boolean value is important i.e runtime image script expects it to be lower case. Seems like doing a case-INsensitive comparison is not straightforward in bash and also seems to depend on the shell too.

src/BuildScriptGeneratorCli/BuildScriptGenerator.cs
```diff
@@ -47,7 +47,9 @@ public bool TryGenerateScript(out string generatedScript)
                     EnableNodeJs = !environment.DisableNodeJs,
                     EnablePython = !environment.DisablePython,
                     EnablePhp = !environment.DisablePhp,
-                    DisableMultiPlatformBuild = environment.DisableMultiPlatformBuild
+                    
+                    // Temporarily disable multi-platform builds
+                    DisableMultiPlatformBuild = true
```
kichalla: @gpcastro  As discussed earlier, I disabled mutli-platform build for now. Would this be the best place for it?

src/startupscriptgenerator/dotnetcore/main.go
```diff
@@ -48,6 +50,20 @@ func main() {
...
 	}
 
+	srcFolder := fullPublishedOutputPath
+	if srcFolder == "" {
+		// The output folder is a sub-directory of this source directory
+		srcFolder = filepath.Join(fullSourcePath, "oryx_publish_output")
+	}
+	buildManifest := common.GetBuildManifest(srcFolder)
+	if buildManifest.ZipAllOutput == "true" {
+		destFolder := "/tmp/output"
```
kichalla: Need to come up with a standard naming pattern for Oryx artifacts

src/startupscriptgenerator/node/main.go
```diff
@@ -39,6 +39,17 @@ func main() {
 
 	common.SetGlobalOperationId(fullAppPath)
 
+	buildManifest := common.GetBuildManifest(fullAppPath)
```
kichalla: @gpcastro As discussed earlier, we check the manifest to see if the output is compressed. If yes, we un-compress it using a separate script and then call into other logic which remains unchanged.

tests/Oryx.Integration.Tests/Oryx.Integration.Tests.csproj
```diff
@@ -20,6 +20,8 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\src\BuildScriptGeneratorCli\BuildScriptGeneratorCli.csproj" />
```
kichalla: To access constants from product code...otherwise the magic strings are getting out of hand and it is easy to make error

src/BuildScriptGenerator/Node/NodeConstants.cs
```diff
@@ -5,26 +5,30 @@
 
 namespace Microsoft.Oryx.BuildScriptGenerator.Node
 {
-    internal static class NodeConstants
+    public static class NodeConstants
```
kichalla: For now I made this public, but we can revisit this when we plan to release this as a library and do not want to expose this. In that case we can do the `InternalsVisibleTo`

src/BuildScriptGenerator/BuildPropertiesHelper.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    internal static class BuildPropertiesHelper
+    {
+        public static bool IsTrue(string propertyKeyName, BuildScriptGeneratorContext context, bool valueIsRequired)
```
arroyc: nit: Can you rename it something like "IsPropertySet" etc?

tests/BuildScriptGenerator.Tests/Node/NodeBuildScriptGenerationTest.cs
```diff
@@ -331,7 +366,7 @@ public void GeneratedScript_ZipsNodeModules_IfZipNodeProperty_IsTarGz()
             repo.AddFile(PackageJsonWithBuildScript, NodeConstants.PackageJsonFileName);
             var context = CreateScriptGeneratorContext(repo);
             context.LanguageVersion = "8.2.1";
-            context.Properties["compress_node_modules"] = "tar-gz";
+            context.Properties[NodeConstants.CompressNodeModulesPropertyKey] = NodeConstants.TarGzNodeModulesOption;
```
arroyc: here.. I was thinking if its better way to just know how you zipped rather than doing what you are doing here, current implementation works no doubt, my point is if my proposal is cleaner way to do it

src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
```diff
@@ -1,25 +1,56 @@
...
-echo Restoring packages ...
+echo Restoring packages...
 echo
 dotnet restore "{{ ProjectFile }}"
 
-echo
-echo "Publishing to directory $outputDir..."
-echo
-dotnet publish "{{ ProjectFile }}" -c Release -o "$outputDir"
+{{ if ZipAllOutput }}
```
gpcastro: Can't we have this zip all output control only in the base script? That is, each snippet controls how it produces the output, but in the end there's a single piece of code that zips all of it?

src/BuildScriptGenerator/Node/NodeConstants.cs
gpcastro: where is it used?

gpcastro: We should already have an internalsVisibleTo the tests... I'd do that now rather than later, since once it is public we might start relying on it and fixing it later will be more expensive

src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
kichalla: That's a good point. Will look into it.

src/BuildScriptGenerator/Node/NodeConstants.cs
kichalla: Sure, will change it

PR #66, opened 2019-04-10
Fixed nightly build failures in DB integration tests

dorfire:


build/test-integration.sh
```diff
@@ -13,10 +13,12 @@ source $REPO_DIR/build/__variables.sh
...
 echo
-echo "Running integration tests..."
+echo "Running integration tests with filter '$testCaseFilter'..."
 echo
 testProjectName="Oryx.Integration.Tests"
 cd "$TESTS_SRC_DIR/$testProjectName"
-# Run non-AKS tests temporarily
-dotnet test --filter E2E!=AKS --test-adapter-path:. --logger:"xunit;LogFilePath=$ARTIFACTS_DIR/testResults/$testProjectName.xml" -c $BUILD_CONFIGURATION
+
+dotnet test --filter $testCaseFilter --test-adapter-path:. --logger:"xunit;LogFilePath=$ARTIFACTS_DIR/testResults/$testProjectName.xml" -c $BUILD_CONFIGURATION
```
kichalla: 👍

tests/Oryx.Integration.Tests/LocalDockerTests/DatabaseTestsBase.cs
```diff
@@ -19,27 +19,25 @@ public abstract class DatabaseTestsBase
...
-        protected readonly int _appPort;
+        protected readonly int _hostPort;
+        protected readonly HttpClient _httpClient = new HttpClient();
 
         protected DatabaseTestsBase(ITestOutputHelper outputHelper, Fixtures.DbContainerFixtureBase dbFixture)
         {
             _output = outputHelper;
             _dbFixture = dbFixture;
-            _appPort = 8080 + _rand.Next(100);
+            _hostPort = 8080 + _rand.Next(100);
```
kichalla: I am wondering how the changes in this PR fixes the issue that we are seeing...I strongly suspect that the issue we might be seeing is due to collisions with the port numbers that are getting generated since uniqueness is not guaranteed here...I think we should just stick to static ports (for now).

dorfire: The errors we've seen were from the DB containers, not the app containers, so I'm compelled to leave this as-is for now. You might be right in general - it's just not something I'd like to tackle in this PR :)

kichalla: Ah I see, got it..

arroyc:
For some reason I can't do review on this PR, can only comment. My comment is "approved"

PR #67, opened 2019-04-11
Added `unzip` to all Node.js runtime images

dorfire:


tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
```diff
@@ -107,6 +107,38 @@ public void NodeVersionMatchesImageName(string nodeTag, string nodeVersion)
...
+        [InlineData("8.8")]
+        [InlineData("8.9")]
+        [InlineData("8.11")]
+        [InlineData("8.12")]
+        [InlineData("9.4")]
+        [InlineData("10.1")]
+        [InlineData("10.10")]
+        [InlineData("10.12")]
+        [InlineData("10.14")]
+        public void NodeImage_Contains_RequiredPrograms(string nodeTag)
```
kichalla: 👍

tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
```diff
@@ -107,6 +107,38 @@ public void NodeVersionMatchesImageName(string nodeTag, string nodeVersion)
...
+        {
+            // Arrange & Act
+            var result = _dockerCli.Run(
+                $"oryxdevms/node-{nodeTag}:latest",
+                "sh", new[] { "-c", "which tar && which unzip" });
```
kichalla: nit: if you have to split the argument into separate line: we can agree on the following approach(which aspnetcore team follows):
Have all the arguments in the same line, if possible within the length limits
OR
Split each argument into its own separate line

dorfire: I wanted to keep it this way on purpose, so that it would resemble the actual command being run: `sh -c which tar && which unzip`.
WDYT?

kichalla: That is a fair point 👍

gpcastro:
I'm completing this PR as I need to work on the same files

PR #71, opened 2019-04-15
Make publish output directory as the apppath for .NET Core runtime images

kichalla:


src/startupscriptgenerator/dotnetcore/main.go
```diff
@@ -14,14 +14,16 @@ import (
 func main() {
 	common.PrintVersionInfo()
 
+	appPathPtr := flag.String(
```
kichalla: Following a similar pattern as other platforms' startup script generators. This simplifies the scenario for dotnetcore too without the need for figuring out the output directory.

PR #72, opened 2019-04-15
DotnetCore Runtime: Use startup file name from build manifest file if present

kichalla:


src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -124,6 +124,15 @@ func (gen *DotnetCoreStartupScriptGenerator) getStartupCommand() string {
 func (gen *DotnetCoreStartupScriptGenerator) getStartupDllFileName() string {
 	logger := common.GetLogger("dotnetcore.scriptgenerator.getStartupDllFileName")
 
+	manifestFilePath := filepath.Join(gen.AppPath, common.ManifestFileName)
+	if common.FileExists(manifestFilePath) {
+		logger.LogInformation("Found build manifest file at '%s'.", manifestFilePath)
+		buildManifest := common.GetBuildManifest(gen.AppPath)
+		return buildManifest.StartupFileName
```
gpcastro: what if the manifest exists, but the property is not there?

kichalla: Good catch, will fix it!

PR #73, opened 2019-04-16
Added further infrastructure to support the Oryx Buildpack

dorfire:


images/pack-builder/stack-base.Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM oryxdevms/build:latest
+
+LABEL io.buildpacks.stack.id=com.microsoft.oryx.stack
+
+# Configure non-root user
+RUN groupadd -g 1002 oryx_group && \
+	useradd -u 1001 -g oryx_group oryx_user && \
```
gpcastro: It is easier to read when the commands are all indented, so it looks like a script

PR #78, opened 2019-04-19
Provide a way to zip all output for dotnet core related builds

kichalla:


src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -72,68 +78,90 @@ public bool TryGenerateBashScript(BuildScriptGeneratorContext context, out strin
...
+                userSuppliedPlatform = selectedPlatform;
+
+                platformVersion = ctx.LanguageVersion;
+                if (string.IsNullOrEmpty(platformVersion))
                 {
-                    _logger.LogDebug("Detecting platform using {platformName}", platform.Name);
-                    var detectionResult = platform.Detect(ctx.SourceRepo);
-                    if (detectionResult != null)
+                    var detectionResult = userSuppliedPlatform.Detect(ctx.SourceRepo);
+                    if (string.IsNullOrEmpty(detectionResult.LanguageVersion))
```
gpcastro: we should probably check for null here. for, say, an empty repo, we won't detect the language or version, but the user might have specified the language version... this also brings the point that we probably want a test for an empty repo, and agree of what we should do in that case (e.g. fail gracefully at least)

src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
```diff
@@ -1,25 +1,182 @@
...
+	"{{ PostBuildScriptPath }}"
+	{{ end }}
+
+	# Zip only the contents and not the parent directory
+	mkdir -p "$ORIGINAL_DESTINATION_DIR"
+	cd "$tmpDestinationPublishDir"
+	tar -zcf ../$zippedOutputFileName .
+	cd "$tmpDestinationDir"
+
+	if [ "$CLEAN_DESTINATION_DIR" == "true" ]
```
gpcastro: is this PR still valid? Are you planning to update it with the changes from your other PR?

src/startupscriptgenerator/common/outputHelper.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+func ExtractZippedOutput(appFolder string) {
+	zipFileName := "oryx_output.tar.gz"
+	scriptPath := "/tmp/extractZippedOutput.sh"
+	scriptBuilder := strings.Builder{}
+	scriptBuilder.WriteString("#!/bin/sh\n")
```
gpcastro: are those standalone scripts? Ideally we should have a single script that can run the app

src/startupscriptgenerator/common/outputHelper.go
```diff
@@ -0,0 +1,68 @@
...
+	scriptBuilder.WriteString("    echo Deleting the file '" + zipFileName + "'...\n")
+	scriptBuilder.WriteString("    rm -f \"" + zipFileName + "\"\n")
+	scriptBuilder.WriteString("    echo \"Done.\"\n")
+	scriptBuilder.WriteString("fi\n\n")
+
+	WriteScript(scriptPath, scriptBuilder.String())
+	ExecuteCommand("/bin/sh", "-c", scriptPath)
+}
+
+func CopyOutputToIntermediateDir(srcFolder string, destFolder string) {
```
gpcastro: sugg: `CopyToFolder`?

src/startupscriptgenerator/dotnetcore/main.go
```diff
@@ -19,6 +21,10 @@ func main() {
 		".",
 		"The path to the published output of the application that is going to be run, e.g. '/home/site/wwwroot/'. " +
 		"Default is current directory.")
+	intermediatePathPtr := flag.String(
+		"intermediatePath",
```
gpcastro: "intermediatePath" sounds more appropriate to build than run. Maybe runFromPath?

tests/BuildScriptGenerator.Tests/DefaultBuildScriptGeneratorTest.cs
```diff
@@ -372,74 +372,6 @@ public void TryGenerateScript_CallsDetector_IfMultiPlatformIsOn_AndLangProvided(
             Assert.True(detector2.DetectInvoked);
         }
 
-        [Fact]
-        public void UsesMaxSatisfyingVersion_WhenOnlyMajorVersion_OfLanguageIsSpecified()
```
gpcastro: q: why are we getting rid of these tests?

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -50,6 +50,12 @@ public bool TryGenerateBashScript(BuildScriptGeneratorContext context, out strin
                 timedEvent.SetProperties(toolsToVersion);
             }
 
+            if (snippets != null && snippets.Count == 1 && snippets[0].IsFullScript)
```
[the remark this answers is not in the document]
kichalla: Sure, sounds good

src/startupscriptgenerator/common/outputHelper.go
kichalla: Fixed this.

tests/BuildScriptGenerator.Tests/DefaultBuildScriptGeneratorTest.cs
kichalla: As discussed offline, removing these tests are they not a common scenario (i.e where we would have multiple instances of platform for the same language name)

tests/BuildScriptGenerator.Tests/DefaultBuildScriptGeneratorTest.cs
```diff
@@ -149,6 +149,25 @@ public void TryGenerateScript_Throws_IfLanguageIsProvidedButNoVersion_AndCannotD
...
+            var context = CreateScriptGeneratorContext(
+                suppliedLanguageName: "test2",
+                suppliedLanguageVersion: null);
+
+            // Act & Assert
+            var exception = Assert.Throws<UnsupportedLanguageException>(
+                () => generator.TryGenerateBashScript(context, out var generatedScript));
+            Assert.Equal("'test2' platform is not supported. Supported platforms are: test1", exception.Message);
+        }
+
```
arroyc: +1

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
kichalla: There was already a test called `TryGenerateScript_Throws_IfLanguageIsProvidedButNoVersion_AndCannotDetectVersion`, but added a new one for language too: `TryGenerateScript_Throws_IfLanguageIsProvided_AndCannotDetectLanguage`

descriptive name are good ;-)

PR #79, opened 2019-04-19
code-less app-insights configuration for node apps

arroyc:
code-less configuration of application insight into oryx when user sets app insight environment variables in app service for node apps


Oryx.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27703.2042
+# Visual Studio Version 16
```
arroyc: I will revert these changes

src/BuildScriptGenerator/Constants.cs
```diff
@@ -10,5 +10,7 @@ public static class Constants
         public const string OryxEnvironmentSettingNamePrefix = "ORYX_";
         public const string BuildEnvironmentFileName = "build.env";
         public const string ManifestFileName = "oryx-manifest.toml";
+        public const string InjectAppInsightBuildProperty = "oryxSetAppInsight";
```
arroyc: will remove this .. no need

arroyc:
Tests are coming

images/build/Dockerfile
```diff
@@ -298,6 +298,10 @@ COPY --from=mcr.microsoft.com/oryx/php-build-7.3:20190325.6 /opt /opt
 COPY --from=buildscriptbinaries /opt/buildscriptgen/ /opt/buildscriptgen/
 RUN ln -s /opt/buildscriptgen/GenerateBuildScript /usr/local/bin/oryx
 
+# Copy applicationinsightloader for node
+COPY images/build/node/oryxappinsightloader.js /tmp
+RUN chmod +x /tmp/oryxappinsightloader.js
+
```
arroyc: I will remove this to create the file dynamically when all conditions are satisfied

images/build/node/oryxappinsightloader.js
```diff
@@ -0,0 +1,16 @@
+// --------------------------------------------------------------------------------------------
```
arroyc: I will make this dynamically and remove this

src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -125,6 +125,15 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 		startupCommand = common.ExtendPathForCommand(startupCommand, gen.SourcePath)
 	}
 
+	scriptBuilder.WriteString("if [ -f ./" + common.ManifestFileName + " ]; then\n")
+	scriptBuilder.WriteString("    isInManifest=$(cat " + common.ManifestFileName + " | grep -c 'NODE_OPTIONS')\n")
+	scriptBuilder.WriteString("    if [ $isInManifest -eq 0 ]; then\n")
+	scriptBuilder.WriteString("        echo \"Oryx has set up Code-less App-Insight...\"\n")
+	scriptBuilder.WriteString("        export NODE_OPTIONS='--require ./oryxappinsightloader.js'\n")
+	scriptBuilder.WriteString("    else\n")
+	scriptBuilder.WriteString("        export NODE_OPTIONS=$NODE_OPTIONS\n")
```
gpcastro: why do we need this line?

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -125,6 +125,26 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
...
+
+            // node_options is only supported in version 8.0.0 or newer and in 6.12.0
+            // so we will be able to set upapp insight only when node version is 6.12.0 or 8.0.0 or newer
+            if (!appInsightDependency
+                && !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(Constants.AppInsightKey))
+                && ( SemanticVersionResolver.CompareVersion(context.NodeVersion, "8.0.0") >= 0
+                || SemanticVersionResolver.CompareVersion(context.NodeVersion, "6.12.0") == 0))
+            {
+                appInsightInjectCommand =
+                string.Concat(NodeConstants.NpmPackageInstallCommand, " --save ", NodeConstants.NodeAppInsightPackageName);
```
gpcastro: indentation

src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -125,6 +125,15 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 		startupCommand = common.ExtendPathForCommand(startupCommand, gen.SourcePath)
 	}
 
+	scriptBuilder.WriteString("if [ -f ./" + common.ManifestFileName + " ]; then\n")
+	scriptBuilder.WriteString("    isInManifest=$(cat " + common.ManifestFileName + " | grep -c 'NODE_OPTIONS')\n")
+	scriptBuilder.WriteString("    if [ $isInManifest -eq 0 ]; then\n")
+	scriptBuilder.WriteString("        echo \"Oryx has set up Code-less App-Insight...\"\n")
+	scriptBuilder.WriteString("        export NODE_OPTIONS='--require ./oryxappinsightloader.js'\n")
```
gpcastro: we should also append existing values of $NODE_OPTIONS the user might have set as an app setting

src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -125,6 +125,15 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 		startupCommand = common.ExtendPathForCommand(startupCommand, gen.SourcePath)
 	}
 
+	scriptBuilder.WriteString("if [ -f ./" + common.ManifestFileName + " ]; then\n")
+	scriptBuilder.WriteString("    isInManifest=$(cat " + common.ManifestFileName + " | grep -c 'NODE_OPTIONS')\n")
```
gpcastro: what does having node_options in the file mean? looks liek we're use a property called NODE_OPTIONS to indicate that app insights should be inject, which is not clear. We could use something like `InjectAppInsights=true` in the manifest.

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -82,6 +82,33 @@ echo "Running '{{ PackageInstallCommand }}'..."
...
+echo "Preparing Envrionment to set up application insight..."
+echo "Running '{{ OryxAppInsightInjectCommand }}'..."
+echo
+{{ OryxAppInsightInjectCommand }}
+echo
+echo "Creating application insight loader file to '$DESTINATION_DIR'"
+echo "// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
```
dorfire: Maybe add another single-line comment like "Added by Oryx"? just to make things clearer

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -82,6 +82,33 @@ echo "Running '{{ PackageInstallCommand }}'..."
...
+
+if (process.env.APPINSIGHTS_INSTRUMENTATIONKEY) {
+    try {
+        appInsights
+            .setup()
+            .start();
+    } catch (e) {
+        console.error(e);
+    }
+}">oryxappinsightloader.js
```
dorfire: this constant name is used in both C# and Go, so it would be a good fit for `build-constants.yaml`

src/BuildScriptGenerator/Node/NodeBashBuildSnippetProperties.cs
```diff
@@ -19,7 +19,8 @@ public class NodeBashBuildSnippetProperties
             string compressNodeModulesCommand,
             string compressedNodeModulesFileName,
             bool configureYarnCache = false,
-            bool pruneDevDependencies = false)
+            bool pruneDevDependencies = false,
+            string oryxAppInsightInjectCommand = null)
```
dorfire: why the `oryx` prefix? we didn't add it to other param names (e.g. `compressNodeModulesCommand`)

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -82,6 +82,33 @@ echo "Running '{{ PackageInstallCommand }}'..."
...
+echo
+{{ OryxAppInsightInjectCommand }}
+echo
+echo "Creating application insight loader file to '$DESTINATION_DIR'"
+echo "// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+var appInsights = require('applicationinsights');
```
dorfire: utilize `NodeAppInsightPackageName`

src/BuildScriptGenerator/SemanticVersionResolver.cs
```diff
@@ -24,5 +24,20 @@ public static string GetMaxSatisfyingVersion(string providedVersion, IEnumerable
 
             return null;
         }
+
+        public static int CompareVersion(string providedVersion, string supportedVersion)
```
dorfire: CompareVersion => CompareVersions

src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -135,6 +135,14 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 		startupCommand = common.ExtendPathForCommand(startupCommand, gen.SourcePath)
 	}
 
+	logger.LogInformation("Checking configuration to setup appinsight auto collection if needed")
```
dorfire: There's no configuration check actually being done here - the only thing that happens is that the Bash script is **written**. It'll be executed at a later point in time

src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -135,6 +135,14 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
...
+	scriptBuilder.WriteString("if [ -f ./" + common.ManifestFileName + " ]; then\n")
+	scriptBuilder.WriteString("    isInManifest=$(cat " + common.ManifestFileName + " | grep -c 'OryxInjectedAppInsight')\n")
+	scriptBuilder.WriteString("    if [ $isInManifest -eq 0 ]; then\n")
+	scriptBuilder.WriteString("        echo \"Oryx has set up Code-less App-Insight...\"\n")
+	scriptBuilder.WriteString("        export NODE_OPTIONS='--require ./oryxappinsightloader.js '$NODE_OPTIONS\n")
```
dorfire: magic string (`oryxappinsightloader.js`)

tests/BuildScriptGenerator.Tests/Node/NodeBuildScriptGenerationTest.cs
```diff
@@ -452,6 +453,116 @@ public void GeneratedScript_DoesNotZipNodeModules_IfZipNodeModulesEnvironmentVar
             Assert.True(scriptGenerator.IsCleanRepo(repo));
         }
 
+        [Fact ]
+        public void GeneratedScript_DoesNotConfigureAppInsight_IfAppInsightEnvironmentVariable_NotSet()
```
dorfire: AppInsight => AppInsights

tests/BuildScriptGenerator.Tests/Node/NodeBuildScriptGenerationTest.cs
```diff
@@ -452,6 +453,116 @@ public void GeneratedScript_DoesNotZipNodeModules_IfZipNodeModulesEnvironmentVar
...
+            Assert.NotNull(snippet);
+            Assert.DoesNotContain("applicationinsights", snippet.BashBuildScriptSnippet);
+            Assert.Equal(
+                TemplateHelpers.Render(TemplateHelpers.TemplateResource.NodeBuildSnippet, expected),
+                snippet.BashBuildScriptSnippet);
+            Assert.True(scriptGenerator.IsCleanRepo(repo));
+        }
+
+        [Fact]
+        public void GeneratedScript_DoesNotConfigureAppInsight_IfNodeVersionCondition_Unsatisfied()
```
dorfire: important test 👍

tests/Oryx.BuildImage.Tests/NodeJSSampleAppsTest.cs
```diff
@@ -130,6 +130,162 @@ public void Build_CopiesOutput_ToNestedOutputDirectory()
...
+        {
+            // Arrange
+            var volume = CreateWebFrontEndVolume();
+            var appDir = volume.ContainerDir;
+            var nestedOutputDir = "/tmp/output";
+            var script = new ShellScriptBuilder()
+                .AddCommand($"oryx build {appDir} -o {nestedOutputDir} --log-file {appDir}/1.log")
+                .AddDirectoryExistsCheck($"{nestedOutputDir}/node_modules")
+                .AddFileExistsCheck($"{nestedOutputDir}/oryxappinsightloader.js")
+                .AddFileExistsCheck($"{nestedOutputDir}/oryx-manifest.toml")
```
dorfire: one is with a dash, one is without - let's align them to the same convention

tests/Oryx.Integration.Tests/LocalDockerTests/NodeEndToEndTests.cs
```diff
@@ -347,6 +347,66 @@ public async Task CanBuildAndRun_NodeApp(string nodeVersion)
...
+                .AddCommand($"oryx build {appDir} -o {appDir} {spcifyNodeVersionCommand} --log-file {appDir}/1.log")
+                .AddDirectoryExistsCheck($"{appDir}/node_modules")
+                .AddFileExistsCheck($"{appDir}/oryxappinsightloader.js")
+                .AddFileExistsCheck($"{appDir}/oryx-manifest.toml")
+                .AddStringExistsInFileCheck("OryxInjectedAppInsight=\"True\"", $"{appDir}/oryx-manifest.toml")
+                .AddStringExistsInFileCheck("Oryx setting up applicationinsight for auto-collection telemetry... ", $"{appDir}/1.log")
+                .ToString();
+
+            var runScript = new ShellScriptBuilder()
+                .AddCommand($"export APPINSIGHTS_INSTRUMENTATIONKEY=asdas")
```
dorfire: utilize aIKey

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -125,6 +125,24 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
...
+            if (!appInsightDependency
+                && !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(Constants.AppInsightKey))
+                && ( SemanticVersionResolver.CompareVersion(context.NodeVersion, "8.0.0") >= 0
+                || SemanticVersionResolver.CompareVersion(context.NodeVersion, "6.12.0") == 0
+                || SemanticVersionResolver.CompareVersion(context.LanguageVersion, "8.0.0") >= 0
+                || SemanticVersionResolver.CompareVersion(context.LanguageVersion, "6.12.0") == 0))
+            {
+                appInsightInjectCommand =
+                    string.Concat(NodeConstants.NpmPackageInstallCommand, " --save ", NodeConstants.NodeAppInsightPackageName);
+                buildProperties[NodeConstants.OryxNodeOptions] = true.ToString();
```
kichalla: Can you use another key name other than `OryxNodeOptions` as it does not indicate what that key is for? something like `loadAppInsightsModule` or something

src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -135,6 +135,14 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 		startupCommand = common.ExtendPathForCommand(startupCommand, gen.SourcePath)
 	}
 
+	logger.LogInformation("Checking configuration to setup appinsight auto collection if needed")
+	scriptBuilder.WriteString("if [ -f ./" + common.ManifestFileName + " ]; then\n")
+	scriptBuilder.WriteString("    isInManifest=$(cat " + common.ManifestFileName + " | grep -c 'OryxInjectedAppInsight')\n")
```
kichalla: just `source` the manifest file (this was the reason we chose to use `key=value` format in the file) and access the variable name

kichalla: Thinking more we would be sourcing this file here as well as in line `62` in this file. I think we should do only once, maybe at the very beginning of the script

tests/BuildScriptGenerator.Tests/Node/NodeBuildScriptGenerationTest.cs
```diff
@@ -452,6 +453,116 @@ public void GeneratedScript_DoesNotZipNodeModules_IfZipNodeModulesEnvironmentVar
...
+            Assert.NotNull(snippet);
+            Assert.DoesNotContain("applicationinsights", snippet.BashBuildScriptSnippet);
+            Assert.Equal(
+                TemplateHelpers.Render(TemplateHelpers.TemplateResource.NodeBuildSnippet, expected),
+                snippet.BashBuildScriptSnippet);
+            Assert.True(scriptGenerator.IsCleanRepo(repo));
+        }
+
+        [Theory]
+        [InlineData("8.0.0")]
```
kichalla: You can add more variations here (like 6.12.0)...is it expected to work for `6.12.1` as its a minor patch update?

tests/BuildScriptGenerator.Tests/Node/NodeBuildScriptGenerationTest.cs
```diff
@@ -452,6 +453,116 @@ public void GeneratedScript_DoesNotZipNodeModules_IfZipNodeModulesEnvironmentVar
...
+            Assert.Equal(
+                TemplateHelpers.Render(TemplateHelpers.TemplateResource.NodeBuildSnippet, expected),
+                snippet.BashBuildScriptSnippet);
+            Assert.True(scriptGenerator.IsCleanRepo(repo));
+        }
+
+        [Fact]
+        public void GeneratedScript_DoesNotConfigureAppInsight_IfNodeVersionCondition_Unsatisfied()
+        {
+            var nodeVersion = "6.9.3";
```
kichalla: I would change this to a Theory test and add more variations for this

tests/BuildScriptGenerator.Tests/Node/NodeBuildScriptGenerationTest.cs
```diff
@@ -452,6 +453,116 @@ public void GeneratedScript_DoesNotZipNodeModules_IfZipNodeModulesEnvironmentVar
...
+            // Arrange
+            var scriptGenerator = GetNodePlatformInstance(defaultNodeVersion: nodeVersion);
+            var repo = new MemorySourceRepo();
+            repo.AddFile(PackageJsonWithBuildScript, NodeConstants.PackageJsonFileName);
+            Environment.SetEnvironmentVariable(Constants.AppInsightKey, "xyz");
```
kichalla: Ideally we do not want unit tests to depend on external source like `System.Environment` here. For this reason we create a layer of indirection through the options class which can be constructed in a test and a value be set on it.
The reason this is bad is because it causes a race condition between tests. Like if a test from a different class(hence would ran in parallel) sets the variable to be empty, then this test can fail (sometimes).

tests/Oryx.BuildImage.Tests/NodeJSSampleAppsTest.cs
```diff
@@ -130,6 +130,162 @@ public void Build_CopiesOutput_ToNestedOutputDirectory()
...
 
+        [Fact]
+        public void BuildNodeApp_ConfigureAppInsight__WithDefaultNodeVersion_AIEnvironmentVariableSet()
+        {
+            // Arrange
+            var volume = CreateWebFrontEndVolume();
+            var appDir = volume.ContainerDir;
+            var nestedOutputDir = "/tmp/output";
+            var script = new ShellScriptBuilder()
+                .AddCommand($"oryx build {appDir} -o {nestedOutputDir} --log-file {appDir}/1.log")
```
kichalla: Based on what you mentioned in scrum meeting, I think we should not rely on the log file to verify in this test as it can make this test flaky and I do not feel its a bug actually since we cannot depend on when NLog decides to flush the logs to a file

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
arroyc: I will create a seperate PR for this, I don't want to do this in this PR

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
arroyc: what do you mean? if you are asking different variable name I followed npm site and there it's been used like this https://www.npmjs.com/package/applicationinsights

src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -151,6 +151,13 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 		startupCommand = common.ExtendPathForCommand(startupCommand, gen.SourcePath)
 	}
 
+	logger.LogInformation("Looking for appinsights loader and to export it to NODE_OPTIONS if needed")
+	scriptBuilder.WriteString("if [ -f ./" + common.ManifestFileName + " ]; then\n")
```
kichalla: Where is the manifest file being sourced? I think the one in line 62 is behind a conditional check, so it would be available always. I think we should always source it if the manifest file is available and we should do this source at the very beginning (may be after the `set -e` command i think)

tests/Oryx.Integration.Tests/LocalDockerTests/NodeEndToEndTests.cs
```diff
@@ -347,6 +347,75 @@ public async Task CanBuildAndRun_NodeApp(string nodeVersion)
                 });
         }
 
+        [Theory]
+        [InlineData("8.0")]
```
gpcastro: we have the class `TestValueGenerator` that can include all these values, perhaps named "AppInsightsNodeVersions`?

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -82,6 +82,34 @@ echo "Running '{{ PackageInstallCommand }}'..."
...
+{{ AppInsightsInjectCommand }}
+echo
+echo "Creating Application Insights loader file in '$DESTINATION_DIR'"
+echo "// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+// Created by Oryx
+var appInsights = require('{{ AppInsightsPackageName }}');
```
gpcastro: what happens if the package is not available?
I'm thinking that here we should always fail gracefully. For example, if the users run `npm install` by themselves and erase the packages that oryx installed, will we break the app?

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -125,6 +140,29 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
...
+                // so we will be able to set up app-insight only when node version is 6.12.0 or 8.0.0 or newer
+                if (!appInsightDependency
+                    && !string.IsNullOrEmpty(appInsightsKey)
+                    && (SemanticVersionResolver.CompareVersions(context.NodeVersion, "8.0.0") >= 0
+                    || SemanticVersionResolver.CompareVersions(context.NodeVersion, "6.12.0") == 0
+                    || SemanticVersionResolver.CompareVersions(context.LanguageVersion, "8.0.0") >= 0
+                    || SemanticVersionResolver.CompareVersions(context.LanguageVersion, "6.12.0") == 0))
+                {
+                    appInsightsInjectCommand =
+                        string.Concat(NodeConstants.NpmPackageInstallCommand, " --save ", NodeConstants.NodeAppInsightsPackageName);
```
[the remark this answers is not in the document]
arroyc: @gpcastro  .. I can see your point of view, but if we don't save it, next time when user trigger a code change, this will need to be executed, do we prefer to do that?

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
arroyc: @gpcastro see my comment above ...

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -125,6 +140,29 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
...
+
+            if (_environment != null)
+            {
+                var appInsightsKey = _environment.GetEnvironmentVariable(Constants.AppInsightsKey);
+
+                // node_options is only supported in version 8.0.0 or newer and in 6.12.0
+                // so we will be able to set up app-insight only when node version is 6.12.0 or 8.0.0 or newer
+                if (!appInsightDependency
+                    && !string.IsNullOrEmpty(appInsightsKey)
+                    && (SemanticVersionResolver.CompareVersions(context.NodeVersion, "8.0.0") >= 0
```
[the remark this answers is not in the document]
arroyc: My test was failing when I pass --language-version, so I added context.languageVersion

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -82,6 +82,35 @@ echo "Running '{{ PackageInstallCommand }}'..."
...
+{
+	var appInsights = require('{{ AppInsightsPackageName }}');
+	if (process.env.APPINSIGHTS_INSTRUMENTATIONKEY) {
+
+        appInsights
+            .setup()
+            .start();
+	}
+}catch (e) {
+        console.log(e);
```
gpcastro: Write some message to indicate what happened, e.g. "Application Insights could not be automatically configured for this application."

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -125,6 +128,29 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
...
+
+            // node_options is only supported in version 8.0.0 or newer and in 6.12.0
+            // so we will be able to set up app-insight only when node version is 6.12.0 or 8.0.0 or newer
+            if (!appInsightsDependency
+                && !string.IsNullOrEmpty(appInsightsKey)
+                && (SemanticVersionResolver.CompareVersions(context.NodeVersion, "8.0.0") >= 0
+                || SemanticVersionResolver.CompareVersions(context.NodeVersion, "6.12.0") == 0
+                || SemanticVersionResolver.CompareVersions(context.LanguageVersion, "8.0.0") >= 0
+                || SemanticVersionResolver.CompareVersions(context.LanguageVersion, "6.12.0") == 0))
+            {
```
gpcastro: sugg: put lines 131-144 in a separate method, e.g. `shouldInjectAppInsights`

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -343,6 +372,21 @@ private static bool ShouldPruneDevDependencies(BuildScriptGeneratorContext conte
             return ret;
         }
 
+        private static bool HasPackageDependencyExist(dynamic packageJson, string packageName)
```
gpcastro: `doesPackageDependencyExist`? or `isPackageDepedency`?

tests/BuildScriptGenerator.Tests/Node/NodeBuildScriptGenerationTest.cs
```diff
@@ -452,14 +453,161 @@ public void GeneratedScript_DoesNotZipNodeModules_IfZipNodeModulesEnvironmentVar
             Assert.True(scriptGenerator.IsCleanRepo(repo));
         }
 
+        [Theory]
+        [InlineData("4.4.7")]
```
gpcastro: move this list of node versions to the GenerateTestVAlues class, so we don't have to duplicate its values and we document what they mean

src/BuildScriptGenerator/Node/NodePlatform.cs
kichalla: or just `HasPackageDependency`

tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
```diff
@@ -275,11 +313,115 @@ public async Task RunNodeAppUsingProcessJson_withDebugging(string nodeVersion)
...
-                    Assert.Contains("devtoolsFrontendUrl", data);
+                    var data = await _httpClient.GetStringAsync($"http://localhost:{hostPort}/");
+                    Assert.Contains("Say It Again", data);
                 },
                 dockerCli: _dockerCli);
 
         }
+
+        [Theory]
+        [MemberData(nameof(TestValueGenerator.GetNodeVersions_SupportDebugging), MemberType = typeof(TestValueGenerator))]
```
gpcastro: are we adding app insights only to the versions that support debugging, or do the versions just coincide?

arroyc: for now those are the versions app insights also supports, so I reused them

arroyc:
ok I got one successful build but for some reason it's not showing here https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2627299 @gpcastro

PR #80, opened 2019-04-19
Cleaning up scripts related to building images

kichalla:


vsts/scripts/dockerCleanup.sh
```diff
@@ -52,6 +52,19 @@ UntagImages "oryxdevmcr.azurecr.io/public/oryx/*:latest"
...
+UntagImages node:6.2.2
+UntagImages node:6.6.0
+UntagImages node:6.9.5
+UntagImages node:6.10.3
+UntagImages node:6.11.5
+UntagImages node:8.0.0
+UntagImages node:8.1.4
+UntagImages node:8.2.1
+UntagImages node:8.8.1
+UntagImages node:8.9.4
```
gpcastro: Why can't we untag node:*?

kichalla: That is because we have the following valid tags where we inherit from where as in the above list we stopped inheriting from the official images:

```
8.11.4-alpine
8.12.0-alpine
9.4.0-alpine
10.1.0-alpine
10.10.0-alpine
10.12.0-alpine
10.14.1-alpine
```

PR #83, opened 2019-04-22
log generated runtime startup script for better troubleshooting

arroyc:
log startup script for runtime images for better troubleshooting


src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -88,8 +88,10 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript() string {
 			return ""
 		}
 	}
-
-	return scriptBuilder.String()
+	var runScript = scriptBuilder.String()
+	logger.LogInformation("Now logging generated startup script.")
```
dorfire: Consolidate these two messages to one, to align with `BuildCommand.cs`:
`logger.LogDebug("Build script content:\n" + scriptContent);`

PR #85, opened 2019-04-22
Enable Python builds to support compress formats zip and tar.gz

kichalla:


build/build-runtimeimages.sh
```diff
@@ -36,23 +36,29 @@ function getTagName()
...
-
-echo
-echo "Generating Dockerfiles for Python runtime images..."
-$REPO_DIR/images/runtime/python/generateDockerfiles.sh
-
-echo
-echo "Generating Dockerfiles for PHP runtime images..."
-$REPO_DIR/images/runtime/php/generate-dockerfiles.sh
-
+runtimeImagesSourceDir="$RUNTIME_IMAGES_SRC_DIR"
```
kichalla: The changes here are to help during development to only build runtime images that a dev asks for rather than try building all the images

src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
```diff
@@ -25,7 +25,7 @@ then
     DESTINATION_DIR=$(pwd -P)
 fi
 
-if [ "$SOURCE_DIR" != "$DESTINATION_DIR" ]
+if [ "$SOURCE_DIR" != "$DESTINATION_DIR" ] && [ "$CLEAN_DESTINATION_DIR" == "true" ]
```
kichalla: @gpcastro What do you think about this? I think we should not have been doing this even for Node and Python because @JennyLawrance would be using an intermediate directory for building and so this condition would have been true and we would be deleting the contents of the output directory there by deleting any persistent data that was in the output folder.

gpcastro: Right. I'd make this a flag though, instead of environment variable, and default to not cleaning it as we dont know what might be in the output directory

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -16,17 +16,24 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Python
     [BuildProperty(
         VirtualEnvironmentNamePropertyKey, "If provided, will create a virtual environment with the given name.")]
     [BuildProperty(
-        ZipVenvDirPropertyKey, "If provided, the virtual environment folder will be zipped to the output folder.")]
+        CompressVirtualEnvPropertyKey,
+        "Indicates how and if virtual environment folder should be compressed into a single file in the output " +
+        "folder. Options are '" + ZipOption + "', and '" + TarGzOption + "'. Default is to not compress. " +
+        "If this option is used, when running the app the virutal environment folder must be extracted from " +
```
gpcastro: typo: virtual

src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
kichalla: Removed this piece of code as per our discussion offline. Users can clean up if they want to using a pre-build script.

PR #86, opened 2019-04-22
Added port parameter to PHP startup script generator

dorfire:


src/startupscriptgenerator/php/scriptgenerator.go
```diff
@@ -26,6 +27,7 @@ func (gen *PhpStartupScriptGenerator) GenerateEntrypointScript() string {
 	scriptBuilder.WriteString("# Enter the source directory to make sure the script runs where the user expects\n")
 	scriptBuilder.WriteString("cd " + gen.SourcePath + "\n")
 	scriptBuilder.WriteString("export APACHE_DOCUMENT_ROOT='" + gen.SourcePath + "'\n")
+	scriptBuilder.WriteString("export APACHE_PORT='" + gen.BindPort + "'\n")
```
kichalla: Take a look at how we do for `Node` startup script generator.

By default our image would have a default port as 8080 and also the default of startup script generator is 8080, so in this case explicitly supplying value to startup is not required as it matches, but anyone could derive from our image and might change the default port to something else, in which case ideally we should honor that setting. So basically the order of precedence is explicitly supplied value over environment value.

src/startupscriptgenerator/common/configuration.go
```diff
@@ -10,12 +10,13 @@ import (
 )
 
 // Order of precedence for value of settings: Explicit command like argument => Environment Variable
-func SetEnvironmentVariableInScript(scriptBuilder *strings.Builder, environmentVariableName string, explicitValue string, defaultValue string){
+func SetEnvironmentVariableInScript(scriptBuilder *strings.Builder, environmentVariableName string, explicitValue string, defaultValue string) {
```
kichalla: Related to my other comment, why are you not calling into this method?

src/startupscriptgenerator/php/scriptgenerator.go
dorfire: so, should I remove the `ENV APACHE_PORT 8080` from the PHP dockerfile?

PR #87, opened 2019-04-22
Expose source and destination directory variables to pre and post build scripts

kichalla:


src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -107,24 +107,28 @@ then
...
-		START_TIME=$SECONDS
-		# Make the contents of the node_modules folder appear in the zip file, not the folder itself
-		cd node_modules
-		{{ CompressNodeModulesCommand }} ../$zippedModulesFileName .
-		ELAPSED_TIME=$(($SECONDS - $START_TIME))
-		echo "Done in $ELAPSED_TIME sec(s)."
+		if [ -f $zippedModulesFileName ]; then
+			echo
+			echo "File '$zippedModulesFileName' already exists under '$SOURCE_DIR'. Deleting it ..."
+			rm -f $zippedModulesFileName
```
gpcastro: I think that the zip tool optimizes for this case when the file is already there, only adding files.

tests/Oryx.BuildImage.Tests/PythonSampleAppsTest.cs
```diff
@@ -859,6 +859,71 @@ public void Build_ExecutesPreAndPostBuildScripts_UsingEnvironmentVariables()
                 result.GetDebugInfo());
         }
 
+        [Fact]
```
arroyc: which version of python does this test target

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
kichalla: I am not doing this change in this PR any more and only exposing the source and destination directories.

tests/Oryx.BuildImage.Tests/PythonSampleAppsTest.cs
kichalla: I think we already discussed about this offline.

PR #92, opened 2019-04-23
Test clean-up: clean ShellScriptBuilder

kichalla:


kichalla:
@arroyc Are these tests needed currently? Are they currently having the similar issue with content not present in log file because of the way NLog flushes the data?

arroyc:
> @arroyc Are these tests needed currently? Are they currently having the similar issue with content not present in log file because of the way NLog flushes the data?

kichalla:
@arroyc Sure, thanks for confirmation

kichalla:
@arroyc I will add the comment in my another PR

PR #93, opened 2019-04-23
Kill SqlServer process from within the sql server container rather than depending on 'docker stop'

kichalla:


tests/Oryx.Integration.Tests/LocalDockerTests/Fixtures/SqlServerDbContainerFixture.cs
```diff
@@ -68,5 +68,19 @@ protected override void InsertSampleData()
...
+
+        protected override void StopContainer()
+        {
+            // We have noticed that 'docker stop' of the sql server container does not work always,
+            // So instead kill the process from within the container itself.
+            var result = _dockerCli.Exec(
+                DbServerContainerName,
+                "/bin/sh",
+                //NOTE: sqlservr is not a typo! It is the name of the process
+                new[] { "-c", "kill -SIGTERM $(pgrep sqlservr)" });
```
kichalla: When testing locally this worked, kicked off a nightly now. let's see

gpcastro: I'm not confident this will actually solve the problem though... since the Docker kill would eventually be translated into a SIGTERM to the process, I'm guessing the sigterm handler on the process is not actually terminating?
I'd do a SIGKILL here, since we have zero interest in the service finishing gracefully, and  it might be faster than sigterm.

kichalla: Sure, didn't know that SIGKILL is more powerful :-)...yes, we don't care if it is shutdown gracefully or not. Will make the change.

kichalla:
> docker kill by default sends sigkill, why are we not leveraging that? why are we specifying --signal?

Good catch. Yes, i have reverted this change now.

PR #94, opened 2019-04-24
Improving support for Node startup files

gpcastro:
This change will make us backwards compatible with the app service startup [script generator](https://github.com/Azure-App-Service/node/blob/master/10.14/startup/generateStartupCommand.js)
It adds support for .json files (other than package.json of course), and [config files](https://pm2.io/doc/en/runtime/guide/ecosystem-file/).

### DIFF
--- a/src/startupscriptgenerator/node/scriptgenerator.go
+++ b/src/startupscriptgenerator/node/scriptgenerator.go
@@ -173,6 +173,41 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 	return scriptBuilder.String()
 }
 
+func getPackageJsonObject(appPath string, userProvidedPath string) (obj *packageJson, filePath string) {
+	logger := common.GetLogger("node.scriptgenerator.GenerateEntrypointScript")
+	defer logger.Shutdown()
+
+	const packageFileName = "package.json"
+	packageJsonPath := ""
+
+	// We prioritize the file the user provided
+	if userProvidedPath != "" {
+		if strings.HasSuffix(userProvidedPath, packageFileName) {
+			logger.LogInformation("Using user-provided path for packageJson: " + userProvidedPath)
+			packageJsonPath = userProvidedPath
+		}
+	} else {
+		logger.LogInformation("Use package.json at the root.")
+		packageJsonPath = packageFileName
+	}
+	if packageJsonPath != "" {
+		packageJsonPath = filepath.Join(appPath, packageJsonPath)
+		if _, err := os.Stat(packageJsonPath); !os.IsNotExist(err) {
+			packageJsonBytes, err := ioutil.ReadFile(packageJsonPath)
+			if err != nil {
+				return nil, ""
+			}
+
+			packageJsonObj := new(packageJson)
+			err = json.Unmarshal(packageJsonBytes, &packageJsonObj)
+			if err == nil {
+				return packageJsonObj, packageJsonPath
+			}
+		}
+	}
+	return nil, ""
+}
+
 // Gets the startup script from package.json if defined. Returns empty string if not found.
 func (gen *NodeStartupScriptGenerator) getPackageJsonStartCommand(packageJsonObj *packageJson, packageJsonPath string) string {
 	if packageJsonObj != nil && packageJsonObj.Scripts != nil && packageJsonObj.Scripts.Start != "" {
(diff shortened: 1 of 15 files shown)


src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -215,6 +241,56 @@ func (gen *NodeStartupScriptGenerator) getPackageJsonMainCommand(packageJsonObj
...
+	}
+
+	if processJsonPath != "" {
+		processJsonFullPath := filepath.Join(gen.SourcePath, processJsonPath)
+		if common.FileExists(processJsonFullPath) {
+			debugCommand := ""
+			if gen.isDebugging() {
+				debugCommand = gen.getNodeWrapperDebugCommand()
+			}
+			return debugCommand + getPm2StartCommand(processJsonPath)
```
kichalla: Do you want to pass `processJsonFullPath` instead?

src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -215,6 +241,56 @@ func (gen *NodeStartupScriptGenerator) getPackageJsonMainCommand(packageJsonObj
...
+		}
+	}
+	return ""
+}
+
+func (gen *NodeStartupScriptGenerator) getConfigYamlCommand(userInputFullPath string) string {
```
kichalla: This method looks different than others where you full path etc....intentional?

tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
```diff
@@ -120,5 +130,157 @@ public void GeneratedScript_CanRunStartupScriptsFromAppRoot()
             // Assert
             RunAsserts(() => Assert.Equal(res.ExitCode, exitCodeSentinel), res.GetDebugInfo());
         }
+
+        [Theory]
+        [MemberData(nameof(TestValueGenerator.GetNodeVersions_SupportPm2), MemberType = typeof(TestValueGenerator))]
+        public async Task RunNodeAppUsingProcessJson(string nodeVersion)
+        {
+
```
kichalla: nit: remove blank line

tests/Oryx.Tests.Common/EndToEndTestHelper.cs
```diff
@@ -117,13 +117,18 @@ public static class EndToEndTestHelper
                output);
 
             // Run
+            await RunAndAssertAppAsync(runtimeImageName, output, volumes, environmentVariables, portMapping, link, runCmd, runArgs, assertAction, dockerCli);
+        }
+
+        public static async Task RunAndAssertAppAsync(string imageName, ITestOutputHelper output, List<DockerVolume> volumes, List<EnvironmentVariable> environmentVariables, string portMapping, string link, string runCmd, string[] runArgs, Func<Task> assertAction, DockerCli dockerCli)
```
kichalla: nit: long lines

src/startupscriptgenerator/node/scriptgenerator.go
kichalla: I think these three methods could be consolidated into one and parameterized..what do you think?

src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -119,6 +119,21 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
...
+			startupCommand = gen.getProcessJsonCommand(userStartupCommandFullPath)
+			commandSource = "ProcessJson"
+		}
+
+		if startupCommand == "" {
+			startupCommand = gen.getConfigJsCommand(userStartupCommandFullPath)
+			commandSource = "ConfigJs"
+		}
+
+		if startupCommand == "" {
```
kichalla: A test in this PR is passing `-userStartupCommand config.yml`...will all these be evaluated or should only the getConfigYaml be executed?...its difficult to follow for me..

src/startupscriptgenerator/node/scriptgenerator.go
gpcastro: No, I wanted to keep the relative path since we're already in the app's directory.

src/startupscriptgenerator/node/scriptgenerator.go
gpcastro: ok, refactored it a bit. Let me know what you think.

src/startupscriptgenerator/node/scriptgenerator.go
gpcastro: only getConfigYaml will be executed, since it is the only one that matches the suffix with the test

PR #100, opened 2019-04-25
Update unzip package in Node runtime images

kichalla:


tests/Oryx.Integration.Tests/LocalDockerTests/NodeEndToEndTests.cs
```diff
@@ -608,7 +608,11 @@ public async Task Node_CreateReactAppSample()
         }
 
         [Theory]
-        [MemberData(nameof(TestValueGenerator.GetNodeVersions), MemberType = typeof(TestValueGenerator))]
+        [InlineData("8.11")]
+        [InlineData("8.12")]
+        [InlineData("10.1")]
```
kichalla: The sample app that we are using seems like cannot be used on all versions. For example, we are getting errors like below for majority of the cases (and not only with this specific `css-loader` package but other packages have version restrictions too)
```
error css-loader@1.0.0: The engine "node" is incompatible with this module. Expected version ">= 6.9.0 <7.0.0 || >= 8.9.0". Got "8.1.4"
error Found incompatible module
```

With unzip update, we were able to run the test successfully on 10.1 and 8.11 which we were unable to earlier.

PR #101, opened 2019-04-26
Upgrading Npm version 6.*.* to 6.9.0

kichalla:


images/build/Dockerfile
```diff
@@ -115,6 +115,18 @@ RUN curl -sL https://git.io/n-install | bash -s -- -ny - \
  && ~/n/bin/n -d 10.15.2 \
  && mv /usr/local/n/versions/node /opt/nodejs \
  && rm -rf /usr/local/n ~/n
+RUN set -e \
+ && for ver in `ls /opt/nodejs`; do \
```
kichalla: So in the build image, we would be looking into _every_ installed Node version and then see if the npm version that comes bundled with it has a major version of `6` and if yes, will upgrade them to `6.9.0`. This restriction is so that we do not make 'major' upgrade which could possibly break apps.

images/runtime/node/Dockerfile.template
```diff
@@ -22,6 +22,13 @@ ENV ORYX_AI_INSTRUMENTATION_KEY=${AI_KEY}
 RUN mkdir -p /node_modules \
  && chmod 777 /node_modules
 
+RUN set -e \
```
kichalla: I did not bother to have this piece of code for older versions of Node runtime images since they are not using `6` version of npm.

tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
```diff
@@ -247,6 +247,42 @@ public void Python3Alias_UsesPythonLatestVersion_ByDefault_WhenNoExplicitVersion
                 result.GetDebugInfo());
         }
 
+
+        [Theory]
+        [InlineData("10.1.0", "5.6.0")]
```
kichalla: This variation to verify that we only upgrade the major version `6`

tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
```diff
@@ -98,6 +98,30 @@ public void NodeVersionMatchesImageName(string nodeTag, string nodeVersion)
                 result.GetDebugInfo());
         }
 
+        [Theory]
+        [InlineData("10.1", "5.6.0")]
```
kichalla: This variation to verify that we only upgrade the major version `6`

kichalla:
cc @JennyLawrance @rramachand21 @gangularamya

images/build/Dockerfile
```diff
@@ -115,6 +115,18 @@ RUN curl -sL https://git.io/n-install | bash -s -- -ny - \
...
+        nodeModulesDir="/opt/nodejs/$ver/lib/node_modules"; \
+        npm_ver=`jq -r .version $nodeModulesDir/npm/package.json`; \
+        if [ ! "$npm_ver" = "${npm_ver#6.}" ]; then \
+            echo "Upgrading node $ver's npm version from $npm_ver to 6.9.0"; \
+            cd $nodeModulesDir; \
+            PATH="/opt/nodejs/$ver/bin:$PATH" \
```
gpcastro: why do you update the PATH here? We'll end up with every node version in the PATH after this loop

images/runtime/node/Dockerfile.template
```diff
@@ -22,6 +22,13 @@ ENV ORYX_AI_INSTRUMENTATION_KEY=${AI_KEY}
...
+ && npm_ver=$(npm --version) \
+ && if [ ! "$npm_ver" = "${npm_ver#6.}" ]; then \
```
gpcastro: what is the purpose of this if?

images/build/Dockerfile
kichalla: The update to the path is local to that one line only(probably the line break is misleading) and it does not affect the final state of the image.

images/runtime/node/Dockerfile.template
gpcastro: couldn't you use `[ ${npm_ver:0:2} = "6." ]` ?

kichalla: Since this change is in the docker template file, it is being used in some Node versions which do not use version `6` of npm, so had to filter

images/runtime/node/Dockerfile.template
```diff
@@ -22,6 +22,13 @@ ENV ORYX_AI_INSTRUMENTATION_KEY=${AI_KEY}
...
+        echo "Upgrading npm version from $npm_ver to 6.9.0"; \
+        npm install -g npm@6.9.0; \
```
gpcastro: what about running this command but passing a version spec that is the latest version of the major version?
e.g. `npm install -g npm@${npm:0:1}`?

images/runtime/node/Dockerfile.template
kichalla: > couldn't you use [ ${npm_ver:0:2} = "6." ] ?

Sure, I will give it a try

images/build/Dockerfile
gpcastro: my bad, I did not pay attention to the line break

images/runtime/node/Dockerfile.template
arroyc: if we use globally I think it gets store in /usr/locl/bin something like that? Will that not cause problem as  our links  are in opt?

tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
arroyc: can you write this comments in the test itself?

images/runtime/node/Dockerfile.template
kichalla: yeah i thought about it but felt it is too broad of a change to do now...what do you think?

kichalla: > if we use globally I think it gets store in /usr/locl/bin something like that? Will that not cause problem as our links are in opt?

This is runtime image and not build image and the global install puts it in where the original npm in installed, so we are good

images/runtime/node/Dockerfile.template
kichalla: Yeah I gave it a try but it does not work and seems like that is because by default `RUN` instruction in dockerfile runs using `/bin/sh` and not bash (when i change the shell to bash though it works, but there are 2 approaches here: one inline 'bash' and other using `SHELL`..both are not great IMO), so I will leave it as it is now.

PR #109, opened 2019-05-01
Update how we probe for ASP.NET Core web applications in a repo

kichalla:


src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
```diff
@@ -56,9 +56,11 @@ public string GetProjectFile(ISourceRepo sourceRepo)
                 if (!sourceRepo.FileExists(projectFile))
                 {
                     _logger.LogWarning($"Could not find the project file '{projectFile}'.");
-                    throw new InvalidOperationException(
+                    throw new InvalidUsageException(
```
kichalla: Changed this type as users can take some action based on this error message.

src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
```diff
@@ -128,47 +133,9 @@ public string GetProjectFile(ISourceRepo sourceRepo)
         // To enable unit testing
         internal static bool IsAspNetCoreWebApplicationProject(XDocument projectFileDoc)
```
kichalla: Changed the probing logic here after input from ASP.NET Core team

src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
```diff
@@ -110,9 +112,12 @@ public string GetProjectFile(ISourceRepo sourceRepo)
 
                 if (webAppProjects.Count > 1)
                 {
+                    var projects = string.Join(", ", webAppProjects);
                     throw new InvalidUsageException(
                         "Ambiguity in selecting an ASP.NET Core web application to build. " +
-                        "Found multiple applications: " + string.Join(", ", webAppProjects));
+                        $"Found multiple applications: {projects}. Use the environment variable " +
```
gpcastro: put the project names under `'` in the message

src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
```diff
@@ -128,47 +133,9 @@ public string GetProjectFile(ISourceRepo sourceRepo)
...
-                })
-                .FirstOrDefault();
-
-            if (aspNetCorePackageReference != null)
-            {
-                return true;
-            }
-
-            return false;
+            var webSdkProjectElement = projectFileDoc.XPathSelectElement(
```
gpcastro: question: won't this change potentially make apps that run fine today break? If yes, we will need a plan to avoid breaking them

arroyc:
is this for a specific of .net (e.g. 2.1) or overall change? I would like to see a theory test for all the .net versions

src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
kichalla: Since all asp.net core versions since 1.0 use this attribute value "microsoft.net.sdk.web", we wouldn't be breaking existing apps.

kichalla:
> is this for a specific of .net (e.g. 2.1) or overall change? I would like to see a theory test for all the .net versions

This is an overall change for all versions. I am not sure if it would add much value as we only look for the `Sdk` attribute on the `Project` element and this has been the same for all the versions

PR #111, opened 2019-05-02
Fix database tests

kichalla:


tests/Oryx.Integration.Tests/LocalDockerTests/Fixtures/DbContainerFixtureBase.cs
```diff
@@ -31,11 +30,32 @@ public abstract class DbContainerFixtureBase : IDisposable
...
+                if (isDbServerUp)
+                {
+                    InsertSampleData();
+                    insertedSampleData = true;
+                }
+            }
+            catch (Exception ex)
             {
-                throw new Exception("Database not ready in time");
+                // Since Dispose method does not get called in case of exceptions, catch the exception here, log it
```
kichalla: This was one of the issues that was happening earlier where dispose wasn't get called (but yeah that does not explain why stopping the containers fails in our uber clean-up script)

tests/Oryx.Integration.Tests/LocalDockerTests/Fixtures/MySqlDbContainerFixture.cs
```diff
@@ -37,19 +37,26 @@ protected override DockerRunCommandResult RunDbServerContainer()
...
+                    // In case of any exception, we consider this retry as a failure and will retry again.
+                    return false;
+                }
             });
         }
 
         protected override void InsertSampleData()
         {
-            const string sqlFile = "/tmp/setup.sql";
+            const string sqlFile = "/tmp/mysql_setup.sql";
```
kichalla: Each database type has this method of its own and all were using the same file location on the agent, so a potential race condition where a sql script intended for one database could have been used for another (since database types run in parallel for different database server types)

tests/Oryx.Integration.Tests/LocalDockerTests/MySqlIntegrationTests.cs
```diff
@@ -18,13 +18,13 @@ public MySqlIntegrationTests(ITestOutputHelper output, Fixtures.MySqlDbContainer
         {
         }
 
-        [Fact(Skip = "Bug 847845")]
+        [Fact]
```
kichalla: I will not check this PR in anytime soon. I am running these tests continuously on linux vm and if they are stable, will do

tests/Oryx.Integration.Tests/LocalDockerTests/Fixtures/DbContainerFixtureBase.cs
```diff
@@ -30,12 +29,44 @@ public abstract class DbContainerFixtureBase : IDisposable
...
+            Console.WriteLine("Done.");
+
+            bool insertedSampleData = false;
+            try
+            {
+                Console.WriteLine("Waiting for database server to be up...");
+                var isDbServerUp = WaitUntilDbServerIsUp();
+                if (isDbServerUp)
+                {
+                    Console.WriteLine("Done.");
```
gpcastro: nit: instead of `Done`, use something more specific like `Database is up.`.

tests/Oryx.Integration.Tests/LocalDockerTests/Fixtures/DbContainerFixtureBase.cs
```diff
@@ -30,12 +29,44 @@ public abstract class DbContainerFixtureBase : IDisposable
...
             {
-                throw new Exception("Database not ready in time");
+                // Since Dispose method does not get called in case of exceptions, catch the exception here, log it
+                // and stop the container.
+                Console.WriteLine("Exception occurred while setting up the database container: " + ex.ToString());
+            }
+
+            if (!insertedSampleData)
+            {
+                Console.WriteLine("Stopping the container...");
```
gpcastro: why don't you add this to the exception handler?

kichalla: Because a database might not be up in which case we would still want to stop the container just to be safe

tests/Oryx.Integration.Tests/LocalDockerTests/MySqlIntegrationTests.cs
kichalla: I have run all the database tests around 30 times on my VM and they are running well without an issue, so I am merging this in.

PR #116, opened 2019-05-06
Updated the marker text that is used to figure out if SQL Server container is up and ready

kichalla:


tests/Oryx.Integration.Tests/LocalDockerTests/Fixtures/SqlServerDbContainerFixture.cs
```diff
@@ -53,7 +53,7 @@ protected override bool WaitUntilDbServerIsUp()
             {
                 try
                 {
-                    var lookUpText = "SQL Server is now ready for client connections";
+                    var lookUpText = "The default language (LCID 0) has been set for engine and full-text services.";
```
kichalla: This is the final message that is printed when the SQL Server is up and ready. Even though the earlier message says 'ready for client connections' I have received errors some times when setting up data.

Also regarding setting up data, one thing we could do is to bake the data into the image itself. This improves the perf a bit, but still does not avoid the problem of making requests to the container before it is ready.

kichalla: FYI:
image

PR #118, opened 2019-05-06
Fix Shapely test

gpcastro:
Build command was missing language version flag.

### DIFF
--- a/tests/Oryx.Integration.Tests/LocalDockerTests/PythonEndToEndTests.cs
+++ b/tests/Oryx.Integration.Tests/LocalDockerTests/PythonEndToEndTests.cs
@@ -566,14 +566,13 @@ namespace Microsoft.Oryx.Integration.Tests.LocalDockerTests
         {
             // Arrange
             var appName = "shapely-flask-app";
-            var virtualEnvName = "antenv";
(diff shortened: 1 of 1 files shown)


tests/Oryx.Integration.Tests/LocalDockerTests/PythonEndToEndTests.cs
```diff
@@ -583,7 +583,7 @@ public async Task CanBuildAndRun_ShapelyFlaskApp_UsingVirtualEnv(string pythonVe
                 _output,
                 volume,
                 "oryx",
-                new[] { "build", appDir },
+                new[] { "build", appDir, "-l", "python", "--language-version", pythonVersion },
```
kichalla: I think this test should fail as this test seems to be passing `virtualEnvName` to startup script path for runtime. Probably it was failing even before your changes?

gpcastro: It's fine to pass the virtual env name to the startup script. It won't actually use it though, as it will take it from the manifest file.
It was passing before, becase using target folder is more forgiving to version differences. Since virtual envs actually contain the python binaries, it created problems when build and runtime are different

PR #123, opened 2019-05-07
Trying to take nightly builds to the tip of parallelism 

arroyc:


tests/Oryx.Integration.Tests/DotNetCoreEndToEndTests.cs
```diff
@@ -13,6 +13,7 @@
 
 namespace Microsoft.Oryx.Integration.Tests
 {
+    [Trait("Category", "DOTNETCORE")]
```
dorfire: Let's keep the case in line with the class names - like `DotNetCore` instead of `DOTNETCORE`

tests/Oryx.Integration.Tests/LocalDockerTests/NotifyEndOfIntegrationTests.cs
```diff
@@ -0,0 +1,27 @@
...
+
+using System.IO;
+using Microsoft.Oryx.Tests.Common;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Oryx.Integration.Tests.LocalDockerTests
+{
+    [Trait("Category", "END")]
+    public class NotifyEndOfIntegrationTests : PlatformEndToEndTestsBase
```
dorfire: What's the purpose of this? Seems a bit "hacky" - maybe add some clarifying documentation?

vsts/pipelines/nightly.yml
```diff
@@ -61,7 +61,74 @@ jobs:
...
       echo "##vso[task.setvariable variable=TestRuntimeImages;]false"
-      echo "##vso[task.setvariable variable=TestIntegrationCaseFilter;]Category!=DB"
+      echo "##vso[task.setvariable variable=TestIntegrationCaseFilter;]Category=PYTHON"
+      echo "##vso[task.setvariable variable=TestIntegration;]true"
+      echo "##vso[task.setvariable variable=PushBuildImages;]false"
+      echo "##vso[task.setvariable variable=PushRuntimeImages;]false"
+      echo "##vso[task.setvariable variable=PushToDockerHub;]false"
+      echo "##vso[task.setvariable variable=EmbedBuildContextInImages;]false"
+    displayName: 'Set variables'
+  - template: _buildParallel.yml
```
dorfire: Instead of running our entire build template, maybe we can just invoke the `test-integration.sh` script?

vsts/pipelines/nightly.yml
```diff
@@ -61,7 +61,74 @@ jobs:
...
+
+- job: Job_NodeIntegrationTests
+  displayName: Test NodeJs Integration Tests
+  dependsOn: 
+    - Job_BuildImage
+    - Job_RuntimeImages
+  pool:
+    name: OryxLinux
+  timeoutInMinutes: 150
```
dorfire: Seems like a lot - the others didn't get such a high timeout

vsts/pipelines/nightly.yml
```diff
@@ -71,7 +138,7 @@ jobs:
   - template: _buildParallel.yml
 
 - job: Job_DbIntegrationTests
-  displayName: Test Database Integration
+  displayName: Test Database Integration Tests
```
dorfire: Test the tests? 😛

vsts/scripts/pullandretag.sh
```diff
@@ -6,22 +6,55 @@
 
 set -o pipefail
 
+declare integrationtestfilter="oryxdevmcr.azurecr.io/public/oryx/"
```
dorfire: This can change, it'll probably be more robust to use the value from `__variables.sh`

vsts/pipelines/nightly.yml
arroyc: yeah node integration tests sometimes taking more than 45 minutes

tests/Oryx.Integration.Tests/DotNetCoreEndToEndTests.cs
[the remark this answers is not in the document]
arroyc: im converting to lower case anyways, but I guess we can stick to one casing

vsts/scripts/pullandretag.sh
arroyc: IMO we should keep vsts build scripts and source code independent. Though we can use some variables by sourcing variables.sh, its better not to do it here. If the name changes, we need to change it everywhere including yaml templates.

PR #124, opened 2019-05-07
Enable supplying commands directly to the pre and post build script envrionment variables

kichalla:


src/BuildScriptGenerator/EnvironmentSettings.cs
```diff
@@ -7,8 +7,13 @@ namespace Microsoft.Oryx.BuildScriptGenerator
 {
     public class EnvironmentSettings
     {
+        // Note: These two keys exist so that we do not break existing users who might still be using them
```
dorfire: To make it clearer:
"these two" => "the following two"
keys => variables

src/BuildScriptGenerator/EnvironmentSettingsKeys.cs
```diff
@@ -7,10 +7,20 @@ namespace Microsoft.Oryx.BuildScriptGenerator
 {
     public static class EnvironmentSettingsKeys
     {
+        // Note: These two keys exist so that we do not break existing users who might still be using them
```
dorfire: Similarly to the previous comment:
"These two" => "the following two"

src/BuildScriptGenerator/PreAndPostCommandScriptHelper.cs
```diff
@@ -0,0 +1,74 @@
...
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using System.IO;
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    internal static class PreAndPostBuildCommandHelper
+    {
+        public static (string preBuildCommand, string postBuildCommand) GetPreAndPostBuildCommands(
```
dorfire: suggestion: move this to be a method of `EnvironmentSettings`?
Or even better, split the code of this method to be getters of `EnvironmentSettings.PreBuildCommand` & `EnvironmentSettings.PostBuildCommand`

tests/Oryx.BuildImage.Tests/DotNetCoreSampleAppsTest.cs
```diff
@@ -472,6 +472,49 @@ public void Build_CopiesContentCreatedByPreAndPostBuildScript_ToImplicitOutputDi
                 result.GetDebugInfo());
         }
 
+        [Fact]
+        public void Build_Executes_InlinePreAndPostBuildCommands()
```
dorfire: I don't see a lot of value from having the same test run on two different platforms - the code is shared among all platforms anyway, isn't it?

tests/Oryx.BuildImage.Tests/DotNetCoreSampleAppsTest.cs
```diff
@@ -472,6 +472,49 @@ public void Build_CopiesContentCreatedByPreAndPostBuildScript_ToImplicitOutputDi
...
+                    script
+                });
+
+            // Assert
+            RunAsserts(
+                () =>
+                {
+                    Assert.True(result.IsSuccess);
+                    Assert.Contains("from pre-build command", result.StdOut);
+                    Assert.Contains("from post-build command", result.StdOut);
```
dorfire: 👍

tests/Oryx.BuildImage.Tests/DotNetCoreSampleAppsTest.cs
kichalla: The thing is DotNetCore is no longer using the common base bash script as other platforms. This is because the difference in how .net core deployment works (yeah I don't like it, but until we find a solution which works for all platforms, we need to have a test to verify there too)

src/BuildScriptGenerator/PreAndPostCommandScriptHelper.cs
kichalla: Yeah, I considered it initially but it felt like having stuff like quotes around the command or not(depending on whether the content is a file or not) seem to be a wrong place/layer for EnvironmentSettings or DefaultEnvironmentSettingsProvider as it is related to how shell script works. Like these apis only give the input (yeah they sanitize a bit, like trimming space and quotes).

tests/Oryx.Tests.Common/EndToEndTestHelper.cs
```diff
@@ -204,11 +216,34 @@ public static class EndToEndTestHelper
...
-                            output.WriteLine(runResult.GetDebugInfo());
+                            string debugInfo = string.Empty;
+
+                            // ToString() on StringBuilder is throwing an exception probably because of a 
+                            // multithreading issue where the container is still writing data into it and we are trying
+                            // to retrieve the content out of it.
+                            try
+                            {
+                                // TO i
+                                // System.ArgumentOutOfRangeException : Index was out of range. Must be non-negative
```
kichalla: cc @arroyc @gpcastro I ran into the following issue again with my PR
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2656104

On searching the error message, seems like it could be a multi-threading issue with stringbuilder

src/BuildScriptGenerator/PreAndPostCommandScriptHelper.cs
dorfire: guiding question - will anyone ever use `EnvironmentSettings.PreBuildCommand`/`EnvironmentSettings.PostBuildCommand` **without** `GetPreAndPostBuildCommands`?

tests/Oryx.BuildImage.Tests/DotNetCoreSampleAppsTest.cs
dorfire: Understood 👍

src/BuildScriptGenerator/PreAndPostCommandScriptHelper.cs
kichalla: Yes, they could...are you implying they wouldn't be getting the logic that is present in `GetPreAndPostBuildCommands`? yes, they would not...but building that logic into those properties seems like a wrong layer as by doing `EnvironmentSettings.PreBuildCommand` you are just trying to get the value supplied by the user (enclosing with quotes or not is specific to the script IMO)

dorfire: ok, got it 👍

PR #126, opened 2019-05-08
add node-sass sample app to master

arroyc:
earlier PR didn't add the sample app


gpcastro:
Add a file `SOURCE.md` to the node-saas sample app specifying where you took the app from.

PR #127, opened 2019-05-09
add more parallelism to nightly builds

arroyc:
I recreated the PR because I was getting numerous conflicts while rebasing on master, so I had to  squash all my commits in a single commit and then it was easier for me rebase .. thanks for the idea @kichalla .. But in that process I had to sacrifice my remote branch


vsts/pipelines/nightly.yml
```diff
@@ -48,8 +48,8 @@ jobs:
     displayName: 'Set variables'
   - template: _buildParallel.yml
 
-- job: Job_IntegrationTests
-  displayName: Test Integration
+- job: Job_PythonIntegrationTests
```
kichalla: Basically you can have a template for a `job` for integration tests itself (related to what you mentioned in scrum meeting today)...job templates can be passed the parameters too

arroyc: yes, similarly I want have template for build image and test image etc etc

tests/Oryx.Tests.Common/DockerCli.cs
```diff
@@ -18,6 +18,8 @@ public class DockerCli
 
         private readonly TimeSpan _waitTimeForExit;
         private readonly IEnumerable<EnvironmentVariable> _globalEnvVars;
+        private readonly Object Lock1 = new Object();
```
gpcastro: let's use more meaningful names, e.g. "outputLogLock"?

tests/Oryx.Tests.Common/DockerCli.cs
```diff
@@ -134,11 +136,17 @@ public DockerCli(TimeSpan waitTimeForExit, IEnumerable<EnvironmentVariable> glob
                     // are called for each line that is written to the output.
                     standardOutputHandler: (sender, args) =>
                     {
-                        outputBuilder.AppendLine(args.Data);
+                        lock (Lock1)
```
gpcastro: as we discussed, let's put the locks in the same scope as the object it is protecting

vsts/scripts/pullandretag.sh
```diff
@@ -5,23 +5,61 @@
 # --------------------------------------------------------------------------------------------
 
 set -o pipefail
+# $1 > buildimage-acr.txt
+# $2 > runtime-images-acr.txt
+declare integrationtestfilter="oryxdevmcr.azurecr.io/public/oryx"
```
gpcastro: I would use a different name here, e.g. `imageFilter`?

tests/Oryx.Tests.Common/DockerCli.cs
arroyc: reverted lock

tests/Oryx.Tests.Common/DockerCli.cs
arroyc: reverted lock

PR #128, opened 2019-05-09
Updated .NET Core related build and runtime to consider the name of runtimeconfig.json file for finding the startup file name

kichalla:


tests/SampleApps/DotNetCore/NetCoreApp21.WebApp/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace NetCoreApp21.WebApp
```
kichalla: An app with `.` in its name..this will generate a startup file with `.` in it...so its `NetCoreApp21.WebApp.dll`

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -87,200 +76,3 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
-	return startupFileName
-}
-
-func (gen *DotnetCoreStartupScriptGenerator) getProjectDetails() projectDetails {
-	logger := common.GetLogger("dotnetcore.scriptgenerator.getProjectDetails")
-	defer logger.Shutdown()
-
-	projDetails := projectDetails{}
-
-	projectEnv := os.Getenv(ProjectEnvironmentVariableName)
```
gpcastro: I think this is still important, for the case where the user has multiple web apps in the same repo (e.g. multiple micro services), but only wants to run one of them in this instance.

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -24,59 +20,52 @@ type DotnetCoreStartupScriptGenerator struct {
...
 	scriptBuilder.WriteString("readonly appPath=\"" + gen.RunFromPath + "\"\n")
 	scriptBuilder.WriteString("userStartUpCommand=\"" + gen.UserStartupCommand + "\"\n")
 	scriptBuilder.WriteString("startUpCommand=\"\"\n")
-	scriptBuilder.WriteString("readonly startupDllFileName=\"" + startupDllFileName + "\"\n")
+	scriptBuilder.WriteString("readonly defaultAppFileDir=\"" + defaultAppFileDir + "\"\n")
 	scriptBuilder.WriteString("readonly defaultAppFilePath=\"" + gen.DefaultAppFilePath + "\"\n")
+	scriptBuilder.WriteString("echo Finding the startup file name...\n")
+	scriptBuilder.WriteString("cd \"$appPath\"\n")
+	scriptBuilder.WriteString("for file in *; do \n")
+	scriptBuilder.WriteString("  if [ -f \"$file\" ]; then \n")
```
gpcastro: shouldn't we first fallback to the previous way, looking for the .csproj file, before only having this? I'm thinking that if we find any issue with this approach in prod, we will have no alternatives

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -24,59 +20,52 @@ type DotnetCoreStartupScriptGenerator struct {
...
+	scriptBuilder.WriteString("        startupDllFileName=\"$startupDllFileNamePrefix.dll\"\n")
+	scriptBuilder.WriteString("        echo \"Found the startup file '$startupDllFileName'\"\n")
+	scriptBuilder.WriteString("        break\n")
+	scriptBuilder.WriteString("      ;;\n")
+	scriptBuilder.WriteString("    esac\n")
+	scriptBuilder.WriteString("  fi\n")
+	scriptBuilder.WriteString("done\n")
 	scriptBuilder.WriteString("if [ ! -z \"$userStartUpCommand\" ]; then\n")
 	scriptBuilder.WriteString("  cd \"$appPath\"\n")
 	scriptBuilder.WriteString("  startUpCommand=\"$userStartUpCommand\"\n")
```
gpcastro: for the other languages, we typically check for the case where the passed in command is not really a command, but the path to the main file (e.g. `server.js`, `app.py`, etc) - in this case the dll file. If the file exists and ends with `.dll` we could just prepend dotnet to it.

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
kichalla: We would be working with an already published app (where the decision of a particular web app among multiple apps was already made...in the case of zip deploy, where the decision is made by an end user or in the case Oryx build was involved, where build expects users to provide this environment variable), so runtime container doesn't need to handle it.

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
kichalla: I have tested this with all versions of SDK and this approach is totally fine.

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
kichalla: I feel it is a bit different for dotnet case though...here we would have one designated (via runtimeconfig.json) dll as startup and where as in case of other languages, it is a source repo with bunch of files and there is no standard way one can know which is the startup file and providing that support for just the file probably makes sense, but I think we need not worry about that case here IMO. We could revisit this later based on feedback.

PR #129, opened 2019-05-10
Fix EndToEndTestHelper to read input and error streams only after process exits

kichalla:


tests/Oryx.Tests.Common/EndToEndTestHelper.cs
```diff
@@ -255,6 +221,12 @@ public static class EndToEndTestHelper
                 {
                     // Stop the container so that shared resources (like ports) are disposed.
                     dockerCli.StopContainer(runResult.ContainerName);
+
+                    // Access the output and error streams after the process has exited
```
kichalla: This solution is different from what I mentioned to you yesterday because even in the scenario where the caller supplies the builders for stdout or stderr streams, the caller still needs to read the builders only after the process no longer writes to them to prevent the exception that we are seeing, so instead of changing the method contract, I just changed this method to make sure to read the output at a point which is safe.

tests/Oryx.Tests.Common/EndToEndTestHelper.cs
```diff
@@ -187,63 +188,28 @@ public static class EndToEndTestHelper
...
+                Assert.False(runResult.Process.HasExited);
 
                 for (var i = 0; i < MaxRetryCount; i++)
                 {
                     await Task.Delay(TimeSpan.FromSeconds(DelayBetweenRetriesInSeconds));
 
                     try
                     {
-                        // Make sure the process is still alive and fail fast if not alive.
-                        await RunAssertsAsync(
```
gpcastro: do we still need this method? If yes, doesn't it have the same problem?

kichalla: We use this method for the scenario when doing a 'build' where we kick off a process and wait for it to exit and write out the output when asserts fail. Since the process already exits, we do not have an issue here.

PR #130, opened 2019-05-10
Fix flaky runtime images unit test related to apinsight injection

arroyc:
increasing timeout for integration tests and changing sampleapp in runtime images unit test  and integration tests for appinsights to avoid concurrency issue.
Total time saved ~ 6.5 minutes
~ 4 minutes by getting rid of extra runtime images test
~ 2.5 minutes by changing sampleapp (lighter sampleApp compared to webfrontend)

### DIFF
--- a/tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
+++ b/tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
@@ -353,7 +353,7 @@ namespace Microsoft.Oryx.RuntimeImage.Tests
             // Arrange
             var imageName = string.Concat("oryxdevms/node-", nodeVersion);
             var hostSamplesDir = Path.Combine(Directory.GetCurrentDirectory(), "SampleApps");
-            var volume = DockerVolume.Create(Path.Combine(hostSamplesDir, "nodejs", "webfrontend"));
+            var volume = DockerVolume.Create(Path.Combine(hostSamplesDir, "nodejs", "linxnodeexpress"));
             var appDir = volume.ContainerDir;
             var manifestFileContent = "injectedAppInsight=\"True\"";
             var aiNodesdkLoaderContent = @"try {
@@ -380,8 +380,8 @@ namespace Microsoft.Oryx.RuntimeImage.Tests
                 .AddDirectoryExistsCheck($"{appDir}/node_modules/applicationinsights")
                 .AddCommand("./run.sh")
                 .ToString();
-           
-            await EndToEndTestHelper.RunAndAssertAppAsync(
+
+           await EndToEndTestHelper.RunAndAssertAppAsync(
                 imageName: $"oryxdevms/node-{nodeVersion}",
                 output: _output,
                 volumes: new List<DockerVolume> { volume },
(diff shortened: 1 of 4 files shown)


tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
```diff
@@ -405,7 +406,7 @@ public async Task GeneratesScript_CanRun_AppInsightsModule_NotFound(string nodeV
             // Arrange
             var imageName = string.Concat("oryxdevms/node-", nodeVersion);
```
kichalla: As we discussed, could you add a comment or change the name of the test to make it clear about what we are testing here?

tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
```diff
@@ -393,9 +393,10 @@ public async Task GeneratesScript_CanRun_AppInsightsModule_Found(string nodeVers
                 assertAction: async () =>
                 {
                     var data = await _httpClient.GetStringAsync($"http://localhost:{hostPort}/");
-                    Assert.Contains("Say It Again", data);
+                    Assert.Contains("Hello World from express!", data);
```
kichalla: how do we know for sure that appinsights module loaded correctly? Should we try accessing any appinsights api from within the app and try doing something with it...if all is well we should not see any errors. Thoughts?

arroyc: so we have an env var NODE_OPTIONS, where we add the appinsightsloader.js. Content of this loader file is basically setup appinsights by calling "appInsights.setup().start()". So in a way we are doing this.

vsts/scripts/pullandretag.sh
```diff
@@ -4,7 +4,7 @@
 # Licensed under the MIT license.
 # --------------------------------------------------------------------------------------------
 
-set -o pipefail
+set -euxo pipefail
```
kichalla: good catch

tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
arroyc: sure

vsts/pipelines/nightly.yml
```diff
@@ -55,6 +55,7 @@ jobs:
     - Job_RuntimeImages
   pool:
     name: OryxLinux
+  timeoutInMinutes: 150
```
arroyc: added this to provide more time to the node-sass integration tests (+37 minutes)

tests/Oryx.RuntimeImage.Tests/NodeImagesTest.cs
```diff
@@ -353,7 +353,7 @@ public async Task GeneratesScript_CanRun_AppInsightsModule_Found(string nodeVers
             // Arrange
             var imageName = string.Concat("oryxdevms/node-", nodeVersion);
             var hostSamplesDir = Path.Combine(Directory.GetCurrentDirectory(), "SampleApps");
-            var volume = DockerVolume.Create(Path.Combine(hostSamplesDir, "nodejs", "webfrontend"));
+            var volume = DockerVolume.Create(Path.Combine(hostSamplesDir, "nodejs", "linxnodeexpress"));
```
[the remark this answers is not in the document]
arroyc: done

PR #131, opened 2019-05-11
Tests: Get the port for running sample apps dynamically

kichalla:


tests/Oryx.Tests.Common/PortHelper.cs
```diff
@@ -0,0 +1,23 @@
...
+// --------------------------------------------------------------------------------------------
+
+using System.Net;
+using System.Net.Sockets;
+
+namespace Oryx.Tests.Common
+{
+    public static class PortHelper
+    {
+        public static int GetNextAvailablePort()
```
kichalla: @arroyc This should help you in your quest for more parallel tests 😄 ...I had tested this in my Windows and Linux machines and it worked fine.

arroyc: @kichalla  … yepp experiment is on https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2669129  😄

tests/Oryx.Tests.Common/PortHelper.cs
```diff
@@ -0,0 +1,23 @@
...
+        {
+            using (var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp))
+            {
+                // Specifying '0' for port here to get any available port
+                socket.Bind(new IPEndPoint(IPAddress.Loopback, port: 0));
+                return ((IPEndPoint)socket.LocalEndPoint).Port;
+            }
```
kichalla: There seems to be a possibility of a race condition where more than one test can get the same port since we are disposing the socket, but I have read somewhere that it takes time for a given port to be re-used by the OS since there is a range of free ports available for it to give. However, we could try limiting the scope of this issue in our case by getting a dynamic port as close as to the place where we actually use it. i.e in some tests we call this method to get port and _then_ do the build of the app and _then_ run. Instead we should move this part of getting port to DockerCli method which runs the container.

arroyc: I would prefer to have a range of dynamic ports dedicated to each category (php, python, node etc.), I'm experimenting that on my dev branch currently. Also I agree that it will be wise to move this call to dockercli. Check my snippet in the teams chat

kichalla: The one you shared does not solve the issue of race condition though. You limited the tests within a specific port range but there is no guarantee that tests running in parallel would get unique ports due to race condition.

kichalla: I found that we have a better way of doing this by depending on docker itself to assign a dynamic host port. This way we wouldn't be having a race condition.

For example: (Note that I only give the port within the container to the `docker run` command and host port is dynamically assigned...we just need to query again with `docker port` to get the dynamically created port.)

```
C:\> docker run -d -p 6379 redis:alpine
4912f472b8176fdf0cdc5424c72165d2f7150803bb6487a0e2ece382895c881f

C:\>  docker ps --format "table {{.ID}}\t{{.Image}}\t{{.Ports}}"
CONTAINER ID        IMAGE               PORTS
4912f472b817        redis:alpine        0.0.0.0:32774->6379/tcp
7bd3d40adceb        redis:alpine        0.0.0.0:32773->6379/tcp
8a84ed6703ef        redis:alpine        0.0.0.0:32772->6379/tcp

C:\> docker port 4912f472b817
6379/tcp -> 0.0.0.0:32774
```

PR #132, opened 2019-05-13
Fixed npm versions symlinks

kichalla:


images/build/Dockerfile
```diff
@@ -199,16 +199,14 @@ RUN set -ex \
  && ln -s 3.9.5 /opt/npm/3.9 \
  && ln -s 3.10.10 /opt/npm/3.10 \
  && ln -s 3.10 /opt/npm/3 \
- && ln -s 4.8.0 /opt/npm/4.8 \
```
kichalla: I checked the node runtime versions 4.4, 4.5 and 4.8 and they all do not use this npm version...seems like this was added by mistake when we added node runtime for 4.8

gpcastro: but this is a npm version, not node, right? So we don't have npm version 4 on our image?

kichalla: yes, exactly...we have node version but not this npm version

PR #138, opened 2019-05-15
.NET Core startup script: Run app executable if present, else fallback to app's dll

kichalla:
This is to support the following scenarios:
1. End users can do a zip deploy of a self-contained app into AppService and our runtime container should use the executable,
   so instead of doing 'dotnet myappname.dll', we should run './myappname'. This is because the whole point of self-contained app
   is to use the packaged runtime which is present in the output.
2. Starting 3.0 .NET Core, a regular 'dotnet publish' creates both the executable and dll.


src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -40,36 +40,61 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
-	scriptBuilder.WriteString("      ;;\n")
-	scriptBuilder.WriteString("    esac\n")
-	scriptBuilder.WriteString("  fi\n")
-	scriptBuilder.WriteString("done\n")
 	scriptBuilder.WriteString("if [ ! -z \"$userStartUpCommand\" ]; then\n")
-	scriptBuilder.WriteString("  cd \"$appPath\"\n")
-	scriptBuilder.WriteString("  startUpCommand=\"$userStartUpCommand\"\n")
-	scriptBuilder.WriteString("  if [ -f \"$userStartUpCommand\" ]; then\n")
-	scriptBuilder.WriteString("    chmod 755 \"$userStartUpCommand\"\n")
+	scriptBuilder.WriteString("    len=${#userStartUpCommand[@]}\n")
```
kichalla: This new block keeps the logic that AppService currently uses

tests/SampleApps/DotNetCore/NetCoreApp30.WebApp/Program.cs
```diff
@@ -7,7 +7,7 @@
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 
-namespace NetCoreApp30WebApp
+namespace NetCoreApp30.WebApp
```
kichalla: Just to verify if having a `.` in the name breaks the tests

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -40,36 +40,61 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
+	scriptBuilder.WriteString("        *.runtimeconfig.json)\n")
+	scriptBuilder.WriteString("          startupDllFileNamePrefix=${file%%.runtimeconfig.json}\n")
+	scriptBuilder.WriteString("          startupExecutableFileName=\"$startupDllFileNamePrefix\"\n")
+	scriptBuilder.WriteString("          startupDllFileName=\"$startupDllFileNamePrefix.dll\"\n")
+	scriptBuilder.WriteString("          break\n")
+	scriptBuilder.WriteString("        ;;\n")
+	scriptBuilder.WriteString("      esac\n")
+	scriptBuilder.WriteString("    fi\n")
+	scriptBuilder.WriteString("  done\n\n")
+	scriptBuilder.WriteString("  if [ -f \"$startupExecutableFileName\" ]; then\n")
```
gpcastro: It just occurred to me that if the app was built for, say, Mac, this won't work. But the existing code will, since it will run the app using `dotnet app.dll`. I think we should leave this for the next iteration so we have time to think about this a bit more.

kichalla: Are you referring to self-contained app? IMO that would be a user error as the user is clearly deploying the app to Linux containers. Like one has to explicitly use osx-x64 or linux-x64 for example. Thoughts?

gpcastro: It's right that the app was built as self contained for one platform and deployed to another, but an app built this way currently _runs_ (we we'll see the `.dll`file  and run it with `dotnet`), but without any user action after this update it will stop working. 

So from the user perspective, they had something working that now doesn't and they will be caught by surprise. We could check the file type and execution permission, for example.

kichalla: I tried the following scenario now - publish the app for Mac and do a `dotnet foo.dll` on linux...
and this failed to execute...so we need not worry about this scenario as it would never have worked.

FYI:
**Publish for Mac but run on Linux**
image

**Publish for Linux and run on Linux**
image

PR #141, opened 2019-05-15
Add --no-cache for building base images

gpcastro:
This is causing issues since we were not reinstalling the latest version of tools that go in the image.

### DIFF
--- a/build/build-runtimeimages-bases.sh
+++ b/build/build-runtimeimages-bases.sh
@@ -85,7 +85,7 @@ for dockerFile in $dockerFiles; do
     cd $REPO_DIR
 
     echo
-    docker build --no-cache -f $dockerFile -t $localImageTagName $labels .
+    docker build -f $dockerFile -t $localImageTagName --build-arg CACHEBUST=$(date +%s) $labels . 
 
     # Retag build image with DockerHub & ACR tags
     if [ -n "$BUILD_NUMBER" ]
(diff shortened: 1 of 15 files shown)


build/build-runtimeimages-bases.sh
```diff
@@ -85,7 +85,7 @@ for dockerFile in $dockerFiles; do
     cd $REPO_DIR
 
     echo
-    docker build -f $dockerFile -t $localImageTagName $labels .
+    docker build --no-cache -f $dockerFile -t $localImageTagName $labels .
```
kichalla: As we discussed in our meeting, should we have this behind a flag so that we do it only when required?

arroyc: I will like that to be adhoc

gpcastro: I've changed the approach now, so only the steps we want to always run won't be cached.

images/runtime/node/4.4/Dockerfile.base
```diff
@@ -33,6 +33,10 @@ RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-
   && tar -xJf "node-v$NODE_VERSION-linux-x64.tar.xz" -C /usr/local --strip-components=1 \
   && rm "node-v$NODE_VERSION-linux-x64.tar.xz" SHASUMS256.txt.asc SHASUMS256.txt
 
+# This is a way to avoid using caches for the next stages, since we want the remaining steps
```
kichalla: :clap:

PR #146, opened 2019-05-16
.NET Core startup script: Check if the startup executable file is meant for Linux

kichalla:


src/startupscriptgenerator/dotnetcore/main.go
```diff
@@ -63,7 +63,7 @@ func main() {
 	}
 
 	scriptBuilder := strings.Builder{}
-	scriptBuilder.WriteString("#!/bin/sh\n")
+	scriptBuilder.WriteString("#!/bin/bash\n")
```
kichalla: Since all our images are based on `stretch`, I changed this

tests/SampleApps/DotNetCore/TodoAppFromMac/Readme.md
```diff
@@ -0,0 +1,3 @@
+This is a regularly(not self-contained) published ASP.NET Core 3.0 app on a Mac machine.
```
kichalla: This is a published app from a MacOS that I am checking in for test verification

images/runtime/dotnetcore/DockerfileWithCurlUpdate.template
```diff
@@ -16,7 +16,12 @@ FROM %DOTNETCORE_BASE_IMAGE%
 # We manually update it here so we can still depend on the original images.
 # This command should be removed once support for deprecated .NET core images is halted.
 RUN sed -i '/jessie-updates/d' /etc/apt/sources.list  # Now archived
-RUN apt-get update && apt-get install curl --yes
```
gpcastro: how about we just update `curl` in all images, since you're adding an apt install command? That would free us from maintaining those two templates.

tests/SampleApps/DotNetCore/TodoAppFromMac/Readme.md
gpcastro: I think we should just build the app using `-r mac` instead of adding binaries to our repo

images/runtime/dotnetcore/DockerfileWithCurlUpdate.template
kichalla: Good point 👍 ..will do that

tests/SampleApps/DotNetCore/TodoAppFromMac/Readme.md
kichalla: I am pasting the following content from another PR for reference...basically we cannot test the scenario that I am testing here when using `-r osx-x64` as it throws:

**Comment from my old PR:**
I tried the following scenario now - publish the app for Mac and do a `dotnet foo.dll` on linux...
and this failed to execute...so we need not worry about this scenario as it would never have worked.

FYI:
**Publish for Mac but run on Linux**
image

**Publish for Linux and run on Linux**
image

images/runtime/dotnetcore/DockerfileWithCurlUpdate.template
kichalla: Oh seems like we are also doing this in the image having the curl update
```
# Older .NET core versions, which have reached end of life and therefore are no longer updated, use
# a version of `curl` that has known issues.
# We manually update it here so we can still depend on the original images.
# This command should be removed once support for deprecated .NET core images is halted.
RUN sed -i '/jessie-updates/d' /etc/apt/sources.list  # Now archived
```
So probably we cannot move it(?)

PR #149, opened 2019-05-16
Introduced message/warning framework

dorfire:


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -132,7 +135,16 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
...
             {
-                var scriptGenerator = new BuildScriptGenerator(console, serviceProvider);
+                var checkerMessages = new List<ICheckerMessage>();
+                var scriptGenerator = new BuildScriptGenerator(serviceProvider, console, checkerMessages);
+
+                if (checkerMessages.Count > 0)
+                {
+                    var messageFormatter = new DefinitionListFormatter();
+                    checkerMessages.Select(msg => messageFormatter.AddDefinition(msg.Level.ToString(), msg.Content));
+                    console.WriteLine(messageFormatter.ToString());
```
kichalla: `console.Error.WriteLine`?

src/BuildScriptGenerator/CheckerExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    public static class CheckerExtensions
+    {
+        public static IEnumerable<IChecker> WhereApplicable(this IEnumerable<IChecker> checkers, IDictionary<string, string> tools)
```
kichalla: nit: long lines..please fix everywhere applicable in this PR

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -175,13 +184,31 @@ public bool TryGenerateBashScript(BuildScriptGeneratorContext context, out strin
...
+            IDictionary<string, string> tools,
+            [NotNull] List<ICheckerMessage> checkerMessageSink)
+        {
+            var checkers = _checkers.WhereApplicable(tools);
+
+            using (var timedEvent = _logger.LogTimedEvent("RunCheckers"))
+            {
+                timedEvent.AddProperty(
+                    "checkersApplied",
+                    string.Join(',', checkers.Select(checker => checker.GetType().Name)));
```
kichalla: nit: `','` => `', '`

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -175,13 +184,31 @@ public bool TryGenerateBashScript(BuildScriptGeneratorContext context, out strin
...
+
+                checkerMessageSink.AddRange(checkers.SelectMany(checker => checker.CheckSourceRepo(ctx.SourceRepo)));
```
kichalla: I would place the calls to `CheckSourceRepo` and `CheckToolVersions` in a try-catch block...we do not want these to fail the build even though the actual build might have succeeded

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
kichalla: Should we have a header or something which shows that these warnings came from Oryx?

src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
```diff
@@ -39,7 +40,7 @@ internal class BuildScriptCommand : CommandBase
 
         internal override int Execute(IServiceProvider serviceProvider, IConsole console)
         {
-            var scriptGenerator = new BuildScriptGenerator(console, serviceProvider);
+            var scriptGenerator = new BuildScriptGenerator(serviceProvider, console, null);
```
kichalla: nit: named parameter for `null`

src/Common/DefinitionListFormatter.cs
```diff
@@ -20,23 +20,32 @@ public class DefinitionListFormatter
...
+        public DefinitionListFormatter AddDefinitions([CanBeNull] IDictionary<string, string> values)
         {
             if (values != null)
             {
                 _rows.AddRange(values
                     .Select(pair => CreateDefTuple(pair.Key, pair.Value))
                     .Where(tuple => tuple != null));
             }
+
+            return this;
```
kichalla: 👍

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
dorfire: this is for our own telemetries, though, NOT for displaying to customers

src/BuildScriptGenerator/Node/NpmDependencyChecker.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Oryx.BuildScriptGenerator.Node
+{
+    [Checker(NodeConstants.NpmName)]
+    public class NpmDependencyChecker : IChecker
+    {
+        // Lists packages that should not be used, but were NOT marked as "deprecated" in npm itself.
+        private static readonly IDictionary<string, string> SupersededPackages = new Dictionary<string, string>
```
arroyc: I think  it will be better to move this dictionary as stand alone and independent of language, every time we add a tool/package etc we can just add it there.

tests/BuildScriptGenerator.Tests/DefaultBuildScriptGeneratorTest.cs
```diff
@@ -15,6 +16,8 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Tests
 {
     public class DefaultBuildScriptGeneratorTest : IClassFixture<TestTempDirTestFixture>
     {
+        private const string TestPlatformName = "test";
```
arroyc: can you also add a test where checkers can fail but we are not breaking the oryx build for that

arroyc:
question: how are we opting in/out ?

dorfire:
> question: how are we opting in/out ?

just added a env var flag :)

src/BuildScriptGenerator/Node/NpmDependencyChecker.cs
dorfire: but it's very dependent of the language - for example, the way we check NPM packages isn't the same as we do for Pip

arroyc: that's true .. but correct me if im  wrong .. what we are doing here is basically scanning the packagejson for each package and checking whether it exists in the dictionary? Will not that be similar for python where we can check the list of packages from reqrmnt.txt and scanning in the dictionary to see if there is any outdated package used by the user?

Also note that it will make our builds longer.

arroyc: question: this checker doesn't care about dev dependencies right?

dorfire: 1. Why would it make the builds longer? It's a simple dict search (O(1))
2. Yes, it checks devDependencies as well (line 32)

arroyc: 1. yes search will be O(1), but scanning through all the dependencies will be O(n), also list addition adds O(n) once it crosses capacity, so approx. O(n) as a whole. Anyway  my point is to scan the dev dependency only when we need it.

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -19,24 +20,30 @@ namespace Microsoft.Oryx.BuildScriptGenerator
...
-        public bool TryGenerateBashScript(BuildScriptGeneratorContext context, out string script)
+        public bool TryGenerateBashScript(
+            BuildScriptGeneratorContext context,
+            out string script,
+            List<ICheckerMessage> checkerMessageSink = null)
         {
             script = null;
 
-            var toolsToVersion = new Dictionary<string, string>();
+            var toolsToVersion = new Dictionary<string, string>(); // To be populated by GetBuildSnippets
```
kichalla: Just to be safe (I know this is built by _us_ and not depended on end user input), I would make the key comparison ignore the case...there is a overload to this dictionary

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
kichalla: I see...just though it will be easy to read...you can leave it

src/BuildScriptGenerator/Node/NodeVersionChecker.cs
```diff
@@ -0,0 +1,36 @@
...
+        public IEnumerable<ICheckerMessage> CheckToolVersions(IDictionary<string, string> tools)
+        {
+            if (tools.ContainsKey(NodeConstants.NodeJsName))
+            {
+                var used = tools[NodeConstants.NodeJsName];
+                if (SemanticVersionResolver.CompareVersions(used, NodeScriptGeneratorOptionsSetup.NodeLtsVersion) < 0)
+                {
+                    return new[]
+                    {
+                        new CheckerMessage($"An outdated version of Node.js was used ({used}). Consider updating.")
```
kichalla: 👍 nice check...but I would like the message to be more clear...like saying current LTS version is this, but you are currently using this, so consider updating etc... (should 'was used' be 'is being used'?)

src/BuildScriptGenerator/Node/NpmDependencyChecker.cs
dorfire: > only when we need it
A user might run into issues with these packages even when building locally (without oryx); thus, IMO, we should let them know about these (potential) problems even if they won't directly affect the build.
To me, it's just another way of reducing customer frustration, not necessarily improving the Oryx build success rate

tests/BuildScriptGenerator.Tests/Node/NodeVersionCheckerTest.cs
```diff
@@ -0,0 +1,59 @@
...
+// --------------------------------------------------------------------------------------------
+
+using System.Linq;
+using System.Collections.Generic;
+using Microsoft.Oryx.BuildScriptGenerator.Node;
+using Xunit;
+
+namespace Microsoft.Oryx.BuildScriptGenerator.Tests.Node
+{
+    public class NodeVersionCheckerTest
```
kichalla: 👍

src/BuildScriptGenerator/Node/NpmDependencyChecker.cs
arroyc: well, I saw we do dev dependencies for npm install all the time, in that case this is not needed. so ignore this

tests/BuildScriptGenerator.Tests/Node/NodeVersionCheckerTest.cs
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        [Fact]
+        public void Checker_DoesNotWarn_WhenCurrentVersionUsed()
+        {
+            // Arrange
+            var checker = new NodeVersionChecker();
+
+            // Act
+            var messages = checker.CheckToolVersions(
```
kichalla: Wouldn't this test fail in future (given how quick Node releases new versions 😄 )...You probably want to dynamically get the LTS version and increment it for this test to be resilient

tests/BuildScriptGeneratorCli.Tests/CliEnvironmentSettingsTest.cs
```diff
@@ -16,6 +16,7 @@ public static IEnumerable<object[]> DisableVariableNamesAndGetters
         {
             get
             {
+                yield return new object[] { "DISABLE_CHECKERS", (Func<CliEnvironmentSettings, bool>)(s => s.DisableCheckers) };
```
kichalla: nit: long line

src/BuildScriptGenerator/CheckerAttribute.cs
```diff
@@ -0,0 +1,19 @@
...
+using System;
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    [AttributeUsage(AttributeTargets.Class)]
+    public class CheckerAttribute : Attribute
+    {
+        private readonly string[] _targetToolNames;
+
+        public CheckerAttribute(params string[] toolNames) => _targetToolNames = toolNames;
```
kichalla: nit: Change the variable & property to something like `RestrictToTools` just to be clear?

src/BuildScriptGenerator/CheckerExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    public static class CheckerExtensions
+    {
+        public static IEnumerable<IChecker> WhereApplicable(
```
kichalla: Can we have some unit tests for this? there are some good scenarios to test

src/BuildScriptGenerator/Contracts/IBuildScriptGenerator.cs
```diff
@@ -17,15 +17,14 @@ public interface IBuildScriptGenerator
         /// The <see cref="BuildScriptGeneratorContext"/> with parameters for the script.
         /// </param>
         /// <param name="script">The generated script if the operation was successful.</param>
+        /// <param name="checkerMessageSink">If specified, messages from checkers will be appended to this list.</param>
         /// <returns><c>true</c> if the operation was successful, <c>false</c> otherwise.</returns>
-        bool TryGenerateBashScript(BuildScriptGeneratorContext scriptGeneratorContext, out string script);
+        bool TryGenerateBashScript(BuildScriptGeneratorContext scriptGeneratorContext, out string script, List<ICheckerMessage> checkerMessageSink = null);
```
kichalla: nit: long line (sorry just want to make it a habit to remember to always do it)

tests/BuildScriptGenerator.Tests/Node/NodeVersionCheckerTest.cs
```diff
@@ -0,0 +1,59 @@
...
+                new Dictionary<string, string> { { NodeConstants.NodeJsName, "1.0.0" } });
+
+            // Assert
+            Assert.Single(messages);
+            Assert.Contains("outdated version of Node.js was used", messages.First().Content);
+        }
+
+        [Fact]
```
arroyc: I would like this to be theory where we verify this for all the lts versions we support

src/BuildScriptGenerator/Node/NodeVersionChecker.cs
dorfire: I didn't want to say "LTS" or "current" because these definition change in Oryx on a different cadence than how they change in Node.js. I can't think of a way to resolve this other than adding a link to the Oryx docs.

tests/BuildScriptGenerator.Tests/Node/NodeVersionCheckerTest.cs
dorfire: It will only fail when we change the value of `NodeScriptGeneratorOptionsSetup.NodeLtsVersion`.

arroyc:
I don't have any new comments but make sure both validation build and nightly build is successful before merging this to master

tests/BuildScriptGenerator.Tests/Node/NodeVersionCheckerTest.cs
dorfire: actually it'll be weird to implement since we don't have any seperation in the `TestValueGenerator` between "old" and "current" versions

tests/BuildScriptGenerator.Tests/CheckerExtensionsTest.cs
```diff
@@ -0,0 +1,48 @@
...
+using Xunit;
+
+namespace Microsoft.Oryx.BuildScriptGenerator.Tests
+{
+    public class CheckerExtensionsTest
+    {
+        [Fact]
+        public void WhereApplicable_Sanity()
+        {
+            var a = new CheckerA();
```
kichalla: can you add a checker which applies globally too

PR #151, opened 2019-05-17
Python runtime images share same structure as build image

gpcastro:
- Unify our python images so the binaries are located in the same place both at build and runtime
- Migrate Python base images to stretch, also for consistency


images/runtime/python/generateDockerfiles.sh
```diff
@@ -8,24 +8,28 @@ set -e
 
 declare -r DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 declare -r PYTHON_VERSIONS_PATH=$DIR/../../../build/__python-versions.sh
-declare -r VERSIONS_FILE="$DIR/pythonVersions.txt"
 declare -r DOCKERFILE_TEMPLATE="$DIR/Dockerfile.template"
-declare -r IMAGE_NAME_PLACEHOLDER="%PYTHON_BASE_IMAGE%"
-declare -r IMAGE_SUFFIX="-slim-stretch"
+declare -r PYTHON_MAJOR_VERSION_PLACEHOLDER="%PYTHON_MAJOR_VERSION%"
+declare -r PYTHON_VERSION_PLACEHOLDER="%PYTHON_VERSION%"
+declare -r PYTHON_FULL_VERSION_PLACEHOLDER="%PYTHON_FULL_VERSION%"
```
kichalla: I am a bit confused with this line and the above...could you add a comment as to how they would like?

images/runtime/python/generateDockerfiles.sh
```diff
@@ -8,24 +8,28 @@ set -e
...
+declare -r ORYX_IAMGE_TAG_PLACEHOLDER="%IMAGE_TAG%"
```
kichalla: typo: IAMGE > IMAGE

images/runtime/python/Dockerfile.template
```diff
@@ -9,7 +9,21 @@ ENV BUILD_NUMBER=${BUILD_NUMBER}
...
+
+# Add python binaries from Oryx base image
+COPY --from=mcr.microsoft.com/oryx/python-build-%PYTHON_VERSION%:%IMAGE_TAG% /opt /opt
+RUN set -ex \
+ && cd /opt/python/ \
+ && ln -s %PYTHON_FULL_VERSION% %PYTHON_VERSION% \
+ && ln -s %PYTHON_VERSION% %PYTHON_MAJOR_VERSION% \
+ && echo /opt/python/%PYTHON_MAJOR_VERSION%/lib >> /etc/ld.so.conf.d/python.conf \
+ && ldconfig
+
```
kichalla: Could you check if any of these are required? https://github.com/docker-library/python/blob/34c9df35e9a69e9f0edde88e861b543edb8bc07a/3.7/stretch/Dockerfile#L62

Also a problem with that official Dockefile is that it is not clear if the environment variables is for doing the building of the source or for anyone to consume when inheriting the docker image.

images/build/Dockerfile
```diff
@@ -228,8 +228,8 @@ RUN apt-get update \
 # https://github.com/docker-library/python/issues/147
 ENV PYTHONIOENCODING UTF-8
 COPY build/__python-versions.sh /tmp
-COPY --from=mcr.microsoft.com/oryx/python-build-2.7:20190322.1 /opt /opt
-COPY --from=mcr.microsoft.com/oryx/python-build-3.6:20190322.1 /opt /opt
+COPY --from=mcr.microsoft.com/oryx/python-build-2.7:20190402.1 /opt /opt
```
arroyc: note: we need to change this again soon

images/runtime/python/Dockerfile.template
```diff
@@ -9,7 +9,21 @@ ENV BUILD_NUMBER=${BUILD_NUMBER}
 # GOPATH is set to "/go" in the base image, so no need to set it again
 RUN ./build.sh python /opt/startupcmdgen/startupcmdgen
 
-FROM python:%PYTHON_BASE_IMAGE%
+FROM buildpack-deps:stretch AS main
+
+ENV PYTHON_VERSION %PYTHON_FULL_VERSION%
+
+# Add python binaries from Oryx base image
+COPY --from=mcr.microsoft.com/oryx/python-build-%PYTHON_VERSION%:%IMAGE_TAG% /opt /opt
```
arroyc: same comment

PR #152, opened 2019-05-17
Fixed ASP.NET Core 'web project' probing logic to consider more scenarios

kichalla:


src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
```diff
@@ -132,19 +133,42 @@ public string GetProjectFile(ISourceRepo sourceRepo)
...
         {
-            var webSdkProjectElement = projectFileDoc.XPathSelectElement(
-                DotnetCoreConstants.WebSdkProjectXPathExpression);
-            return webSdkProjectElement != null;
+            // For reference
+            // https://docs.microsoft.com/en-us/visualstudio/msbuild/project-element-msbuild?view=vs-2019
+
+            // Look for the attribute value on Project element first as that is more common
+            // Example: <Project Sdk="Microsoft.NET.Sdk.Web">
+            var expectedWebSdkName = DotnetCoreConstants.WebSdkName.ToLowerInvariant();
```
gpcastro: why not just use `OrdinalIgnoreCase` in both if statements and skip this `ToLowerInvariant()` call?

src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
```diff
@@ -132,19 +133,42 @@ public string GetProjectFile(ISourceRepo sourceRepo)
...
+            var sdkAttributeValue = projectFileDoc.XPathEvaluate(
+                DotnetCoreConstants.ProjectSdkAttributeValueXPathExpression);
+            var sdkName = sdkAttributeValue as string;
+            if (!string.IsNullOrEmpty(sdkName))
+            {
+                return sdkName.ToLowerInvariant().StartsWith(expectedWebSdkName);
```
gpcastro: If this is false, shouldn't we let it proceed to the next check, with the `Sdk` node?
Thinking of this case:
```
<Project Sdk="bla">
   <Sdk Name="Microsoft.NET.Sdk.Web... />
...
```

kichalla: Sure, sounds good

PR #154, opened 2019-05-17
Added App Insights source prefix

dorfire:


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -68,6 +68,20 @@ internal class BuildCommand : CommandBase
             Description = "Additional information used by this tool to generate and run build scripts.")]
         public string[] Properties { get; set; }
 
+        public static string BuildOperationName(IEnvironment env)
```
kichalla: Did we find out how it is generally recommended for this kind of use-case? Like this is pretty basic case...here they seem to be recommending custom dimensions https://stackoverflow.com/a/41705558/1184056

src/Common/LoggingConstants.cs
```diff
@@ -12,12 +13,16 @@ public static class LoggingConstants
         public const string ApplicationInsightsInstrumentationKeyEnvironmentVariableName
             = "ORYX_AI_INSTRUMENTATION_KEY";
 
-        public const string AppServiceAppNameEnvironmentVariableName = "APPSETTING_WEBSITE_SITE_NAME";
-        public const string ContainerRegistryAppNameEnvironmentVariableName = "REGISTRY_NAME";
         public const string DefaultOperationName = ".oryx";
 
         public const string DefaultLogPath = "/tmp/oryx.log";
 
+        public static readonly IDictionary<string, string> OperationNameSourceEnvVars = new Dictionary<string, string>
```
kichalla: We need changes in runtime startup scripts too?

tests/BuildScriptGeneratorCli.Tests/BuildCommandTest.cs
```diff
@@ -362,6 +364,19 @@ public void IsValid_IsFalse_IfLanguageVersionSpecified_WithoutLanguageName()
                 testConsole.StdError);
         }
 
+        public static IEnumerable<object[]> GetOperationNameEnvVarNames() =>
+            Microsoft.Oryx.Common.LoggingConstants.OperationNameSourceEnvVars.Select(e => new[] { e.Key, e.Value });
+
+        [Theory]
+        [MemberData(nameof(GetOperationNameEnvVarNames))]
+        public void BuildOperationName_Sanity(string envVarName, string opNamePrefix)
```
kichalla: Prefer descriptive comment so that it is clear what the test is doing

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
kichalla: Based on our offline discussion, ignore this comment

src/Common/LoggingConstants.cs
dorfire: yes, in a separate PR

kichalla: sound good

PR #156, opened 2019-05-17
Rebuild all node base images on stretch

gpcastro:
We need to save overall disk space. I'm moving all our node images to be built on top of latest stretch, so they can share its bits.


build/build-runtimeimages-bases.sh
```diff
@@ -49,18 +49,6 @@ fi
 
 labels="--label com.microsoft.oryx.git-commit=$GIT_COMMIT --label com.microsoft.oryx.build-number=$BUILD_NUMBER"
 
-generateDockerFiles=$(find $runtimeImagesSourceDir -type f -name "generateDockerfiles.sh")
```
kichalla: I am not clear why this change is made

gpcastro: It's actually pretty clear... I screwed up :)

PR #161, opened 2019-05-20
converting ci builds to parallel

arroyc:
this also includes a fix that use signed binaries baked to the build image


vsts/pipelines/_buildParallel.yml
```diff
@@ -111,7 +108,9 @@ steps:
   inputs:
     type: InlineScript
     script: 'echo "success">$(Build.SourcesDirectory)/artifacts/$(Build.BuildNumber).txt'
-  condition: and(succeeded(), in(variables['Build.Reason'], 'Schedule', 'Manual'), eq(variables['Build.DefinitionName'], 'Oryx-CI'))
+  condition: >
```
gpcastro: this indentation is confusing. We could split the statements inside `and` one per line.

vsts/pipelines/_integration.yml
```diff
@@ -0,0 +1,160 @@
+parameters:
+  isCIBuild: 'false'
+
+jobs:
+- job: Job_RuntimeImages
+  displayName: Build and Test Runtime Images
+  condition: succeeded()
+  pool:
+    name: OryxLinux
+  # Building runtime images can take a long time due our PHP images
```
gpcastro: nit: due _to_ our PHP...

vsts/pipelines/_integration.yml
```diff
@@ -0,0 +1,160 @@
...
+    - Job_RuntimeImages
+    - Job_PythonIntegrationTests
+    - Job_PhpIntegrationTests
+    - Job_NodeIntegrationTests
+    - Job_DotNetCoreIntegrationTests
+    - Job_DbIntegrationTests
+  pool:
+    name: OryxLinux
+  steps:
+  - task: ms-devlabs.utilitytasks.task-Shellpp.Shell++@0
```
gpcastro: what do we need this job for?

vsts/pipelines/_signbinary.yml
```diff
@@ -5,29 +5,34 @@ steps:
...
+- script: |
+    echo $(setSignTypeVariable.SignType)
+  name: SignType
+
 - task: NuGetToolInstaller@0
   displayName: 'Use NuGet 4.6.2'
   inputs:
     versionSpec: 4.6.2
-  condition: and(succeeded(), eq(variables['SignType'], 'real'))
+  condition: and(succeeded(), eq(variables['setSignTypeVariable.SignType'], 'real'))
```
gpcastro: Didn't we have a test to check if the dll was actually signed, case in which we add a .txt file to the build output that is then checked in the release definition?

vsts/pipelines/nightly.yml
```diff
@@ -23,160 +22,14 @@ jobs:
...
-      echo "##vso[task.setvariable variable=PushRuntimeImages;]true"
-      echo "##vso[task.setvariable variable=PushBuildImages;]false"
-      echo "##vso[task.setvariable variable=PushToDockerHub;]false"
       echo "##vso[task.setvariable variable=EmbedBuildContextInImages;]true"
     displayName: 'Set variables'
   - template: _buildParallel.yml
+    parameters:
+      pushToDockerHub: 'false'
 
-- job: Job_PythonIntegrationTests
```
gpcastro: So... the nightly builds won't be parallel anymore? Where did these changes go?

vsts/pipelines/_integration.yml
```diff
@@ -0,0 +1,160 @@
...
+      echo "##vso[task.setvariable variable=TestRuntimeImages;]true"
+      echo "##vso[task.setvariable variable=PushRuntimeImages;]true"
+      echo "##vso[task.setvariable variable=PushBuildImages;]false"
+      echo "##vso[task.setvariable variable=EmbedBuildContextInImages;]true"
+    displayName: 'Set variables'
+  - template: _buildParallel.yml
+    parameters:
+      pushToDockerHub: ${{ parameters.isCIBuild }}
+
+- job: Job_PythonIntegrationTests
```
gpcastro: q: Why can't we move these tests tasks to the template so we can reuse them on every build? that is, make the tests parallel everywhere we run them.

arroyc: yes im working on this .. I created this PR on a stable point where I know it's working :)  … I will be cleaning up this now and update few more things

vsts/pipelines/nightly.yml
arroyc: its still parallel

vsts/pipelines/_signbinary.yml
arroyc: that txt file was for integration test ran successfully

vsts/pipelines/templates/_buildParallel.yml
```diff
@@ -1,15 +1,12 @@
 parameters:
```
kichalla: I feel this `_buildParallel.yml` is still monolithic having different tasks that we have in different build definitions...I think we should probably break this down more into reusable jobs?...something to think about...i might have missed few things...does you think this makes it better?

```
1. build, test and push build image (job type 1)

2. build, test and push runtime images (job type 2)

3. integration tests - <runtime-name> (job type 3)
   pull and retag build image
   pull and retag runtime images - <runtime-name>

4. signing job (job type 4)

validation.yml:
1. job type 1 (do not push)
2. job type 2 (do not push)

night.yml:
1. job type 1 (push)
2. job type 2 (push)
3. job type 3 (python)
4. job type 3 (php)
5. job type 3 (node)
6. job type 3 (dotnetcore)

ci.yml:
night.yml:
0. job type 4 (sign)
1. job type 1 (push)
2. job type 2 (push)
3. job type 3 (python)
4. job type 3 (php)
5. job type 3 (node)
6. job type 3 (dotnetcore)



```

arroyc: considering this as future work as there will be too many moveable parts

tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
```diff
@@ -27,12 +27,17 @@ public VersionInformationTest(ITestOutputHelper output)
         }
 
         [SkippableFact]
-        public void OryxBuildImage_Contains_VersionAndCommit_Information()
+        public void OryxBuildImage_Contains_VersionCommitAndSign_Information()
```
arroyc: added a condition for sign check

vsts/pipelines/templates/_signbinary.yml
```diff
@@ -0,0 +1,87 @@
...
+- task: CopyFiles@2
+  displayName: 'Copy Files to: $(Build.ArtifactStagingDirectory)/binaries'
+  inputs:
+    SourceFolder: 'src/BuildScriptGeneratorCli/bin/$(BuildConfiguration)/linux-x64/publish'
+    Contents: '**\*'
+    TargetFolder: '$(Build.ArtifactStagingDirectory)/binaries'
+    CleanTargetFolder: true
+  condition: and(succeeded(), eq(variables['setSignTypeVariable.SignType'], 'real'))
+
+- powershell: |
```
arroyc: this task is to add a simple log file that will be copied to final build image when we sign oryx binaries @gpcastro

tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
arroyc: reverted

vsts/pipelines/templates/_signbinary.yml
arroyc: reverted

vsts/pipelines/templates/_buildParallel.yml
gpcastro: I wonder if we should call this file _buildTemplate.tml now, since parallel is an implementation detail, and build template is what it really is?

arroyc: I agree … I will be sending another PR soon to address this

PR #162, opened 2019-05-21
Fix .NET Core build script to build from intermediate directory if provided

kichalla:


src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
```diff
@@ -32,49 +33,47 @@ internal class DefaultAspNetCoreWebAppProjectFileProvider : IAspNetCoreWebAppPro
...
+                        $" '{EnvironmentSettingsKeys.Project}' with value '{projectFileWithRelativePath}'. " +
                         "Make sure the path to the project file is relative to the root of the repo. " +
                         "For example: PROJECT=src/Dashboard/Dashboard.csproj");
                 }
 
                 // NOTE: Do not check if the project file specified by the end user is a web application since this
                 // can be a escape hatch for end users if our logic to determine a web app is incorrect.
-                return projectFile;
+                return projectFileWithRelativePath;
             }
```
gpcastro: nit: I'd leave the `else` here to make it clearer that this is only when options.Project is null

src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
```diff
@@ -171,6 +169,25 @@ internal static bool IsAspNetCoreWebApplicationProject(XDocument projectFileDoc)
...
+        internal static string GetRelativePathToRoot(string projectFilePath, string repoRoot)
+        {
+            var repoRootDir = new DirectoryInfo(repoRoot);
+            var projectFileInfo = new FileInfo(projectFilePath);
+            var currDir = projectFileInfo.Directory;
+            var parts = new List<string>();
+            parts.Add(projectFileInfo.Name);
+
+            // Since directory names are case sensitive on non-Windows OSes, try not to use ignore case
+            while (!string.Equals(currDir.FullName, repoRootDir.FullName, StringComparison.Ordinal))
```
gpcastro: I'm concerned that this might throw an exception if those will never be equal, e.g. when `projectFilePath` is not in a subdir of `repoRoot`... I think we should handle this error case here just to make sure it fails gracefully if that ever happens

src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
```diff
@@ -171,6 +169,25 @@ internal static bool IsAspNetCoreWebApplicationProject(XDocument projectFileDoc)
...
+            {
+                parts.Insert(0, currDir.Name);
```
gpcastro: as we discussed, we could append and then reverse once for efficiency

src/BuildScriptGenerator/DotNetCore/DefaultAspNetCoreWebAppProjectFileProvider.cs
```diff
@@ -171,6 +169,25 @@ internal static bool IsAspNetCoreWebApplicationProject(XDocument projectFileDoc)
             return string.Equals(sdkName, expectedWebSdkName, StringComparison.OrdinalIgnoreCase);
         }
 
+        // To enable unit testing
+        internal static string GetRelativePathToRoot(string projectFilePath, string repoRoot)
+        {
+            var repoRootDir = new DirectoryInfo(repoRoot);
```
gpcastro: how about getting the full path for both, and just getting the `repoRoot` substring starting at `projectFilePath.Length +1`?

PR #163, opened 2019-05-21
Added a basic buildpack test & fixed checker discovery

dorfire:


src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -483,10 +488,8 @@ private string GetNpmVersion(dynamic packageJson)
             if (!string.IsNullOrWhiteSpace(npmVersionRange))
             {
                 var supportedNpmVersions = _nodeVersionProvider.SupportedNpmVersions;
-                npmVersion = SemanticVersionResolver.GetMaxSatisfyingVersion(
-                    npmVersionRange,
-                    supportedNpmVersions);
-                if (string.IsNullOrWhiteSpace(npmVersion))
+                npmVersion = SemanticVersionResolver.GetMaxSatisfyingVersion(npmVersionRange, supportedNpmVersions);
+                if (string.IsNullOrEmpty(npmVersion))
```
gpcastro: why this change?

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -208,16 +216,24 @@ private static string GetBenvArgs(IDictionary<string, string> benvArgsMap)
             IDictionary<string, string> tools,
             [NotNull] List<ICheckerMessage> checkerMessageSink)
         {
-            var checkers = _checkers.WhereApplicable(tools);
+            var checkers = _checkers.WhereApplicable(tools).ToArray();
```
gpcastro: all of these parameters should be `[NotNull]`

src/BuildScriptGenerator/Node/NodeDependencyChecker.cs
```diff
@@ -19,11 +20,19 @@ public class NpmDependencyChecker : IChecker
...
+        {
+            _logger = logger;
+        }
+
         public IEnumerable<ICheckerMessage> CheckSourceRepo(ISourceRepo repo)
         {
             dynamic packageJson = NodePlatform.GetPackageJsonObject(repo, null);
             if (packageJson == null)
             {
+                _logger.LogDebug("packageJson is null");
```
gpcastro: It would be helpful to have more context on this message as we read through the logs, to provide a semantic value rather than pure state. e.g. "Node checker skipping repo since packageJson is null"

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -203,12 +203,17 @@ public bool IsEnabledForMultiPlatformBuild(BuildScriptGeneratorContext scriptGen
...
+                string npmVersion = GetNpmVersion(packageJson);
+                _logger.LogDebug("GetNpmVersion returned {npmVersion}", npmVersion);
                 if (!string.IsNullOrEmpty(npmVersion))
                 {
                     toolsToVersion[NodeConstants.NpmToolName] = npmVersion;
                 }
             }
+            else
+            {
+                _logger.LogDebug("packageJson is null");
```
gpcastro: adding more context would be help, e.g. "Not settings Node tools version since packageJson is null"

src/BuildScriptGenerator/Node/NodeVersionChecker.cs
```diff
@@ -6,26 +6,33 @@
...
-                if (SemanticVersionResolver.CompareVersions(used, NodeScriptGeneratorOptionsSetup.NodeLtsVersion) < 0)
+                return new[]
                 {
-                    return new[]
-                    {
-                        new CheckerMessage($"An outdated version of Node.js was used ({used}). Consider updating. " +
-                                           $"Versions supported in Oryx: {Constants.OryxGitHubUrl}")
-                    };
-                }
+                    new CheckerMessage($"An outdated version of Node.js was detected ({used}). Consider updating.\n" +
```
gpcastro: we have a resource file on this project to put labels in. We should store those strings there so its easier to review them.

tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
```diff
@@ -252,14 +251,47 @@ public async Task CanBuildAndRun_NodeApp(string nodeVersion)
                 });
         }
 
+        [Fact]
+        public async Task CanBuildAndRun_NodeApp_WithBuildpack()
+        {
+            // Arrange
+            var appName = "webfrontend";
+            var appVolume = CreateAppVolume(appName);
+            var dockerPort = new DockerVolume("/var/run/docker.sock", "/var/run/docker.sock");
```
kichalla: Curious, what does this do?

tests/Oryx.Tests.Common/EndToEndTestHelper.cs
```diff
@@ -130,7 +161,7 @@ public static class EndToEndTestHelper
             // Build
             var buildAppResult = dockerCli.Run(new DockerRunArguments
             {
-                ImageId = Settings.BuildImageName,
+                ImageId = buildImage,
```
kichalla: 👍 I wanted to make this change for a while now

tests/Oryx.Tests.Common/DockerVolume.cs
```diff
@@ -22,6 +22,15 @@ public class DockerVolume
 
         public const string ContainerDirRoot = "/oryxtests";
 
+        // TODO: rename Dir -> Path
+        public DockerVolume(
```
kichalla: Not making this public was deliberate since end users (us who write tests) do not hit into errors, so I am wondering if this change is really required for your use case.

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -140,16 +140,22 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
...
                 var scriptGenerator = new BuildScriptGenerator(serviceProvider, console, checkerMessages);
 
+                var generated = scriptGenerator.TryGenerateScript(out scriptContent);
+                stopwatch.AddProperty("generateSucceeded", generated.ToString());
+
                 if (checkerMessages.Count > 0)
                 {
                     var messageFormatter = new DefinitionListFormatter();
-                    checkerMessages.Select(msg => messageFormatter.AddDefinition(msg.Level.ToString(), msg.Content));
+                    checkerMessages.ForEach(msg => messageFormatter.AddDefinition(msg.Level.ToString(), msg.Content));
```
kichalla: Do we need tests for these lines of code?

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -255,13 +271,14 @@ private static string GetBenvArgs(IDictionary<string, string> benvArgsMap)
...
                 {
-                    _logger.LogDebug("Script generator {scriptGenType} was used", platform.GetType());
+                    _logger.LogDebug("Platform {platformType} was used", platform.GetType());
                     snippets.Add(snippet);
                     platform.SetRequiredTools(context.SourceRepo, targetVersion, toolsToVersion);
                 }
                 else
                 {
-                    _logger.LogDebug("Script generator {scriptGenType} cannot be used", platform.GetType());
+                    _logger.LogWarning("{platformType}.GenerateBashBuildScriptSnippet() returned null",
```
kichalla: Split the arguments to a method into separate lines if you are not on a single line

src/BuildScriptGenerator/Node/NodePlatform.cs
dorfire: because it fits in less than 120 characters

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
dorfire: correct, but i wanted to be over-explicit and avoid confusion with `IBuildScriptGenerator.GenerateBashScript` which can receive a null checkerMessageSink.

src/BuildScriptGenerator/Node/NodeDependencyChecker.cs
dorfire: Don't forget this message will have a `LoggerName` costume dimension with it - do you think the extra context is still worth adding?

tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
dorfire: Allows the `pack` binary to use the Docker daemon

tests/Oryx.Tests.Common/DockerVolume.cs
dorfire: Well, I don't really need the whole mirror-copy thing that happens in `DockerVolume.Create`.. so the only other choice is introducing another static initializer (like renaming the current method to `CreateMirror` and then adding a no-op `DockerVolume.Create`)

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
dorfire: We can modify an existing end-to-end test to cover them

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
dorfire: I changed the styling rule file so that this doesn't trigger a warning anymore

tests/Oryx.Tests.Common/DockerVolume.cs
kichalla: Yes, i was thinking the same thing...we should rather create another static member to abstract the user from this detail.

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
kichalla: Some teams (ex: ASP.NET Core) follow this pattern of either having all args on a single line or split individually...we could all agree on one approach and stick to it...

src/BuildScriptGenerator/Node/NodeDependencyChecker.cs
gpcastro: Yes. It's easier to see what's going on when reading through the files. Besides, this message depends on the existing version of the code to see what that means. If we change this code at some point, we may not be able to understand the right context in which this was presented.

PR #165, opened 2019-05-22
Added a checker for global NPM package installations

dorfire:
Should be reviewed only after PR #163 is merged


dorfire:
The build seems to hit a weird issue:https://github.com/dotnet/coreclr/issues/20236

src/BuildScriptGenerator/Node/NodePackageScriptsChecker.cs
```diff
@@ -0,0 +1,60 @@
...
+using JetBrains.Annotations;
+
+namespace Microsoft.Oryx.BuildScriptGenerator.Node
+{
+    [Checker(NodeConstants.NodeToolName)]
+    public class NodePackageScriptsChecker : IChecker
+    {
+        public IEnumerable<ICheckerMessage> CheckSourceRepo(ISourceRepo repo)
+        {
+            dynamic packageJson = NodePlatform.GetPackageJsonObject(repo, null);
```
kichalla: nit: named parameter for `null`

src/BuildScriptGenerator/Node/NodePackageScriptsChecker.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        private void CheckScript(IDictionary<string, string> scripts, string scriptKey, List<ICheckerMessage> result)
+        {
+            if (!scripts.ContainsKey(scriptKey))
+            {
+                return;
+            }
+
+            string script = scripts[scriptKey];
+            if (script?.Contains("npm install -g") == true || script?.Contains("npm i -g") == true)
```
kichalla: Just to be safe, I would split the command by white space and look for presence of `-g` or `--global` since user could use this switch anywhere

tests/BuildScriptGenerator.Tests/Node/NodePackageScriptsCheckerTest.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class NodePackageScriptsCheckerTest
+    {
+        [Fact]
+        public void Checker_DetectsGlobalNpmInstalls()
+        {
+            // Arrange
+            var checker = new NodePackageScriptsChecker();
+            var scripts = new Dictionary<string, string>
+            {
+                { "preinstall",     "npm i -g casperjs # checked and problematic" },
```
kichalla: Make sure to add tests for the scenario mentioned in my above comment.

src/BuildScriptGenerator/Node/NodePackageScriptsChecker.cs
gpcastro: in some environments this is actually OK. We need to show this message only for app service

src/BuildScriptGenerator/Node/NodePackageScriptsChecker.cs
```diff
@@ -0,0 +1,60 @@
...
+            {
+                result.Add(new CheckerMessage(
+                    $"The script '{scriptKey}', defined in {NodeConstants.PackageJsonFileName}, seems to be trying " +
+                    $"to install packages globally. This is unsupported by Oryx.", Extensions.Logging.LogLevel.Error));
```
gpcastro: instead of "this is unsupported by Oryx", say something like "This might not work as expected since global packages might not be available when running the application"

src/BuildScriptGenerator/DefaultEnvironment.cs
```diff
@@ -7,11 +7,28 @@
...
 
 namespace Microsoft.Oryx.BuildScriptGenerator
 {
     internal class DefaultEnvironment : IEnvironment
     {
+        public EnvironmentType Type
+        {
+            get
+            {
+                foreach (var entry in LoggingConstants.OperationNameSourceEnvVars)
```
gpcastro: nit: consider caching it. How big can this IEnumerable be?

src/BuildScriptGenerator/Node/NodePackageScriptsChecker.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        private void CheckScript(IDictionary<string, string> scripts, string scriptKey, List<ICheckerMessage> result)
+        {
+            if (!scripts.ContainsKey(scriptKey))
+            {
+                return;
+            }
+
+            string script = scripts[scriptKey];
+            if (script?.Contains("-g") == true || script?.Contains("--global") == true)
```
gpcastro: this seems fragile. You could do that, per line, it if starts with `npm`.

src/BuildScriptGenerator/Node/NodePackageScriptsChecker.cs
```diff
@@ -0,0 +1,73 @@
...
+            {
+                result.Add(new CheckerMessage(
+                    $"The script '{scriptKey}', defined in {NodeConstants.PackageJsonFileName}, seems to be trying " +
+                    $"to install packages globally. This is unsupported by Oryx."));
```
gpcastro: Did you decide to keep Oryx in the message? I don't think users will make sense of what this means, and we can make it more actionable, saying that local packages should work fine in this environment.

src/BuildScriptGenerator/DefaultEnvironment.cs
dorfire: not big at all (should always stay less than 10 in my opinion) - but cache won't hurt

src/BuildScriptGenerator/Node/NodePackageScriptsChecker.cs
dorfire: the problem is that the line could start with something else altogether - for example, `true && npm i -g bla`.
that's why the wording of the warning message is delicate - it says "_seems to be trying to install_" rather than "_installing_"

src/BuildScriptGenerator/Node/NodePackageScriptsChecker.cs
dorfire: I was sure I changed it. Must have been swallowed by a bad merge or a VS update 😊

dorfire: changed it + moved to a label in a resource file

src/BuildScriptGenerator/Resources/Labels.resx
```diff
@@ -121,6 +121,10 @@
     <value>The package '{0}', specified in package.json's {1},\nis known to have been superseded by '{2}'. Consider switching over.</value>
     <comment>{0} - detected package; {1} - key of object in package.json; {2} - alternative package</comment>
   </data>
+  <data name="NodePackageGlobalInstallMessageFormat" xml:space="preserve">
+    <value>The script '{0}', defined in package.json, seems to be trying to install packages globally. This could have unexpected results, as global packages might not be available in the runtime container.</value>
```
dorfire: @gpcastro

src/BuildScriptGenerator/Node/NodePackageScriptsChecker.cs
gpcastro: Right, but at the same time a warning message that the user cannot resolve is annoying. We can use a regex e.g. `\bnpm.+\-?\-g(lobal)?\b`

tests/BuildScriptGenerator.Tests/Node/NodePackageScriptsCheckerTest.cs
```diff
@@ -0,0 +1,49 @@
...
+            Assert.Single(checker.CheckScriptsForGlobalInstallationAttempts(scripts));
+        }
+
+        [Theory]
+        [InlineData("npm install -g pkg",       true)]
+        [InlineData("npm install --global pkg", true)]
+        [InlineData("npm i -g pkg",             true)]
+        [InlineData("npm i --global pkg",       true)]
+        [InlineData("npm install pkg",                  false)]
+        [InlineData("npm install pkg && grep -g bla",   false)]
```
gpcastro: "cd abc && npm install -g"

tests/BuildScriptGenerator.Tests/Node/NodePackageScriptsCheckerTest.cs
```diff
@@ -0,0 +1,49 @@
...
+        [InlineData("npm install endswith-g",           false)]
+        [InlineData("npm install endswith-global",      false)]
```
gpcastro: how about `npm install --globally`?

src/BuildScriptGenerator/Node/NodePackageScriptsChecker.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Text.RegularExpressions;
+using JetBrains.Annotations;
+
+namespace Microsoft.Oryx.BuildScriptGenerator.Node
+{
+    [Checker(NodeConstants.NodeToolName)]
+    public class NodePackageScriptsChecker : IChecker
+    {
+        public static readonly Regex NpmGlobalPattern =
+            new Regex(@"npm[^&|;#]+\s\-?\-g(lobal)?");
```
gpcastro: you probably need `\b` in both ends to bound this expression.

tests/BuildScriptGenerator.Tests/Node/NodePackageScriptsCheckerTest.cs
```diff
@@ -0,0 +1,49 @@
...
+            // Act & Assert
+            Assert.Single(checker.CheckScriptsForGlobalInstallationAttempts(scripts));
+        }
+
+        [Theory]
+        [InlineData("npm install -g pkg",       true)]
+        [InlineData("npm install --global pkg", true)]
+        [InlineData("npm i -g pkg",             true)]
+        [InlineData("npm i --global pkg",       true)]
+        [InlineData("npm install pkg",                  false)]
```
gpcastro: "npm install\ngrep -g bla"

dorfire: scripts in `package.json` can't span multiple lines

tests/BuildScriptGenerator.Tests/Node/NodePackageScriptsCheckerTest.cs
```diff
@@ -36,10 +36,12 @@ public void Checker_DetectsGlobalNpmInstalls()
         [InlineData("npm install --global pkg", true)]
```
kichalla: Add a case like `npm install foo -g` and `npm install foo --global` (basically package name before the global switch)

PR #169, opened 2019-05-24
[DO NOT MERGE]enable docker system prune

arroyc:
enabling docker clean up by running docker system prune

### DIFF
--- a/vsts/scripts/dockerCleanup.sh
+++ b/vsts/scripts/dockerCleanup.sh
@@ -5,7 +5,6 @@
 # --------------------------------------------------------------------------------------------
 
 declare -r BUILD_NUMBER="$BUILD_BUILDNUMBER"
-declare -r DOCKER_SYSTEM_PRUNE="${ORYX_DOCKER_SYSTEM_PRUNE:-false}"
 
 echo
 echo "Printing all running containers and stopped containers"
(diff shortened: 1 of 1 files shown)


vsts/scripts/dockerCleanup.sh
```diff
@@ -5,7 +5,6 @@
 # --------------------------------------------------------------------------------------------
 
 declare -r BUILD_NUMBER="$BUILD_BUILDNUMBER"
```
gpcastro: what was the motivation for this change?

vsts/scripts/dockerCleanup.sh
```diff
@@ -81,13 +80,9 @@ echo
 docker images
 
 echo
-echo "Cleanup: Run 'docker system prune': $DOCKER_SYSTEM_PRUNE"
```
gpcastro: instead of removing it, we could just set `declare -r DOCKER_SYSTEM_PRUNE="${ORYX_DOCKER_SYSTEM_PRUNE:-true}"`

arroyc: true, but i didn't find any valid use case, do we use this variable during build?

gpcastro:
@arroyc should we close this PR, or is anything pending?

PR #170, opened 2019-05-25
Adding major-only images for Node

gpcastro:


gpcastro:
I update the base image tags once the build completes

images/runtime/node/4.4/Dockerfile.base
```diff
@@ -33,6 +33,10 @@ RUN curl -SLO "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE_VERSION-linux-
   && tar -xJf "node-v$NODE_VERSION-linux-x64.tar.xz" -C /usr/local --strip-components=1 \
   && rm "node-v$NODE_VERSION-linux-x64.tar.xz" SHASUMS256.txt.asc SHASUMS256.txt
 
+CMD [ "node" ]
+
+FROM node-original AS main
```
kichalla: motivation for this change? was it to have a clear separation of what the official image content was vs. what we added more?

images/runtime/node/6/Dockerfile.base
```diff
@@ -0,0 +1,88 @@
...
+  && tar -xzf yarn-v$YARN_VERSION.tar.gz -C /opt/ \
+  && ln -s /opt/yarn-v$YARN_VERSION/bin/yarn /usr/local/bin/yarn \
+  && ln -s /opt/yarn-v$YARN_VERSION/bin/yarnpkg /usr/local/bin/yarnpkg \
+  && rm yarn-v$YARN_VERSION.tar.gz.asc yarn-v$YARN_VERSION.tar.gz
+
+CMD [ "node" ]
+
+FROM node-original AS main
+
+RUN apt-get update && apt-get install unzip --yes
```
kichalla: we should ideally do this too `&& rm -rf /var/lib/apt/lists/*` as per docker guidelines to reduce space...we can handle it as a separate issue though

images/runtime/node/README.md
```diff
@@ -14,13 +14,11 @@ ones used in the official Node image for each particular version.
 # Generate images from template
 
 Since our node images are identical except for the base image, which determines the node version,
-some of our node images are generated from a template file, `./Dockerfile.template`, which contains the placeholder
-for the base image name. File `./nodeVersions.txt` has a list of the supported node versions as their tags for the official node images.
+our final node images are generated from a template file, `./Dockerfile.template`, which contains the placeholder
```
dorfire: isn't it wrong for some of the images?

gpcastro: not anymore, since we're rebuilding them all from scratch now.

images/runtime/node/4.4/Dockerfile.base
gpcastro: Yes, to separate the official image

images/runtime/node/6/Dockerfile.base
gpcastro: I've created a work item to investigate this in all our images

PR #173, opened 2019-05-29
Control Node LTS versions using constants file

gpcastro:


build/build-runtimeimages-bases.sh
```diff
@@ -88,7 +89,13 @@ for dockerFile in $dockerFiles; do
     cd $REPO_DIR
 
     echo
-    docker build -f $dockerFile -t $localImageTagName --build-arg CACHEBUST=$(date +%s) $labels . 
+    docker build -f $dockerFile \
+        -t $localImageTagName \
+        --build-arg CACHEBUST=$(date +%s) \
+        --build-arg NODE6_VERSION=$NODE6_VERSION \
```
kichalla: These feel out of place since this script is used to build runtime images of all platforms. Can you narrow it down to only the folders which deal with these versions?

gpcastro: The goal was to keep it simple, and the variables are specific to a platform + version won't cause confusion and will simply be ignored by the other platforms. By separating it, we either would have to split the script into one per platform, or create a logic to create the variables dynamically - adding unnecessary complexity IMO.

Let me know what you think

kichalla: Sure, sound good

PR #174, opened 2019-05-29
Build build pack only on non-Windows platforms

kichalla:


build/build-buildimages.sh
```diff
@@ -81,7 +81,10 @@ touch $ACR_BUILD_IMAGES_ARTIFACTS_FILE
 > $ACR_BUILD_IMAGES_ARTIFACTS_FILE
 
 # Build buildpack images
-source $REPO_DIR/build/build-buildpacks-images.sh
+# 'pack' is not supported on Windows
+if [[ "$OSTYPE" == "linux-gnu" ]] || [[ "$OSTYPE" == "darwin"* ]]; then
```
kichalla: @dorfire I ran into this while trying to build the build image on my windows machine

dorfire: the only problem is with `pack create-builder` - it isn't implemented on Windows.
did you get an error from a different part?

kichalla: Yeah, the error I got was `ERROR: `pack create-builder` is not implemented on Windows.`

build/build-buildimages.sh
```diff
@@ -81,7 +81,13 @@ touch $ACR_BUILD_IMAGES_ARTIFACTS_FILE
...
+	source $REPO_DIR/build/build-buildpacks-images.sh
+else
+	echo
+	echo "Skipping building 'build packs' as this platform '$OSTYPE' is not supported."
```
dorfire: well then, if we add an OS gate to `build-buildimages.sh` -
shouldn't we remove the existing OS gate from `install-pack.sh`?
And only leave the `Could not detect compatible pack binary platform` condition?

build/build-buildimages.sh
dorfire: well then, if we add an OS gate to `build-buildimages.sh` -
shouldn't we remove the existing OS gate from `install-pack.sh`?
And only leave the Could not detect compatible pack binary platform condition?

kichalla: But you can still run `build-buildpacks-images.sh` by itself though right, in which case you might still want to keep the OS gate in the other file

dorfire: `install-pack.sh` doesn't need to check if it's running on Windows. It only needs to check if it's on Linux or Mac, so it can download the correct binary.
Thus, I'm suggesting to remove the special Windows treatment for it, and leave it only in `build-buildpacks-images.sh`

kichalla: Updated.

build/build-buildimages.sh
```diff
@@ -81,7 +81,13 @@ touch $ACR_BUILD_IMAGES_ARTIFACTS_FILE
 > $ACR_BUILD_IMAGES_ARTIFACTS_FILE
 
 # Build buildpack images
-source $REPO_DIR/build/build-buildpacks-images.sh
+# 'pack' is not supported on Windows
```
dorfire: this is incorrect

build/build-buildimages.sh
```diff
@@ -81,7 +81,13 @@ touch $ACR_BUILD_IMAGES_ARTIFACTS_FILE
...
+if [[ "$OSTYPE" == "linux-gnu" ]] || [[ "$OSTYPE" == "darwin"* ]]; then
+	source $REPO_DIR/build/build-buildpacks-images.sh
+else
+	echo
+	echo "Skipping building 'build packs' as platform '$OSTYPE' is not supported."
```
dorfire: `'build packs'` => `Buildpacks images` (with a capital B, as this is a name)

images/pack-builder/install-pack.sh
```diff
@@ -6,18 +6,17 @@
...
-	exit 1
-else
-	echo 'ERROR: Could not detect compatible pack binary platform.'
-	exit 1
+packPlatform="$1"
+
+if [ -z "$packPlatform" ]; then
+	echo "Platform type cannot be empty."
+	echo "Usage		: install-pack.sh <platform-name>"
+	echo "Example	: install-pack.sh linux"
```
dorfire: indentation seems a bit off?

images/pack-builder/install-pack.sh
```diff
@@ -6,18 +6,17 @@
...
-	packPlatform='linux';
-elif [[ "$OSTYPE" == "darwin"* ]]; then
-	packPlatform='macos';
-elif [[ "$OSTYPE" == "cygwin" || "$OSTYPE" == "msys" ]]; then
-	echo 'ERROR: `pack create-builder` is not implemented on Windows.'
-	exit 1
-else
-	echo 'ERROR: Could not detect compatible pack binary platform.'
-	exit 1
+packPlatform="$1"
```
dorfire: why not rely on the Bash variable OSTYPE?

build/build-buildimages.sh
kichalla: You mean should it be `package create-builder`?

images/pack-builder/install-pack.sh
kichalla: We did not want to duplicate these checks?

build/build-buildimages.sh
kichalla: I have reverted my changes in all other files to keep it simple since I found that even though I need to have checks in OS type in `build-buildpacks-images` for quick exit, I also needed the same checks again in `install-pack.sh` (because I cannot use the OSTYPE to get the tar file i.e using `linux-gnu` vs `linux` when getting the tar file).

build/build-buildimages.sh
```diff
@@ -81,7 +81,13 @@ touch $ACR_BUILD_IMAGES_ARTIFACTS_FILE
 > $ACR_BUILD_IMAGES_ARTIFACTS_FILE
 
 # Build buildpack images
-source $REPO_DIR/build/build-buildpacks-images.sh
+# 'pack create-builder' is not supported on Windows
+if [[ "$OSTYPE" == "linux-gnu" ]] || [[ "$OSTYPE" == "darwin"* ]]; then
+	source $REPO_DIR/build/build-buildpacks-images.sh
+else
+	echo
+	echo "Skipping building 'Buildpacks images' as platform '$OSTYPE' is not supported."
```
dorfire: why the quotes around `Buildpacks images`?

PR #176, opened 2019-05-29
Write operation id to build manifest file instead of a separate file

kichalla:


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -121,21 +120,6 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
 
             console.WriteLine(buildInfo.ToString());
 
-            // Try writing the ID to a file in the source directory
```
kichalla: Just FYI...I am making this change because the file containing the operation id was being put in the source directory and not the destination directory. This breaks scenario where the source directory is not the same as the destination directory (ex: dotnetcore and also if a different directory is provided as output for node, python or php). In this PR we store this information in the manifest file which the runtime startup scripts read to get the operation id.

I have tested this E2E (using our test AI instance) for all platforms and the correlation happens end-to-end.

build/build-constants.yaml
```diff
@@ -81,15 +81,6 @@
     - type: shell
       directory: build
       file-name-prefix: __
-- name: file-paths
```
dorfire: why remove the whole section? we can use it to hold the value `oryx-manifest.toml`, as it is used in both C# and Go code

kichalla: Good point. Will change it.

tests/Oryx.Tests.Common/DockerVolume.cs
```diff
@@ -42,30 +41,35 @@ private DockerVolume(string originalHostDir, string mountedHostDir, string conta
         /// </summary>
         public string ContainerDir { get; }
 
-        public static DockerVolume Create([NotNull] string hostPath, [NotNull] string containerPath)
+        /// <summary>
+        /// Allows a container to use host's docker engine.
+        /// </summary>
+        /// <returns></returns>
+        public static DockerVolume CreateDockerSocket()
```
kichalla: @dorfire Probably I was not clear but I actually meant this when I mentioned in your earlier PR.

tests/Oryx.Tests.Common/DockerVolume.cs
```diff
@@ -42,30 +41,35 @@ private DockerVolume(string originalHostDir, string mountedHostDir, string conta
...
         }
 
         /// <summary>
         /// Creates a copy of a local directory, and returns a DockerVolume instance for mounting that copy in a
         /// container.
         /// </summary>
-        /// <param name="originalDir">local directory to be used in a container</param>
+        /// <param name="hostDir">local directory to be used in a container</param>
         /// <returns>DockerVolume instance that can be used to mount the new copy of `originalDir`.</returns>
-        public static DockerVolume CreateMirror(string originalDir)
```
kichalla: I reverted the earlier changes because I do not understand what a 'mirror' here is.

tests/Oryx.Tests.Common/DockerVolume.cs
dorfire: this name could be a little confusing - after all, no "DockerSockets" are being created here

tests/Oryx.Tests.Common/DockerVolume.cs
dorfire: a mirror is a directory that is a copy of another directory - that's what this method does:
1. copies source directory to a temporary location
2. mounts that temporary location to the destination path in the container

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
dorfire: We could have just moved the build ID file to the destination directory, no?

src/startupscriptgenerator/common/logging.go
```diff
@@ -32,12 +30,12 @@ type Logger struct {
...
-		rawId, err := ioutil.ReadFile(path.Join(appRootPath, consts.BuildIdFileName))
-		if err == nil { // Silently ignore errors
-			buildOpId = strings.TrimSpace(string(rawId))
-		}
+	buildManifest := GetBuildManifest(appRootPath)
+
+	if buildManifest.OperationId != "" {
+		buildOpId = strings.TrimSpace(buildManifest.OperationId)
 	}
+	fmt.Println("Operation Id: " + buildOpId)
```
dorfire: Id => ID

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
kichalla: Yes, that idea crossed my mind, but we already have a way to describe a built state with the manifest file so we need not have another one specifically for build id

src/startupscriptgenerator/node/scriptgenerator.go
```diff
@@ -48,7 +48,7 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 
 	logger.LogInformation("Generating script for source at '%s'", gen.SourcePath)
 
-	const oryxManifestFile string = "oryx-manifest.toml"
+	const oryxManifestFile string = consts.BuildManifestFileName
```
dorfire: why do we even need this other const, `oryxManifestFile`, if we can just use `consts.BuildManifestFileName`?

src/startupscriptgenerator/python/scriptgenerator.go
```diff
@@ -92,7 +93,7 @@ func logReadDirError(logger *common.Logger, path string, err error) {
 // Builds the commands to setup the Python packages, using virtual env or a package folder.
 func (gen *PythonStartupScriptGenerator) getPackageSetupCommand() string {
 	scriptBuilder := strings.Builder{}
-	const buildManifestFile string = "oryx-manifest.toml"
+	const buildManifestFile string = consts.BuildManifestFileName
```
dorfire: same here

tests/Oryx.BuildImage.Tests/DotNetCoreSampleAppsTest.cs
```diff
@@ -25,7 +25,7 @@ public DotNetCoreSampleAppsTest(ITestOutputHelper output) : base(output)
         }
 
         private DockerVolume CreateSampleAppVolume(string sampleAppName) =>
-            DockerVolume.CreateMirror(Path.Combine(_hostSamplesDir, "DotNetCore", sampleAppName));
+            DockerVolume.Create(Path.Combine(_hostSamplesDir, "DotNetCore", sampleAppName));
```
dorfire: Can you introduce this change in a separate PR, so we can clear out all these changed files from this PR?

tests/Oryx.Tests.Common/DockerVolume.cs
kichalla: Any suggestions? I initially wanted to say something like `MountDockerSocket` but that felt incorrect as you are not really mounting it at the point you might be calling this method.

tests/Oryx.Tests.Common/DockerVolume.cs
kichalla: Sure I get it but it doesn't feel something like a term commonly used in docker lingo, agree?

tests/Oryx.Tests.Common/DockerVolume.cs
dorfire: Maybe just add a static member `DockerVolume.DockerDaemonSocket`?

kichalla: Sounds good 👍

tests/Oryx.BuildImage.Tests/DotNetCoreSampleAppsTest.cs
kichalla: Sure, sounds good

src/startupscriptgenerator/node/scriptgenerator.go
kichalla: yeah, i know...i wanted to minimize the changes...will change it now..

tests/Oryx.Tests.Common/DockerVolume.cs
dorfire: It might not be commonly used, but it's OK to introduce terms relevant only to our realm;
we did it with Platform, with Snippet, with Checker, etc.

PR #178, opened 2019-05-30
Fixed SqlServerIntegrationTests to use a sql server database on Azure

kichalla:


tests/Oryx.Integration.Tests/SqlServerIntegrationTests.cs
```diff
@@ -3,50 +3,176 @@
...
-    //            Path.Combine(HostSamplesDir, "php", "sqlsrv-example"),
-    //            8080,
-    //            specifyBindPortFlag: false);
-    //    }
-    //}
+    [Trait("category", "db")]
+    [Trait("db", "sqlserver")]
+    public class SqlServerIntegrationTests : PlatformEndToEndTestsBase
+    {
+        private const string DbServerHostnameEnvVarName = "SQLSERVER_DATABASE_HOST";
```
kichalla: I am going to set these values on the build definitions to avoid storing sensitive data

tests/Oryx.Integration.Tests/SqlServerIntegrationTests.cs
```diff
@@ -3,50 +3,176 @@
...
+                ContainerPort,
+                "/bin/bash",
+                new[]
+                {
+                    "-c",
+                    script
+                },
+                async (hostPort) =>
+                {
+                    var data = await _httpClient.GetStringAsync($"http://localhost:{hostPort}/");
```
kichalla: One hiccup that I ran into was firewalls on Azure do not allow requests from random IP addresses to connect to the database server. For now, I gave a range of IP addresses which essentially is like having no firewall, but ideally we should have a restricted range.

tests/Oryx.Integration.Tests/SqlServerIntegrationTests.cs
dorfire: can you store them in a Build Variable Group? this way, it can be shared among different build definitions, can't it?

kichalla: I haven't checked that yet, but will do now.

kichalla: Seems like Elena had created a variable group called 'Oryx' for our team already, we just need to sync it up with these new variables. Working on it...

vsts/pipelines/templates/_buildParallel.yml
```diff
@@ -109,6 +109,11 @@ steps:
   inputs:
     scriptPath: ./build/test-integration.sh
     args: $(TestIntegrationCaseFilter)
+  env:
```
kichalla: This works, but this is not ideal as we do not want to expose the secrets as plain text in environment variables (it defeats the purpose of holding secrets in keyvault). Seems like the recommended way to access keyvault is through Manage Service Identity which does not work for our scenario of using them in tests. Are we OK with this risk vs. the earlier idea that I had about running a dedicated sql server container in the build machine? cc gpcastro

tests/Oryx.Integration.Tests/Fixtures/SqlServerDbContainerFixture.cs
```diff
@@ -1,94 +0,0 @@
...
-
-using System;
-using System.Collections.Generic;
-using Microsoft.Oryx.Tests.Common;
-using Polly;
-using Xunit;
-
-namespace Microsoft.Oryx.Integration.Tests.Fixtures
-{
-    public class SqlServerDbContainerFixture : DbContainerFixtureBase
```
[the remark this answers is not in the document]
kichalla: I am not convinced that we need to have this kind of support. It is a one time setup and if anyone does the changes incorrectly the tests would fail and we go and fix it on Azure and forget it. As of now I do not think we will see the need to modify the data anytime soon.

tests/Oryx.Integration.Tests/DatabaseTestsBase.cs
```diff
@@ -68,7 +69,7 @@ protected DatabaseTestsBase(ITestOutputHelper outputHelper, Fixtures.DbContainer
                 {
                     var data = await _httpClient.GetStringAsync($"http://localhost:{hostPort}/");
                     Assert.Equal(
-                        _dbFixture.GetSampleDataAsJson(),
+                        DbContainerFixtureBase.GetSampleDataAsJson(),
```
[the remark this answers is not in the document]
kichalla: Because this test class does not use the fixture now

dorfire:
Do we still need the const `MicrosoftSQLServerImageName`?

kichalla:
> Do we still need the const MicrosoftSQLServerImageName?

Nope, we do not..removed it now...

build/test-integration.sh
```diff
@@ -9,11 +9,30 @@ set -e
 declare -r REPO_DIR=$( cd $( dirname "$0" ) && cd .. && pwd )
 source $REPO_DIR/build/__variables.sh
 
+# When this script is run in CI agent these environment variables are already set
+if [ -z "$SQLSERVER_DATABASE_HOST" ]; then
+    function getValueFromKeyVault() {
+        local secretName="$1"
+        getValueFromKeyVault_Result=`az.cmd keyvault secret show --name "$secretName" --vault-name "oryx" | grep value | awk '{print $NF}' | tr -d '"'`
+    }
```
arroyc: this is what we discussed in stand up, thanks

kichalla:
@dorfire let me know if you have any additional comments otherwise I think this is good to go

tests/Oryx.Integration.Tests/Fixtures/SqlServerDbContainerFixture.cs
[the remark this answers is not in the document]
dorfire: It's not about modifying the data - it's about "self healing".
Imagine we need to re-create that server in a few months - not too unlikely - wouldn't it be nice if we'll be able to just update the connection string in our Key Vault, and have the test data repopulated without any intervention?

PR #183, opened 2019-06-01
Fix integration tests (python & db) and handle adding duplicate OperationId property in case of multi-platform build

kichalla:


images/build/Dockerfile
```diff
@@ -277,9 +277,9 @@ RUN apt-get update \
 # https://github.com/docker-library/python/issues/147
 ENV PYTHONIOENCODING UTF-8
 COPY build/__python-versions.sh /tmp
-COPY --from=mcr.microsoft.com/oryx/python-build-base:2.7-20190530.1 /opt /opt
-COPY --from=mcr.microsoft.com/oryx/python-build-base:3.6-20190530.1 /opt /opt
-COPY --from=mcr.microsoft.com/oryx/python-build-base:3.7-20190530.1 /opt /opt
+COPY --from=mcr.microsoft.com/oryx/python-build-2.7:20190402.1 /opt /opt
```
kichalla: @arryoc I reverted this change of yours because it is failing the Shapely integration tests. Please investigate why they fail on moving to the new tags.

kichalla: Ignore my comment. Seems like the error is happening in runtime image of Python

kichalla: @gpcastro FYI..this could be a regression from your changes to Python runtime images having similar structure as the build image. Could you investigate it?

arroyc: @kichalla I think we need to add the virtualenv command back to this test … commit beb4961a
I will send a PR, please take a look

arroyc: @kichalla I think we need to remove this test ( as I see another test for shapely is running correctly). Another thing  I noticed the path in run.sh is wrong export PYTHONPATH=$PYTHONPATH:"$virtualEnvDir/lib/python$PYTHON_VERSION/site-packages", it should be export PYTHONPATH=$PYTHONPATH:"$virtualEnvDir/opt/python$PYTHON_VERSION/site-packages"
looking more into it

arroyc: as per the current logic in the python script generator, we assume virtualenv is used whenever there is manifest file, but when we use packagedir we are not using virtualenv though manifest file is still getting generated which has operation_id. so it's throwing error. When I deleted manifest file, it worked for me. This seems to be a bug

arroyc: bug 903955 is already created it seems

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -321,9 +321,10 @@ private void LogScriptIfGiven(string type, string scriptPath)
                 .Where(s => s.BuildProperties != null)
                 .SelectMany(s => s.BuildProperties)
                 .ToDictionary(p => p.Key, p => p.Value);
+            buildProperties[ManifestFilePropertyKeys.OperationId] = context.OperationId;
```
dorfire: the `BuildScriptGeneratorContext` already has a member `IDictionary<string, string> Properties`;
so why do we need the new `BuildScriptGeneratorContext.OperationId`?

images/build/Dockerfile
kichalla: The bug is with Python startup script IMO as it needs to check if virtual environment was used or not correctly. I am fixing it and will send out an update.

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
kichalla: The thing is those Properties are user supplied i.e coming from `-p` supplied at command line. We currently do not just use those properties directly as properties that go into manifest file (i think that makes sense as we might want to write or do something differently). I think the names are getting us confused. I would rather change this line of code to say `manifestFileProperties` instead.

src/startupscriptgenerator/python/scriptgenerator.go
```diff
@@ -93,75 +94,84 @@ func logReadDirError(logger *common.Logger, path string, err error) {
...
-	return scriptBuilder.String()
-}
-
-func getVirtualEnvironmentCommand() string {
-	scriptBuilder := strings.Builder{}
-	scriptBuilder.WriteString("PYTHON_VERSION=$(python -c \"import sys; print(str(sys.version_info.major) + '.' + str(sys.version_info.minor))\")\n")
-	scriptBuilder.WriteString("echo \"Using packages from virtual environment '$virtualEnvName' located at '$virtualEnvDir'.\"\n")
-	virtualEnvFolder := "\"$virtualEnvDir/lib/python$PYTHON_VERSION/site-packages\""
-	scriptBuilder.WriteString("export PYTHONPATH=$PYTHONPATH:" + virtualEnvFolder + "\n")
+	scriptBuilder.WriteString("sourcePath=\"" + gen.SourcePath + "\"\n")
```
gpcastro: why are you changing the entire block to bash? The final script should be as simple as possible.

images/build/Dockerfile
gpcastro: Can you explain what the issue is? there's too many changes it's hard to find out.

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -67,26 +68,31 @@ public LanguageDetectorResult Detect(ISourceRepo sourceRepo)
...
         {
-            var buildProperties = new Dictionary<string, string>();
-            buildProperties[ManifestFilePropertyKeys.OperationId] = context.OperationId;
+            var manifestFileProperties = new Dictionary<string, string>();
 
             var packageDir = GetPackageDirectory(context);
             var virtualEnvName = GetVirtualEnvironmentName(context);
-            if (string.IsNullOrEmpty(packageDir))
+
+            if (!string.IsNullOrWhiteSpace(packageDir) && !string.IsNullOrWhiteSpace(virtualEnvName))
```
gpcastro: if neither is provided, we default to virtual env with its default name

kichalla: Right, it does not change that(line 83). I just rearranged to code to make it more readable.

src/startupscriptgenerator/python/scriptgenerator.go
kichalla: I changed it to make it more readable as before this we shift between golang and bash which I felt a bit hard to follow. the code structure that I have here is similar to what I did earlier for dotnetcore. The declarations of these variable can be viewed as inputs to a bash script which follows these declarations.
I also ran nightly build against this PR to make sure I did not break any tests.

tests/Oryx.Integration.Tests/PythonEndToEndTests.cs
```diff
@@ -83,7 +128,7 @@ public async Task CanBuildAndRun_Python27App_UsingVirtualEnv()
                 .AddCommand("pip install gunicorn")
                 .AddCommand("pip install flask")
                 .AddCommand($"cd {appDir}")
-                .AddCommand($"oryx -appPath {appDir} -bindPort {ContainerPort} -virtualEnvName={virtualEnvName}")
```
kichalla: Supplying virtual env is not required as this info is present in manifest file

images/build/Dockerfile
kichalla: The issue was that Python startup script was only considering virtualenv if a manifest is just present on disk. We had few tests which use package directory which were failing due to it. This issue came to light because of my change related to OperationId. Since we always write OperationId to manifest file now, we are going to have this file created always which caused this issue to pop-up.

src/startupscriptgenerator/python/scriptgenerator.go
```diff
@@ -39,15 +38,16 @@ func (gen *PythonStartupScriptGenerator) GenerateEntrypointScript() string {
 	scriptBuilder.WriteString("cd " + gen.SourcePath + "\n\n")
 
 	common.SetEnvironmentVariableInScript(&scriptBuilder, "PORT", gen.BindPort, DefaultBindPort)
-	packageSetupBlock := gen.getPackageSetupCommand()
+	buildManifest := common.GetBuildManifest(gen.SourcePath)
```
kichalla: @gpcastro I refactored this based on our offline discussion. Locally all integration tests passed, so I think my changes are good. I also kicked off a Nightly build now.

kichalla:
@gpcastro @dorfire I have updated the PR. Can you guys take a look? Nightly build also ran successfully.

src/startupscriptgenerator/python/main.go
```diff
@@ -18,7 +18,7 @@ func main() {
 	defaultAppFilePathPtr := flag.String("defaultApp", "", "[Optional] Path to a default file that will be executed if the entrypoint is not found. Ex: '/opt/defaultsite'")
 	defaultAppModulePtr := flag.String("defaultAppModule", "application:app", "Module of the default application, e.g. 'application:app'.")
 	virtualEnvironmentNamePtr := flag.String("virtualEnvName", "", "Name of the virtual environment for the app")
-	packagesFolderPtr := flag.String("packagedir", "__oryx_packages__", "Directory where the python packages were installed, if no virtual environment was used.")
+	packagesFolderPtr := flag.String("packagedir", "", "Directory where the python packages were installed, if no virtual environment was used.")
```
kichalla: @gpcastro I think a default value here is not required, take a look at my other .go file related to this

src/startupscriptgenerator/python/scriptgenerator.go
```diff
@@ -91,68 +91,85 @@ func logReadDirError(logger *common.Logger, path string, err error) {
...
+				// We add the virtual env site-packages to PYTHONPATH instead of activating it to be backwards compatible with existing
+				// app service implementation. If we activate the virtual env directly things don't work since it has hardcoded references to
+				// python libraries including the absolute path. Since Python is installed in different paths in build and runtime images,
+				// the libraries are not found.
+				scriptBuilder.WriteString("virtualEnvDir=\"" + virtualEnvDir + "\"\n")
+				virtualEnvCommand := getVirtualEnvironmentCommand()
+				scriptBuilder.WriteString(virtualEnvCommand)
+
+			} else {
+				packageDirName = "__oryx_packages__"
```
kichalla: @gpcastro I will add a comment here but basically this is the place where we set the packagedir to the old default package directory and line 156 gets evaluated for this case.

tests/Oryx.Integration.Tests/PythonEndToEndTests.cs
```diff
@@ -3,61 +3,81 @@
 // Licensed under the MIT license.
 // --------------------------------------------------------------------------------------------
 
+using Microsoft.Oryx.BuildScriptGenerator.Python;
```
kichalla: Seems like my VS settings messed the order here (should have been System usings before Microsoft ones). Will fix it.

tests/Oryx.Integration.Tests/PythonEndToEndTests.cs
```diff
@@ -3,61 +3,81 @@
...
 using System.Threading.Tasks;
-using Microsoft.Oryx.Tests.Common;
 using Xunit;
 using Xunit.Abstractions;
 
 namespace Microsoft.Oryx.Integration.Tests
 {
-    [Trait("category", "python")]
-    public class PythonEndToEndTests : PlatformEndToEndTestsBase
+    public abstract class PythonEndToEndTestsBase : PlatformEndToEndTestsBase
```
kichalla: I just re-arranged the tests into different classes to enable to run them in parallel and was able to cut down the time from 30 to 11 minutes on my machine. I needed this to get quick feedback on my changes.

tests/Oryx.Integration.Tests/PythonEndToEndTests.cs
```diff
@@ -3,61 +3,81 @@
...
+        public PythonEndToEndTestsBase(ITestOutputHelper output, TestTempDirTestFixture testTempDirTestFixture)
         {
             _output = output;
             _hostSamplesDir = Path.Combine(Directory.GetCurrentDirectory(), "SampleApps");
             _tempRootDir = testTempDirTestFixture.RootDirPath;
         }
+    }
+
+    [Trait("category", "python")]
+    public class PythonEndToEndTests_BackCompat : PythonEndToEndTestsBase
```
kichalla: @gpcastro you would be interested in tests of this class

tests/Oryx.Integration.Tests/PythonEndToEndTests.cs
```diff
@@ -3,61 +3,81 @@
...
+            var appName = "flask-app";
             var hostDir = Path.Combine(_hostSamplesDir, "python", appName);
             var volume = DockerVolume.CreateMirror(hostDir);
             var appDir = volume.ContainerDir;
-            var startupFile = "/tmp/startup.sh";
-            var script = new ShellScriptBuilder()
+            var virtualEnvName = "antenv";
+            var buildScript = new ShellScriptBuilder()
+                // App should run fine even with manifest file not present
+                .AddCommand($"oryx build {appDir} -p packagedir={PythonConstants.DefaultTargetPackageDirectory}")
```
kichalla: @gpcastro This test simulates our yesterday's talk. This app is built with package directory and when running a virtual environment name is always provided at line 64

src/startupscriptgenerator/python/scriptgenerator.go
```diff
@@ -162,21 +179,22 @@ func getVirtualEnvironmentCommand() string {
...
 	return scriptBuilder.String()
 }
 
 // Checks if the app is based on Django, and returns a startup command if so.
-func (gen *PythonStartupScriptGenerator) getDjangoStartupModule() string {
+func (gen *PythonStartupScriptGenerator) getDjangoStartupModule(buildManifest common.BuildManifest) string {
 	logger := common.GetLogger("python.scriptgenerator.getDjangoStartupModule")
 	defer logger.Shutdown()
-
+
```
gpcastro: nit: extra tab

kichalla: Yeah, I will fix in a follow up PR.

src/startupscriptgenerator/python/scriptgenerator.go
dorfire: This value repeats in both a C# project and this Go project - maybe consolidate them in the YAML? (not necessarily in this PR)

kichalla: yes, i have a bug filed earlier for this which will handle this scenario

tests/Oryx.Integration.Tests/PythonEndToEndTests.cs
dorfire: Let's not use underscores in class names

PR #184, opened 2019-06-03
Added a buildpack test with the Heroku builder

dorfire:


tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
```diff
@@ -253,8 +253,10 @@ public async Task CanBuildAndRun_NodeApp(string nodeVersion)
                 });
         }
 
-        [Fact]
-        public async Task CanBuildAndRun_NodeApp_WithBuildpack()
+        [Theory]
+        [InlineData("oryxdevms/pack-builder")]
+        [InlineData("heroku/buildpacks")]
```
kichalla: What does testing heroku buildpacks mean here? and why are we having a test for it?

tests/SampleApps/nodejs/webfrontend/server.js
```diff
@@ -10,7 +10,7 @@ app.get('/api', function (req, res) {
     res.send('Hello from webfrontend');
 });
 
-var port = process.env.PORT || 80;
+var port = process.env.PORT || 8080;
```
kichalla: Curious, why this change? Could you paste the contents of `docker inspect` of the final runtime image that the build package produces for this test?

tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
dorfire: I want to make sure our `pack` runner image is compatible with other buildpacks, not just the Oryx buildpack that we maintain.

tests/SampleApps/nodejs/webfrontend/server.js
dorfire: this change is because Heroku's generated launch script doesn't set the PORT env var, so **this app would start on port 80, and fail** (_Access Denied_)

tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
kichalla: Nice!

PR #186, opened 2019-06-04
Added method EqualsIgnoreCase

dorfire:


src/Common/Extensions/StringExtensions.cs
```diff
@@ -17,6 +17,11 @@ public static class StringExtensions
 
         private static readonly string UrlUserInfoReplacement = "***";
 
+        public static bool EqualsIgnoreCase(string a, string b)
+        {
+            return string.Equals(a, b, StringComparison.OrdinalIgnoreCase);
```
dorfire: Used OrdinalIgnoreCase over a culture-aware comparison for performance.

src/Common/Extensions/StringExtensions.cs
```diff
@@ -17,6 +17,11 @@ public static class StringExtensions
 
         private static readonly string UrlUserInfoReplacement = "***";
 
+        public static bool EqualsIgnoreCase(string a, string b)
```
kichalla: Instead of `StringExtensions.EqualsIgnoreCase(a, b)`, we can support `a.EqualsIgnoreCase(b)`...it is more concise

Also a subtle thing to note here: `a.Equals(b)` throws NRE if `a` is null, but we need not check if `a` is null in our extension case and delegate that responsibility to `string.Equals` method which correctly handles the equality here

dorfire: but then we'd need to check for `null` before using it.. the way it's implemented now, it behaves exactly like `string.Equals`

tests/Oryx.Common.Tests/StringExtensionsTest.cs
```diff
@@ -9,6 +9,16 @@ namespace Microsoft.Oryx.Common.Extensions
 {
     public class StringExtensionsTest
     {
+        [Fact]
+        public void EqualsIgnoreCase_Sanity()
```
dorfire: @arroyc

src/Common/Extensions/StringExtensions.cs
kichalla: I mean we need _not_ check for null. the current usage feels verbose and not I was originally expecting. @gpcastro what do you think?

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -143,6 +144,71 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
             };
         }
 
+<<<<<<< HEAD
```
kichalla: bad merge

PR #191, opened 2019-06-04
publish vsts scripts in build templates for base images as well

arroyc:


dorfire:
why do we need this?

arroyc:
> why do we need this?

so that we can access the scripts from release artifacts

PR #193, opened 2019-06-04
Added buildpacks tests

dorfire:


tests/Oryx.Tests.Common/EndToEndTestHelper.cs
```diff
@@ -264,6 +266,40 @@ public static class EndToEndTestHelper
             }
         }
 
+        public static Task RunPackAndAssertAppAsync(
+            ITestOutputHelper output,
+            string appName,
+            DockerVolume appVolume,
+            string appImageName,
```
kichalla: Is this the name of the final runtime image that gets built as part of this build pack process?

tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
```diff
@@ -254,35 +254,18 @@ public async Task CanBuildAndRun_NodeApp(string nodeVersion)
         }
 
         [Theory]
-        [InlineData("oryxdevms/pack-builder")]
-        [InlineData("heroku/buildpacks")]
+        [InlineData(Constants.OryxBuildpackBuilderImageName)]
```
kichalla: Dotnetcore and Python end-to-end tests?

tests/Oryx.Tests.Common/EndToEndTestHelper.cs
kichalla: To prevent tests running in parallel trumping on same names, should we generate unique names for these images?

build/build-constants.yaml
```diff
@@ -85,6 +85,7 @@
 - name: file-paths
   constants:
     build-manifest-file-name: oryx-manifest.toml
+    run-script-generator-dir: /opt/startupcmdgen
```
kichalla: Change this PR's base branch to the branch you created for PR # 194...this way you can only see test changes in this PR

tests/Oryx.Tests.Common/EndToEndTestHelper.cs
kichalla: Think about untagging these images too as otherwise they would consume disk space in view of you might be writing more tests in this area. You could follow a naming pattern (like we do for container names) like `oryxtests_*` for these images and can untag them in existing `dockerCleanup.sh` script

build/build-constants.yaml
dorfire: Done

tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
dorfire: They're both dependent on AB#896178

tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
```diff
@@ -255,35 +255,18 @@ public async Task CanBuildAndRun_NodeApp(string nodeVersion)
...
+        [InlineData(Constants.HerokuBuildpackBuilderImageName)]
         public async Task CanBuildAndRun_NodeApp_WithBuildpack(string builder)
```
arroyc: move these two tests (CanBuildAndRun_NodeApp_WithBuildpack and TwigExample_WithBuildpack ) under a new class something like 
```
public class NodeBuildPackTests: NodeEndToEndTestsBase
{
}
```

Node integration tests takes very long, this way we can optimize a little

tests/Oryx.Integration.Tests/PythonEndToEndTests.cs
```diff
@@ -272,6 +272,24 @@ public async Task CanBuildAndRunPythonApp_UsingPython37()
                 });
         }
 
+        [Fact(Skip = "AB#896178")]
+        public async Task CanBuildAndRunPythonApp_WithBuildpack()
```
arroyc: though php integration tests are fastest today, but it will be nice to follow the same pattern as mentioned above, i.e. move these buildpack tests under a new class

images/build/Dockerfile
```diff
@@ -340,6 +340,7 @@ RUN ./build.sh node       /opt/startupcmdgen/nodejs
 RUN ./build.sh php        /opt/startupcmdgen/php
 RUN ./build.sh python     /opt/startupcmdgen/python
 
+
```
arroyc: nit: extra line/space?

vsts/scripts/dockerCleanup.sh
```diff
@@ -48,6 +48,7 @@ docker images
 #	b. oryxdevms/node-<major.minor>:latest
 #	b. oryxdevms/dotnetcore-<major.minor>:latest
 # - We should untag these images only after they have been pushed to a remote repository.
+UntagImages "test-*"
```
arroyc: should we also get rid of heroku images? how big are they?

arroyc:
also run a nightly build from your branch before merging it ...

vsts/scripts/dockerCleanup.sh
dorfire: we're gonna need them for every test run, so performance will be better if we don't get rid of them

tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
arroyc: yeah, I meant TwigExample_WithBuildpack  to move to a new class in php ernd to end tests

PR #195, opened 2019-06-05
Code clean-up in tests

kichalla:


tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
```diff
@@ -120,12 +119,10 @@ public async Task CanBuildAndRunNodeApp_OnSecondBuild_AfterZippingNodeModules_In
             var appName = "webfrontend";
             var volume = CreateAppVolume(appName);
             var appDir = volume.ContainerDir;
-            var script = new ShellScriptBuilder()
-                .AddCommand($"cd {appDir}")
```
kichalla: Many of our tests were doing this, which should not be required and if the tests fail then it would be a bug. I will kick off a nightly build against this PR branch now.

PR #196, opened 2019-06-05
.NET Core: Warn if multiple 'runtimeconfig.json' files are present and fix user startup command logic

kichalla:


src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -36,7 +36,7 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
 	defaultAppFileDir := filepath.Dir(gen.DefaultAppFilePath)
 
 	scriptBuilder.WriteString("readonly appPath=\"" + gen.RunFromPath + "\"\n")
-	scriptBuilder.WriteString("userStartUpCommand=\"" + gen.UserStartupCommand + "\"\n")
+	scriptBuilder.WriteString("userStartUpCommand=(" + gen.UserStartupCommand + ")\n")
```
kichalla: @gpcastro I found this bug while writing tests for the mutli-runtimeconfig.json files. Basically a startup command like `dotnet foo.dll` does not work currently 😞 as it is being considered as a single value(so the above change is to make as an array of values instead) and we check if a file with that name exists or not. 
A workaround for this would be to create a script file(which does `dotnet foo.dll` in it and drop it in the app output folder and use this script as a startup command.
I had tests for the user startup command scenario but we have a fallback logic where we try to figure out the command (using runtimeconfig.json hueristics) and so the tests were always able to find the file successfully.

tests/Oryx.Integration.Tests/DotNetCoreEndToEndTests.cs
```diff
@@ -1173,17 +1279,16 @@ public async Task CanRun_NetCore30App_PublishedOnMacMachine_ButRunOnNetCore30Run
                 new DockerCli());
         }
 
-        [Theory]
-        [InlineData("NetCoreApp30.WebApp")]
```
kichalla: I removed support for making this scenario work as users could always do `./NetCoreApp30.WebApp` and supporting this feature was making the script more complicated

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -53,43 +53,56 @@ isLinuxExecutable() {
...
-		  echo "Could not find the file '${userStartUpCommand[1]}' on disk."
+			echo "WARNING: Error in user provided startup command '${userStartUpCommand[@]}'."
+			echo "WARNING: Could not find the file '${userStartUpCommand[1]}' on disk."
 		fi
 	else
-		startUpCommand="$userStartUpCommand"
+		startUpCommand="${userStartUpCommand[@]}"
+		echo "Using the startup command '$startUpCommand' provided by the user to start the app."
 	fi
 fi
```
kichalla: TODO: this should be fixed so that we directly go to using the default app file.

tests/Oryx.Integration.Tests/DotNetCoreEndToEndTests.cs
gpcastro: we solve this for the other languages by using `common.ExtendPathForCommand` from our common library.
It would be great if the .NET startup gen followed the pattern of the other languages and had its code in Go, so we could benefit from those things we already have in Go

src/startupscriptgenerator/common/buildManifestHelper.go
```diff
@@ -6,42 +6,43 @@
 package common
 
 import (
-    "startupscriptgenerator/common/consts"
-    "github.com/BurntSushi/toml"
-    "log"
-    "path/filepath"
+	"log"
```
kichalla: not sure why this got changed, but you can ignore viewing white space changes in Github if it is too distracting

tests/Oryx.Integration.Tests/DotNetCoreEndToEndTests.cs
kichalla: As discussed offline, I will work on making majority of the dotnetcore startupscript generation in Golang code and will use the functionality you mention at that time.

src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
```diff
@@ -93,6 +93,7 @@ fi
 mkdir -p "$DESTINATION_DIR"
 
 {{ if PreBuildCommand | IsNotBlank }}
+echo "Running the pre-build script..."
```
dorfire: Shouldn't this be `PreBuildCommandPrologue`?

src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
```diff
@@ -113,12 +114,16 @@ dotnet restore "{{ ProjectFile }}"
 	publishToDirectory "$tmpDestinationPublishDir"
 
 	{{ if PostBuildCommand | IsNotBlank }}
+	echo
+	echo "Running the post-build script..."
```
dorfire: Same here - should it print the value of `PostBuildCommandPrologue`?

src/startupscriptgenerator/dotnetcore/scriptgenerator.go
```diff
@@ -53,43 +53,57 @@ isLinuxExecutable() {
...
 
+	fileCount=${#runtimeConfigJsonFiles[@]}
+	if [ $fileCount -eq 1 ]; then
+		file=${runtimeConfigJsonFiles[0]}
+		startupDllFileNamePrefix=${file%%.runtimeconfig.json}
+		startupExecutableFileName="$startupDllFileNamePrefix"
+		startupDllFileName="$startupDllFileNamePrefix.dll"
+	else
+		echo "WARNING: Unable to find the startup dll file name."
+		echo "WARNING: Expected to find only one file with extension 'runtimeconfig.json' but found $fileCount"
```
dorfire: This could be fit in an `IChecker` implementation - your call

PR #198, opened 2019-06-07
.NET Core: Save app startup dll file name in manifest file

kichalla:


src/BuildScriptGenerator/DotNetCore/DotNetCoreManifestFilePropertyKeys.cs
```diff
@@ -0,0 +1,12 @@
+﻿// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+namespace Microsoft.Oryx.BuildScriptGenerator.DotNetCore
+{
+    internal static class DotNetCoreManifestFilePropertyKeys
+    {
+        internal const string StartupDllFileName = nameof(StartupDllFileName);
```
dorfire: our prop key names are spread across multiple files - maybe we should (at least aim to) consolidate them?

tests/Oryx.Integration.Tests/DotNetCoreEndToEndTests.cs
```diff
@@ -920,6 +921,55 @@ public async Task CanRunApp_UsingDefaultApp_WhenStartupCommand_IsNotValid(string
                     Assert.Equal("Running default web app", data);
                 });
         }
+
+        [Fact]
+        public async Task CanRunCorrectApp_WhenOutputHasMultipleRuntimeConfigJsonFiles_DueToProjectFileRenaming()
```
dorfire: 👍

src/BuildScriptGenerator/DotNetCore/DotNetCoreManifestFilePropertyKeys.cs
kichalla: These property keys are specific to DotNetCore, so I put them in a separate file. I think we should do similarly for other platforms. Yes, keys which are common across all platforms (ex: OperationID) can be put in a common file. Do you agree?

src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
```diff
@@ -93,6 +93,7 @@ fi
 mkdir -p "$DESTINATION_DIR"
 
 {{ if PreBuildCommand | IsNotBlank }}
+echo "Running the pre-build script..."
```
kichalla: You can ignore these since these will use the common constants file using the prologue and epilogue

PR #200, opened 2019-06-07
Rename members

dorfire:


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -48,13 +48,13 @@ internal class BuildCommand : CommandBase
             OptionTemplates.Platform,
             CommandOptionType.SingleValue,
             Description = "The name of the programming language being used in the provided source directory.")]
```
kichalla: can you update the descriptions too?

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -48,13 +48,13 @@ internal class BuildCommand : CommandBase
...
-        public string Language { get; set; }
+        public string PlatformName { get; set; }
 
         [Option(
             OptionTemplates.PlatformVersion,
             CommandOptionType.SingleValue,
             Description = "The version of programming language being used in the provided source directory.")]
```
kichalla: same here

PR #205, opened 2019-06-12
Added support for Python 3.8

dorfire:


build/__python-versions.sh
```diff
@@ -1,6 +1,9 @@
 # This file was auto-generated from 'build-constants.yaml'. Changes may be overridden.
 
+PYTHON_BUILD_BASE_TAG='20190610.2'
+PYTHON_RUNTIME_BASE_TAG='20190402.1'
 PIP_VERSION='19.0.3'
 PYTHON27_VERSION='2.7.16'
 PYTHON36_VERSION='3.6.8'
 PYTHON37_VERSION='3.7.3'
+PYTHON38_VERSION='3.8.0b1'
```
kichalla: are you planning to add some E2E tests?

dorfire: added one for now. will add more in a separate PR.

build/build-constants.yaml
```diff
@@ -48,10 +48,13 @@
       file-name-prefix: __
 - name: python-versions
   constants:
+    python-build-base-tag: 20190610.2
+    python-runtime-base-tag: 20190610.2
```
gpcastro: we don't have a runtime-base for Python, everything comes from the same base (which is currently called "build", admittedly a misleading name"

PR #207, opened 2019-06-12
make release as part of multi stage build pipeline (oryx-ci)

arroyc:
goal:

1. make release as part of multi stage ci build pipeline
2. release scripts can be versioned from now on. this will ensure breaking changes can be rolled back in future.

todo:
1. Similar needs to done for base image release
2. Similar needs to be done for build pack image release


vsts/pipelines/templates/_releaseTemplate.yml
```diff
@@ -0,0 +1,120 @@
...
+  displayName: 'Download Build Artifacts'
+  inputs:
+    buildType: specific
+    project: '0bdbc590-a062-4c3f-b0f6-9383f67865ee'
+    pipeline: 11021
+    specificBuildWithTriggering: true
+    artifactName: drop
+
+- task: CopyFiles@2
+  displayName: 'Copy Files to: $(Build.ArtifactStagingDirectory)'
```
kichalla: Usually we copy files to staging directory so that when we really want to publish artifacts, we can copy the contents from the staging directory to the final directory, but here we do not do that, so do we still need this copy step? Can't we access the System.ArtifactsDirectory directly?

vsts/pipelines/ci.yml
```diff
@@ -4,75 +4,91 @@ resources:
...
-- job: Job_BuildImage
-  displayName: Build and Test Build Image
-  dependsOn: Job_SignBinaries
-  condition: succeeded()
-  pool:
-    name: OryxLinux
-  variables:
-    SignType: $[ dependencies.Job_SignBinaries.outputs['setSignTypeVariable.SignType'] ]  # map in the signtype variable
+    - job: Job_BuildImage
+      displayName: Build and Test Build Image
```
kichalla: There is something weird going on with these display names...can you investigate more?

image

vsts/pipelines/templates/_releaseTemplate.yml
```diff
@@ -0,0 +1,120 @@
...
+    enforceDockerNamingConvention: false
+  condition: and(succeeded(), eq(variables['ReleaseRuntimeImages'], 'true'))
+
+- task: ShellScript@2
+  displayName: 'Clean up Docker containers and images'
+  inputs:
+    scriptPath: ./vsts/scripts/dockerCleanup.sh
+  condition: true
+
+- task: ShellScript@2
```
kichalla: This task should not be required here as we are not running any integration tests

vsts/pipelines/templates/_releaseTemplate.yml
kichalla: Also I see that there are 2 `drop` folders now...I am wondering if the second one got created because you are copying files to staging directory here and hence it is automatically copied to the final directory?

image

vsts/pipelines/templates/_releaseTemplate.yml
```diff
@@ -0,0 +1,120 @@
...
+  displayName: 'Clean up Docker mounted directories'
+  inputs:
+    scriptPath: ./vsts/scripts/dockerArtifactsRemoval.sh
```
kichalla: Since you are now using the scripts folder from the sources itself, we should no longer require to copy the contents of the scripts to the artifacts directory. I still see these:

image

vsts/pipelines/ci.yml
```diff
@@ -4,75 +4,91 @@ resources:
...
+          echo "##vso[task.setvariable variable=PushRuntimeImages;]true"
+          echo "##vso[task.setvariable variable=PushBuildImages;]false"
+          echo "##vso[task.setvariable variable=EmbedBuildContextInImages;]true"
+        displayName: 'Set variables'
+
+      - template: templates/_buildTemplate.yml
+        parameters:
+          pushToDockerHub: 'true'
+
+    - template: templates/_integration.yml
```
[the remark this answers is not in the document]
kichalla: Also something is wrong with these integration tests, I do not see the tests results file here...you can file a bug and visit this later:

image

vsts/pipelines/templates/_releaseTemplate.yml
```diff
@@ -0,0 +1,120 @@
...
+  displayName: Container registry login
+  inputs:
+    command: login
+    azureSubscriptionEndpoint: ${{ parameters.ascName }}
+    azureContainerRegistry: ${{ parameters.acrName }}
+
+- task: DownloadBuildArtifacts@0
+  displayName: 'Download Build Artifacts'
+  inputs:
+    buildType: specific
```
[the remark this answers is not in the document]
kichalla: From this snapshot, it talks about `Latest` which is a bit worrying...we want the current build's artifacts:

image

vsts/pipelines/templates/_releaseTemplate.yml
```diff
@@ -0,0 +1,120 @@
...
+
+- task: ms-devlabs.utilitytasks.task-Shellpp.Shell++@0
+  displayName: 'Pull and create release tags for runtime images'
+  inputs:
+    type: FilePath
+    scriptPath: ./vsts/scripts/create-releasetags-runtimeimages.sh
+  condition: and(succeeded(), eq(variables['ReleaseRuntimeImages'], 'true'))
+
+- task: Docker@0
+  displayName: 'Push Build image to DockerHub'
```
kichalla: Are these `push` tasks correct? Here you are supplying a image names path file...where is this file coming from? in the earlier step we create tags, why can't we directly call `docker push` on those tags itself (and within that same script)

vsts/pipelines/ci.yml
```diff
@@ -4,75 +4,91 @@ resources:
...
+          echo "##vso[task.setvariable variable=PushRuntimeImages;]false"
+          echo "##vso[task.setvariable variable=EmbedBuildContextInImages;]true"
+        displayName: 'Set variables'
       
-  - template: templates/_buildTemplate.yml
-    parameters:
-      pushToDockerHub: 'true'
+      - template: templates/_buildTemplate.yml
+        parameters:
+          pushToDockerHub: 'true'
```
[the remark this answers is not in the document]
arroyc: removed …

vsts/pipelines/templates/_releaseTemplate.yml
```diff
@@ -0,0 +1,120 @@
...
+    command: login
+    azureSubscriptionEndpoint: ${{ parameters.ascName }}
+    azureContainerRegistry: ${{ parameters.acrName }}
+
+- task: DownloadBuildArtifacts@0
+  displayName: 'Download Build Artifacts'
+  inputs:
+    buildType: specific
+    project: '0bdbc590-a062-4c3f-b0f6-9383f67865ee'
+    pipeline: 11021
```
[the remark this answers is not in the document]
arroyc: we dont need it as we will be downloading from current build.

vsts/pipelines/templates/_releaseTemplate.yml
arroyc: as discussed .. this is something vsts needs to address ... I will address this in the community

vsts/pipelines/templates/_releaseTemplate.yml
arroyc: removed

vsts/pipelines/templates/_releaseTemplate.yml
arroyc: removed

vsts/pipelines/templates/_releaseTemplate.yml
arroyc: yepp its correct, pushing images from build-images-mcr.txt which I create in the release scripts

vsts/pipelines/ci.yml
arroyc: its concatenating job names with stage name when using step  template … i will address this to vsts community

PR #210, opened 2019-06-17
Added basic infrastructure for "install-runtime" command + consolidated subcommand names

dorfire:


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -19,9 +19,11 @@
 
 namespace Microsoft.Oryx.BuildScriptGeneratorCli
 {
-    [Command("build", Description = "Generate and run build scripts.")]
+    [Command(BuildCommand.Name, Description = "Generate and run build script.")]
```
gpcastro: how about "Builds an application" for description?

src/BuildScriptGenerator/Contracts/IProgrammingPlatform.cs
```diff
@@ -62,6 +62,13 @@ public interface IProgrammingPlatform
         /// <c>null</c> otherwise.</returns>
         BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorContext scriptGeneratorContext);
 
+        /// <summary>
+        /// Generate a bash script that can install the required runtime bits for the application's platforms.
+        /// </summary>
+        /// <param name="options">Options to generate the installation script with.</param>
+        /// <returns>The bash installation script.</returns>
+        string GenerateBashRunTimeInstallationScript(RunTimeInstallationScriptGeneratorOptions options);
```
gpcastro: each programming platform only needs to now the target directory and target version. `RunTimeInstallationScriptGeneratorOptions` also has platform name, which is redundant here since the platform object already knows which platform it is.

src/BuildScriptGeneratorCli/Commands/InstallRunTimeCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+        [Option(
+            OptionTemplates.PlatformVersion,
+            CommandOptionType.SingleValue,
+            Description = "The version of the platform for which the runtime components should be installed.")]
+        public string PlatformVersion { get; set; }
+
+        [Option(
+            "--installation-prefix <prefix>",
+            CommandOptionType.SingleValue,
+            Description = "The prefix into which the components will be installed.")]
```
kichalla: Can you give an example of what this looks like? and maybe add it in the description too?

src/BuildScriptGeneratorCli/Commands/InstallRunTimeCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+using System;
+using System.IO;
+using McMaster.Extensions.CommandLineUtils;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Oryx.BuildScriptGenerator;
+using Microsoft.Oryx.Common;
+
+namespace Microsoft.Oryx.BuildScriptGeneratorCli
+{
+    [Command(InstallRunTimeCommand.Name, Description = "Install the required runtime components for a platform.")]
```
kichalla: For my own understanding, what is this command meant to do? an E2E scenario would be helpful

src/BuildScriptGeneratorCli/Commands/InstallRunTimeCommand.cs
dorfire: it expects a path, and follows the `./configure` paradigm - i'll change the option name and its description

src/BuildScriptGeneratorCli/Commands/InstallRunTimeCommand.cs
dorfire: It'll install the actual platform - but just the runtime components (not SDKs)

kichalla: So this command generates a bash script which installs the runtime components? say for example if .NET Cor runtime 2.1 needs to be installed, does the generated bash script install it by downloading the content?

src/BuildScriptGeneratorCli/Commands/InstallRunTimeCommand.cs
kichalla: how is this path used? is it like a base path where the runtime bits are installed to like a `opt/dotnet` for example if we are installing the .NET Core 2.1 runtime?

src/BuildScriptGeneratorCli/Commands/InstallRunTimeCommand.cs
dorfire: Correct

PR #215, opened 2019-06-20
add missing category for node integration tests

arroyc:
Thanks @dorfire  for catching this


tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
```diff
@@ -34,6 +34,7 @@ public NodeEndToEndTestsBase(ITestOutputHelper output, TestTempDirTestFixture te
             DockerVolume.CreateMirror(Path.Combine(_hostSamplesDir, "nodejs", appName));
     }
 
+    [Trait("category", "node")]
```
kichalla: Oh, so these tests were not running for few days/weeks now...could you please check the same for other platform tests too?

build/test-integration.sh
```diff
@@ -36,7 +36,9 @@ integrationTestPlatform=""
 
 if [ -n "$1" ]; then
     testCaseFilter="--filter $1"
-    integrationTestPlatform="."$(echo $1 | cut -d'=' -f 2)
+	if [ -n "$BUILD_NUMBER" ]; then
```
gpcastro: this check by itself doesn't make sense to me. Are you using BUILD_NUMBER as a proxy for "running on an agent as opposed to dev box"? If yes, let's add a variable in __variables.sh to express this directly, and use that variable instead.

tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
gpcastro: is there a way to run tests in a "catch-all" fashion, that is, those not in any category? This is to avoid not running tests that aren't categorized, like this case

build/test-integration.sh
arroyc: We currently use this in our other scripts. (e.g. build scipts for build and runtime images)
Curious: what is the advantage of doing that?  We can change other scripts also then

build/__variables.sh
```diff
@@ -40,9 +40,15 @@ declare -r ACR_RUNTIME_IMAGES_REPO="$ACR_PUBLIC_PREFIX"
...
 
 # Flag to add information to images through labels (example: build number, commit sha)
 declare -r EMBED_BUILDCONTEXT_IN_IMAGES="${EMBEDBUILDCONTEXTINIMAGES:-false}"
 declare -r GIT_COMMIT=$(git rev-parse HEAD)
 
 declare -r DOCKER_SYSTEM_PRUNE="${ORYX_DOCKER_SYSTEM_PRUNE:-false}"
+
+# If build_number has value that means we are building in build agent and not locally
+if [ -n "$BUILD_NUMBER" ]; then
+    AGENT_BUILD="true"
```
kichalla: `declare -r AGENT_BUILD="true"`

build/build-buildimages.sh
```diff
@@ -100,7 +100,7 @@ else
 fi
 
 # Retag build image with DockerHub and ACR tags
-if [ -n "$BUILD_NUMBER" ]
+if [ -n "$AGENT_BUILD" ]
```
kichalla: 👍

build/__variables.sh
arroyc: @kichalla I declared it as "" by default and only for build's in build-agent it will be true. 
`declare -r AGENT_BUILD="" `

build/__variables.sh
```diff
@@ -40,9 +40,15 @@ declare -r ACR_RUNTIME_IMAGES_REPO="$ACR_PUBLIC_PREFIX"
 declare -r ACR_PACK_BUILDER_IMAGE_REPO="$ACR_PUBLIC_PREFIX/$PACK_BUILDER_IMAGE_NAME"
 declare -r ACR_PACK_IMAGE_REPO="$ACR_PUBLIC_PREFIX/$PACK_IMAGE_NAME"
 declare -r ACR_PACK_STACK_BASE_IMAGE_REPO="$ACR_PUBLIC_PREFIX/$PACK_STACK_BASE_IMAGE_NAME"
+declare -r AGENT_BUILD=""
```
dorfire: why declare it in this case? our scripts are run **without** `set -u`

build/__variables.sh
dorfire: still, it would probably be better if it was read-only in this case as well

build/test-integration.sh
```diff
@@ -32,11 +32,13 @@ fi
...
 echo
 
-integrationTestPlatform=""
+integrationTestPlatform=".default"
 
 if [ -n "$1" ]; then
     testCaseFilter="--filter $1"
-    integrationTestPlatform="."$(echo $1 | cut -d'=' -f 2)
+	if [ -n "$AGENT_BUILD" ]; then
+		integrationTestPlatform="."$(echo $1 | cut -d'=' -f 2)
```
dorfire: add documentation? What’s the purpose of this variable?

build/test-integration.sh
```diff
@@ -55,4 +57,13 @@ dotnet test \
     $testCaseFilter \
     --test-adapter-path:. \
     --logger:"xunit;LogFilePath=$ARTIFACTS_DIR/testResults/$testProjectName$integrationTestPlatform.Tests.xml" \
-    -c $BUILD_CONFIGURATION
\ No newline at end of file
+    -c $BUILD_CONFIGURATION
+
+if [ "$testCaseFilter"=="db" ];then
```
dorfire: spaces around operator

build/test-integration.sh
```diff
@@ -55,4 +57,13 @@ dotnet test \
...
+		echo "Running integration tests with no categories (tests missed to be tagged in their categories)..."
```
dorfire: This line could be simplified, e.g. "Running uncategorized tests".

Also - point to consider - do we even **want** to have such tests? Maybe the "right thing" to do is to fail the build if **there are any** uncategorized tests? @gpcastro @kichalla

build/test-integration.sh
```diff
@@ -55,4 +57,13 @@ dotnet test \
...
+		dotnet test \
+			--filter "category!=node&category!=python&category!=php&category!=dotnetcore&category!=db" \
```
dorfire: This list of categories can, and will, change. It should be moved to some central location (i.e. an enum of the possible categories)

arroyc:
as discussed, just enabling the node categtoris @gpcastro @dorfire

build/testIntegration.sh
```diff
@@ -32,11 +32,18 @@ fi
 
 echo
 
-integrationTestPlatform=""
+# This is needed because when we are running tests in multiple agent machines
+# this variable will be used to name the testresults file and that way we can
+# avoid test results file.
```
dorfire: "avoid test results file"?

build/testIntegration.sh
```diff
@@ -32,11 +32,18 @@ fi
...
+# This is needed because when we are running tests in multiple agent machines
+# this variable will be used to name the testresults file and that way we can
+# avoid overwriting test results file.
+integrationTestPlatform=".default"
 
 if [ -n "$1" ]; then
     testCaseFilter="--filter $1"
-    integrationTestPlatform="."$(echo $1 | cut -d'=' -f 2)
+	if [ -n "$AGENT_BUILD" ]; then
+        # Extract platform name for which the integration tests are running
```
dorfire: indentation

tests/Oryx.Integration.Tests/NodeEndToEndTests.cs
```diff
@@ -583,7 +584,7 @@ public async Task Node_CreateReactAppSample_zippedNodeModules(string nodeVersion
                 });
         }
 
-        [Fact]
+        [Fact(Skip = "Single image tests are still failing #938129")]
```
dorfire: are they really still failing?

PR #216, opened 2019-06-20
 Fixing Golang code workspace and other minor things

kichalla:


.gitignore
```diff
@@ -250,23 +250,16 @@ paket-files/
...
 
-# Go packages, if Oryx is added to GOPATH (which it should)
-pkg/
-
 # Go packages from GitHub
-src/github.com/
-src/startupscriptgenerator/vendor/
-src/golang.org/
\ No newline at end of file
+src/startupscriptgenerator/src/github.com/
```
kichalla: I moved all directories which were previously under `startupscriptgenerator` to a sub-directory under it now called `src`. My aim was to keep the structure as per Golang conventions.

.vscode/launch.json
```diff
@@ -0,0 +1,36 @@
+{
+    // Use IntelliSense to find out which attributes exist for C# debugging
+    // Use hover for the description of the existing attributes
+    // For further information visit https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md
+    "version": "0.2.0",
+    "configurations": [
+        {
+			"name": "Go Launch",
```
kichalla: This enables us to debug go code. I have tested this and it works fine. Also, will fix the spacings here

.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
+{
+    "files.exclude": {
```
kichalla: This will cause these folders to be NOT visible in VS Code UI...these were distracting

.vscode/tasks.json
```diff
@@ -0,0 +1,46 @@
+{
```
kichalla: Auto generated content. I just let VS Code do it automagic and wanted to have the default behavior which is good for most of us.

build/test-startupscriptgenerators.sh
```diff
@@ -4,20 +4,27 @@
...
+
 # When volume mounting a directory from the host machine, we host it as a readonly folder because any modifications by a
 # container in that folder would be owned by 'root' user(as containers run as 'root' by default). Since CI build agents
 # run as non-root cleaning these files would be a problem. So we copy the mounted directory in the container
 # to a different directory within the container itself and run tests on it.
-declare -r GEN_DIR_CONTAINER_RO="/src"
-declare -r GEN_DIR_CONTAINER="/go/src/startupscriptgenerator"
-declare -r MODULE_TO_TEST="startupscriptgenerator/..."
+declare -r GEN_DIR_CONTAINER_RO="/startupscriptgenerator"
+declare -r GEN_DIR_CONTAINER="/go/src"
```
kichalla: So in a go container, now the structure would need to mimic what our local workspace looks like:
So if your local path was `c:\oryx\src\startupscriptgenerator` (under which you would have `src`, `bin`, `pkg` as child folders), then the path in the container would be `/go` (under which you would have `src`, `bin`, `pkg` as child folders)

build/test-startupscriptgenerators.sh
```diff
@@ -4,20 +4,27 @@
...
+declare -r MODULE_TO_TEST="..."
 declare -r CONTAINER_NAME="oryxtests_$RANDOM"
 
+runTests="echo && echo Running tests..."
+runTests="$runTests && cd $GEN_DIR_CONTAINER/common && go test -v "
+runTests="$runTests && cd $GEN_DIR_CONTAINER/dotnetcore && go test -v "
+runTests="$runTests && cd $GEN_DIR_CONTAINER/node && go test -v "
+runTests="$runTests && cd $GEN_DIR_CONTAINER/python && go test -v "
+runTests="$runTests && cd $GEN_DIR_CONTAINER/php && go test -v "
+
```
kichalla: I really do not like this but I am do not have another alternative. The thing is since _now_ on we are going to have folders like `github.com`, `gitlab.org` etc. I cannot run a `got test` on all files under `src` since we would be running tests under `github.com` too

.vscode/settings.json
```diff
@@ -0,0 +1,11 @@
...
+        "**/startupscriptgenerator/bin": true,
+        "**/startupscriptgenerator/pkg": true,
+        "**/startupscriptgenerator/**/vendor": true,
+        "**/startupscriptgenerator/src/github.com": true,
+        "**/startupscriptgenerator/src/gotest.tools": true,
+        "**/startupscriptgenerator/src/golang.org": true
+    },
+    "go.toolsGopath": "${env:HOME}/go"
```
kichalla: When working on Go code in VS Code, it asks to install bunch of Go related tools(linters etc.) which can help us in development process. Since these tools can be use in any Go project (not only Oryx), I made this change so that it installs in a location or workspace other than our Oryx workspace. This way when you want to do `git clean -xdf` you do not unnecessarily clean-up these tools too.

src/startupscriptgenerator/src/restorePackages.sh
```diff
@@ -0,0 +1,25 @@
...
+    if [ -f "$pkgDir/$tomlFileName" ]; then
+            echo "Running 'dep ensure' under '$pkgDir'..."
+            cd $pkgDir
+            dep ensure
+        else
+            echo "Cound not find '$tomlFileName' under '$pkgDir'. Not running 'dep ensure'"
+    fi
+}
+
+runDepEnsure $WORKSPACE_DIR/src/common
```
kichalla: restoring packages under each folder that we care (i say this because we can other folders under `src`...for example `github.com` `gitlang.org` etc.)

src/startupscriptgenerator/src/php/main.go
```diff
@@ -21,7 +21,7 @@ func main() {
 
 	fullAppPath := common.GetValidatedFullPath(*appPathPtr)
 
-	common.SetGlobalOperationId(fullAppPath)
+	common.SetGlobalOperationID(fullAppPath)
```
kichalla: Just some changes made to satisfy `lint` tool complaints

.gitignore
dorfire: these should go under `src/startupscriptgenerator/src/vendor`, no?

dorfire: also, it feels a bit odd to have `src` under `src`

src/startupscriptgenerator/src/common/logging.go
```diff
@@ -6,36 +6,36 @@
...
-	OperationId string
+	OperationID string
 }
 
 // Represents unique identifier that can be used to correlate messages with the build logs
-var buildOpId string
+var buildOpID string
 
-func SetGlobalOperationId(appRootPath string) {
+func SetGlobalOperationID(appRootPath string) {
```
dorfire: please move these styling changes to a separate PR to make this one smaller and more focused on the re-org

src/startupscriptgenerator/src/restorePackages.sh
```diff
@@ -0,0 +1,25 @@
...
+echo "Installing dep..."
+go get -u github.com/golang/dep/cmd/dep
+
+function runDepEnsure() {
+    pkgDir="$1"
+    if [ -f "$pkgDir/$tomlFileName" ]; then
+            echo "Running 'dep ensure' under '$pkgDir'..."
+            cd $pkgDir
+            dep ensure
+        else
```
dorfire: indentation

src/startupscriptgenerator/src/restorePackages.sh
dorfire: this wouldn't be a problem if we stick them all under `vendor` :)

src/startupscriptgenerator/src/common/logging.go
kichalla: I think this is fine for this PR since for me at this point I want to make sure that the development experience is good and that I did not do anything wrong, but are we not doing these kind of changes? ;-) let's all agree to do these kind of changes in separate PRs instead

.gitignore
kichalla: > these should go under src/startupscriptgenerator/src/vendor, no?
 
but I did not see that happening in my workspace...let me investigate further

> also, it feels a bit odd to have src under src

yes, I agree, but that is the nature of our repo where we have different platforms (.net core and go)..any suggestions?

kichalla: Regarding vendor folder, yes you are right...i was earlier doing a `go get` which created these folders `github.com` etc. but if we use `dep` tool, it puts under a vendor folder. Will update

.gitignore
```diff
@@ -250,23 +250,18 @@ paket-files/
...
-# Go packages, if Oryx is added to GOPATH (which it should)
-pkg/
-
 # Go packages from GitHub
-src/github.com/
-src/startupscriptgenerator/vendor/
-src/golang.org/
\ No newline at end of file
+src/startupscriptgenerator/bin/
+src/startupscriptgenerator/src/github.com/
```
dorfire: is this still needed even after moving every external packages to the `vendor` folder?

src/startupscriptgenerator/src/restorePackages.sh
```diff
@@ -0,0 +1,23 @@
...
+declare -r WORKSPACE_DIR=$( cd $( dirname "$0" ) && cd .. && pwd )
+tomlFileName="Gopkg.toml"
+
+echo "Restoring packages..."
+echo "Installing dep..."
+go get -u github.com/golang/dep/cmd/dep
+# Delete the dep sources so that we do not use it when running tests etc.
+rm -rf $WORKSPACE_DIR/src/github.com/golang/dep
+
+tomlFileName="Gopkg.toml"
```
dorfire: already declared above

src/startupscriptgenerator/src/restorePackages.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
dorfire: 1. add copyright notice
2. consider renaming file to match our more common convention (`restore-packages.sh`)

.gitignore
kichalla: yes, because we do a `go get` for the `dep` tool itself which puts its stuff in this `github.com` folder

.dockerignore
```diff
@@ -2,10 +2,10 @@
...
+artifacts/
 tests/
 vsts/
-src/golang.org/
-src/github.com/
-pkg/
-bin/
-artifacts/
\ No newline at end of file
+**/github.com/
```
dorfire: this is going to be under `vendor` anyway, no?

.gitignore
dorfire: maybe we better document this in the .gitignore file, to help prevent future confusion

.dockerignore
kichalla: This is for the same reason I mentioned in my earlier comment (i.e we install dep using `go get`)

.gitignore
kichalla: Sure, I agree. Will add a comment in my next PR related to making dotnetcore startup script use go

PR #217, opened 2019-06-20
Fixed Python 3.8 directory discovery

dorfire:


arroyc:
is  this a fix to the python integration test failure

StdErr          : The 'python' version '3.8' is not supported. Supported versions are: 2.7.16, 3.6.8, 3.7.3

dorfire:
> is this a fix to the following python integration test failure in nightly builds?

Yes

PR #219, opened 2019-06-21
Added `oryx exec` command

dorfire:


src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,104 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Oryx.BuildScriptGenerator;
+using Microsoft.Oryx.Common;
+using Microsoft.Oryx.Common.Extensions;
+
+namespace Microsoft.Oryx.BuildScriptGeneratorCli
+{
+    [Command(Name, Description = "Execute an arbitrary command in the default shell, in an environment " +
+        "with the best-matching platform tool versions.")]
+    internal class ExecCommand : CommandBase
```
kichalla: Tests? There are fair number of cases we need to test here. commands with quotes, multiple commands separated by `&&`, a script file itself

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        [Argument(0, Description = "The source directory.")]
+        public string SourceDir { get; set; }
+
+        [Argument(1, Description = "The command to execute in an app-specific environment.")]
+        public string Command { get; set; }
+
+        internal override int Execute(IServiceProvider serviceProvider, IConsole console)
+        {
+            var logger = serviceProvider.GetRequiredService<ILogger<BuildCommand>>();
```
kichalla: `ILogger<ExecCommand>`

src/BuildScriptGenerator/CheckerMessage.cs
```diff
@@ -4,24 +4,19 @@
 // --------------------------------------------------------------------------------------------
 
 using Microsoft.Extensions.Logging;
-using Microsoft.Oryx.BuildScriptGenerator;
 
 namespace Microsoft.Oryx.BuildScriptGenerator
 {
     public class CheckerMessage : ICheckerMessage
     {
-        private readonly LogLevel _level;
```
kichalla: As we discussed earlier in my PR, let's avoid making unrelated changes. A separate clean-up PR would be better for these.

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -241,8 +241,8 @@ private static string GetBenvArgs(IDictionary<string, string> benvArgsMap)
         private IList<BuildScriptSnippet> GetBuildSnippets(
             BuildScriptGeneratorContext context,
             Dictionary<string, string> toolsToVersion,
-            List<string> directoriesToExcludeFromCopyToIntermediateDir,
-            List<string> directoriesToExlcudeFromCopyToBuildOutputDir)
+            [CanBeNull] List<string> directoriesToExcludeFromCopyToIntermediateDir,
```
kichalla: Please revert all these type of unrelated changes!

src/BuildScriptGenerator/SourceRepo/DefaultSourceRepoProvider.cs
```diff
@@ -39,42 +31,5 @@ public ISourceRepo GetSourceRepo()
             _sourceRepo = new LocalSourceRepo(_sourceDirectory, _loggerFactory);
             return _sourceRepo;
         }
-
-        private void CopyDirectories(string sourceDirectory, string destinationDirectory, bool recursive)
```
kichalla: Was this dead code?

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -24,7 +24,7 @@ internal abstract class CommandBase
         public string LogFilePath { get; set; }
 
         [Option("--debug", Description = "Print stack traces for exceptions.")]
```
kichalla: I am not suggesting to make a change here, but I think this should probably be `--debugInfo`

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
dorfire: yeah, fixed already but did not push. thanks!

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
dorfire: They aren't unrelated - i changed the way this function works, so that it can accept `null`s in these two parameters.
The new method `GetRequiredToolVersions` relies on this change.

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
dorfire: working on them still; didn't want to block the review of this PR in the meantime

src/BuildScriptGenerator/SourceRepo/DefaultSourceRepoProvider.cs
dorfire: yep

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -241,8 +241,8 @@ private static string GetBenvArgs(IDictionary<string, string> benvArgsMap)
...
+            [CanBeNull] List<string> directoriesToExlcudeFromCopyToBuildOutputDir)
```
gpcastro: can you also fix the typo: directoriesToEx*cl*ude

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,113 @@
...
+{
+    [Command(Name, Description = "Execute an arbitrary command in the default shell, in an environment " +
+        "with the best-matching platform tool versions.")]
+    internal class ExecCommand : CommandBase
+    {
+        public const string Name = "exec";
+
+        public const string DefaultBashPath = "/bin/bash";
+
+        public const string SrcDirDoesNotExistErrorMessageFmt = "Could not find the source directory '{0}'.";
```
gpcastro: move those to the resource file with the labels

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,113 @@
...
+        public const string CommandMissingErrorMessage = "A command is required.";
+        public const string NoToolsDetectedErrorMessage = "No usable tools detected for source directory.";
+
+        [Argument(0, Description = "The source directory.")]
+        public string SourceDir { get; set; }
```
gpcastro: I would have this is a named parameter, not as positional argument. e.g. `oryx exec -sourceDir /app <command>`

gpcastro: and default to current directory

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            return exitCode;
+        }
+
+        internal override bool IsValidInput(IServiceProvider serviceProvider, IConsole console)
+        {
+            var logger = serviceProvider.GetRequiredService<ILogger<ExecCommand>>();
+            var options = serviceProvider.GetRequiredService<IOptions<BuildScriptGeneratorOptions>>().Value;
+
+            if (!Directory.Exists(options.SourceDir))
```
gpcastro: how about the [DirectoryExists] attribute?

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,113 @@
...
+            {
+                logger.LogError("Could not find the source directory {srcDir}", SourceDir);
+                console.WriteErrorLine(string.Format(SrcDirDoesNotExistErrorMessageFmt, SourceDir));
+                return false;
+            }
+
+            if (string.IsNullOrWhiteSpace(Command))
```
gpcastro: I wouldn't error out in this case, just do a no-op. When writing a script that calls this command, it's easier to just no-op when there is no command so the script doesn't fail. I don't think `benv` for instance fails with an empty command

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
dorfire: I wanted to keep this in-line with `BuildCommand`. WDYT?

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
kichalla: I see you added only 1 E2E test. Can you add more scenarios as mentioned above?

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,92 @@
...
+                logger.LogInformation("Running {shell} -c {cmd}", shellPath, cmd);
+                if (DebugMode)
+                {
+                    console.WriteLine($"> {shellPath} -c '{cmd}'");
+                }
+
+                exitCode = serviceProvider.GetRequiredService<IScriptExecutor>().ExecuteScript(
+                    shellPath,
+                    new[] { "-c", cmd },
+                    SourceDir,
```
kichalla: Doesn't this fail if the SourceDir is not provided by the user? I think you should pick it out of the `options` which resolves the source directory to the current directory if user does not provide it.

tests/Oryx.BuildImage.Tests/ExecCommandTest.cs
```diff
@@ -0,0 +1,57 @@
...
+                .ToString();
+
+            // Act
+            var result = _dockerCli.Run(Settings.BuildImageName, "/bin/bash", "-c", script);
+
+            // Assert
+            RunAsserts(
+                () =>
+                {
+                    Assert.Contains($"{expectedBashPath} -c '{FilePaths.Benv} node={NodeConstants.NodeLtsVersion} {cmd}'", result.StdOut);
```
kichalla: nit: long line

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
dorfire: `SourceDir` defaults to the current directory, like in `BuildCommand`

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,109 @@
...
+            }
+
+            int exitCode;
+            using (var timedEvent = logger.LogTimedEvent("ExecCommand"))
+            {
+                var benvCmd = $"{FilePaths.Benv} {StringExtensions.JoinKeyValuePairs(tools)}";
+
+                // Build envelope script
+                var script = new ShellScriptBuilder("\n")
+                    .AddShebang(shellPath)
```
kichalla: I think we should add a `set -e` too so that we propagate the right exit code. Add a test for this failure case too.

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,109 @@
...
+                    .Source(benvCmd)
+                    .AddCommand(Command)
+                    .ToString();
+                logger.LogDebug("Script content:\n{script}", script);
+
+                // Create temporary file to store script
+                var tempScriptPath = Path.GetTempFileName();
+                File.WriteAllText(tempScriptPath, script);
```
kichalla: Do we need to set any permissions on the generated script so that it can execute without any problem?

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,109 @@
...
+                    console.WriteLine($"Temporary script @ {tempScriptPath}:");
+                    console.WriteLine("---");
+                    console.WriteLine(script);
+                    console.WriteLine("---");
+                }
+
+                exitCode = ProcessHelper.RunProcess(
+                    shellPath,
+                    new[] { tempScriptPath },
+                    serviceProvider.GetRequiredService<IOptions<BuildScriptGeneratorOptions>>().Value.SourceDir,
```
kichalla: nit: get the options into a variable (after line 36) and use it here instead.

tests/Oryx.BuildImage.Tests/ExecCommandTest.cs
```diff
@@ -0,0 +1,92 @@
...
+                    Assert.Contains(nodeArg, result.StdOut);
+                    Assert.True(result.IsSuccess);
+                    // Actual output from `node --version` starts with a 'v'
+                    Assert.Contains($"v{NodeConstants.NodeLtsVersion}", result.StdOut);
+                },
+                result.GetDebugInfo());
+        }
+
+        [Fact]
+        public void Exec_NodeVersionAndPhpVersion()
```
kichalla: Rename the test so that it indicates what the pivot here actually is...CanExecuteMultipleCommands may be?

tests/Oryx.BuildImage.Tests/ExecCommandTest.cs
```diff
@@ -0,0 +1,92 @@
...
+using Microsoft.Oryx.Tests.Common;
+using Xunit;
+using Xunit.Abstractions;
+using Microsoft.Oryx.BuildScriptGenerator.Node;
+using Microsoft.Oryx.Common;
+using Microsoft.Oryx.BuildScriptGenerator.Php;
+
+namespace Microsoft.Oryx.BuildImage.Tests
+{
+    public class ExecCommandTest : SampleAppsTestBase
```
kichalla: Add a test where the user provided command exits with a non successful exit code

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -0,0 +1,109 @@
...
+            int exitCode;
+            using (var timedEvent = logger.LogTimedEvent("ExecCommand"))
+            {
+                var benvCmd = $"{FilePaths.Benv} {StringExtensions.JoinKeyValuePairs(tools)}";
+
+                // Build envelope script
+                var script = new ShellScriptBuilder("\n")
+                    .AddShebang(shellPath)
+                    .Source(benvCmd)
+                    .AddCommand(Command)
```
kichalla: I think this command should be executed in the context of the source directory, i.e we should probably `cd` into the source directory and then execute the command. This way any user provided relative paths can work fine. Add a test for this scenario too.

tests/BuildScriptGeneratorCli.Tests/ExecCommandTest.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            // Act
+            var exitCode = cmd.OnExecute(new CommandLineApplication(console), console);
+
+            // Assert
+            Assert.Equal(ProcessConstants.ExitSuccess, exitCode);
+        }
+
+        [Fact]
+        public void OnExecute_ShowsErrorAndExits_WhenNoUsableToolsAreDetected()
```
kichalla: Since our build image has LTS versions as default, IMO we should not throw this error and should run the provided command without erroring out. Unfortunately I do not know the usage of VS saas team here, but i will give a scenario: Let's say i start out with a directory which is empty and then provide a command like `dotnet new empty`...this should cd into the source directory and run this command and this will create a empty asp.net core project

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
dorfire: No - it's not executed, it's just passed as a parameter an read by Bash

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
dorfire: that's the purpose of this? I preferred to have one less variable because this isn't used anywhere else

kichalla: It just looks ugly IMO. You could argue the same thing for a `using` statement. We do not put the whole namespace at the point we use because we use it in one place.

PR #225, opened 2019-06-25
add extra check for version information tests

arroyc:
This is one way of getting rid of unwanted test failure where we see commit-id is mismatching. We can also follow other ways where we don't run this test for PR validation. But for now I'm just curious if this works.

### DIFF
--- a/tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
+++ b/tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
@@ -31,7 +31,17 @@ namespace Microsoft.Oryx.BuildImage.Tests
         public void OryxBuildImage_Contains_VersionAndCommit_Information()
         {
             var agentOS = Environment.GetEnvironmentVariable("AGENT_OS");
-            var gitCommitID = Environment.GetEnvironmentVariable("BUILD_SOURCEVERSION");
+
+
+            // On a pull request, Azure DevOps does not build the exact version of the code that has
+            // been pushed, but rather a version merged with the target branch; When the target branch 
+            // pointer has advanced, it creates different commit-ids and this test fails. This only 
+            // happens in PR-validation build, so this is a check for that. For more details, refer to
+            // https://github.com/microsoft/azure-pipelines-tasks/issues/9801
+
+            var gitCommitID = Environment.GetEnvironmentVariable("SYSTEM_PULLREQUEST_SOURCECOMMITID") ??
+                Environment.GetEnvironmentVariable("BUILD_SOURCEVERSION");
+
             var buildNumber = Environment.GetEnvironmentVariable("BUILD_BUILDNUMBER");
             var expectedOryxVersion = string.Concat(Settings.OryxVersion, buildNumber);
 


tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
```diff
@@ -32,6 +32,19 @@ public void OryxBuildImage_Contains_VersionAndCommit_Information()
...
+
+            // On a pull request, Azure DevOps does not build the exact version of the code that has
+            // been pushed, but rather a version merged with the target branch; When the target branch 
+            // pointer has advanced, it creates different commit-ids and this test fails. This only 
+            // happens in PR-validation build, so this is a check for that. For more details, refer to
+            // https://github.com/microsoft/azure-pipelines-tasks/issues/9801
+
+            if (currentBuildDefinition.ToLower().Contains("oryx-validation-github"))
+            {
+                gitCommitID = Environment.GetEnvironmentVariable("SYSTEM_PULLREQUEST_SOURCECOMMITID");
```
dorfire: could we do something like

```csharp
GetEnvVar("SYSTEM_PULLREQUEST_SOURCECOMMITID") ?? GetEnvVar("BUILD_SOURCEVERSION")
```

and not check the `DEFINITIONNAME` at all?

tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
```diff
@@ -32,6 +32,19 @@ public void OryxBuildImage_Contains_VersionAndCommit_Information()
         {
             var agentOS = Environment.GetEnvironmentVariable("AGENT_OS");
             var gitCommitID = Environment.GetEnvironmentVariable("BUILD_SOURCEVERSION");
+            var currentBuildDefinition = Environment.GetEnvironmentVariable("BUILD.DEFINITIONNAME");
+
+            // On a pull request, Azure DevOps does not build the exact version of the code that has
+            // been pushed, but rather a version merged with the target branch; When the target branch 
+            // pointer has advanced, it creates different commit-ids and this test fails. This only 
+            // happens in PR-validation build, so this is a check for that. For more details, refer to
+            // https://github.com/microsoft/azure-pipelines-tasks/issues/9801
```
dorfire: good find!

tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
kichalla: Yes, I agree with @dorfire ...we should rather depend on the merged commit id instead of build definition name

arroyc:
this is interesting when I rebuilt the PR (by choosing PR number instead of branch name) from azure devops, it didn't acknowledge 
here is it: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2798950 or https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2799718

but it does acknowledge when you rebuild it from the branch. https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2799724

Is it normal behavior? @kichalla @gpcastro @dorfire

PR #229, opened 2019-06-26
fix release scripts

arroyc:


vsts/scripts/tagRunTimeImagesForRelease.sh
```diff
@@ -7,7 +7,7 @@
 set -o pipefail
 
 outFileMCR="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-mcr.txt"
-outFileDocker="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-mcr.txt"
+outFileDocker="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-dockerhub.txt"
```
dorfire: isn't it about time we stopped releasing to DockerHub? :)

arroyc: i will be happy to scratch this out .. one less step to fail :P ..

PR #230, opened 2019-06-26
 .NET Core startup script: Refactored the code to generate minimal shell script and doing most of the logic in golang

kichalla:


src/BuildScriptGenerator/DotNetCore/DotnetCoreBashBuildSnippetProperties.cs
```diff
@@ -32,6 +33,8 @@ public class DotNetCoreBashBuildSnippetProperties
 
         public bool ZipAllOutput { get; set; }
 
+        public string ZippedOutputFileName { get; set; } = FilePaths.CompressedOutputFileName;
```
kichalla: Zipped and Compressed...ideally it should probably say compressed everywhere...i kept as zipped for now and will create a separate pr for cleaning this up...also need to see how to not break existing customers with a change to the property name `zip_all_output`

src/startupscriptgenerator/src/common/buildManifestHelper.go
```diff
@@ -47,3 +47,8 @@ func GetBuildManifest(appPath string) BuildManifest {
 
 	return _buildManifest
 }
+
+func ManifestFileExists(appPath string) bool {
+	tomlFile := filepath.Join(appPath, consts.BuildManifestFileName)
+	return FileExists(tomlFile)
```
gpcastro: in general, it's good to store the function return to a variable and then return the variable to facilitate debugging when needed

src/startupscriptgenerator/src/common/outputHelper.go
```diff
@@ -6,39 +6,44 @@
...
+		intermediateDir = appDir
+	}
+
+	if FileExists(fullZipFilePath) {
+		scriptBuilder.WriteString("echo Extracting '" + consts.CompressedOutputFileName + "' contents...\n")
+		scriptBuilder.WriteString("cd \"" + intermediateDir + "\"\n")
+		scriptBuilder.WriteString("tar -xzf \"" + consts.CompressedOutputFileName + "\"\n")
+		scriptBuilder.WriteString("echo Done.\n")
+		scriptBuilder.WriteString("echo Deleting the file '" + consts.CompressedOutputFileName + "'...\n")
+		scriptBuilder.WriteString("rm -f \"" + consts.CompressedOutputFileName + "\"\n")
```
gpcastro: why do we delete the file? what if the container restarts, the app will have to rebuilt?

src/BuildScriptGenerator/DotNetCore/DotnetCoreBashBuildSnippetProperties.cs
dorfire: we can still support `zip_all_output` but mark it as deprecated.
then, we can add a Checker that warns when deprecated properties are used.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -33,132 +35,120 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
+			scriptBuilder.WriteString(gen.UserStartupCommand + " || EXIT_CODE=$?\n")
+			scriptBuilder.WriteString("if [ $EXIT_CODE != 0 ]; then\n")
+			scriptBuilder.WriteString("    echo \"WARNING: Startup command execution failed with exit code $EXIT_CODE\"\n")
+			scriptBuilder.WriteString("    echo \"Running the default application instead...\"\n")
+			scriptBuilder.WriteString("    cd \"" + defaultAppFileDir + "\"\n")
+			scriptBuilder.WriteString("    dotnet \"" + gen.DefaultAppFilePath + "\"\n")
+			scriptBuilder.WriteString("fi\n")
+		}
+	} else {
+		if !builtByOryx {
```
gpcastro: nit: reverse the if/else blocks so we don't use reverse logic

src/BuildScriptGenerator/DotNetCore/DotnetCoreBashBuildSnippetProperties.cs
kichalla: Good point about using Checker here.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -33,132 +35,120 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
 }
+
+func (gen *DotnetCoreStartupScriptGenerator) getRuntimeConfigJsonFiles() []string {
+	logger := common.GetLogger("dotnetcore.scriptgenerator.getRuntimeConfigJsonFiles")
+	defer logger.Shutdown()
+
+	fileList := make([]string, 0)
+
+	appDir, err := os.Open(gen.AppPath)
+	if err != nil {
```
gpcastro: instead of duplicating the err != nil case, you can use move forward `if err == nil`, and add a single block `if `err != nil` at the end

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -33,132 +35,120 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
+			RuntimeConfigJsonExtension,
+			appDir.Name(),
+			err)
+		return fileList
+	}
+
+	for _, file := range files {
+		if file.Mode().IsRegular() {
+			fileName := file.Name()
+			if strings.HasSuffix(fileName, RuntimeConfigJsonExtension) {
```
gpcastro: should we also check that the corresponding dll file really exists?

src/startupscriptgenerator/src/common/outputHelper.go
kichalla: Good point. I think I should fix some code in another place too.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
kichalla: yeah, I agree...will update

PR #231, opened 2019-06-26
Removed command line option `--language` & styling/cleanup

dorfire:


src/BuildScriptGeneratorCli/Commands/OptionTemplates.cs
```diff
@@ -7,8 +7,8 @@ namespace Microsoft.Oryx.BuildScriptGeneratorCli
 {
     internal static class OptionTemplates
     {
-        public const string Platform = "--platform <name>|-l|--language <name>";
-        public const string PlatformVersion = "--platform-version <version>|--language-version <version>";
+        public const string Platform = "-l|--platform <name>";
```
dorfire: As of commit 40a99094d03756e2e26458da2cdaae7b8dddfe74, **KuduLite still uses `-l` and `--language-version`**

gpcastro: cc @patricklee2 
we can submit a PR to their repo

kichalla: So I think @gpcastro is saying you can go ahead and change everything to be `platform` and `platform-version` and we can send out a reaction-PR to KuduLite

PR #234, opened 2019-06-27
[WIP] Run script generator: Added a script builder for golang code

kichalla:


kichalla:
Sent out this PR just to get any initial feedback before I proceed to making changes everywhere

src/startupscriptgenerator/src/common/outputHelper.go
```diff
@@ -23,27 +22,25 @@ func AppendScriptToExtractZippedOutput(scriptBuilder *strings.Builder, appDir st
 	}
 
 	if FileExists(fullZipFilePath) {
-		scriptBuilder.WriteString("\n")
-		scriptBuilder.WriteString("echo Extracting '" + consts.CompressedOutputFileName + "' contents...\n")
-		scriptBuilder.WriteString("cd \"" + intermediateDir + "\"\n")
-		scriptBuilder.WriteString("tar -xzf \"" + consts.CompressedOutputFileName + "\"\n")
-		scriptBuilder.WriteString("echo Done.\n")
+		scriptBuilder.AppendEmptyLine()
```
kichalla: What do you think of this API?

src/startupscriptgenerator/src/common/outputHelper.go
```diff
@@ -23,27 +22,25 @@ func AppendScriptToExtractZippedOutput(scriptBuilder *strings.Builder, appDir st
...
-	scriptBuilder.WriteString("    rm -rf \"$destDir\"\n")
-	scriptBuilder.WriteString("    echo Done.\n")
-	scriptBuilder.WriteString("fi\n")
-	scriptBuilder.WriteString("mkdir -p \"$destDir\"\n")
-	scriptBuilder.WriteString("echo \"Copying content from '$srcDir' to directory '$destDir'...\"\n")
-	scriptBuilder.WriteString("cp -rf \"$srcDir\"/* \"$destDir\"\n")
-	scriptBuilder.WriteString("echo Done.\n")
+func AppendScriptToCopyToDir(scriptBuilder *ScriptBuilder, srcDir string, destDir string) {
+	scriptBuilder.AppendEmptyLine()
+	scriptBuilder.AppendLine("if [ -d \"" + destDir + "\" ]; then")
```
kichalla: Few complexities related to blocks like `if`:
1. Expressing conditions: condition1 or condition2, condition1 and condition 2
2. Indentation of statements below this if condition. Ideally we might want to use api like `ChangeToDirectory` under this condition too, but the problems are: a). we would loose indentation when generating the script b). Even if we can fix the indentation (keep track of these kind of blocks using a stack) in the generated script, when reading this code in our code base, have indentation in golang code is also needed

src/startupscriptgenerator/src/common/outputHelper.go
dorfire: I like the idea! would probably go for even shorter names (`builder.Chdir`, `builder.ExtractArchive`).
great initiative 👍

src/startupscriptgenerator/src/common/scriptBuilder.go
```diff
@@ -0,0 +1,50 @@
...
+// --------------------------------------------------------------------------------------------
+
+package common
+
+import (
+	"strings"
+)
+
+type ScriptBuilder struct {
+	scriptBuilder strings.Builder
```
dorfire: member `scriptBuilder` in type `ScriptBuilder` seems redundant - maybe just `script`?

src/startupscriptgenerator/src/common/outputHelper.go
gpcastro: one thing we could do is to add a `scriptBuilder.AddIfBlock(condition: string)` that returns a `ifBlock` object, in a fluent design. We could do this:
```
scriptBuilder.AddIfBlock("-d \"" + destDir + "\"")
    .appendLine(...)
    .end()
```

dorfire: The aforementioned IfBlock object itself could probably be another `ScriptBuilder` instance

PR #235, opened 2019-06-27
Embedded commit information: Fix the commit id used to embed into binaries and images

kichalla:


kichalla:
cc @arroyc

build/__variables.sh
```diff
@@ -43,6 +43,11 @@ declare -r ACR_PACK_STACK_BASE_IMAGE_REPO="$ACR_PUBLIC_PREFIX/$PACK_STACK_BASE_I
 
 # Flag to add information to images through labels (example: build number, commit sha)
 declare -r EMBED_BUILDCONTEXT_IN_IMAGES="${EMBEDBUILDCONTEXTINIMAGES:-false}"
-declare -r GIT_COMMIT=$(git rev-parse HEAD)
+
+# On a pull request, Azure DevOps does not build the exact version of the code that has
+# been pushed, but rather a version merged with the target branch; When the target branch
+# pointer has advanced, it creates different commit-ids.
+# For more details, refer to https://github.com/microsoft/azure-pipelines-tasks/issues/9801
+declare -r GIT_COMMIT="${SYSTEM_PULLREQUEST_SOURCECOMMITID:-$BUILD_SOURCEVERSION}"
```
arroyc: oh ok …

gpcastro: how about we fall back to `$(git rev-parse HEAD)` if `$BUILD_SOURCEVERSION` is also empty? so we also get the commit info in local builds

kichalla: Ideally we do not want the commit id in our local builds because it causes a cache bust every time you make a commit locally and then build images. So when you expect image layers being built no-op, you would sometimes see the cache busted and this is because of this commit id change (as you might have committed as your normal workflow)

arroyc:
runtime image tests also needs to be addressed, dotnetcore, nodeimages, php and python

PR #236, opened 2019-06-27
[DO NOT MERGE] experimenting on validation build for uncategorized integration tests

arroyc:
This is as per my understanding to find uncategorized integration tests during validation build. What's your thoughts?


build/testBuildImages.sh
```diff
@@ -23,6 +23,15 @@ else
     $buildBuildImagesScript "$@"
 fi
 
+testResult=$(dotnet test --filter "$MISSING_CATEGORY_FILTER" "$TESTS_SRC_DIR/$INTEGRATION_TEST_PROJECT/$INTEGRATION_TEST_PROJECT.csproj")
```
arroyc: these tests can fail or can be successful. I think we should let It run instead of forcing them to fail

build/testBuildImages.sh
```diff
@@ -23,6 +23,15 @@ else
...
+
+if [ "$testResult" == "No test matches the given testcase filter" ]; then 
+    echo
+    echo "All integration tests have category: No missing category tests found..."  
+else 
+    echo "$testResult"
```
arroyc: after showing the output here can do exit 1 or something like that

arroyc:
this validation build for PR will always fail because there are some uncategorized node integration tests .. To make sure I didn't break anything I am running another build https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2807283.  Here I made sure I have all the categories. So for no missing categories validation build should be successfull

src/Common/IntegrationTestCategory.cs
```diff
@@ -0,0 +1,13 @@
+// This file was auto-generated from 'constants.yaml'. Changes may be overridden.
+
+namespace Microsoft.Oryx.Common
+{
+    public static class IntegrationTestCategory
```
dorfire: I thought we could enhance the SharedCodeGenerator so that it actually outputs an `enum`, not a class with `const string`s :)

PR #239, opened 2019-06-27
Use 'latest' tag only for images produced out of master branch builds

kichalla:


arroyc:
Will this be same for buildpack and base images?

kichalla:
> Will this be same for buildpack and base images?
yeah, at least it should be the same for buildpack images i think...base images anyway users are not expected to use them

arroyc:
> Will this be same for buildpack and base images?
> 
> yeah, at least it should be the same for buildpack images i think...base images anyway users are not expected to use them

Can you please make necessary changes in 
```
create-releasetags-buildpacks.sh
create-releasetags-baseimages.sh
```

PR #241, opened 2019-06-28
update baseimage build definition as per renamed bash scripts

arroyc:


arroyc:
@gpcastro  please see this sample build https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2810570

vsts/pipelines/buildImageBases.yml
```diff
@@ -10,41 +10,41 @@ stages:
...
       displayName: Build and push Yarn Cache base image
-      scriptPath: ./build/build-buildimage-bases.sh
+      scriptPath: ./build/buildBuildImageBases.sh
       imageDir: yarn-cache
       artifactsFileName: yarn-cache-buildimage-bases.txt
 
 - stage: Release
   displayName: Release Stage
-  dependsOn: Build
+  condition: or(succeeded(), failed())
```
arroyc: this will make the release stage to run always, but it will wait for the build stage to finish first

Oryx.sln
```diff
@@ -26,15 +24,19 @@ EndProject
...
+		build\buildRunTimeImageBases.sh = build\buildRunTimeImageBases.sh
+		build\buildRunTimeImages.sh = build\buildRunTimeImages.sh
+		build\buildSln.sh = build\buildSln.sh
+		build\buildTestBuildImage.sh = build\buildTestBuildImage.sh
+		build\constants.yaml = build\constants.yaml
+		build\testBuildImages.sh = build\testBuildImages.sh
+		build\testBuildScriptGenerator.sh = build\testBuildScriptGenerator.sh
+		build\testIntegration.sh = build\testIntegration.sh
+		build\testRunTimeImages.sh = build\testRunTimeImages.sh
+		build\testStartupScriptGenerators.sh = build\testStartupScriptGenerators.sh
```
arroyc: these changes are probably not that important, but being a vs user when you open oryx.sln and you try to open some files (bash or yaml) in your repo and it keeps giving you annyoying error that file has been moved or deleted. I only included what we had earlier. We have many more, we can add them  if we want. But I guess this is as minimum as it can be.

PR #242, opened 2019-07-01
Renamed command `languages` to `platforms`

dorfire:


src/BuildScriptGeneratorCli/Commands/LanguagesCommand.cs
```diff
@@ -17,16 +17,16 @@
 namespace Microsoft.Oryx.BuildScriptGeneratorCli
 {
     [Command(Name, Description = "Show a list of supported platforms along with their versions and build properties.")]
-    internal class LanguagesCommand : CommandBase
+    internal class PlatformsCommand : CommandBase
```
kichalla: Change the file name too

PR #243, opened 2019-07-01
Consolidated version provider logic

dorfire:


arroyc:
it will be nice to add a small test for version providerhelper

PR #244, opened 2019-07-01
Re-added deprecated option names for a smoother transition

dorfire:


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -254,6 +288,18 @@ internal override bool IsValidInput(IServiceProvider serviceProvider, IConsole c
             var options = serviceProvider.GetRequiredService<IOptions<BuildScriptGeneratorOptions>>().Value;
             var logger = serviceProvider.GetRequiredService<ILogger<BuildCommand>>();
 
+            if (_languageWasSet)
+            {
+                logger.LogWarning("Deprecated option '--language' used");
+                console.WriteLine("Warning: the deprecated option '--language' was used.");
```
kichalla: Should we also recommend users to use the new name?

tests/BuildScriptGeneratorCli.Tests/BuildCommandTest.cs
```diff
@@ -87,6 +87,30 @@ public void IsValidInput_IsTrue_EvenIfDestinationDirDoesNotExist()
...
+            var serviceProvider = new ServiceProviderBuilder()
+                .ConfigureScriptGenerationOptions(o =>
+                {
+                    o.SourceDir = _testDir.CreateChildDir();
+                    o.DestinationDir = _testDir.GenerateRandomChildDirPath();
+                })
+                .Build();
+            var testConsole = new TestConsole();
+            var expectedPlatformName = "test";
+            var buildCommand = new BuildCommand { LanguageName = expectedPlatformName };
```
kichalla: Add the LanguageVersion too here and its corresponding assert

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
dorfire: I don't think we have that many users dependent on the old name. AFAIK this PR is solely for App Service

PR #245, opened 2019-07-01
Added a timed event for each command execution

dorfire:

### DIFF
--- a/src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
+++ b/src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Oryx.BuildScriptGeneratorCli
             new TextSpan(
                 "RunPostBuildScript",
                 Oryx.BuildScriptGenerator.Constants.PostBuildCommandPrologue,
-                Oryx.BuildScriptGenerator.Constants.PostBuildCommandEpilogue)
+                Oryx.BuildScriptGenerator.Constants.PostBuildCommandEpilogue),
         };
 
         [Argument(0, Description = "The source directory.")]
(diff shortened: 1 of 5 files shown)


src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -45,7 +49,11 @@ public int OnExecute(CommandLineApplication app, IConsole console)
                     console.WriteLine("Debug mode enabled");
                 }
 
-                return Execute(_serviceProvider, console);
+                using (var timedEvent = logger.LogTimedEvent(GetType().Name))
```
dorfire: This will generate events with names like `PlatformsCommand`

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -34,7 +34,7 @@ internal class BuildCommand : CommandBase
             new TextSpan(
                 "RunPostBuildScript",
                 Oryx.BuildScriptGenerator.Constants.PostBuildCommandPrologue,
-                Oryx.BuildScriptGenerator.Constants.PostBuildCommandEpilogue)
+                Oryx.BuildScriptGenerator.Constants.PostBuildCommandEpilogue),
```
arroyc: is that new stylecop rule? :)

src/BuildScriptGeneratorCli/Program.cs
```diff
@@ -11,13 +11,13 @@
...
-    [Subcommand(InstallRunTimeCommand.Name,  typeof(InstallRunTimeCommand))]
-    [Subcommand(BuildpackDetectCommand.Name, typeof(BuildpackDetectCommand))]
-    [Subcommand(BuildpackBuildCommand.Name,  typeof(BuildpackBuildCommand))]
+    [Subcommand(typeof(BuildCommand))]
+    [Subcommand(typeof(PlatformsCommand))]
+    [Subcommand(typeof(BuildScriptCommand))]
+    [Subcommand(typeof(RunScriptCommand))]
+    [Subcommand(typeof(InstallRunTimeCommand))]
+    [Subcommand(typeof(BuildpackDetectCommand))]
+    [Subcommand(typeof(BuildpackBuildCommand))]
```
arroyc: what is this change for? we don't need descriptions anymore?

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
dorfire: yep

src/BuildScriptGeneratorCli/Program.cs
dorfire: the previous constructor we used is now deprecated

PR #247, opened 2019-07-02
Custom manifest file location: Added a switch to build and runtime script generators

kichalla:


src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -52,15 +53,12 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
...
-	scriptBuilder.WriteString("if [ -f ./" + consts.BuildManifestFileName + " ]; then\n")
-	scriptBuilder.WriteString("    echo \"Found '" + consts.BuildManifestFileName + "'\"\n")
-	scriptBuilder.WriteString("    . ./" + consts.BuildManifestFileName + "\n")
-	scriptBuilder.WriteString("fi\n\n")
 
 	// Expose the port so that a custom command can use it if needed.
 	common.SetEnvironmentVariableInScript(&scriptBuilder, "PORT", gen.BindPort, DefaultBindPort)
 
 	if !gen.SkipNodeModulesExtraction {
+		scriptBuilder.WriteString("compressedNodeModulesFile=\"" + gen.Manifest.CompressedNodeModulesFile + "\"\n")
```
kichalla: Filed an issue now to fix this as we can minimize the amount of script that we generate by doing majority of logic in go

src/BuildScriptGenerator/BaseBashBuildScriptProperties.cs
```diff
@@ -49,5 +49,10 @@ public class BaseBashBuildScriptProperties
         /// Gets or sets the name of the manifest file.
         /// </summary>
         public string ManifestFileName { get; set; }
+
+        /// <summary>
+        /// Gets or sets the path to the directory where the manifest file needs to be put.
+        /// </summary>
+        public string ManifestFileDir { get; set; }
```
dorfire: nit: consider shortening to `ManifestDir`

src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
```diff
@@ -148,13 +148,20 @@ dotnet restore "{{ ProjectFile }}"
 
 {{ if ManifestFileName | IsNotBlank }}
 MANIFEST_FILE={{ ManifestFileName }}
+
+MANIFEST_FILE_DIR={{ ManifestFileDir }}
```
dorfire: (same here)

src/BuildScriptGenerator/DotNetCore/DotnetCoreBashBuildSnippetProperties.cs
```diff
@@ -37,6 +37,8 @@ public class DotNetCoreBashBuildSnippetProperties
 
         public string ManifestFileName { get; set; }
 
+        public string ManifestFileDir { get; set; }
```
dorfire: maybe this class could inherit from `BaseBashBuildScriptProperties`, or from a new common base class, to avoid this declaration's repetition?

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -70,6 +70,12 @@ internal class BuildCommand : CommandBase
             Description = "Additional information used by this tool to generate and run build scripts.")]
         public string[] Properties { get; set; }
 
+        [Option(
+            OptionTemplates.ManifestFileDir,
+            CommandOptionType.SingleValue,
+            Description = "The path to a directory where build manifest file is to be put into.")]
+        public string ManifestFileDir { get; set; }
```
dorfire: Annotate with `DirectoryExists`

src/BuildScriptGeneratorCli/Commands/OptionTemplates.cs
```diff
@@ -10,5 +10,6 @@ internal static class OptionTemplates
         public const string Platform = "-l|--platform <name>";
         public const string PlatformVersion = "--language-version <version>";
         public const string Property = "-p|--property <key-value>";
+        public const string ManifestFileDir = "--manifest-file-dir <directory-path>";
```
dorfire: could be shortened to `--manifest-dir` (regardless of shortening the corresponding C# member)

src/startupscriptgenerator/src/common/buildManifestHelper.go
```diff
@@ -7,49 +7,81 @@ package common
...
+	if *manifestFileDirPathPtr == "" {
+		manifestFileFullPath = filepath.Join(fullAppPath, consts.BuildManifestFileName)
+	} else {
+		providedPath := *manifestFileDirPathPtr
+		absPath, err := filepath.Abs(providedPath)
+		if err != nil || !PathExists(absPath) {
+			fmt.Printf(
+				"Error: Provided manifest file directory path '%s' is not valid or does not exist.\n",
+				providedPath)
+			os.Exit(1)
```
dorfire: Move magic val `1` to a const

src/startupscriptgenerator/src/common/buildManifestHelper.go
```diff
@@ -7,49 +7,81 @@ package common
...
+		}
+
+		manifestFileFullPath = filepath.Join(absPath, consts.BuildManifestFileName)
+		if !FileExists(manifestFileFullPath) {
+			fmt.Printf("Error: Could not file manifest file '%s' at '%s'.\n", consts.BuildManifestFileName, absPath)
+			os.Exit(1)
```
dorfire: same here

src/startupscriptgenerator/src/common/logging.go
```diff
@@ -61,7 +59,7 @@ func (logger *Logger) logTrace(sev contracts.SeverityLevel, format string, a ...
 	message := fmt.Sprintf(format, a...)
 
 	// Uncomment the follwing line to see the trace messages on standard out
-	//fmt.Println(message)
+	fmt.Println(message)
```
dorfire: was this intended?

src/startupscriptgenerator/src/dotnetcore/main.go
```diff
@@ -27,6 +27,11 @@ func main() {
 		"runFromPath",
 		"",
 		"The path to the directory where the output is copied and run from there.")
+	manifestFileDirPathPtr := flag.String(
+		"manifestFileDirPath",
```
dorfire: same as my suggestion for the C# parameter: `manifestFileDirPath` -> `manifestDir`

src/startupscriptgenerator/src/common/buildManifestHelper.go
```diff
@@ -7,49 +7,81 @@ package common
...
-	VirtualEnvName           string
-	PackageDir               string
-	CompressedVirtualEnvFile string
-	StartupDllFileName       string
-}
-
-var _buildManifest = BuildManifest{}
-var _readManifest = false
+	// Flag to indicate of the manifest file exists or not
+	Exists bool
```
dorfire: this seems (IMO) like a rather non-standard way to signal existence. Why not just set the `BuildManifest` value to `nil`?

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -65,7 +60,7 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
 			scriptBuilder.WriteString("fi\n")
 		}
 	} else {
-		if builtByOryx {
+		if gen.Manifest.Exists {
```
dorfire: consider my comment above: this could have been `if gen.Manifest != nil {`

src/startupscriptgenerator/src/php/main.go
```diff
@@ -14,19 +14,26 @@ func main() {
 	common.PrintVersionInfo()
 
 	appPathPtr := flag.String("appPath", ".", "The path to the application folder, e.g. '/home/site/wwwroot/'.")
+	manifestFileDirPathPtr := flag.String(
+		"manifestFileDirPath",
```
dorfire: this flag is essentially the same across generators - the `flag.String` can moved into our `common` package and used by them. this will decrease repetition and make it easier to change this flag down the road

src/startupscriptgenerator/src/common/buildManifestHelper.go
kichalla: Yeah, actually that was the first thing I tried too..i.e accessing it as a pointer and checking for `nil`, but the thing is it caused the code to be awkward impacting readability..so where ever we did `manifest.VirtualEnvironmentName` now we need to null check for `manifest` and then do it...golang does not have `manfiest?.VirtualEnvironmentName` like in C#.

src/startupscriptgenerator/src/common/logging.go
kichalla: Nope. good catch. will remove.

src/startupscriptgenerator/src/php/main.go
kichalla: Hmm, I will see how this can be achieved, but I agree that this is becoming repetitive and is error-prone

src/startupscriptgenerator/src/common/buildManifestHelper.go
dorfire: possible solution:
1. create a BuildManifest [interface](https://gobyexample.com/interfaces)
2. create a `TomlFileBuildManifest` implementation
3. create a `NilBuildManifest` implementation that just returns `nil`s or empty strings for every field

src/startupscriptgenerator/src/common/buildManifestHelper.go
kichalla: For now seems like I cannot do it since constants values are currently only 'strings'. Once we fix the constants generating tool to provide ability to specify the type of value, we could fix these.

src/startupscriptgenerator/src/common/buildManifestHelper.go
kichalla: See my comment above

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
kichalla: see my comment above regarding using `nil` check

src/startupscriptgenerator/src/common/buildManifestHelper.go
kichalla: That would involve using reflection on the interface to check whether a method received a `TomlFileBuildManifest` or `NilFileBuildManifest` to figure if a manifest file exists to not. It feels a bit heavy for that IMO. I now updated this code to create a wrapper type instead. Let me know how it looks. It simulates a Nullable type in c# (but not exactly though). In C#, we have `HasValue` and `Value` properties

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
kichalla: I cannot do it here because we can create the directory if it does not already exist. It is just a matter of convenience to the end user.

src/BuildScriptGenerator/DotNetCore/DotnetCoreBashBuildSnippetProperties.cs
kichalla: The script templates themselves do not inherit each other, so I am not too keen on adding an inheritance hierarchy for their supporting classes. Let me know what you think. Inheritance creates a  rigid model IMO (which was the reason originally for me to move away dotnet core separately as it did not fit nicely with the interpreted languages which the base build script is mostly suited for)

dorfire: fair enough 👍

src/startupscriptgenerator/src/common/buildManifestHelper.go
kichalla: I simplified this in latest commit. Instead of checking for Exists, I just check if a value is empty or not. Take a look and let me know.

src/startupscriptgenerator/src/common/buildManifestHelper.go
dorfire: even still, this could be a const in this module (at the top)

src/startupscriptgenerator/src/common/commandLineSwitches.go
```diff
@@ -0,0 +1,16 @@
+// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+package common
+
+import "flag"
+
+func GetManifestDirSwitch() *string {
```
dorfire: 1. does this have to be a function? can it just be an exported variable?
2. these are called Flags, not Switches, in the Go jargon - let's keep aligned to it

src/startupscriptgenerator/src/common/buildManifestHelper.go
```diff
@@ -7,49 +7,76 @@ package common
...
 	"common/consts"
-	"log"
+	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/BurntSushi/toml"
 )
 
 type BuildManifest struct {
```
dorfire: Some members in this struct end with `File`, some end with `FileName` - can we align the names to the same convention?

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -65,7 +60,8 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
 			scriptBuilder.WriteString("fi\n")
 		}
 	} else {
-		if builtByOryx {
+		if gen.Manifest.StartupDllFileName != "" {
+			scriptBuilder.WriteString("echo Found startup dll name from manifest file\n")
```
dorfire: dll -> DLL

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -65,7 +60,8 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
 			scriptBuilder.WriteString("fi\n")
 		}
 	} else {
-		if builtByOryx {
+		if gen.Manifest.StartupDllFileName != "" {
```
dorfire: 👍

src/startupscriptgenerator/src/common/buildManifestHelper.go
kichalla: I agree regarding the naming differences, but I do not want to handle that in this PR because it requires a change in the build phase too and we need to consider backcompat scenarios too.

PR #248, opened 2019-07-02
Moved Docker's executable name to a const

dorfire:


tests/Oryx.Tests.Common/DockerCli.cs
```diff
@@ -16,6 +16,7 @@ namespace Microsoft.Oryx.Tests.Common
     public class DockerCli
     {
         private const string CreatedContainerPrefix = "oryxtests_";
+        private const string DockerExe = "docker";
```
arroyc: nit: DockerExecutable instead of DockerExe, exe sounds like platform oriented (windows)

PR #249, opened 2019-07-03
adding support for nodejs 12.5.0 to build and runtime images

arroyc:


images/build/Dockerfile
```diff
@@ -164,9 +164,11 @@ RUN chmod a+x /tmp/__nodeVersions.sh \
  && ~/n/bin/n -d 10.1.0 \
  && ~/n/bin/n -d 10.10.0 \
  && ~/n/bin/n -d 10.14.2 \
+ && ~/n/bin/n -d 12.5.0 \
```
gpcastro: remove? Why do we need to install this version hardcoded as well?

images/build/Dockerfile
```diff
@@ -235,13 +237,16 @@ RUN set -ex \
  && ln -s 10.1.0 /opt/nodejs/10.1 \
  && ln -s 10.10.0 /opt/nodejs/10.10 \
  && ln -s 10.14.2 /opt/nodejs/10.14 \
+ && ln -s 12.5.0 /opt/nodejs/12.5 \
```
gpcastro: Remove?

images/runtime/node/12.5/Dockerfile.base
```diff
@@ -0,0 +1,88 @@
+FROM oryx-run-base AS node-original
```
gpcastro: Let's just add the file under "12", since this will be a major-version only option

PR #250, opened 2019-07-03
Updated Python 3.6

dorfire:


build/constants.yaml
```diff
@@ -51,7 +51,7 @@
     python-base-tag: 20190610.2
     pip-version: 19.0.3
     python27-version: 2.7.16
-    python36-version: 3.6.8
+    python36-version: 3.6.9
```
kichalla: run the constants generator?

dorfire: yeah, not done with this PR yet, should have marked it as WIP

PR #251, opened 2019-07-03
 Introduced convenience constructor in UnsupportedVersionException

dorfire:


src/BuildScriptGeneratorCli/Commands/InstallRunTimeCommand.cs
```diff
@@ -59,12 +49,7 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
 
         internal override bool IsValidInput(IServiceProvider serviceProvider, IConsole console)
         {
-            if (!Directory.Exists(InstallationPrefix))
-            {
-                console.WriteErrorLine($"Could not find the installation prefix directory '{InstallationPrefix}'.");
-                return false;
-            }
-
+            // TODO: validate?
```
kichalla: You need not override this method. I think it defaults to true?

dorfire: it does - i just wanted to leave its body here for now, in case new options are necessary

PR #253, opened 2019-07-04
 Add apt-https package to actually fetch MSODBC package info

gpcastro:


images/runtime/python/install-dependencies.sh
```diff
@@ -16,6 +16,7 @@ apt-get update \
         default-libmysqlclient-dev \
         unzip \
         libodbc1 \
+        apt-transport-https \
```
dorfire: i presume the order doesn't matter?

gpcastro: correct

PR #258, opened 2019-07-05
Do not add common runtime image to the list of images built

gpcastro:
This prevents the image from being added to the list of images built, which is used to push the images to the repos.


gpcastro:
> I would actually rather this image pushed into MCR (https://dev.azure.com/devdiv/DevDiv/_workitems/edit/943285)

Why? It is built as part of every build to make sure we get the latest OS updates. If we were to push this to MCR, we'd need to build it + update the tags, so we need to justify this extra steps - build and release definition + auto tag update, and account for the fact that they might fail to and need maintenance.

PR #262, opened 2019-07-08
Removed unused build script & locked Debian versions to stretch

dorfire:


images/build/python/prereqs/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Install Python build prerequisites
-FROM buildpack-deps:stable
+FROM buildpack-deps:stretch
```
kichalla: Should we even update the run-base image too? From `buildpack-deps:curl` to `buildpack-deps:stretch-curl`? cc @gpcastro

dorfire: good point - done

PR #263, opened 2019-07-08
Logging `pip` download events

dorfire:


src/BuildScriptGeneratorCli/ITextStreamProcessor.cs
```diff
@@ -0,0 +1,12 @@
+﻿// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    public interface ITextStreamProcessor
```
gpcastro: does it have to be public? can we make it internal?

src/BuildScriptGeneratorCli/PipDownloadEventLogger.cs
```diff
@@ -0,0 +1,57 @@
...
+// --------------------------------------------------------------------------------------------
+
+using System;
+using System.Linq;
+using Microsoft.Extensions.Logging;
+using Microsoft.Oryx.Common;
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    internal class PipDownloadEventLogger : ITextStreamProcessor
```
gpcastro: add some description for this is supposed to do

src/BuildScriptGeneratorCli/PipDownloadEventLogger.cs
```diff
@@ -0,0 +1,57 @@
...
+        public PipDownloadEventLogger(ILogger logger)
+        {
+            _logger = logger;
+        }
+
+        public void ProcessLine(string line)
+        {
+            if (_currentDownload != null)
+            {
+                _currentDownload.AddProperty("nextLine", line.Trim());
```
gpcastro: can `line` contain PII?

src/BuildScriptGeneratorCli/PipDownloadEventLogger.cs
```diff
@@ -0,0 +1,57 @@
...
+                _currentDownload.Dispose();
+                _currentDownload = null;
+            }
+
+            if (line.Contains(PipDownloadMarker))
```
gpcastro: "Downloading" seems too broad... isn't there a regex that we can use instead?

dorfire: This seems to be enough - especially considering *this checker only runs for Python builds*.
We could come up with a regex, but my guess is that it'll be difficult to write and to maintain

PR #267, opened 2019-07-08
Utilized option attributes

dorfire:


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -38,6 +38,7 @@ internal class BuildCommand : CommandBase
         };
 
         [Argument(0, Description = "The source directory.")]
+        [DirectoryExists]
```
kichalla: The question is whether this works for the cases like `.`, `c:\foo\..\dir1` etc.

dorfire: Yeah, it works with relative paths as well.

kichalla: Currently we support doing `oryx build` without supplying any value - in which case current directory is taken. This change will break it?

dorfire: it'll probably break it - i'm surprised we don't have a test for that - adding one.

kichalla: I think we do have integration tests for it

tests/Oryx.BuildImage.Tests/OryxCommandTest.cs
```diff
@@ -6,14 +6,36 @@
...
+                CommandToExecuteOnRun = "oryx",
+                CommandArguments = new[] { "build" },
+                WorkingDirectory = "/tmp"
+            });
+
+            // Assert
+            RunAsserts(
+                () =>
+                {
+                    Assert.Contains("Error: Could not detect", result.StdErr);
```
kichalla: Instead of an error case, we could just build a sample app of any platform and test for sucess

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
kichalla: Revert this change then?

src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
```diff
@@ -19,6 +19,7 @@ internal class BuildScriptCommand : CommandBase
         public const string Name = "build-script";
 
         [Argument(0, Description = "The source directory.")]
+        [DirectoryExists]
```
kichalla: Same here. Revert this change.

PR #269, opened 2019-07-09
Python startup script: Increase the default timeout value for gunicorn

kichalla:


src/startupscriptgenerator/src/python/scriptgenerator.go
```diff
@@ -240,19 +240,29 @@ func (gen *PythonStartupScriptGenerator) getFlaskStartupModule() string {
 
 // Produces the gunicorn command to run the app
 func (gen *PythonStartupScriptGenerator) getCommandFromModule(module string, appDir string) string {
-	args := ""
+
+	// Default to AppService's timeout value (in seconds)
+	args := "--timeout 600"
```
arroyc: this long timeout?

kichalla: yes, that is the default that AppService currently has

PR #270, opened 2019-07-09
[DO NOT MERGE] experiment semver comparison

arroyc:
This PR is just to validate a potential bug in our code/semver api that we consume to compare versions


tests/BuildScriptGenerator.Tests/SemanticVersionResolverTest.cs
```diff
@@ -56,11 +56,13 @@ public void MajorAndMinorVersionsProvided_MatchesMajorMinorAndLatestPatchVersion
         [InlineData("1")]
         [InlineData("1.x")]
         [InlineData("1.x.x")]
+        [InlineData("3")]
+        [InlineData("4")]
```
arroyc: @kichalla  I added these two test cases where I'm seeing failure, if we decide to have only major versions this can be a problem

PR #272, opened 2019-07-10
remove remaining references to oryxdevms docker repo

arroyc:
remove some more references to oryxdevms

### DIFF
--- a/tests/Oryx.Integration.Tests/DotNetCoreEndToEndTests.cs
+++ b/tests/Oryx.Integration.Tests/DotNetCoreEndToEndTests.cs
@@ -806,7 +806,7 @@ namespace Microsoft.Oryx.Integration.Tests
                     "-c",
                     buildImageScript
                 },
-                $"oryxdevms/dotnetcore-{dotnetcoreVersion}",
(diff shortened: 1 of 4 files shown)


tests/Oryx.Integration.Tests/DotNetCoreEndToEndTests.cs
```diff
@@ -806,7 +806,7 @@ public async Task CanBuildAndRunApp_WhenOutputIsZipped_AndIntermediateDir_IsUsed
                     "-c",
                     buildImageScript
                 },
-                $"oryxdevms/dotnetcore-{dotnetcoreVersion}",
+                $"oryxdevmcr.azurecr.io/public/oryx/dotnetcore-{dotnetcoreVersion}",
```
kichalla: Using such a long name when working/testing in local environment is really inconvenient. We should have a shorter tag too IMO. For exmample `oryx/build`, `oryx/node-8.1` etc. What do you guys think?

dorfire: I agree. In addition, we should put that name in a variable so that it's easier to change later

arroyc: sure ..  I will do the clean up in a separate work-item I created

PR #277, opened 2019-07-12
fix buildpack integreation tests for node

arroyc:


build/testIntegration.sh
```diff
@@ -54,9 +54,10 @@ echo
 testProjectName="Oryx.Integration"
 cd "$TESTS_SRC_DIR/$testProjectName.Tests"
 
-# These two images are used in Buildpacks-related integration tests
+# These three images are used in Buildpacks-related integration tests
 docker pull "heroku/buildpacks:18"
 docker pull "heroku/pack:18"
+docker pull "mcr.microsoft.com/oryx/pack-stack-base"
```
arroyc: we need to pull this  since builder.toml has been changed @dorfire

PR #278, opened 2019-07-12
Refactoring: Pass generator context instead of just ISourceRepo

kichalla:


src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -147,7 +147,7 @@ internal class DefaultBuildScriptGenerator : IBuildScriptGenerator
                 platformVersion = ctx.LanguageVersion;
                 if (string.IsNullOrEmpty(platformVersion))
                 {
-                    var detectionResult = userSuppliedPlatform.Detect(ctx.SourceRepo);
+                    var detectionResult = userSuppliedPlatform.Detect(ctx);
```
kichalla: We need this access to context for example when we want to get information from the build properties.

gpcastro: can you give an example of where we'd need the entire context here?

kichalla: Build properties for example, in case of dotnet core platform trying to get information from a build property like `project` which is then used by a component called projectfileprovider. If the provider finds the 'project' property, it will use it.

gpcastro: Makes sense. Thanks

PR #279, opened 2019-07-12
added some more tests

arroyc:
add some more tests for semanticversionresolver

### DIFF
--- a/tests/BuildScriptGenerator.Tests/SemanticVersionResolverTest.cs
+++ b/tests/BuildScriptGenerator.Tests/SemanticVersionResolverTest.cs
@@ -2,6 +2,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 // --------------------------------------------------------------------------------------------
+
 using Xunit;
 
 namespace Microsoft.Oryx.BuildScriptGenerator.Tests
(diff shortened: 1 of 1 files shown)


tests/BuildScriptGenerator.Tests/SemanticVersionResolverTest.cs
```diff
@@ -94,5 +95,21 @@ public void SpecificVersionSpecified()
...
         }
+
+        [Theory]
+        [InlineData("1", "1.2.3", int.MinValue)]
+        [InlineData("1.2", "1.2.3", -1)]
+        [InlineData("1.3.4", "1.2.3", 1)]
+        [InlineData("3.5", "2.1.1", 1)]
+        [InlineData("3.5", "4.1", -1)]
+        [InlineData("3.5", "4", int.MinValue)]
+        public void CompareVersionsTest(string providedVersion, string supportedVersion, int resultExpected)
```
kichalla: What is the pivot (scenario) that you are trying to test here? And I would put that pivot as part of the test name.

arroyc: I'm trying to compare versions as many scenarios as I could think of. I will try to add some comment or rename the test to be more meaningful

tests/BuildScriptGenerator.Tests/SemanticVersionResolverTest.cs
```diff
@@ -94,5 +95,40 @@ public void SpecificVersionSpecified()
...
+        {
+            // Arrange and Act 
+            int comparisonResult= SemanticVersionResolver.CompareVersions(providedVersion, supportedVersion);
+
+            // Assert 
+            Assert.Equal(comparisonResult, resultExpected);
+        }
+
+        [Theory]
+        [InlineData("1", "1.2.3", int.MinValue)]
```
kichalla: Looking at code of `CompareVersions`, it appears that the value of `int.MinValue` would occur when the parsing of the version like let's say `1` here would throw an exception, right?
Do you think we should instead assume `1` to be `1.0.0` and then do a comparison to get the value?

arroyc: actually I wanted to test the exception, that's the reason for this test. Usually if the provided version is "1" then we convert it to max satisfying version for example 1.2.4 and then this method gets called.

PR #290, opened 2019-07-19
Revert back php image base versions to avoid libssl error

kichalla:


build/__phpVersions.sh
```diff
@@ -1,7 +1,7 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
-PHP_BUILD_BASE_TAG='20190716.3'
-PHP_RUNTIME_BASE_TAG='20190716.3'
+PHP_BUILD_BASE_TAG='20190710.4'
```
kichalla: Between these versions the `libssl1.0.2` library is missing. I have meanwhile started a base image build but wanted to put this change in to kick off a CI build. Once I test successfully with the latest base build image i will create a new PR.

image

tests/SampleApps/php/sqlsrv-example/index.php
```diff
@@ -1,12 +1,22 @@
...
-echo json_encode($rows);
+	$rows = [];
+	while ($row = $query->fetch(PDO::FETCH_ASSOC))
+		$rows[] = $row;
+
+	echo json_encode($rows);
+}
+catch (PDOException $e) {
+	print("Error connecting to SQL Server.");
+	die(print_r($e));
```
dorfire: `die($e->getTraceAsString())`

kichalla:
> Do we have a user story to fix the base images and be able to update?

Could you elaborate more on this? You mean like the no-cache thing that we have been discussing? No, not yet.

PR #292, opened 2019-07-22
Php runtime image: Install libssl1.0 for fixing Sql Server driver connectivity

kichalla:


images/runtime/php/Dockerfile.base.template
```diff
@@ -80,7 +80,7 @@ RUN apt-get update \
 #  - https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server
 RUN set -eux; \
     if [[ $PHP_VERSION == 7.1.* || $PHP_VERSION == 7.2.* || $PHP_VERSION == 7.3.* ]]; then \
-        apt-get update && apt-get install -y gnupg2 apt-transport-https && \
+        apt-get update && apt-get install -y libssl1.0 gnupg2 apt-transport-https && \
```
arroyc: curious why do we need to install these libraries now? is it because the php base images do not have them anymore?

kichalla: Yes, right. Recently a change was made at their end which caused a regression where the SQL Server drivers were unable to connect to a database.

PR #293, opened 2019-07-22
support jamspell

arroyc:

### DIFF
--- a/images/build/Dockerfile
+++ b/images/build/Dockerfile
@@ -289,9 +289,15 @@ RUN apt-get update \
  && apt-get install -y --no-install-recommends \
         tk-dev \
         uuid-dev \
+        libpcre3 \
+        libpcre3-dev \
+        swig \
+        locales \
  && rm -rf /var/lib/apt/lists/*
+
 # https://github.com/docker-library/python/issues/147
 ENV PYTHONIOENCODING UTF-8
+ENV PYTHONPATH /opt/python/2.7.16/lib:/opt/python/3.6.9/lib:/opt/python/3.7.4/lib:/opt/python/3.8.0b2/lib
 COPY build/__pythonVersions.sh /tmp
 COPY --from=py27-build-base /opt /opt
 COPY --from=py36-build-base /opt /opt
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -49,12 +49,25 @@ python_bin=python
 
 {{ else }}
 
+# We need to use the python binary selected by benv
+python_bin=$python
+
+# Detect the location of the cd /tmp/-packages to add the .pth file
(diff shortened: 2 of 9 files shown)


images/build/Dockerfile
```diff
@@ -289,9 +289,15 @@ RUN apt-get update \
...
         uuid-dev \
+        libpcre3 \
+        libpcre3-dev \
+        swig \
+        locales \
  && rm -rf /var/lib/apt/lists/*
+
 # https://github.com/docker-library/python/issues/147
 ENV PYTHONIOENCODING UTF-8
+ENV PYTHONPATH /opt/python/2.7.16/lib:/opt/python/3.6.9/lib:/opt/python/3.7.4/lib:/opt/python/3.8.0b2/lib
```
kichalla: Is this change still required even if virtualenv is used?

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -49,12 +49,25 @@ python_bin=python
 
 {{ else }}
 
+# We need to use the python binary selected by benv
```
kichalla: Should we revert these changes and only support virtualenv for now?

tests/Oryx.Integration.Tests/Python/PythonJamSpellAppTests.cs
```diff
@@ -0,0 +1,110 @@
...
+            // Arrange
+            var appName = "jamspell-flask-app";
+            var volume = CreateAppVolume(appName);
+            var appDir = volume.ContainerDir;
+            var buildScript = new ShellScriptBuilder()
+               .AddCommand($"oryx build {appDir} --platform python --platform-version {pythonVersion}")
+               .ToString();
+            var runScript = new ShellScriptBuilder()
+                .AddCommand($"oryx -appPath {appDir} -bindPort {ContainerPort}")
+                .AddCommand(DefaultStartupFilePath)
```
kichalla: As discussed let's see if setting the locale related environment variables here (simulating how an AppService end user would can do it through app settings) fixes the issue

tests/Oryx.Integration.Tests/Python/PythonJamSpellAppTests.cs
```diff
@@ -0,0 +1,110 @@
...
+                {
+                    var data = await _httpClient.GetStringAsync($"http://localhost:{hostPort}/");
+                    Assert.Contains("[I am the begt spell cherken!]", data);
+                    Assert.Contains("I am the best spell checker!", data);
+                });
+        }
+
+        [Theory(Skip = "locale bug, https://github.com/bakwc/JamSpell/issues/17")]
+        [MemberData(nameof(TestValueGenerator.GetPythonVersions), MemberType = typeof(TestValueGenerator))]
+        public async Task CanBuildAndRun_JamSpellFlaskApp_PackageDir(string pythonVersion)
```
kichalla: revert?

PR #295, opened 2019-07-24
Node runtime image: Pin PM2 to a specific version

kichalla:


images/runtime/node/10.1/Dockerfile.base
```diff
@@ -83,4 +83,4 @@ ARG CACHEBUST=0
 
 # PM2 is supported as an option when running the app,
 # so we need to make sure it is available in our images.
-RUN npm install -g pm2
+RUN npm install -g pm2@3.5.1
```
gpcastro: how about we move this to a script that is run by every image, so we don't have to change it everywhere when we update?

kichalla:
> Should we move apt-get install unzip to this new script?

I considered it initially but did not do it because that 'line' having tools to be installed probably wouldn't change much and so we can keep using that cached layer. But let's say if we put `unzip` in the script then anytime the script changes we would be bust the cache and would be reinstalling `unzip` again. To be honest I am not too strong on my decision either.

PR #296, opened 2019-07-24
python pip cache to be  located at /usr/local/share/pip-cache

arroyc:
python package cache to be located at /usr/local/share/pip-cache

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -1,6 +1,11 @@
 declare -r TS_FMT='[%T%z] '
 declare -r REQS_NOT_FOUND_MSG='Could not find requirements.txt; Not running pip install'
 echo "Python Version: $python"
+PIP_CACHE_DIR=/usr/local/share/pip-cache
+
+if [ ! -d "$PIP_CACHE_DIR" ];then
+	mkdir -p $PIP_CACHE_DIR
+fi
 
 {{ if VirtualEnvironmentName | IsNotBlank }}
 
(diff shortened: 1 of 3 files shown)


src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -1,6 +1,11 @@
 declare -r TS_FMT='[%T%z] '
 declare -r REQS_NOT_FOUND_MSG='Could not find requirements.txt; Not running pip install'
 echo "Python Version: $python"
+PIP_CACHE_DIR=/usr/local/share/pip-cache
```
kichalla: Curious, why this particular location and not something like let's say `/tmp/pip-cache`?

kichalla: Can you kick off a Nightly build against this change? I am curious if this stuff works with `oryxtests/build` image? You probably already tried it?

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -34,7 +39,7 @@ then
 	echo "Done in $ELAPSED_TIME sec(s)."
 
 	echo "Running pip install..."
-	pip install --prefer-binary -r requirements.txt | ts $TS_FMT
+	pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r requirements.txt | ts $TS_FMT
```
kichalla: How is the performance of a second build? do you see a similar experience as before this change?

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
arroyc: sure .. I know what you are trying to get .. but I added permission as we have for yarn cache, but yet. nightly-build is a good idea. Answering your first question, I tried to keep the location/path similar to the yarn-cache

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
arroyc: locally I didn't see any difference (i.e. running integration tests etc on local containers)

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
kichalla: Cool

PR #302, opened 2019-07-27
Added built binaries to GitHub releases

dorfire:


kichalla:
> what use is this? Our binaries currently do not work outside our images since it depends on the SDKs being installed in the expected locations. Adding the binaries would only add confusion, since users would (rightfully) expect them to work and file issues.

I believe @dorfire was trying to work on the updating binaries dynamically by AppService? But I think we need a bit more thought here than just having binaries in drop share. I have some thoughts about this. Let's discuss in today's scrum meeting.

dorfire:
Need to check with GitHub if they use a CDN

PR #317, opened 2019-08-15
update node 12 to 12.8.1

arroyc:
node runtime base image reference for 12.8  needs to be updated once I have a successful build/release

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=2947236


images/runtime/node/12.8/Dockerfile.base
```diff
@@ -0,0 +1,64 @@
+FROM oryx-run-base AS node-original
```
kichalla: Delete the 12.7 Dockerfile?

PR #318, opened 2019-08-15
[.NET Core]Do not publish output if target directory is not supplied

kichalla:


src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -61,43 +66,273 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
...
+                BashBuildScriptSnippet = scriptBuilder.ToString(),
+                IsFullScript = true,
             };
-            var script = TemplateHelper.Render(
-                TemplateHelper.TemplateResource.DotNetCoreSnippet,
-                templateProperties,
-                _logger);
-            return new BuildScriptSnippet { BashBuildScriptSnippet = script, IsFullScript = true };
+
+            void AddScriptToCopyToIntermediateDirectory()
```
gpcastro: having those inline methods make things harder to read IMO. How about we add this to a separate builder class with fluid API, without the `Add` in each method? Or as extension methods within a "dotnet" child namespace?
```
buildScript
     .RunPreBuildCommand()
     .ZipAllOutput()
     .RunPostBuildCommand()
     .CreateManifestFile()
...
```

PR #321, opened 2019-08-20
move appinsight codeless configuration into runtime images

arroyc:
this is pending on a node run base image publish to mcr .. once that image is in mcr images/runtime/node/generateDockerfiles.sh base_build_number needs to be changed accordingly


src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -44,6 +44,49 @@ const NodeWrapperPath = "/opt/node-wrapper/"
 const LocalIp = "0.0.0.0"
 const inspectParamVariableName = "ORYX_NODE_INSPECT_PARAM"
 
+// Checks if the application insights needs to be enabled for the current runtime
+func shouldApplicationInsightsBeConfigured() bool {
```
arroyc: @markwolff, for the node apps that have applicationinsights as dependency, should we assume they will have their own applicationinsightloader.js? or the loader will always be injected by oryx?

markwolff: Discussed/resolved this offline with Arjun. For apps which are deployed with app insights as a dependency, they will not provide their own loader. We should not start app insights for them because they are either doing it themselves or removed app insights (and forgot to remove the dependency)

images/runtime/node/installDependencies.sh
```diff
@@ -20,3 +20,7 @@ fi
 # PM2 is supported as an option when running the app,
 # so we need to make sure it is available in our images.
 npm install -g pm2@3.5.1
+
+# Application-Insights is supported as an option for telemetry when running the app,
+# so we need to make sure it is available in our images.
+npm install -g applicationinsights@1.4.1
```
markwolff: Just for clarification, when is this run? When the images are checked in to the Docker store; aka before exposed to users?

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -44,6 +44,49 @@ const NodeWrapperPath = "/opt/node-wrapper/"
...
+}
+
+// Create application insights loader fiule
+func createApplicationInsightsLoaderFile(appInsightsLoaderFilePath string) {
+	appInsightsLoaderFileContent := `// --------------------------------------------------------------------------------------------
+	// Copyright (c) Microsoft Corporation. All rights reserved.
+	// Licensed under the MIT license.
+	// --------------------------------------------------------------------------------------------
+
+	// Created by Oryx
```
markwolff: nit: since this is duplicated from the build image loader, is it possible to reuse it? I imagine this will be more built-out in the future, specifically to report to the user why attach failed for their app (via UX, as windows currently does with Kudu). It would be ideal to not have to maintain this in 2 places

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -44,6 +44,49 @@ const NodeWrapperPath = "/opt/node-wrapper/"
...
+			try {
+				var appInsightsSDKVersion = appInsights.defaultClient.context.keys.internalSdkVersion;
+				envelope.tags[appInsightsSDKVersion] = "ad_" + envelope.tags[appInsightsSDKVersion];
+			} catch(e) {}
+
+			return true;
+		}
+		if (process.env.APPINSIGHTS_INSTRUMENTATIONKEY && process.env.APPLICATIONINSIGHTS_CODELESS_ENABLED) {
+			appInsights
+				.setup()
```
markwolff: We may want to add `.setSendLiveMetrics(true)` here since that is how it being configured for AKS. I will circle back on this.
@AlexKlimovMS @Mikhail-Pranovich

src/startupscriptgenerator/src/node/scriptgenerator.go
arroyc: that's a good point .. I'm now thinking to get rid of this logic in build image, because I feel we don't need that there anymore, so this will only stay in runtime

src/startupscriptgenerator/src/node/scriptgenerator.go
arroyc: okie,  is that confirmed then? I can add it here

images/runtime/node/installDependencies.sh
arroyc: yes(kind  of), 
From user prospective: when user run his app on the specific run image, app service pull that specific image from docker store (read mcr), and creates a runtime container which contains this

src/startupscriptgenerator/src/node/scriptgenerator.go
markwolff: Yes confirmed you can add it

tests/Oryx.Integration.Tests/Node/NodeTestWithAppInsightsConfigured.cs
```diff
@@ -22,33 +22,26 @@ public NodeTestWithAppInsightsConfigured(ITestOutputHelper output, TestTempDirTe
...
         public async Task CanBuildAndRun_NodeApp_WithAppInsights_Configured(string nodeVersion)
         {
             // Arrange
             var appName = "linxnodeexpress";
             var volume = CreateAppVolume(appName);
             var appDir = volume.ContainerDir;
-            var spcifyNodeVersionCommand = "--platform nodejs --language-version=" + nodeVersion;
+            var spcifyNodeVersionCommand = "--platform nodejs --platform-version=" + nodeVersion;
             var aIKey = "APPINSIGHTS_INSTRUMENTATIONKEY";
+            var aIEnabled = "APPLICATIONINSIGHTS_CODELESS_ENABLED";
```
markwolff: This will likely be renamed. Putting a note here for when I know for sure what it will be instead.

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -44,6 +44,50 @@ const NodeWrapperPath = "/opt/node-wrapper/"
 const LocalIp = "0.0.0.0"
 const inspectParamVariableName = "ORYX_NODE_INSPECT_PARAM"
 
+// Checks if the application insights needs to be enabled for the current runtime
+func shouldApplicationInsightsBeConfigured() bool {
+	nodeAppInsightsKeyEnv := os.Getenv("APPINSIGHTS_INSTRUMENTATIONKEY")
+	nodeAppInsightsEnabledEnv := os.Getenv("APPLICATIONINSIGHTS_CODELESS_ENABLED")
+
+	if nodeAppInsightsEnabledEnv != "" && strings.ToLower(nodeAppInsightsKeyEnv) == "true" {
```
kichalla: Is this condition correct? the 'true' check should be with the 'enabled' key?

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -177,13 +221,16 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
...
-	logger.LogInformation("Looking for App-Insights loader injected by Oryx and export to NODE_OPTIONS if needed")
-	if gen.Manifest.InjectedAppInsights != "" {
+	logger.LogInformation("Looking for App-Insights loader generated by Oryx and export to NODE_OPTIONS if needed")
+	if shouldApplicationInsightsBeConfigured() {
 		loaderFile := filepath.Join(gen.SourcePath, consts.NodeAppInsightsLoaderFileName)
-		if common.FileExists(loaderFile) {
-			var nodeOptions = "'--require ./" + consts.NodeAppInsightsLoaderFileName + " '$NODE_OPTIONS"
-			scriptBuilder.WriteString("export NODE_OPTIONS=" + nodeOptions + "\n")
+
+		if common.FileExists(loaderFile) !=true {
```
kichalla: `!common.FileExists(loaderFile)`?

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -177,13 +221,16 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
...
+			createApplicationInsightsLoaderFile(loaderFile)			
 		}
+
+		var nodeOptions = "'--require ./" + consts.NodeAppInsightsLoaderFileName + " '$NODE_OPTIONS"
```
kichalla: space after the ending single quote?

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -44,6 +44,50 @@ const NodeWrapperPath = "/opt/node-wrapper/"
 const LocalIp = "0.0.0.0"
 const inspectParamVariableName = "ORYX_NODE_INSPECT_PARAM"
 
+// Checks if the application insights needs to be enabled for the current runtime
+func shouldApplicationInsightsBeConfigured() bool {
+	nodeAppInsightsKeyEnv := os.Getenv("APPINSIGHTS_INSTRUMENTATIONKEY")
+	nodeAppInsightsEnabledEnv := os.Getenv("APPLICATIONINSIGHTS_CODELESS_ENABLED")
```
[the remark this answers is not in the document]
markwolff: A user can enable/disable application insights via the Portal. Many who instrument their code directly with app insights are already using the ikey env var. If attach were to be enabled by default (as it is Windows), there must be a way to disable attach without breaking an app which is using it already in-code.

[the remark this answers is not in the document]
gpcastro: we can inject AI if:
1) `APPINSIGHTS_INSTRUMENTATIONKEY` is defined
2) We can't find the app insights sdk as a dependency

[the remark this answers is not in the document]
arroyc: As per my understanding, new requirement is ..
we will inject AI if
APPINSIGHTS_INSTRUMENTATIONKEY and APPLICATIONINSIGHTS_CODELESS_ENABLED are defined  as env variables

If user has a dependency of application insight already in package.json that's great, if not  globally installed application insight sdk will be referenced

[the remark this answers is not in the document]
markwolff: From the Portal's perspective, it has no idea whether or not the app insights sdk is present, so it needs a way to alert the runtime instance the the user opted out of app insights. This can either be done with this new enablement flag or with the ikey flag. If we opt to just delete the ikey flag, it will impact users with stuff already deployed the using the ikey flag (they would just need to add it back I suppose). I think it is just more clear to separate the definitions of the two, but happy to discuss this further if you disagree.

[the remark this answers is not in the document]
markwolff: I will also add that windows app services instances use `ApplicationInsightsAgent_EXTENSION_VERSION` to accomplish the same thing.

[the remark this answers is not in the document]
gpcastro: OK, I'm fine with it then if it makes things easier for the portal. We just have a few more cases to consider, e.g. what if `APPLICATIONINSIGHTS_CODELESS_ENABLED` is true but `APPINSIGHTS_INSTRUMENTATIONKEY` doesn't exist?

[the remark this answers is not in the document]
arroyc: @markwolff please confirm .. as per  my understanding oryx will configure appinsight if and only if both the environment variables are defined, if only one of it is defined we are simply ignoring it.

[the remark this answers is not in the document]
markwolff: You would not want to do the injection. This should only occur if someone directly changed these environment variables. I believe in the UX we would render app insights as not enabled if this is the case.

tests/Oryx.Integration.Tests/Node/NodeTestWithAppInsightsConfigured.cs
markwolff: Can you rename this to `ApplicationInsightsAgent_EXTENSION_ENABLED`?

arroyc: I will, but waiting for the confirmation about the casing .. new variable's casing looks bit unorthodox :)

markwolff: It can be made all uppercase

PR #322, opened 2019-08-20
 Fix benv to ignore named versions (#320)

kichalla:


images/build/benv.sh
```diff
@@ -50,118 +34,93 @@ benv-resolve() {
   local value=$(echo $1 | sed 's/^.*=//')
 
   # Resolve node versions
-  if [ "$name" == "node" -o "${name::5}" == "node_" ] && [ "${value::1}" != "/" ]; then
+  if [ "$name" == "node" ] && [ "${value::1}" != "/" ]; then
```
arroyc: xunit?

kichalla:
> I saw the PR, :) can you please add some tests … there are some complicated string parsing logic which is difficult to follow

@arroyc I added a few tests to this PR now. Please note that we already had other tests cases related to `benv` in the same test class.

kichalla:
> but I think we should also include a failed test (that we discussed) which gets failed with current benv implementation, so that we don't re-investigate in future the cause

Yup, added it in latest commit. Thanks.

PR #323, opened 2019-08-21
Create a slim version of build image

kichalla:


build/__variables.sh
```diff
@@ -9,8 +9,10 @@ declare -r BUILD_CONFIGURATION="${BUILDCONFIGURATION:-Debug}"
 
 declare -r BUILD_IMAGES_BUILD_CONTEXT_DIR="$__REPO_DIR/"
 declare -r BUILD_IMAGES_DOCKERFILE="$__REPO_DIR/images/build/Dockerfile"
+declare -r BUILD_IMAGES_SLIM_DOCKERFILE="$__REPO_DIR/images/build/slim.Dockerfile"
 declare -r PACK_IMAGE_DOCKERFILE="$__REPO_DIR/images/pack-builder/pack-runner.Dockerfile"
 declare -r ORYXTESTS_BUILDIMAGE_DOCKERFILE="$__REPO_DIR/tests/images/build/Dockerfile"
+declare -r ORYXTESTS_SLIM_BUILDIMAGE_DOCKERFILE="$__REPO_DIR/tests/images/build/slim.Dockerfile"
```
kichalla: `slim.Dockerfile` => named it this way to enable syntax highlighting in VS Code which looks for `Dockerfile` extension

images/build/slim.Dockerfile
```diff
@@ -0,0 +1,286 @@
...
+ && ln -s $PYTHON37_VERSION /opt/python/3.7 \
+ && ln -s 3.7 /opt/python/3
+RUN set -ex \
+ && cd /usr/local/bin \
+ && cp -sn /opt/python/3/bin/* . \
+ # Make sure the alias 'python' always refers to Python 3 by default
+ && ln -sf /opt/python/3/bin/python python
+
+# This stage is used only when building locally
+FROM dotnet-install AS buildscriptbuilder
```
kichalla: FYI: BuildScriptGenerator needs to be part of the slim image (understandably), but the side effect of having this here is that if you are frequently making changes to the BuildScriptGenerator and want to test the change, now the building of the full image would be slow every time as the change in content of the slim image would cause a cache bust resulting in the full build image to be built again.

build/buildBuildImages.sh
```diff
@@ -48,46 +48,92 @@ fi
```
gpcastro: We finally have a reason to call this file buildBuildImage*s*.sh

arroyc:
are we releasing the slim image to acr as well?

arroyc:
curious: any modification needs to be done in pullandtag script for nightly integration test on this image?

build/buildBuildImages.sh
kichalla: LOL..yes

kichalla:
> curious: any modification needs to be done in pullandtag script for nightly integration test on this image?

I did make a change in the file `tagBuildImagesForRelease.sh` for it.

PR #324, opened 2019-08-21
Benv: Make variable name match case-insensitive and consider variable names of type platform_version (ex: dotnet_version=1)

kichalla:
Following would work now:
dotnet=1
Dotnet=1
DOTNET_VERSION=1
Dotnet_Version=1

### DIFF
--- a/images/build/benv.sh
+++ b/images/build/benv.sh
@@ -4,17 +4,54 @@
 # Licensed under the MIT license.
 # --------------------------------------------------------------------------------------------
 
+# Perform case-insensitive comparison
+matchesName() {
+  local expectedName="$1"
+  local providedName="$2"
+  local result=
+  shopt -s nocasematch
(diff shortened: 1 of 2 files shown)


images/build/benv.sh
```diff
@@ -4,17 +4,54 @@
...
 #   source benv dotnet=3
 #   dotnet --version (This should print version 3)
-[ -n "$python" ] && set -- "python=$python" "$@"
-[ -n "$php" ] && set -- "php=$php" "$@"
-[ -n "$npm" ] && set -- "npm=$npm" "$@"
-[ -n "$node" ] && set -- "node=$node" "$@"
-[ -n "$dotnet" ] && set -- "dotnet=$dotnet" "$@"
+while read benvvar; do
+  set -- "$benvvar" "$@"
+done < <(set | grep -i '^php=')
```
gpcastro: should we stick to just one pattern? I'd just keep the `_version` option

PR #328, opened 2019-08-23
fix appinsights loader content

arroyc:


src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -75,7 +75,7 @@ func createApplicationInsightsLoaderFile(appInsightsLoaderFilePath string) {
 
 			return true;
 		}
-		if (process.env.APPINSIGHTS_INSTRUMENTATIONKEY && process.env.APPLICATIONINSIGHTSAGENT_EXTENSION_ENABLED) {
+		if (process.env.APPINSIGHTS_INSTRUMENTATIONKEY && process.env.APPLICATIONINSIGHTSAGENT_EXTENSION_ENABLED == "true") {
```
markwolff: use `===` since it behaves more intuitively than `==` in JS

PR #329, opened 2019-08-24
Tag build image with the right tags

kichalla:


vsts/scripts/tagBuildImagesForRelease.sh
```diff
@@ -6,48 +6,34 @@
...
-    outFileDocker="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/build-images-dockerhub.txt"
 
     echo "Pulling the source image $buildImageName ..."
     docker pull "$buildImageName" | sed 's/^/     /'
-        
-    acrProdRepo="oryxmcr.azurecr.io/public/oryx/build"
-    acrLatest="$acrProdRepo:latest"
-    acrSpecific="$acrProdRepo:$buildNumber"    
-        
-    dockerHubRepoName="oryxprod/build"
```
kichalla: I removed this since we are not publishing to DockerHub anymore

vsts/scripts/tagBuildImagesForRelease.sh
```diff
@@ -6,48 +6,34 @@
...
     fi
 
     echo -------------------------------------------------------------------------------
 }
 
-tagBuildImage "oryxdevmcr.azurecr.io/public/oryx/build"
-tagBuildImage "oryxdevmcr.azurecr.io/public/oryx/build-slim"
\ No newline at end of file
+tagBuildImage "latest" "$buildNumber"
+tagBuildImage "slim" "slim-$buildNumber"
```
kichalla: In view of MCR team's comment here I modified it to be like this. What do you guys think of it?
https://github.com/microsoft/mcr/pull/712#pullrequestreview-279210913

so it is going to be `mcr.microsoft.com/oryx/build:slim` (indicating a moving tag or `latest`) and `mcr.microsoft.com/oryx/build:slim-20190823.1`

This feels a bit odd...like the pattern usually with debian image is `debian:stretch` and `debian:stretch-slim`, so it should be like `oryx:build` and `oryx:build-slim`, but I do not know if this is good

PR #340, opened 2019-09-04
remove explicit "git checkout" step from oryx buildstep template

arroyc:


arroyc:
this is not the issue .. not needed anymore

PR #341, opened 2019-09-04
[WIP] Publish platform sdks to Azure Storage

kichalla:


build/buildBuildImageBases.sh
```diff
@@ -5,8 +5,6 @@
 # --------------------------------------------------------------------------------------------
 #
 # This script builds some base images that are needed for the build image:
```
kichalla: This is the gist of this PR:

- As part of the `build base images` definition (or we can come up another build definition for it) each platform's sdk is either built(ex: php, python) or just downloaded (ex: for node and dotnet). They are only built or downloaded if there is no version already present on Azure Storage. The output of this is a `tar.gz` file copied to the 'artifacts' folder. A pipeline task will push the content of this artifacts folder to Azure Blob storage.
- The full and slim build image consume the `.tar.gz` file of each platform to extract them in respective `/opt` folders.
- There is a script called `prepEnv.sh` which is added to the build images to enable installing sdks. Ex: `prepEnv dotnet=4 node=15` should install these sdks from Azure Storage if they are not installed already.
- Also a command called `oryx prep-env` would do the same as the above point except that it proves the source repo to find the platform versions that are needed to be installed and then invokes `prepEnv` respectively.

Note that most of the functionality is not yet complete or tested(due to Pipelines issue). This PR is to give an idea of how things are currently.
TBD: 
- A pattern for end users to request installing native libraries
- Currently this PR's logic loops through versions file for each platform to build/download sdks. We have typically used `constants.yml` file for maintaining versions. So, I am yet to figure out how this would fit in with the constants file.
- Testing a change locally instead of relying on remote azure storage.

kichalla:
Closing this to create a new Draft PR (to avoid triggering builds...thanks @arroyc )

PR #343, opened 2019-09-05
Upgraded .NET Core 3 from Preview8 to Preview9 and Python 3.8 from 0b3 to 0b4

kichalla:


build/constants.yaml
```diff
@@ -53,7 +53,7 @@
     python27-version: 2.7.16
     python36-version: 3.6.9
     python37-version: 3.7.4
-    python38-version: 3.8.0b3
+    python38-version: 3.8.0b4
```
kichalla: This requires the base image to be built which as you are currently broken. So this PR will have to wait to be merged until that issue is fixed.

kichalla:
@arroyc Can you take a look again?

PR #347, opened 2019-09-11
add condition for allowing telemetry for appinsights sdk

arroyc:


src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -49,7 +49,8 @@ func shouldApplicationInsightsBeConfigured() bool {
 	nodeAppInsightsKeyEnv := os.Getenv(consts.UserAppInsightsKeyEnv)
 	nodeAppInsightsEnabledEnv := os.Getenv(consts.UserAppInsightsEnableEnv)
 
-	if nodeAppInsightsKeyEnv != "" && strings.ToLower(nodeAppInsightsEnabledEnv) != "disabled" {
+	// Check if the application insights environment variables are present
+	if nodeAppInsightsKeyEnv != "" && nodeAppInsightsEnabledEnv != "" {
```
kichalla: I would add an additional check for `disabled` too here..it does not make sense to inject this module and not use it when we find out that the value was indeed `disabled`

PR #351, opened 2019-09-13
Changed the named from 'nodejs' to 'node' all across

kichalla:


arroyc:
we should wait this till ant-85 gets delivered to app service, too many moving parts

PR #355, opened 2019-09-17
Added additional logs to save the platform name and version used

kichalla:


src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -156,6 +156,17 @@ internal class DefaultBuildScriptGenerator : IBuildScriptGenerator
                     }
 
                     platformVersion = detectionResult.LanguageVersion;
+                    _logger.LogDebug(
```
kichalla: Since `platformName` and `platformVersion` are standard names that we have been using all across our code base, it might become problematic to write queries just based on them. IMO we should start using `eventid` too. So it is like `get me traces where eventid=100 and platformName is dotnet` (here eventid 100 should represent the event of selecting a platform and version for the repo). Thoughts?

kichalla: cc @dorfire @gpcastro

gpcastro: what is an event ID?

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -156,6 +156,17 @@ internal class DefaultBuildScriptGenerator : IBuildScriptGenerator
...
+                        "Using user supplied platform {platformName} and detected version {platformVersion}.",
```
gpcastro: Do you mean detected platform?

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -156,6 +156,17 @@ internal class DefaultBuildScriptGenerator : IBuildScriptGenerator
...
+                        userSuppliedPlatform,
```
gpcastro: will we log those as dimensions as well? It makes it easier to write queries IMO

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
kichalla: No, this is where user supplied the platform but not the version. We already have logging statements for auto-detection of a repo when explicit info was provided by the user

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
kichalla: It can uniquely identify a trace message all across an app. I believe it will also get logged as a dimension

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
kichalla: yes, all those names in `{` and `}` get logged in dimensions

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
kichalla: Example:
image

gpcastro: those will very easily become confusing though; how are those event IDs generated? Is it AI that maintains them?

kichalla: No, these event IDs are maintained by us. We need to make sure we do not change them randomly so that our queries would not get broken. It requires discipline but I think it will make querying easy. Let's discuss this tomorrow and then I will merge it.

PR #372, opened 2019-10-01
Add info about Node app in Oryx's build manifest file

kichalla:


src/BuildScriptGenerator/Node/NodeManifestFilePropertyKeys.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.Oryx.BuildScriptGenerator.Node
+{
+    public class NodeManifestFilePropertyKeys
+    {
+        /// <summary>
+        /// The name of the property whose value indicates the location of the output directory of a Node app.
+        /// For example, 'dist' or 'public' folder.
+        /// </summary>
+        public const string OutputDirPath = nameof(OutputDirPath);
```
gpcastro: is it really the output, or just some specific component like static content?

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -168,6 +170,53 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
             };
         }
 
+        private void GetAppOutputDirPath(dynamic packageJson, Dictionary<string, string> buildProperties)
```
gpcastro: nit: this method is actually getting the value but also setting it in the build properties. Should we split this in two? Or rename it as it's not called `Get`?

src/BuildScriptGenerator/Node/NodeManifestFilePropertyKeys.cs
kichalla: As I mentioned in scrum today, this has been the typical naming pattern by different frameworks, so I kept it this way.

src/BuildScriptGenerator/Node/NodePlatform.cs
kichalla: Sure I agree. Will change it in a different PR.

PR #374, opened 2019-10-03
Fixed Node integration tests

kichalla:


tests/Oryx.Integration.Tests/Node/NodeNextJsAppTest.cs
```diff
@@ -67,11 +65,10 @@ public async Task CanBuildAndRun_BlogStarterNextJsApp_WithoutZippingNodeModules(
                 });
         }
 
-        [Theory]
-        [InlineData("10")]
-        [InlineData("12")]
```
gpcastro: why are we not testing this for Node 12 anymore?

kichalla: We added two samples related to NextJS and NuxtJs to add some coverage. I felt it is not necessary to do it for multiple versions.

PR #375, opened 2019-10-03
Reverted the change related to dynamic setting of gunicorn worker count

kichalla:
We are reverting because some customers were depending on in-memory session which does not persist when an app runs on multiple workers.

### DIFF
--- a/src/startupscriptgenerator/src/python/scriptgenerator.go
+++ b/src/startupscriptgenerator/src/python/scriptgenerator.go
@@ -11,27 +11,24 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


src/startupscriptgenerator/src/python/scriptgenerator.go
```diff
@@ -11,27 +11,24 @@ import (
...
 	"os"
 	"path/filepath"
-	"runtime"
-	"strconv"
 	"strings"
 )
 
-
 type PythonStartupScriptGenerator struct {
-	AppPath                     string
```
kichalla: All these whitespace changes were done by VS Code automatically

PR #381, opened 2019-10-07
[714233] Add --output parameter to build-script command

cormacpayne:
Resolves issue 714233 -- following the same pattern as the `run-script` command, added an `--output` parameter to the `build-script` command to allow users to write the generated build script to a given path.

### DIFF
--- a/src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
+++ b/src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
@@ -48,6 +48,12 @@ namespace Microsoft.Oryx.BuildScriptGeneratorCli
             Description = "Additional information used by this tool to generate and run build scripts.")]
         public string[] Properties { get; set; }
 
+        [Option(
+            "--output",
+            CommandOptionType.SingleValue,
+            Description = "The path that the build script will be written to. If not specified, the result will be written to STDOUT.")]
+        public string OutputPath { get; set; }
+
         internal override int Execute(IServiceProvider serviceProvider, IConsole console)
         {
             var scriptGenerator = new BuildScriptGenerator(
@@ -61,7 +67,18 @@ namespace Microsoft.Oryx.BuildScriptGeneratorCli
                 return ProcessConstants.ExitFailure;
             }
 
-            console.WriteLine(generatedScript);
+            if (string.IsNullOrEmpty(OutputPath))
+            {
+                console.WriteLine(generatedScript);
+            }
+            else
+            {
+                File.WriteAllText(OutputPath, generatedScript);
+                console.WriteLine($"Script written to '{OutputPath}'");
+
+                // Try making the script executable
+                ProcessHelper.TrySetExecutableMode(OutputPath);
+            }
 
             return ProcessConstants.ExitSuccess;
         }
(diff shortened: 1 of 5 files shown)


src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
```diff
@@ -48,6 +48,12 @@ internal class BuildScriptCommand : CommandBase
             Description = "Additional information used by this tool to generate and run build scripts.")]
         public string[] Properties { get; set; }
 
+        [Option(
+            "--output",
+            CommandOptionType.SingleValue,
+            Description = "The path that the build script will be written to. If not specified, the result will be written to STDOUT.")]
```
kichalla: nit: We try to keep the width of the text <=120 characters to avoid scrolling in GitHub. I use the VS extension `Guidelines` with which you can set the limit to 120.

src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
```diff
@@ -61,7 +67,18 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
                 return ProcessConstants.ExitFailure;
             }
 
-            console.WriteLine(generatedScript);
+            if (string.IsNullOrEmpty(OutputPath))
```
kichalla: Get the full path of this string since users can supply relative (`./foo.sh` or `../foo.sh`) or full paths

src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
```diff
@@ -61,7 +67,18 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
...
+            {
+                console.WriteLine(generatedScript);
+            }
+            else
+            {
+                File.WriteAllText(OutputPath, generatedScript);
```
kichalla: Can you check if this api still works if a user supplies a path which does not have all the directories in the path? if not, we should create it IMO

cormacpayne: Added an extensions class for file system operations that includes a method for this.

tests/Oryx.Common.Tests/FileExtensionsTest.cs
```diff
@@ -0,0 +1,50 @@
...
+        private string _testDirPath;
+
+        public FileExtensionsTest(TestTempDirTestFixture testFixture)
+        {
+            _testDir = testFixture;
+            _testDirPath = testFixture.RootDirPath;
+        }
+
+        [Fact]
+        public void SafeWriteAllText_Sanity()
```
kichalla: This is grouping multiple test scenarios in one method...It is better to have separate methods because a failure in one test blocks running other tests in this method (sure you can run the build again to find the next failure but that is not ideal) and also if we start looking at a test method's run history we would need to consider mixing multiple scenarios

src/Common/Extensions/FileExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                Directory.CreateDirectory(parentPath);
+            }
+
+            File.WriteAllText(outputPath, contents);
+        }
+
+        /// <summary>
+        /// This method will recursively enumerate the files under a given path since the
+        /// Directory.EnumerateFiles call does not check to see if a directory exists before
```
kichalla: I am curious when does this actually happen. I would argue it is the job of this .NET Core api to do these checks and give us the results. If we want to do safe enumeration, it should be enough to just check if supplied argument's directory exists?

Issue #90, opened 2019-04-23
oryx run-script - Platform name is required


derekbekoe:
### Bug Report

- At what date and time did you most recently experience the problem? =
**Apr 22 2019**

- Where did you experience the problem? E.g. Azure Web Apps, Azure Functions, Azure Container Registry, or offline use.
**Local**

- If your repo is publicly available please share its URL:
**https://github.com/vsls-contrib/guestbook**

- What happened?
Platform name is required while [the oryx README](https://github.com/Microsoft/Oryx/blob/master/README.md#build-and-run-an-app) suggests it's not required.

- What did you expect or want to happen?
It runs without `--platform nodejs` being required.

- How can we reproduce it?
```bash
$ docker run -it oryxprod/build:latest bash
# git clone https://github.com/vsls-contrib/guestbook
# cd guestbook/
# oryx build
# oryx run-script --appPath .
Platform name is required.
```

gpcastro:
Thanks Derek, we will work on this.

cormacpayne:
The fix for this has been merged and will be available in the next release of the Oryx build image.

PR #383, opened 2019-10-08
[854302] Platform name is no longer required for run-script command

cormacpayne:
Resolves issue 854302 (or GitHub issue https://github.com/microsoft/Oryx/issues/90) -- the `run-script` command currently fails if the user does not provide a value for the `--platform` parameter. These changes will use the same detection algorithm as the `build*` commands to determine the platform of the given repository (or the current directory if a source is not given).

A `ScriptGeneratorContext` abstraction was introduced to allow `build*` and `run*` commands to have their own information about the generator context, but not disrupt what the platform and detector classes need to determine if their platform is compatible.

### DIFF
--- a/src/BuildScriptGenerator/RunScriptGeneratorContext.cs
+++ b/src/BuildScriptGenerator/RunScriptGeneratorContext.cs
@@ -11,16 +11,16 @@ namespace Microsoft.Oryx.BuildScriptGenerator
     public class RunScriptGeneratorContext : ScriptGeneratorContext
     {
         /// <summary>
-        /// Gets or sets the name of the main programming language used in the repo.
-        /// If none is given, a language detection algorithm will attemp to detect it.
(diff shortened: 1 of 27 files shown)


src/BuildScriptGenerator/RunScriptGeneratorContext.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// Options for the run script generation process.
+    /// </summary>
+    public class RunScriptGeneratorContext : ScriptGeneratorContext
+    {
+        /// <summary>
+        /// Gets or sets the name of the main programming language used in the repo.
+        /// If none is given, a language detection algorithm will attemp to detect it.
+        /// </summary>
+        public string Language { get; set; }
```
gpcastro: we like to use "platorm" instead of "language", although we might still have occurrences of "language" in the code. The reason for this is node for instance is not really a language but a "programming platform".

PR #387, opened 2019-10-14
Enable diagnostic logging when running tests to catch random test failures on CI agents

kichalla:


build/testBuildImages.sh
```diff
@@ -6,6 +6,11 @@
 
 set -e
 
+echo
```
kichalla: Temporary code. We can remove it once we have identified the root cause of failed builds.

build/testBuildImages.sh
```diff
@@ -26,4 +31,14 @@ fi
...
 cd "$TESTS_SRC_DIR/$testProjectName"
-dotnet test --test-adapter-path:. --logger:"xunit;LogFilePath=$ARTIFACTS_DIR\testResults\\$testProjectName.xml" -c $BUILD_CONFIGURATION
\ No newline at end of file
+
+artifactsDir="$REPO_DIR/artifacts"
+mkdir -p "$artifactsDir"
+diagnosticFileLocation="$artifactsDir/$testProjectName-log.txt"
+
+dotnet test \
+    --diag "$diagnosticFileLocation" \
```
kichalla: So apparently the upgrade from .NET Core 2.2.200 to 2.2.401 did not fix the issue we have been seeing with our tests, so I am enabling diagnostic logging here to get as much detail as possible to enable us to debug whenever we see a failure.

PR #389, opened 2019-10-15
Updated .NET Core runtime images to install LLDB package and also install new dotnetcore tools in .NET Core 3.0 image

kichalla:


images/runtime/dotnetcore/3.0/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# GOPATH is set to "/go" in the base image
+WORKDIR /go/src
+COPY src/startupscriptgenerator/src .
+ARG GIT_COMMIT=unspecified
+ARG BUILD_NUMBER=unspecified
+ENV GIT_COMMIT=${GIT_COMMIT}
+ENV BUILD_NUMBER=${BUILD_NUMBER}
+RUN ./build.sh dotnetcore /opt/startupcmdgen/startupcmdgen
+
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0 AS tools-install
```
kichalla: Since currently these tools are only available as part of SDK, we install the tools in a directory and later copy them to the final runtime image

images/runtime/dotnetcore/2.2/Dockerfile
```diff
@@ -11,11 +11,12 @@ ENV GIT_COMMIT=${GIT_COMMIT}
 ENV BUILD_NUMBER=${BUILD_NUMBER}
 RUN ./build.sh dotnetcore /opt/startupcmdgen/startupcmdgen
 
-FROM %DOTNETCORE_BASE_IMAGE%
+FROM mcr.microsoft.com/dotnet/core/aspnet:2.2.7
 
 RUN apt-get update \
   && apt-get install -y \
      file \
+     lldb \
```
kichalla: For .NET Core versions prior to 3.0, LLDB debugger + SOS (which is part of the runtime folder in this image) is the way to debug the applications.

images/runtime/dotnetcore/3.0/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-sos
+RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-trace
+RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-dump
+RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-counters
+
+FROM mcr.microsoft.com/dotnet/core/aspnet:3.0.0
```
kichalla: In this PR we are only installing the necessary tools/packages but are NOT enabling capturing dumps. The idea is for AppService to provide the values for the following .NET Core settings when they derive from our image.
For example, AppService can give the default directory for dumps to be the `$HOME/Logs/CoreDumps`

```
ENV COMPlus_DbgEnableMiniDump="1" (This enables automatic capturing of coredump before a program crashes finally)
ENV COMPlus_DbgMiniDumpName="$HOME/Logs/Dumps/coredump.%d" (This is for the location of the dumps)
```

images/runtime/dotnetcore/3.0/Dockerfile
arroyc: nit:can you add this comment in the dockerfile also?

PR #390, opened 2019-10-15
Added --blame flag to 'dotnet test' to find crashing test

kichalla:


build/testBuildImages.sh
```diff
@@ -43,6 +43,7 @@ export MSBUILDDEBUGPATH="$msbuildDebugLogsDir"
 export MSBUILDDISABLENODEREUSE=1
 
 dotnet test \
+    --blame \
```
kichalla: On a recent test run, my build failed with the error `The active test run was aborted. Reason: Test host process crashed`, so adding the following flag to see if we can get more information:

image

build/testBuildImages.sh
```diff
@@ -34,17 +29,26 @@ cd "$TESTS_SRC_DIR/$testProjectName"
...
 artifactsDir="$REPO_DIR/artifacts"
 mkdir -p "$artifactsDir"
-diagnosticFileLocation="$artifactsDir/$testProjectName-log.txt"
 
 # Create a directory to capture any debug logs that MSBuild generates
 msbuildDebugLogsDir="$artifactsDir/msbuildDebugLogs"
 mkdir -p "$msbuildDebugLogsDir"
 export MSBUILDDEBUGPATH="$msbuildDebugLogsDir"
-export MSBUILDDISABLENODEREUSE=1
+export COMPlus_DbgEnableMiniDump="1"
```
gpcastro: should we add a comment with what this thing is about?

kichalla: Sure, will do

PR #391, opened 2019-10-15
Added .NET Core 3.1 preview1 version

kichalla:


build/__dotNetCoreRunTimeVersions.sh
```diff
@@ -6,3 +6,5 @@ NET_CORE_APP_20='2.0.9'
 NET_CORE_APP_21='2.1.13'
 NET_CORE_APP_22='2.2.7'
 NET_CORE_APP_30='3.0.0'
+NET_CORE_APP_31='3.1.0'
```
kichalla: This was released today and is going to be LTS version in future, so thought it would be better to have some coverage for it.

cormacpayne: @kichalla since this is going to be LTS, should we add it to the slim build? I know that we have an existing item to add 3.0 to the slim image, should we change that to be 3.1 instead?

tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
```diff
@@ -103,6 +103,7 @@ public void DotNetAlias_UsesLtsVersion_ByDefault(string buildImageName)
         [InlineData("2.2", DotNetCoreSdkVersions.DotNetCore22SdkVersion)]
         [InlineData("3", DotNetCoreSdkVersions.DotNetCore30SdkVersion)]
         [InlineData("3.0", DotNetCoreSdkVersions.DotNetCore30SdkVersion)]
+        [InlineData("3.1", DotNetCoreSdkVersions.DotNetCore31SdkVersionPreviewName)]
```
cormacpayne: @kichalla do we want to add a test for `DotNetCore31SdkVersion` as well? If not, is there any reason we're just testing the preview version here?

tests/Oryx.Integration.Tests/DotNetCore/DotNetCoreRuntimeVersion30Tests.cs
```diff
@@ -329,5 +329,49 @@ public async Task CanBuildAndRun_NetCore30WebApp_HavingExplicitAssemblyName()
                     Assert.Contains("Hello World!", data);
                 });
         }
+
+        [Fact]
+        public async Task CanBuildAndRun_NetCoreApp31WebApp()
+        {
+            // Arrange
+            var dotnetcoreVersion = "3.1";
```
cormacpayne: @kichalla do we want to test both stable and preview in this test?

tests/Oryx.RuntimeImage.Tests/DotNetCore/DotnetCoreImageVersionsTest.cs
```diff
@@ -88,6 +88,7 @@ public void RuntimeImage_HasExecptedDotNetVersion_NetCoreApp10Versions(string ve
         [InlineData("2.1", "Version: " + DotNetCoreRunTimeVersions.NetCoreApp21)]
         [InlineData("2.2", "Version: " + DotNetCoreRunTimeVersions.NetCoreApp22)]
         [InlineData("3.0", "Version: " + DotNetCoreRunTimeVersions.NetCoreApp30)]
+        [InlineData("3.1", "Version: " + DotNetCoreRunTimeVersions.NetCoreApp31PreviewName)]
```
cormacpayne: @kichalla same question here about whether or not we should add a test using `DotNetCore31SdkVersion` here

build/__dotNetCoreRunTimeVersions.sh
kichalla: Since this is still `Preview1`, we should wait until it is finally released(which will happen in December) for us to move it to slim image.

tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
kichalla: The reason is we depend on the output from the `dotnet --version` command which prints out the preview name as it is still in preview

tests/Oryx.Integration.Tests/DotNetCore/DotNetCoreRuntimeVersion30Tests.cs
kichalla: Stable wouldn't be available until December, so for now we can only test preview.

tests/Oryx.RuntimeImage.Tests/DotNetCore/DotnetCoreImageVersionsTest.cs
kichalla: Same here, we depend on the output of `dotnet --info` to verify which still prints out preview name. We will modify this test once the released version is available in December.

PR #394, opened 2019-10-17
[999795] Updates to logging

cormacpayne:


cormacpayne:
TODO: Add `--verbosity` flag to commands

src/Common/Extensions/StringExtensions.cs
```diff
@@ -74,5 +75,29 @@ public static IList<string> Chunkify(this string str, int maxLength)
...
+        /// <param name="str">The string to hash.</param>
+        /// <returns>The SHA-256 hash of the given string.</returns>
+        public static string Hash(this string str)
+        {
+            if (string.IsNullOrEmpty(str))
+            {
+                return str;
+            }
+
+            var sha = SHA256.Create();
```
kichalla: We might need to call `Dispose` on this instance https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.sha256?view=netcore-2.1

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -319,7 +319,7 @@ internal static dynamic GetPackageJsonObject(ISourceRepo sourceRepo, ILogger log
                 // This prevents Oryx from erroring out when Node.js itself might be able to tolerate the file.
                 logger.LogWarning(
                     exc,
-                    $"Exception caught while trying to deserialize {NodeConstants.PackageJsonFileName}");
+                    $"Exception caught while trying to deserialize {NodeConstants.PackageJsonFileName.Hash()}");
```
kichalla: `package.json` is platform defined file name (nodejs), so should we hash these too?

cormacpayne: Jennie mentioned that for these platform-defined files that we know the name of, we can create a separate table that maps each file name to a hash that we use when looking at the logs.

Issue #395, opened 2019-10-18
Support finding executables in package manager bin directories


derekbekoe:
### Bug Report

- Where did you experience the problem? E.g. Azure Web Apps, Azure Functions, Azure Container Registry, or offline use.

Offline use.

- What happened?

Executable installed with `pip install` not found on PATH.

The executable can be found in /opt/python/3/bin/pylint but it is not on PATH.

- What did you expect or want to happen?

The executable was found and ran as expected.

- How can we reproduce it?

```bash
$ docker run -it mcr.microsoft.com/oryx/build:20191015.2 bash
root@fbc1f71d1acc:/# pip install pylint
Collecting pylint
...
Successfully installed astroid-2.3.2 isort-4.3.21 lazy-object-proxy-1.4.2 mccabe-0.6.1 pylint-2.4.3 six-1.12.0 typed-ast-1.4.0 wrapt-1.11.2

root@fbc1f71d1acc:/# pylint
bash: pylint: command not found
```

--

Should the directory `/opt/python/3/bin` be added to `/opt/oryx/defaultversions`. So as long as I have `/opt/oryx/defaultversions` on PATH, it'll be found?

derekbekoe:
Also happens with npm.

```bash
$ docker run -it mcr.microsoft.com/oryx/build:20191015.2 bash
root@d8084c3b4d00:/# npm install -g serve
/opt/nodejs/10.16.3/bin/serve -> /opt/nodejs/10.16.3/lib/node_modules/serve/bin/serve.js
+ serve@11.2.0
added 78 packages from 39 contributors in 6.213s
root@d8084c3b4d00:/# serve
bash: serve: command not found 
root@d8084c3b4d00:/# /opt/nodejs/lts/bin/serve
ERROR: Cannot copy to clipboard: Couldn't find the required `xsel` binary. On Debian/Ubuntu you can install it with: sudo apt install xsel

   ┌────────────────────────────────────────────────┐
   │                                                │
   │   Serving!                                     │
   │                                                │
   │   - Local:            http://localhost:5000    │
   │   - On Your Network:  http://172.17.0.2:5000   │
   │                                                │
   └────────────────────────────────────────────────┘

```

derekbekoe:
Is this a scenario you'd expect to work or you expect users to add the required paths themselves?

kichalla:
Thanks @derekbekoe 

> Is this a scenario you'd expect to work or you expect users to add the required paths themselves?

[...]

PR #396, opened 2019-10-20
benv: Removed 'defaultversions' directory and added LTS version's bin directory directly to path

kichalla:


images/build/slim.Dockerfile
```diff
@@ -218,9 +213,9 @@ RUN chmod a+x /opt/buildscriptgen/GenerateBuildScript
 FROM python AS final
 WORKDIR /
 
-ENV PATH=$PATH:/opt/oryx/defaultversions
-COPY images/build/benv.sh /opt/oryx/defaultversions/benv
-RUN chmod +x /opt/oryx/defaultversions/benv
+ENV PATH="$PATH:/opt/oryx:/opt/nodejs/lts/bin:/opt/dotnet/sdks/lts:/opt/python/latest/bin:/opt/yarn/stable/bin"
```
kichalla: Related to https://github.com/microsoft/Oryx/issues/395

tests/Oryx.BuildImage.Tests/BenvTest.cs
```diff
@@ -29,34 +29,29 @@ public BenvTest(ITestOutputHelper output)
 
         [Theory]
         // DotNet
-        [InlineData("dotnet")]
+        [InlineData("dotnet", "/opt/dotnet/")]
```
gpcastro: Shouldn't we also check that those tools are in the path?

kichalla: Shouldn't this test be enough though? This test verifies that these tools are accessible out of the box.

PR #397, opened 2019-10-22
.NET Core runtime images: create custom base images which share common layers (ex: LLDB etc.)

kichalla:


images/runtime/dotnetcore/1.0/Dockerfile.base
```diff
@@ -0,0 +1,15 @@
...
+
+# Older .NET core versions, which have reached end of life and therefore are no longer updated, use
+# a version of `curl` that has known issues.
+# We manually update it here so we can still depend on the original images.
+# This command should be removed once support for deprecated .NET core images is halted.
+RUN sed -i '/jessie-updates/d' /etc/apt/sources.list  # Now archived
+
+RUN apt-get update \
+  && apt-get upgrade -y \
+  && apt-get install -y \
```
kichalla: Not installing LLDB in the out-of-support versions 1.0, 1.1 and 2.0

images/runtime/dotnetcore/2.1/Dockerfile.base
```diff
@@ -0,0 +1,46 @@
+# dotnet tools are currently available as part of SDK so we need to create them in an sdk image
+# and copy them to our final runtime image
+FROM mcr.microsoft.com/dotnet/core/sdk:2.1.802 AS tools-install
+RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-sos
+
+FROM oryx-run-base AS dotnetcore-original
```
kichalla: We could have based on .NET Core team's `runtime-deps` image (so we can avoid the content from lines 8 to 25), but we wanted to keep the base image consistent across our stack. .NET Core team's base images are based on either `stretch-slim` or `buster-slim`

images/runtime/dotnetcore/3.0/Dockerfile.base
```diff
@@ -0,0 +1,57 @@
+# dotnet tools are currently available as part of SDK so we need to create them in an sdk image
+# and copy them to our final runtime image
+FROM mcr.microsoft.com/dotnet/core/sdk:3.0.100 AS tools-install
+RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-sos
+RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-trace
+RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-dump
+RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-counters
+
+FROM debian:buster-slim
```
kichalla: Note that we are NOT  using `oryx-run-base` here since it is based on `debian:stretch`

images/build/yarn-cache/Dockerfile
```diff
@@ -2,7 +2,7 @@
 # It is never used directly, but rather as a base for our build image.
 # Yarn cache will be placed in /usr/local/share/yarn-cache, that is the folder
 # that must be copied to the build image.
-FROM node:lts
+FROM node:10.16.3
```
kichalla: This was changed because `lts` now points to `12.13.0` which is not part of this PR. I am sending out one more PR for updating all versions.

build/constants.yaml
```diff
@@ -108,6 +109,8 @@
       directory: src/startupscriptgenerator/src/common/consts
 - name: node-versions
   constants:
+    node-runtime-base-tag: 20191023.4
```
gpcastro: is there a way to avoid repeating this value here (from `build/__nodeVersions.sh `)

dorfire: `__nodeVersions.sh` is generated from this .yaml

dorfire:
maybe also run a nightly build?

build/constants.yaml
kichalla: You mean repeating between this line here and the next (yarn-cache-base-tag)? I wanted it to be flexible here because sometimes we have seen these tag numbers being different between different platforms depending on how successful a build went. For example, we could be having a failure in building Node runtime images but could have been successful in case of Yarn in which case we should be able to use the tag for Yarn cache separately

kichalla:
> maybe also run a nightly build?

Sure, sounds good

images/runtime/node/generateDockerfiles.sh
```diff
@@ -6,11 +6,13 @@
 
 set -e
 
+declare -r REPO_DIR=$( cd $( dirname "$0" ) && cd .. && cd .. && cd .. && pwd )
```
gpcastro: nit: `cd ..\..\..`?

src/BuildScriptGenerator/PythonVersions.cs
```diff
@@ -4,7 +4,7 @@ namespace Microsoft.Oryx.Common
 {
     public static class PythonVersions
     {
-        public const string PythonBaseTag = "20191018.1";
+        public const string PythonBaseTag = "20191023.4";
```
gpcastro: why do we need this?

kichalla: We really do not need this but we put all python constants in one file in `constants.yml` file and so this just gets generated along with other used properties

images/runtime/node/generateDockerfiles.sh
kichalla: Sure, will do...but will do in next PR which is dealing with versions.

PR #401, opened 2019-10-24
Added back support to log build script output and with extra info as properties

kichalla:


kichalla:
cc @dorfire @gpcastro

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -275,7 +277,16 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
 
             if (exitCode != ProcessConstants.ExitSuccess)
             {
-                logger.LogError("Build script exited with {exitCode}", exitCode);
+                logger.LogLongMessage(
```
kichalla: Currently following is the query I need to run to get details of failed build(I already saved this query for our team if you want to use):
```
traces 
| where message startswith "Build script exited with"
| project operation_ParentId
// Using leftouter here as build script output could be broken into multiple trace records based on its size
| join kind= leftouter (
   traces
   | where message startswith "Build script output" 
) on operation_ParentId
```

With my change here, it would be
```
traces
| where customDimensions.buildExitCode != 0
```

src/Common/Extensions/LoggerAiExtensions.cs
```diff
@@ -64,15 +62,20 @@ public static void LogTrace(this ILogger logger, string message, IDictionary<str
...
         /// <param name="level">log level. will apply to all chunks</param>
         /// <param name="header">chunk header. will be followed by chunk index, a colon, and a line break</param>
-        /// <param name="message">long message to be chunkified and logged</param>
-        public static void LogLongMessage(this ILogger logger, LogLevel level, [NotNull] string header, string message)
+        /// <param name="nonFormattedMessage">long message to be chunkified and logged</param>
+        public static void LogLongMessage(
+            this ILogger logger,
+            LogLevel level,
+            [NotNull] string header,
+            string nonFormattedMessage,
```
dorfire: 👍

src/Common/ILoggerExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+    public static class ILoggerExtensions
+    {
+        /// <summary>
+        /// Logs message and the supplied properties. The message must not have formatted strings as they are not processed.
+        /// </summary>
+        /// <param name="logger"></param>
+        /// <param name="logLevel"></param>
+        /// <param name="nonFormattedMessage">The message without any formatting strings in it.</param>
+        /// <param name="properties"></param>
+        public static void Log(
```
dorfire: is this used anywhere?

kichalla: It is used by the `LogLongMessage` extension...the `properties` argument should make it clear..

dorfire: oh, totally missed it

PR #402, opened 2019-10-29
Enabled new PHP extensions

kichalla:


images/build/php/prereqs/build.sh
```diff
@@ -102,6 +102,21 @@ fi
 		--with-mhash \
 		--enable-ftp \
 		--enable-mbstring \
+		--enable-exif \
+		--enable-gd \
```
kichalla: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3192987

cc @dorfire

PR #403, opened 2019-10-29
Disabled Composer platform validation

dorfire:


images/runtime/php/Dockerfile.base.template
```diff
@@ -50,7 +50,7 @@ RUN apt-get update \
     && docker-php-ext-configure gd --with-png-dir=/usr --with-jpeg-dir=/usr \
     && docker-php-ext-configure imap --with-kerberos --with-imap-ssl \
     && docker-php-ext-configure pdo_odbc --with-pdo-odbc=unixODBC,/usr \
```
gpcastro: why don't we need to install gd in the runtime image, only in the build?

src/BuildScriptGenerator/Php/PhpBashBuildSnippet.sh.tpl
```diff
@@ -6,7 +6,7 @@ declare -r composer='/opt/php-composer/composer.phar'
 echo "Composer archive: $composer"
 echo "Running composer install..."
 echo
-php $composer install
+php $composer install --ignore-platform-reqs --no-interaction
```
kichalla: Add a comment as to why we are doing this here

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -149,7 +149,6 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 		scriptBuilder.WriteString("if [ -d node_modules ] || [ -L node_modules ]; then\n")
 		// We move the directory/link first to prevent node from start using it
 		scriptBuilder.WriteString("    mv -f node_modules _del_node_modules || true\n")
-		scriptBuilder.WriteString("    nohup rm -fr _del_node_modules &> /dev/null &\n")
```
kichalla: I am wondering if you should revert this change in this PR and just keep the PHP change...i will make this change in a different PR

tests/SampleApps/php/exif-example/composer.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "oryx/exif-example",
+  "require": {
+    "php": ">=5.3.0",
+    "ext-exif": "*"
```
kichalla: I suppose this breaks in the build image prior to your `--ignore-platform-reqs` change?

src/BuildScriptGenerator/Php/PhpBashBuildSnippet.sh.tpl
kichalla: Also it would be nice to print out that we are running this command, like we do for npm and yarn install...can give end user an idea of what we are doing...

tests/SampleApps/php/exif-example/composer.json
dorfire: correct

PR #406, opened 2019-10-30
Refactor context and add abstraction over platform detection

cormacpayne:

- Rename `BuildScriptGeneratorContext` to `RepositoryContext` and refactor the class to include additional properties.
- Add `ICompatiblePlatformDetector` abstraction to encapsulate logic for detecting the valid platform(s) for a repository (removing it from the `DefaultBuildScriptGenerator` class)

### DIFF
--- a/src/BuildScriptGenerator/Contracts/ICompatiblePlatformDetector.cs
+++ b/src/BuildScriptGenerator/Contracts/ICompatiblePlatformDetector.cs
@@ -3,8 +3,6 @@
 // Licensed under the MIT license.
 // --------------------------------------------------------------------------------------------
 
-using System;
-using System.Collections;
 using System.Collections.Generic;
 
 namespace Microsoft.Oryx.BuildScriptGenerator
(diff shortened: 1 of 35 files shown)


src/BuildScriptGenerator/Contracts/ICompatiblePlatformDetector.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// Returns a bool specifying whether or not the given platform is compatible with the provided repository.
+        /// </summary>
+        /// <param name="ctx">The <see cref="RepositoryContext"/>.</param>
+        /// <param name="platformName">The name of the platform.</param>
+        /// <param name="platformResult">If the given platform is compatible, this will be a tuple that maps
+        /// the platform object to a compatible version.</param>
+        /// <returns>True if the platform is compatible, false otherwise.</returns>
+        bool IsCompatiblePlatform(RepositoryContext ctx,
+                                  string platformName,
+                                  out Tuple<IProgrammingPlatform, string> platformResult);
```
dorfire: why use an out param here?

dorfire: also, instead of returning a tuple, it could be easier to work with a specialized data class (`ProgrammingPlatformWithVersion`?)

src/BuildScriptGenerator/DefaultCompatiblePlatformDetector.cs
```diff
@@ -0,0 +1,150 @@
...
+                    var languages = string.Join(", ", _programmingPlatforms.Select(p => p.Name));
+                    var exec = new UnsupportedLanguageException($"'{platformName}' platform is not supported. " +
+                        $"Supported platforms are: {languages}");
+                    _logger.LogError(exec, $"Exception caught, provided platform '{platformName}' is not supported.");
+                    throw exec;
+                }
+
+                if (!selectedPlatform.IsEnabled(ctx))
+                {
+                    var exec = new UnsupportedLanguageException($"Platform '{selectedPlatform.Name}' has been disabled.");
```
dorfire: exec => exc?

cormacpayne: Fixed.

src/BuildScriptGenerator/Contracts/ICompatiblePlatformDetector.cs
cormacpayne: Made the two `IsCompatible` methods private within the `DefaultCompatiblePlatformDetector` class -- the `out` parameter was just for convenience, but I can change it to the output of this method and scrap the `bool` if you think that'll be better.

dorfire: but then why even have them in the interface if you hide them in the implementation? :)

PR #408, opened 2019-10-30
Disabled Composer platform validation (#403)

kichalla:
This ports changes from `master` branch


src/BuildScriptGenerator/Php/PhpBashBuildSnippet.sh.tpl
```diff
@@ -4,9 +4,12 @@ echo "PHP executable: $phpBin"
 {{ if ComposerFileExists }}
 declare -r composer='/opt/php-composer/composer.phar'
 echo "Composer archive: $composer"
-echo "Running composer install..."
+echo "Running 'composer install --ignore-platform-reqs --no-interaction'..."
 echo
-php $composer install
+# `--ignore-platform-reqs` ensures Composer won't fail a build when
+# an extension is missing from the build image (it could exist in the
```
gpcastro: this is a bit odd. Why are we ok with the installation failing? If we don't need it in the build image, do we have to even install the extensions? How can we know if the failure is actually ok?

dorfire: [full context in Teams discussion]
in short - because App Service allows people to install custom extensions in the runtime image.

kichalla: FYI: https://blogs.msdn.microsoft.com/azureossds/2019/01/29/azure-app-service-linux-adding-php-extensions/

PR #412, opened 2019-11-04
Node runtime startup script: Avoid deleting existing node modules folder

kichalla:


src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -149,14 +149,12 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
...
 		scriptBuilder.WriteString("    mv -f node_modules _del_node_modules || true\n")
-		scriptBuilder.WriteString("    nohup rm -fr _del_node_modules &> /dev/null &\n")
 		scriptBuilder.WriteString("fi\n\n")
 		scriptBuilder.WriteString("if [ -d /node_modules/.bin ]; then\n")
-		// We move the directory/link first to prevent node from start using it
-		scriptBuilder.WriteString("\n# creating symbolic links for local node_module's directory\n")
-		scriptBuilder.WriteString("ln -s /node_modules ./node_modules \n\n")
-		scriptBuilder.WriteString("\n# creating symbolic links for local node_module's bin directory\n")
-		scriptBuilder.WriteString("ln -s /node_modules/.bin ./node_modules/.bin \n\n")
+		// Create a symlink to extracted nodemodules directory so that binaries mentioned in package.json(ex: 'ng serve')
```
kichalla: porting fix from patch branch

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -149,14 +149,12 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
...
+		// can still find the nodemodules.
+		scriptBuilder.WriteString("    ln -s /node_modules ./node_modules \n")
+		scriptBuilder.WriteString("    ln -s /node_modules/.bin ./node_modules/.bin \n")
```
gpcastro: is this line really needed?

src/startupscriptgenerator/src/node/scriptgenerator.go
gpcastro: the mv with `-f` this case overrides the current `_del_node_modules` if it exists?

tests/Oryx.Integration.Tests/Node/NodeAngularAppTest.cs
```diff
@@ -126,6 +126,31 @@ public async Task CanBuildAndRunAngular6App_WithDevAndProdDependencies_UsingZipp
                     var data = await _httpClient.GetStringAsync($"http://localhost:{hostPort}/");
                     Assert.Contains("Angular6app", data);
                 });
+
+            // Re-run the runtime container multiple times against the same output to catch any issues.
```
kichalla: I think we should make a default in our test framework to re-run the runtime image multiple times against an output for ALL tests. This will simulate some scenarios of AppService and hopefull catch any issues.

src/startupscriptgenerator/src/node/scriptgenerator.go
kichalla: This is not needed. I removed it

src/startupscriptgenerator/src/node/scriptgenerator.go
kichalla: Should we be worried about if it overrides as we anyway not going to use it?

PR #413, opened 2019-11-04
Reduce size of images: Common base for PHP runtime images, .NET Core tools installation changes

kichalla:


build/__phpVersions.sh
```diff
@@ -1,15 +1,15 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
-PHP_BUILD_BASE_TAG='20191025.2'
-PHP_RUNTIME_BASE_TAG='20191025.2'
+PHP_BUILD_BASE_TAG='20191104.7'
+PHP_RUNTIME_BASE_TAG='20191104.7'
```
kichalla:
```
-------------------------------------BEFORE-----------------------------------------

$ docker system df -v
Images space usage:

REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE                SHARED SIZE         UNIQUE SIZE         CONTAINERS
oryx/php-7.3        latest              9a9ff9a4034f        2 minutes ago       900.8MB             342.4MB             558.5MB             0
oryx/php-7.2        latest              8321845b6c9b        2 minutes ago       899.9MB             342.4MB             557.6MB             0
oryx/php-7.0        latest              744c743f50c7        3 minutes ago       969.8MB             0B                  969.8MB             0
oryx/php-5.6        latest              a8ddd8f77326        3 minutes ago       952.2MB             0B                  952.2MB             0

$ docker system df
TYPE                TOTAL               ACTIVE              SIZE                RECLAIMABLE
Images              4                   0                   3.38GB              3.38GB (100%)
Containers          0                   0                   0B                  0B
Local Volumes       0                   0                   0B                  0B
Build Cache         0                   0                   0B                  0B


-------------------------------------AFTER-----------------------------------------
$ docker system df -v
Images space usage:

REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE                SHARED SIZE         UNIQUE SIZE         CONTAINERS
oryx/php-7.3        latest              102f97536594        2 minutes ago       1.109GB             857.3MB             251.6MB             0
oryx/php-7.2        latest              e425956a40d3        2 minutes ago       1.108GB             857.3MB             250.5MB             0
oryx/php-7.0        latest              734dd49cbde6        2 minutes ago       1.034GB             822.7MB             211.7MB             0
oryx/php-5.6        latest              d2e194e2cea0        3 minutes ago       960.3MB             0B                  960.3MB             0

$ docker system df
TYPE                TOTAL               ACTIVE              SIZE                RECLAIMABLE
Images              4                   0                   2.531GB             2.531GB (100%)
Containers          0                   0                   0B                  0B
Local Volumes       0                   0                   0B                  0B
Build Cache         0                   0                   0B                  0B

```

kichalla: Notice that 5.6 image does not share the layers...this is expected as we are currently unable to use common layers since building it results in errors (possible because of dependency on stale packages which are no longer available)

kichalla: cc @sanchitmehta

build/buildBuildImages.sh
```diff
@@ -151,6 +151,9 @@ echo
 echo "List of images tagged (from '$ACR_BUILD_IMAGES_ARTIFACTS_FILE'):"
 cat $ACR_BUILD_IMAGES_ARTIFACTS_FILE
 
+echo
+showDockerImageSizes
```
arroyc: is this permanent or was for debugging?

images/runtime/php/7.0/Dockerfile.7.0
```diff
@@ -0,0 +1,235 @@
+# Commit b4319e8f767f1200c9013e08baf2c34b9c84e301 from repo https://github.com/docker-library/php.git
+FROM php-run-base
```
arroyc: nit: add the comment like source for php-7.0 as it's not available in ://github.com/docker-library/php.git anymore

images/runtime/php/7.0/docker-php-ext-install
```diff
@@ -0,0 +1,122 @@
+#!/bin/sh
+set -e
```
arroyc: did you verify manually if this works for php 7?

images/runtime/php/7.0/docker-php-ext-enable
```diff
@@ -0,0 +1,114 @@
+#!/bin/sh
+set -e
```
arroyc: did you verify manually if this works for php 7?

images/runtime/php/7.2/docker-php-ext-enable
```diff
@@ -0,0 +1,114 @@
+#!/bin/sh
+set -e
+
+extDir="$(php -d 'display_errors=stderr' -r 'echo ini_get("extension_dir");')"
```
arroyc: same question here .. as per my experience I have seen sometimes these scripts doesn't work, so it will be nice if we can test them as well, either manually , better if automated

src/BuildScriptGenerator/PythonVersions.cs
```diff
@@ -4,7 +4,7 @@ namespace Microsoft.Oryx.Common
 {
     public static class PythonVersions
     {
-        public const string PythonBaseTag = "20191025.2";
+        public const string PythonBaseTag = "20191104.7";
```
arroyc: what is the purpose for this change?

src/BuildScriptGenerator/Node/NodeVersions.cs
```diff
@@ -4,8 +4,8 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
 {
     public static class NodeVersions
     {
-        public const string NodeRuntimeBaseTag = "20191025.2";
-        public const string YarnCacheBaseTag = "20191025.2";
+        public const string NodeRuntimeBaseTag = "20191104.7";
```
arroyc: are we also updating the node runtimes with this? I would prefer not to change node and python runtime with this pr, you may have better reasoning though

build/buildBuildImages.sh
kichalla: Yes I wanted it to be there since it felt useful to see those details

images/runtime/php/7.0/docker-php-ext-enable
kichalla: The Nightly build passed after the changes in this PR https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3212379

images/runtime/php/7.2/docker-php-ext-enable
kichalla: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3212379

src/BuildScriptGenerator/Node/NodeVersions.cs
kichalla: It is important to rely on the same base layers 'across' runtimes too..for example if Node base layer 'oryx-run-base' is old and PHP uses an updated(some new versions of packages were recently released) oryx-run-base, then we would not be sharing these common layers. Also since Nightly build has passed, I would not worry much about it..

src/BuildScriptGenerator/PythonVersions.cs
kichalla: check my above comment

kichalla:
> nit: huge change in php images, I would like to add more tests in php as we don't have much in php as of now, don't know what may go wrong. please add a workitem for adding variety of sampleapps for php to increase test coverage

As per our offline discussion we do have some sample apps which test some scenarios

PR #416, opened 2019-11-05
Remove .NET Core 3.1 preview SDK and fix how language detector determines runtime version

kichalla:


src/BuildScriptGenerator/DotNetCore/DotnetCoreLanguageDetector.cs
```diff
@@ -91,30 +91,18 @@ public LanguageDetectorResult Detect(RepositoryContext context)
...
         {
-            switch (targetFramework)
+            // Ex: "netcoreapp2.2" => "2.2"
+            targetFramework = targetFramework.Replace(
+                "netcoreapp",
+                string.Empty,
+                StringComparison.OrdinalIgnoreCase);
+
+            // Ex: "2.2" => 2.2
+            if (double.TryParse(targetFramework, out var val))
```
cormacpayne: @kichalla are we verifying the version that we detect anywhere else? It seems like this was the best place to determine whether or not the .NET version used is supported. For example, if someone ran `oryx build` with a .NET Core 3.1 app, we would be able to parse the string and return a value, but in the previous implementation, we wouldn't have a constant for 3.1 and it would return `null`.

kichalla: tests/BuildScriptGenerator.Tests/DotnetCore/DotnetCoreLanguageDetectorTest.cs

PR #418, opened 2019-11-05
Added more logging and fixed runtime startup script generation to include ReleaseTagName

kichalla:


src/startupscriptgenerator/src/common/versionInfo.go
```diff
@@ -5,14 +5,19 @@
...
-	Commit      = "unspecified"
+	BuildNumber    = "unspecified"
+	Version        = "0.2" + "." + BuildNumber
+	Commit         = "unspecified"
+	ReleaseTagName = "unspecified"
 )
 
 func PrintVersionInfo() {
-	println("Oryx Version : " + Version + ", Commit: " + Commit + "\n")
+	fmt.Printf("Oryx Version: %s, Commit: %s, ReleaseTagName: %s\n", Version, Commit, ReleaseTagName)
```
kichalla: tidbit: it seems `println` writes to `stderr` where as `fmt.Printf` here writes to `stdout`

src/startupscriptgenerator/src/common/logging.go
```diff
@@ -121,3 +129,13 @@ func (logger *Logger) Shutdown() {
 		// to complete, so let's just exit.
 	}
 }
+
+func (logger *Logger) StartupScriptRequested() {
```
kichalla: This log message can enable us to find out how many times Oryx run script was invoked

src/startupscriptgenerator/src/common/logging.go
```diff
@@ -39,6 +39,14 @@ func SetGlobalOperationID(buildManifest BuildManifest) {
 	} else {
 		buildOpID = strings.TrimSpace(buildManifest.OperationID)
 	}
+
+	logger := GetLogger("logging.SetGlobalOperationID")
+	defer logger.Shutdown()
+
+	if buildManifest.OperationID == "" {
+		logger.LogInformation("No operation ID found in manifest.")
```
kichalla: This can tell us non-Oryx built code being run in our run images

PR #420, opened 2019-11-06
Create image for Oryx CLI to use in GitHub Actions

cormacpayne:

Generate a new image `oryx/cli:latest` that includes the Oryx CLI and .NET SDK 2.2 to be used in GitHub Actions when producing a Dockerfile.


images/build/cli.Dockerfile
```diff
@@ -0,0 +1,123 @@
+FROM buildpack-deps:stretch
```
kichalla: Is the purpose of this image to use Oryx's build script generator alone so that it can generate a dockerfile by looking at a repo? If yes, we need not install .NET Core SDK at all since we are doing a self-contained deployment of the build script generator.

kichalla: I mean you can use a multi-stage approach where you install 2.2 SDK in an image, publish Oryx app as standalone and then copy back the content of this stage to the final image (as we do in our other build images)

images/build/cli.Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM oryxdevmcr.azurecr.io/public/oryx/build-slim:latest AS build
```
arroyc: do we intend to build from the oryxdevmcr build-slim image? or we should also consider to build it from production

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
```diff
@@ -103,5 +103,15 @@ private string GenerateRuntimeTag(string version)
 
             return $"{split[0]}.{split[1]}";
         }
+
+        private string ConvertToRuntimeName(string platformName)
```
arroyc: im bit confused on the reason of having this function instead of having a key value pair  for dotnetcore, we can discuss offline as well

vsts/scripts/tagCliImagesForRelease.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
arroyc: nit: we need to create a PR in MCR repo for these new images

images/build/cli.Dockerfile
cormacpayne: The intention of the CLI image is to only have the artifacts needed to run the `oryx` CLI -- since we're copying the build artifacts from the Oryx project to the new image, I don't think it matters if we're using `slim` or `latest`. Since this image contains only the bare minimum needed to run the CLI, specifically the `oryx dockerfile` command (which doesn't need any dev tools), we want to keep the image as small as possible and copy over just the artifacts, another reason I don't believe it matters which tag we're using. Let me know what you think.

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
cormacpayne: This part is tricky because the name of the platform in our project is called `dotnet`, but the runtime image is called `dotnetcore`, so we need to add in a small layer to transform the name. I'm not sure there's an easy way for us to change the platform name to `dotnetcore` without breaking existing users in some way, but if there's a way to do this, or if you have idea of how we should resolve this issue, let me know.

vsts/scripts/tagCliImagesForRelease.sh
cormacpayne: Done ☺️

images/build/cli.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM oryxdevmcr.azurecr.io/public/oryx/build-slim:latest AS build
+FROM mcr.microsoft.com/oryx/build:latest as build
```
arroyc: thanks !

cormacpayne: Updated, thanks.

cormacpayne:
@kichalla @arroyc any additional comments? 😃

arroyc:
As we discussed, that we will be controlling the cli image version in a way that we know what version GitHub action is using, so that if we have any breaking change in oryx tooling we can control it better

PR #424, opened 2019-11-08
fix missing node modules during rsync copy

arroyc:
when we do rsync copy we exclude node_modules folder, but somehow we were excluding all the node_modules folder in all sub directories, rather we should exclude the node_modules folder only on the root.


tests/SampleApps/nodejs/kudu-bug/another-directory/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "kudu-bug",
```
kichalla: please rename this app to be more appropriate for the scenario being tested here...something which indicates `nested package.json`

tests/Oryx.Integration.Tests/Node/NodeTestUsingZippedNodeModules.cs
```diff
@@ -164,5 +164,48 @@ public async Task BuildsAndRunsNodeApp_WhenPruneDevDependenciesIsTrue_AndNodeMod
                     Assert.Contains("Say It Again", data);
                 });
         }
+
+        [Fact]
+        public async Task CopiesNodeModulesInSubDirectory_ToDestination_WhenNodeModulesAreCompressed()
```
kichalla: Should we even test when compression is not used?

tests/Oryx.Integration.Tests/Node/NodeTestUsingZippedNodeModules.cs
```diff
@@ -164,5 +164,48 @@ public async Task BuildsAndRunsNodeApp_WhenPruneDevDependenciesIsTrue_AndNodeMod
...
+                .AddCommand($"oryx -appPath {appOutputDir} -bindPort {ContainerPort}")
+                .AddCommand(DefaultStartupFilePath)
+                .AddDirectoryExistsCheck($"{appOutputDir}/another-directory/node_modules")
+                .ToString();
+            var buildScript = new ShellScriptBuilder()
+               .AddCommand(
+                $"oryx build {appDir} -i /tmp/int -o {appOutputDir} --platform nodejs " +
+                $"--platform-version {nodeVersion} -p {NodePlatform.CompressNodeModulesPropertyKey}=tar-gz" +
+                $" -p {NodePlatform.PruneDevDependenciesPropertyKey}=true")
+               //.AddCommand($"cp -rf /tmp/out/* {appOutputDir}")
```
kichalla: remove?

tests/Oryx.Integration.Tests/Node/NodeTestUsingZippedNodeModules.cs
```diff
@@ -164,5 +164,48 @@ public async Task BuildsAndRunsNodeApp_WhenPruneDevDependenciesIsTrue_AndNodeMod
...
+            var runAppScript = new ShellScriptBuilder()
+                .AddCommand($"oryx -appPath {appOutputDir} -bindPort {ContainerPort}")
+                .AddCommand(DefaultStartupFilePath)
+                .AddDirectoryExistsCheck($"{appOutputDir}/another-directory/node_modules")
+                .ToString();
+            var buildScript = new ShellScriptBuilder()
+               .AddCommand(
+                $"oryx build {appDir} -i /tmp/int -o {appOutputDir} --platform nodejs " +
+                $"--platform-version {nodeVersion} -p {NodePlatform.CompressNodeModulesPropertyKey}=tar-gz" +
+                $" -p {NodePlatform.PruneDevDependenciesPropertyKey}=true")
```
kichalla: Current app service does not use this, so i would say to to have multiple test cases here to test different scenarios here

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -300,7 +300,7 @@ public IEnumerable<string> GetDirectoriesToExcludeFromCopyToIntermediateDir(Buil
                 NodeConstants.AllNodeModulesDirName,
                 NodeConstants.ProdNodeModulesDirName,
                 NodeConstants.NodeModulesDirName,
-                NodeConstants.NodeModulesToBeDeletedName,
+                "/"+NodeConstants.NodeModulesToBeDeletedName,
```
kichalla: Can you add a comment as to why we are doing this

src/BuildScriptGenerator/Node/NodeConstants.cs
```diff
@@ -17,7 +17,7 @@ internal static class NodeConstants
         internal const string NpmStartCommand = "npm start";
         internal const string YarnStartCommand = "yarn run start";
         internal const string YarnCommand = "yarn";
-        internal const string NpmPackageInstallCommand = "npm install";
+        internal const string NpmPackageInstallCommand = "npm install --unsafe-perm";
```
kichalla: Does the test in this PR fail without this flag?

tests/Oryx.Integration.Tests/Node/NodeTestUsingZippedNodeModules.cs
arroyc: good idea

src/BuildScriptGenerator/Node/NodeConstants.cs
arroyc: as discussed offline

kichalla: can you try and see if `yarn` also needs something like this?

PR #429, opened 2019-11-11
[1006079] Remove hard-coded image from build image tests

cormacpayne:

Resolves 1006079 -- similar to runtime tests, remove the hard-coded image that was being used for tests and set it with environment variables defined in the executing test scripts.


tests/Oryx.BuildImage.Tests/BenvTest.cs
```diff
@@ -93,7 +91,8 @@ public BenvTest(ITestOutputHelper output)
                 .ToString();
 
             // Act
-            var result = _dockerCli.Run(SlimBuildImageName, "/bin/bash", "-c", script);
+            var image = _imageHelper.GetTestImage("build-slim", "latest");
```
kichalla: Isn't `build-slim` not being renamed to `build:slim` as part of this PR?

cormacpayne: This will be handled in a separate PR once these changes are merged.

tests/Oryx.BuildImage.Tests/BenvTest.cs
```diff
@@ -55,7 +52,8 @@ public BenvTest(ITestOutputHelper output)
                 .ToString();
 
             // Act
-            var result = _dockerCli.Run(FullBuildImageName, "/bin/bash", "-c", script);
+            var image = _imageHelper.GetTestImage("build", "latest");
```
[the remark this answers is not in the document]
cormacpayne: As mentioned offline, for tags, the helper only helps with the suffix -- a base tag will still need to be provided so we can add the tag suffix if one was provided. We'll let CI figure out if there's anything wrong with this approach ☺️

PR #432, opened 2019-11-12
Push base images to a new repo: 'public/oryx/base'

kichalla:
For examle:  oryxmcr.azurecr.io/public/oryx/base:php-7.3-20191105.2


build/__variables.sh
```diff
@@ -44,6 +44,8 @@ declare -r ACR_SLIM_BUILD_IMAGE_REPO="$ACR_DEV_NAME/public/oryx/build-slim"
 declare -r ACR_RUNTIME_IMAGES_REPO="$ACR_PUBLIC_PREFIX"
 declare -r ACR_PACK_IMAGE_REPO="$ACR_PUBLIC_PREFIX/$PACK_IMAGE_NAME"
 
+declare -r BASE_IMAGES_REPO="$ACR_DEV_NAME/public/oryx/base"
```
kichalla: Reason for this PR is that we will just have one repo name to handle all our base images (some new ones are going to come) and to avoid explosion of repo names in MCR.

arroyc: But the prod acr needs to maintain the standard naming provided by mcr team. Make sure we follow that. I understand what you are trying here, but I'm afraid if we can do that in final/prod image. We can discuss his offline

arroyc: I read the example again, seems to be fine .. but I would still like to discuss other edge cases (may be)

kichalla: I already got this naming confirmed in MCR repo so I’m not clear what is the actual concern here?

build/buildBuildImageBases.sh
```diff
@@ -10,72 +10,62 @@
...
 
 # Load all variables
 source $REPO_DIR/build/__variables.sh
 
-IMAGE_DIR_TO_BUILD=$1
-IMAGE_TAG="${BUILD_NUMBER:-latest}"
-BUILD_IMAGE_PREFIX="$__REPO_DIR/images/build"
+# Example: dontetcore, python
+IMAGE_DIR_TO_BUILD="$1"
+UNIQUE_TAG=""
```
cormacpayne: nit: should we renamed this to `UNIQUE_TAG_SUFFIX`?

kichalla: I will have one more PR in this area where I will fix it (just want to avoid a whole new build for this change).

PR #433, opened 2019-11-12
Created a new .NET Core install directory having flattened versions of SDK

kichalla:
This is for VSO's use case where it is preferred for dotnet muxer to choose between different sdk versions


build/__dotNetCoreRunTimeVersions.sh
```diff
@@ -2,7 +2,7 @@
 
 DOT_NET_CORE_RUNTIME_BASE_TAG='20191105.2'
 NET_CORE_APP_10='1.0.16'
-NET_CORE_APP_11='1.1.14'
+NET_CORE_APP_11='1.1.13'
```
cormacpayne: @kichalla what's the reason for the downgrade?

images/build/createFlattenedDotNetCoreInstall.sh
```diff
@@ -0,0 +1,51 @@
...
+source "$DIR/__dotNetCoreSdkVersions.sh"
+source "$DIR/__dotNetCoreRunTimeVersions.sh"
+
+splitSdksDir="/opt/dotnet/sdks"
+
+allSdksDir="/opt/dotnet/all"
+mkdir -p "$allSdksDir"
+
+# Copy latest muxer and license files
+cp -f "$splitSdksDir/3/dotnet" "$allSdksDir"
```
cormacpayne: @kichalla should we parameterize the folder that we're getting this content from? This could be something that's easy to overlook if/when we move to .NET Core 4, .NET Core 5, etc. and we would be copying over potentially outdated content.

build/__dotNetCoreRunTimeVersions.sh
kichalla: Actually we had it wrong in our repo. There is no runtime version `1.1.14` (however there is sdk version `1.1.14`).

images/build/createFlattenedDotNetCoreInstall.sh
kichalla: I agree. Totally valid point. Will update it.

gpcastro: We could use the `lts` link

images/build/createFlattenedDotNetCoreInstall.sh
```diff
@@ -0,0 +1,51 @@
...
+cp -f "$splitSdksDir/3/LICENSE.txt" "$allSdksDir"
+cp -f "$splitSdksDir/3/ThirdPartyNotices.txt" "$allSdksDir"
+
+function createLinks() {
```
gpcastro: add a comment with that it does?

images/build/createFlattenedDotNetCoreInstall.sh
kichalla: I considered `lts` but currently lts points to `2.1` but we already have `3.0` available, so ideally we want the latest version

kichalla: @cormacpayne Can you take at this PR again? I made the change where we always get the latest version of SDK dynamically.

PR #434, opened 2019-11-14
chore(AppInsights): add support for connectionstrings

markwolff:
- Bumps version to 1.7.2 (Connection strings supported as of [1.5.0](https://github.com/microsoft/ApplicationInsights-node.js/releases/tag/1.5.0))
- Changes `sdkVersion` telemetry processor prefix from `ad_` to `alr_`
- Uses SDK built-in loader script instead to setup & start the SDK


src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -71,14 +71,15 @@ func createApplicationInsightsLoaderFile(appInsightsLoaderFilePath string) {
 		var prefixInternalSdkVersion = function (envelope, contextObjects) {
 			try {
 				var appInsightsSDKVersion = appInsights.defaultClient.context.keys.internalSdkVersion;
-				envelope.tags[appInsightsSDKVersion] = "ad_" + envelope.tags[appInsightsSDKVersion];
+				envelope.tags[appInsightsSDKVersion] = "alr_" + envelope.tags[appInsightsSDKVersion];
```
kichalla: We seems to skip creating this loader file if there is one already on disk (it could have been created in previous release) and might be incompatible with these new changes(?). We should recreate this file. cc @arroyc

arroyc: IMO, it will be better if appinsights provides us this file  with the sdk, we would then update the path in the script. what do you think @kichalla @markwolff.

tests/SampleApps/nodejs/linxnodeexpress-appinsights/server.js
```diff
@@ -3,15 +3,16 @@ var app = express();
...
+var setupString = process.env.APPLICATIONINSIGHTS_CONNECTION_STRING || process.env.APPINSIGHTS_INSTRUMENTATIONKEY;
 
-console.log(process.env.APPINSIGHTS_INSTRUMENTATIONKEY)
+console.log(setupString);
 console.log(process.env.ApplicationInsightsAgent_EXTENSION_VERSION)
 
-if (process.env.APPINSIGHTS_INSTRUMENTATIONKEY && process.env.ApplicationInsightsAgent_EXTENSION_VERSION
+if (setupString && process.env.ApplicationInsightsAgent_EXTENSION_VERSION
     && process.env.ApplicationInsightsAgent_EXTENSION_VERSION !== "disabled") {
 	console.log("hello world here")
```
arroyc: nit: can be removed? :)

src/startupscriptgenerator/src/node/scriptgenerator.go
kichalla: Hmm, that is a good idea actually. @markwolff As Arjun mentioned here, if the loader file is going to be a standard one which does not have any other external dependencies it could be part of the applicationinsights module itself?

markwolff: Seems good. An issue is that eventually the loader script would contain each compute platform's internal telemetry & self-troubleshooting logic. Maybe the loader abstracts it out and lets it be set by each platform before executing the rest of the loader.

Mikhail-msft: @markwolff, in future - this should be part of your agent. in java case we use special marker file in order to determine the prefix

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -229,13 +189,9 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 
 	logger.LogInformation("Looking for App-Insights loader generated by Oryx and export to NODE_OPTIONS if needed")
 	if shouldApplicationInsightsBeConfigured() {
-		loaderFile := filepath.Join(gen.SourcePath, consts.NodeAppInsightsLoaderFileName)
-
-		if !common.FileExists(loaderFile) {
-			createApplicationInsightsLoaderFile(loaderFile)
-		}
+		loaderFile := filepath.Join(gen.GlobalModulesPath, consts.NodeAppInsightsLoaderFileName)
```
arroyc: I think it's better to use the constant instead of using globalmodulespath in scriptgenerator structure

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -229,13 +189,9 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
...
 
-		var nodeOptions = "'--require ./" + consts.NodeAppInsightsLoaderFileName + " ' $NODE_OPTIONS"
```
arroyc: just confirming .. we dont need ./ anymore?

tests/SampleApps/nodejs/linxnodeexpress-appinsights/package.json
```diff
@@ -4,7 +4,10 @@
   "description": "A sample Node.js app using Express 4 and application insights",
   "main": "server.js",
   "scripts": {
-    "start": "node server.js"
+    "start": "cross-env GLOBAL_PATH=/usr/local/lib/node_modules/ node server.js"
```
arroyc: why?? what happens when user doesn't provvide cross-env GLOBAL_PATH

tests/SampleApps/nodejs/linxnodeexpress-appinsights/package.json
```diff
@@ -4,7 +4,10 @@
...
+  },
+  "devDependencies": {
+    "cross-env": "^5.2.1"
```
arroyc: what does it do?

markwolff: Since this is only running on linux -- nothing. Removed it

tests/SampleApps/nodejs/linxnodeexpress-appinsights/package.json
markwolff: Removed cross-env and the env var

arroyc:
reopening to kick the ci

markwolff:
Closing this and reopening in #501

PR #435, opened 2019-11-14
Clean-up: Renamed files to get intellisense, receive Gpg keys only once for all Node runtime base images etc.

kichalla:


build/buildRunTimeImageBases.sh
```diff
@@ -36,12 +36,19 @@ docker build \
     -t "$RUNTIME_BASE_IMAGE_NAME" \
     $REPO_DIR
 
+if [ "$runtimeSubDir" == "node" ]; then
```
kichalla: These new lines of code are temporary here as I would be sending one more PR (related to pushing runtime bases to MCR) which would have different content.

build/buildRunTimeImageBases.sh
```diff
@@ -36,12 +36,19 @@ docker build \
...
+        $REPO_DIR
+fi
+
 labels="--label com.microsoft.oryx.git-commit=$GIT_COMMIT"
 labels="$labels --label com.microsoft.oryx.build-number=$BUILD_NUMBER"
 
 execAllGenerateDockerfiles "$runtimeImagesSourceDir"
 
-dockerFileName="Dockerfile.base"
+dockerFileName="base.Dockerfile"
```
kichalla: `Dockerfile` extension is understood by tools like VS and VS Code which provide highlights and intellisense

images/receiveGpgKeys.sh
```diff
@@ -6,19 +6,19 @@
 
 set -ex
 
-KEYS="$@"
+KEY_IDS="$@"
 
-if [ "$KEYS" == "" ]; then
+if [ "$KEY_IDS" == "" ]; then
```
kichalla: What this script gets as input are really ids of the actual public keys that we are trying to receive.

images/runtime/commonbase/nodeRuntimeBase.Dockerfile
```diff
@@ -0,0 +1,38 @@
+FROM oryx-run-base
+
+RUN apt-get update \
+  && apt-get install -y \
+    unzip \
+  && rm -rf /var/lib/apt/lists/*
+  
+# NOTE: This is a list of keys for ALL Node versions and also Yarn package installs.
```
kichalla: This and the Yarn install below should speed up our Node runtime base image builds and also we will be consistent with the Yarn version that we install for all Node runtime versions.

kichalla: cc @arroyc

kichalla: Base images build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3245498

PR #438, opened 2019-11-27
Use rsync with checksum instead of modification times to prevent failures

kichalla:


src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
```diff
@@ -104,7 +106,9 @@ then
 	{{ for excludedDir in DirectoriesToExcludeFromCopyToBuildOutputDir }}
 	excludedDirectories+=" --exclude {{ excludedDir }}"
 	{{ end }}
-	rsync -rtE --links $excludedDirectories . "$DESTINATION_DIR"
+
+	rsync -rEc --links $excludedDirectories . "$DESTINATION_DIR"
```
kichalla: @gpcastro Following are my findings based on which I sent out this PR:
- `--omit-dir-times` alone is not fixing the issue as I am still seeing the `Operation not permitted` errors.
- Removing the `-t` entirely (which includes removing the `omit-dir-times` switch), while does not run into the `Operation not permitted` errors, is worse of all as it is taking lot of time to copy to the destination directory (wwwroot here, which is mounted directory) for 2nd, 3rd builds etc. also
- Checksum is providing better performance(> 50%) compared to the above for 2nd, 3rd builds etc.

cc @sanchitmehta

kichalla: Just FYI...the nightly builds also passed on these changes.

gpcastro: Let's add a comment to this file with our findings regarding the timestamp usage and why we switched to checksums, so we don't go over this research again in the future

src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
```diff
@@ -104,7 +106,9 @@ then
 	{{ for excludedDir in DirectoriesToExcludeFromCopyToBuildOutputDir }}
 	excludedDirectories+=" --exclude {{ excludedDir }}"
 	{{ end }}
-	rsync -rtE --links $excludedDirectories . "$DESTINATION_DIR"
+
```
gpcastro: should we remove that copying to a separate directory in our tests to see if this fix helps those issues as well?

src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
```diff
@@ -46,7 +46,9 @@ then
 	{{ for excludedDir in DirectoriesToExcludeFromCopyToIntermediateDir }}
 	excludedDirectories+=" --exclude {{ excludedDir }}"
 	{{ end }}
-	rsync --delete -rt $excludedDirectories . "$INTERMEDIATE_DIR"
+
+	rsync --delete -rc $excludedDirectories . "$INTERMEDIATE_DIR"
```
gpcastro: do we need to use check sum here as well? Can't we use timestamps since this is inside the container?

src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
kichalla: Sure, will do

src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
kichalla: yeah, we could probably, but just thought it will be consistent across. I will change it to use timestamp instead

src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
kichalla: yes, good point. We should now be able to fix those issues as well. I will make that change as part of this PR.

PR #445, opened 2019-12-11
Add checks for Hexo and Hugo apps, add sample test apps for each

cormacpayne:


src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -79,6 +79,12 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
                 configureYarnCache = true;
                 packageInstallerVersionCommand = NodeConstants.YarnVersionCommand;
             }
+            else if (IsHugoSite(ctx.SourceRepo))
```
kichalla: I am wondering if this would break existing apps in AppService or in general new apps' experience. For example, `config.json` is a very common file that any app could have. Here we would be considering it as a hugo app instead.

images/build/slim.Dockerfile
```diff
@@ -120,6 +120,10 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends \
         jq \
     && rm -rf /var/lib/apt/lists/*
+RUN curl -fsSLO --compressed "https://github.com/gohugoio/hugo/releases/download/v0.59.1/hugo_0.59.1_Linux-64bit.tar.gz" \
```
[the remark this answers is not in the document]
cormacpayne: Added a variable for the Hugo version that we're using in this file.

src/BuildScriptGenerator/Node/NodeLanguageDetector.cs
```diff
@@ -149,5 +153,23 @@ private string DetectNodeVersion(dynamic packageJson)
...
+        {
+            return IsHugoSite(sourceRepo);
+        }
+
+        private bool IsHugoSite(ISourceRepo sourceRepo)
+        {
+            if (sourceRepo.FileExists(NodeConstants.HugoJsonFileName) ||
+                sourceRepo.FileExists(NodeConstants.HugoTomlFileName) ||
+                sourceRepo.FileExists(NodeConstants.HugoYamlFileName) ||
+                sourceRepo.DirExists(NodeConstants.HugoConfigFolderName))
```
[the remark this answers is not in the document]
cormacpayne: I updated the logic in this block: we check for `config.toml` at the root, a `config` folder with a `config.toml` file in it, or a `config` folder with a `*.toml` file anywhere in it. I scrapped the checks for `config.json` and `config.yml` since they don't seem to be the streamlined scenarios (I couldn't find any examples online of apps being built with these two files), so I stuck with the other scenarios outlined in the [Hugo configuration doc](https://gohugo.io/getting-started/configuration/). Let me know what you think.

src/BuildScriptGenerator/Node/NodePlatform.cs
cormacpayne: I removed this scenario -- see above comment.

src/BuildScriptGenerator/Node/NodeLanguageDetector.cs
```diff
@@ -149,5 +154,51 @@ private string DetectNodeVersion(dynamic packageJson)
...
+            if (sourceRepo.FileExists(NodeConstants.HugoTomlFileName))
+            {
+                lines = sourceRepo.ReadAllLines(NodeConstants.HugoTomlFileName);
+            }
+            else if (sourceRepo.FileExists(NodeConstants.HugoConfigFolderName, NodeConstants.HugoTomlFileName))
+            {
+                lines = sourceRepo.ReadAllLines(NodeConstants.HugoConfigFolderName, NodeConstants.HugoTomlFileName);
+            }
+
+            if (lines.Any(l => hugoVariables.Any(v => l.TrimStart(' ').StartsWith(v))))
```
kichalla: There should be Toml parser libraries available. I think we should use that to determine this.

src/BuildScriptGenerator/Node/NodeLanguageDetector.cs
```diff
@@ -149,5 +154,51 @@ private string DetectNodeVersion(dynamic packageJson)
 
             return nodeVersion;
         }
+
+        private bool IsStaticSite(ISourceRepo sourceRepo)
+        {
+            return IsHugoSite(sourceRepo);
+        }
+
+        private bool IsHugoSite(ISourceRepo sourceRepo)
```
kichalla: Let's make this `static` and call from NodePlatform to avoid duplicating logic

src/BuildScriptGenerator/Node/NodeLanguageDetector.cs
```diff
@@ -149,5 +154,51 @@ private string DetectNodeVersion(dynamic packageJson)
...
+        {
+            // Hugo configuration variables: https://gohugo.io/getting-started/configuration/#all-configuration-settings
```
kichalla: Seems like there are 3 different formats for this: json, yaml and toml. We should handle all these?

src/BuildScriptGenerator/Node/NodeLanguageDetector.cs
```diff
@@ -149,5 +154,51 @@ private string DetectNodeVersion(dynamic packageJson)
...
+                lines = sourceRepo.ReadAllLines(NodeConstants.HugoConfigFolderName, NodeConstants.HugoTomlFileName);
+            }
+
+            if (lines.Any(l => hugoVariables.Any(v => l.TrimStart(' ').StartsWith(v))))
+            {
+                return true;
+            }
+
+            // Search for config/*.toml.
+            if (sourceRepo.DirExists(NodeConstants.HugoConfigFolderName))
```
kichalla: nit: just a general guideline for readability...to avoid deep nested code, you can short-circuit the code early (ex: return false if directory does not exist)

src/BuildScriptGenerator/Node/NodeLanguageDetector.cs
cormacpayne: Included a Toml and Yaml parsers to help extract the information we want out of these files.

src/BuildScriptGenerator/Node/NodeLanguageDetector.cs
cormacpayne: Moved this functionality into its own helper class so it's not coupled to the `NodeLanguageDetector`

src/BuildScriptGenerator/Node/NodeLanguageDetector.cs
cormacpayne: Added checks for each file.

src/BuildScriptGenerator/Helpers/ParserHelper.cs
```diff
@@ -0,0 +1,58 @@
...
+        /// Parse a .yaml file into a YamlMappingNode from the YamlDotNet library.
+        /// See https://github.com/aaubry/YamlDotNet for more information.
+        /// </summary>
+        /// <param name="sourceRepo">Source repo for the application.</param>
+        /// <param name="filePath">The path to the .yaml file.</param>
+        /// <returns>A YamlMappingNode object containing information about the .yaml file.</returns>
+        public static YamlMappingNode ParseYamlFile(ISourceRepo sourceRepo, string filePath)
+        {
+            var yamlContent = sourceRepo.ReadFile(filePath);
+            var yamlStream = new YamlStream();
```
kichalla: Check if you need to dispose this stream

src/BuildScriptGenerator/Helpers/StaticSiteGeneratorHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+using System.IO;
+using System.Linq;
+using Microsoft.Oryx.BuildScriptGenerator.Node;
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    /// <summary>
+    /// Helper class for functions around static site generators.
+    /// </summary>
+    internal static class StaticSiteGeneratorHelper
```
kichalla: 👍

src/BuildScriptGenerator/Helpers/StaticSiteGeneratorHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+                tomlFilePaths.AddRange(configSourceRepo.EnumerateFiles("*.toml", true));
+
+                // Search for config/*.yaml
+                yamlFilePaths.AddRange(configSourceRepo.EnumerateFiles("*.yaml", true));
+
+                // Search for config/*.json
+                jsonFilePaths.AddRange(configSourceRepo.EnumerateFiles("*.json", true));
+            }
+
+            if (tomlFilePaths.Any())
```
kichalla: nit: need not have these conditions as execution would not go into the for loop

src/BuildScriptGenerator/Helpers/StaticSiteGeneratorHelper.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            return IsHugoApp(sourceRepo);
+        }
+
+        /// <summary>
+        /// Checks whether or not the given repository is a Hugo application.
+        /// </summary>
+        /// <param name="sourceRepo">Source repo for the application.</param>
+        /// <returns>True if the app is a Hugo app, false otherwise.</returns>
+        public static bool IsHugoApp(ISourceRepo sourceRepo)
```
kichalla: Seems like we can also look into environment variables to determine if it is a hugo app. We should probably check that first and then fall back to config files? https://gohugo.io/getting-started/configuration/#configure-with-environment-variables

src/BuildScriptGenerator/Helpers/ParserHelper.cs
cormacpayne: This type doesn't have a `.Dispose()` or `.Close()` call (I don't know why this type is called a stream, it inherits from an enumerable)

src/BuildScriptGenerator/Helpers/StaticSiteGeneratorHelper.cs
```diff
@@ -32,6 +33,16 @@ public static bool IsStaticSite(ISourceRepo sourceRepo)
         /// <returns>True if the app is a Hugo app, false otherwise.</returns>
         public static bool IsHugoApp(ISourceRepo sourceRepo)
         {
+            // Check for Hugo environment variables
+            var environmentVariables = Environment.GetEnvironmentVariables();
```
kichalla: Make this method take in `IEnvironment` so that if we want to unit test, it can be easy

src/BuildScriptGenerator/Helpers/StaticSiteGeneratorHelper.cs
```diff
@@ -32,6 +33,16 @@ public static bool IsStaticSite(ISourceRepo sourceRepo)
...
+            foreach (var key in environmentVariables?.Keys)
+            {
+                if (key.ToString().StartsWith("HUGO"))
```
kichalla: `StartsWith("HUGO_")`?

src/BuildScriptGenerator/Helpers/StaticSiteGeneratorHelper.cs
```diff
@@ -70,39 +81,30 @@ public static bool IsHugoApp(ISourceRepo sourceRepo)
                 jsonFilePaths.AddRange(configSourceRepo.EnumerateFiles("*.json", true));
             }
 
-            if (tomlFilePaths.Any())
+            foreach (var path in tomlFilePaths)
             {
-                foreach (var path in tomlFilePaths)
+                var tomlTable = ParserHelper.ParseTomlFile(sourceRepo, path);
+                if (tomlTable.Keys.Any(k => hugoVariables.Contains(k)))
```
kichalla: Are the keys case sensitive?

src/BuildScriptGenerator/Helpers/StaticSiteGeneratorHelper.cs
```diff
@@ -32,6 +33,16 @@ public static bool IsStaticSite(ISourceRepo sourceRepo)
...
+            foreach (var key in environmentVariables?.Keys)
+            {
+                if (key.ToString().StartsWith("HUGO"))
+                {
+                    return true;
+                }
+            }
+
             // Hugo configuration variables: https://gohugo.io/getting-started/configuration/#all-configuration-settings
             var hugoVariables = new[] { "baseURL", "title", "languageCode", "theme" };
```
kichalla: Could you add few more variables at the beginning of this array which look more specific to Hugo maybe? (languageCode seems good but others i am not sure)...just want to avoid hitting false positives when probing.

src/BuildScriptGenerator/Helpers/StaticSiteGeneratorHelper.cs
cormacpayne: Just verified they aren't case sensitive, so I had a comparer to ignore case sensitivity.

src/BuildScriptGenerator/Helpers/StaticSiteGeneratorHelper.cs
cormacpayne: Added a few more configuration variables to this list.

PR #448, opened 2019-12-13
Foundation PR for adopting PHP-FPM-NGINX into ORYX

arroyc:

PR #451, opened 2019-12-19
Foundation PR for adopting PHP-FPM-NGINX into ORYX

arroyc:
Created new and closed #448 . For some reason the commits are not showing correctly there


images/runtime/php-fpm/.gitignore
```diff
@@ -0,0 +1,2 @@
+Dockerfile
+base.Dockerfile
```
dorfire: do we need this file? doesn't the root .gitignore contain similar entries?

build/constants.yaml
```diff
@@ -4,6 +4,7 @@
     application-insights-instrumentation-key-env-var-name: ORYX_AI_INSTRUMENTATION_KEY
     app-service-app-name-env-var-name: APPSETTING_WEBSITE_SITE_NAME
     scm-commit-id-env-var-name: SCM_COMMIT_ID
+    php-origin-env-var-name: PHP_ORIGIN
```
dorfire: add docs - what is it used for? who defines it? who consumes it?

src/startupscriptgenerator/src/php/scriptgenerator.go
```diff
@@ -25,12 +27,25 @@ func (gen *PhpStartupScriptGenerator) GenerateEntrypointScript() string {
 
 	logger.LogInformation("Generating script for source at '%s'", gen.SourcePath)
 
+	var _phpOrigin = os.Getenv(consts.PhpOriginEnvVarName)
```
dorfire: do we need our own env var for that? can we / should we ask App Service to export some variable FOR us?

images/runtime/php-fpm/7.3/7.3.Dockerfile
```diff
@@ -0,0 +1,248 @@
+# From https://github.com/docker-library/php.git
```
dorfire: this file looks like an almost identical copy of the 7.2 dockerfile - can we use a template for the two?

images/runtime/php-fpm/.gitignore
arroyc: yeah I will remove it ..

build/constants.yaml
arroyc: sure ..

src/startupscriptgenerator/src/php/scriptgenerator.go
arroyc: there can be multiple ways, at this moment I am just trying to create a foundation image for app service team, once they are happy with it, we can accommodate these changes.

images/runtime/php-fpm/7.3/7.3.Dockerfile
arroyc: templating will be good, but as of now the scope is provide appservice something with a stable state. Going forward we will refactor these things as a seperate workitem

tests/Oryx.RuntimeImage.Tests/Php-fpm/PhpFpmImageTest.cs
```diff
@@ -0,0 +1,122 @@
...
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net.Http;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Oryx.RuntimeImage.Tests
```
kichalla: Can you write some integrations tests too?

arroyc: yes, that will be next step, integration tests will require adjusting build and release pipeline , and I don't want to address those in this PR. I have planned to create a separate PR with workitem for that.

tiagoefreitas:
Is nginx usable? Any image is using it? I wanted to create a php-fpm-nginx image for App Service since the examples in App Service are all outdated and don't use Oryx

PR #457, opened 2020-01-14
Add mechanism to disable telemetry in CLI, add GitHub Action check

cormacpayne:
- Allow users to opt out of Oryx CLI telemetry by setting the environment variable `DISABLE_ORYX_TELEMETRY` to `true`
- Add check for `GITHUB_ACTIONS` environment variable (as outlined [here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables#default-environment-variables))


src/Common/LoggingConstants.cs
```diff
@@ -13,6 +13,9 @@ public static class LoggingConstants
         public const string ApplicationInsightsInstrumentationKeyEnvironmentVariableName
             = "ORYX_AI_INSTRUMENTATION_KEY";
 
+        public const string DisableOryxTelemetryEnvironmentVariableName
+            = "DISABLE_ORYX_TELEMETRY";
```
gpcastro: our environment variables usually start with 'Oryx', which makes it easy to find all variables related to us. So I'd call it 'ORYX_DISABLE_TELEMETRY' instead.

cormacpayne:
@arroyc I was able to test this manually with the `oryx-test-other` Application Insights instrumentation key and nothing is sent to the logs when `ORYX_DISABLE_TELEMETRY` is set to true and passed as an environment variable to the `docker run` command.

PR #467, opened 2020-01-27
Write .npmrc with a custom registry URL

dorfire:


src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -4,6 +4,13 @@ node --version
 echo
 {{ PackageInstallerVersionCommand }}
 
+{{ if PackageRegistryUrl | IsNotBlank }}
+echo
+echo "Adding package registry to .npmrc: {{ PackageRegistryUrl }}"
+echo "registry={{ PackageRegistryUrl }}" > ~/.npmrc
```
kichalla: Just trying to understand the scenario here...here `registry` value is being written to a file in home directory. If a repo has an npmrc file, I believe that file's registry value takes precedence over the value in this home directory, Is that expected?
But for all other scenarios where there is no .npmrc file in the repo, this is good since this is non-intrusive solution to get our secure packages?

dorfire: Correct - if someone has an .npmrc for some reason (most don't), it will take precedence

kichalla: Sound good. Thanks.

PR #471, opened 2020-01-30
fix ci build by adding tag files to the images that doesn't follow existing convention (image repo has '-' in the name like php-fpm)

arroyc:


build/__functions.sh
```diff
@@ -22,7 +22,20 @@ function getTagName()
 	fi
 
 	local replacedPath="$RUNTIME_IMAGES_SRC_DIR/"
+	echo "Runtime Image Source Directory: "$RUNTIME_IMAGES_SRC_DIR
 	local remainderPath="${1//$replacedPath/}"
+	echo "RemainderPath: "$remainderPath
+	tagNameFile="$RUNTIME_IMAGES_SRC_DIR/$remainderPath/tag.txt"
+	echo "TagFileName: "$tagNameFile
+	ls -l $replacedPath
```
kichalla: remove. below line too

kichalla: nit: Change PR name to something descriptive of what we are doing

PR #472, opened 2020-02-02
fix upload to azure storage task in release template

arroyc:
fix upload to azure storage task in release template


arroyc:
@kichalla took care of it

PR #475, opened 2020-02-03
Install sdks dynamically-part1

kichalla:


arroyc:
nit: buildimage test failures

.gitignore
```diff
@@ -267,4 +267,5 @@ src/startupscriptgenerator/**/vendor/
 
 # Go debugging related generated files
 src/startupscriptgenerator/src/**/__debug_bin
-src/startupscriptgenerator/src/**/run.sh
\ No newline at end of file
+src/startupscriptgenerator/src/**/run.sh
+!**/dynamic/Dockerfile
```
arroyc: curious: what does ! stands for?

kichalla: We currently ignore all files which are dynamically generated in runtime folder (i.e the ones generated using `generateDockerfiles.sh`), but this also ignores the static Dockerfile created under `dynamic` folder, so this `!` says to not ignore that file and allow us to checkin

PR #477, opened 2020-02-06
remove unused variable

qianz2:

### DIFF
--- a/tests/BuildScriptGenerator.Tests/BuildScriptGenerator.Tests.csproj
+++ b/tests/BuildScriptGenerator.Tests/BuildScriptGenerator.Tests.csproj
@@ -11,6 +11,10 @@
     <RootNamespace>Microsoft.Oryx.BuildScriptGenerator.Tests</RootNamespace>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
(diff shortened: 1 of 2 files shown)


tests/BuildScriptGenerator.Tests/BuildScriptGenerator.Tests.csproj
```diff
@@ -11,6 +11,10 @@
     <RootNamespace>Microsoft.Oryx.BuildScriptGenerator.Tests</RootNamespace>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
gpcastro: This shouldn't be required; does the tests run on your machine without this block?

kichalla: I suggested this change to @qianz2. This is a temporary workaround suggested by the VS team to enable running tests in VS currently. This started happening after `16.4` version (if you recall I sent out an email to the team regarding this few weeks ago).  This happens because we are using 2.1 version in Oryx whereas Terrapin uses 3.1 and does not hit the same issue. I would suggest to check this in until VS team resolves it.
https://developercommunity.visualstudio.com/content/problem/831442/test-explorer-wont-run-tests-too-particular-about.html

PR #485, opened 2020-02-13
fix build image builds temporarily

arroyc:
please take a look, temporarily disabled apt-get upgrades to make the build work

### DIFF
--- a/images/build/Dockerfile
+++ b/images/build/Dockerfile
@@ -15,7 +15,7 @@ ENV LANG C.UTF-8
 
 # Install basic build tools
 RUN apt-get update \
-    && apt-get upgrade -y \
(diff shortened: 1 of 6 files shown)


images/build/Dockerfile
```diff
@@ -15,7 +15,7 @@ ENV LANG C.UTF-8
 
 # Install basic build tools
 RUN apt-get update \
-    && apt-get upgrade -y \
+#    && apt-get upgrade -y \
```
kichalla: Why these changes?

kichalla: just FYI...we added the upgrade to avoid getting flagged by S360 for missing security updates...

arroyc:
closing on behalf of #486

PR #486, opened 2020-02-14
fix build image build break

arroyc:


tests/SampleApps/python/python2-flask-app/requirements.txt
```diff
@@ -1 +1,2 @@
 Flask==0.9
+Werkzeug==0.9
```
arroyc: this is to get rid of python integration test failure …
File "/oryxtests/python2-flask-app/pythonenv2.7/lib/python2.7/site-packages/flask/sessions.py", line 14, in <module>
                from werkzeug.contrib.securecookie import SecureCookie
            ImportError: No module named contrib.securecookie

PR #487, opened 2020-02-15
Gets started time and completed time for running jobs within a Github…

qianz2:
Pass in data as two optional parameters (or we could do as environment variables?)


actions/entrypoint.sh
```diff
@@ -0,0 +1,60 @@
...
+startTime=$(curl -X GET "${url}" | sed 's/,/\n/g' | grep "started_at" | awk '{print $2}' | sed -n '3p')
+endTime=$(curl -X GET "${url}" | sed 's/,/\n/g' | grep "completed_at" | awk '{print $2}' | sed -n '3p')
+
+echo "Started time is ${startTime}."
+echo "Completed time is ${endTime}."
+
+#pass time data as two parameters (or as environment variables?)
+#oryxCommand="${oryxCommand} --github-build-start ${startTime} --github-build-end ${endTime}"
+
+eval $oryxCommand
```
kichalla: I am confused by what does `Completed time` here means. How can we get the completed time even though the build has not happened yet?

qianz2: The completed time (format example: "2020-02-15T02:51:50.000Z") is the time when GitHub action finished downloading images from mcr.microsoft.com/oryx/build:latest. This entrypoint.sh will run afterwards.

kichalla: Ah gotcha. Thanks for clarifying. I think the name and description of the options should clearly say that. Currently it says `build` which is kind of misleading.

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -43,6 +51,14 @@ public int OnExecute(CommandLineApplication app, IConsole console)
...
                 {
                     logger?.LogInformation("The current Oryx command is being run from within a GitHub Action.");
+                    var githubActionBuildStartTime = GitHubBuildStartTime;
+                    var githubActionBuildEndTime = GitHubBuildEndTime;
+                    var buildEventProps = new Dictionary<string, string>()
+                    {
+                        { "githubActionBuildStartTime", githubActionBuildStartTime},
+                        { "githubActionBuildEndTime", githubActionBuildEndTime},
+                    };
+                    logger.LogEvent("GithubActionTimeLog", buildEventProps);
```
kichalla: We should have appropriate name for this event too

actions/entrypoint.sh
qianz2: How about "GITHUB_BUILDIMAGE_START" and "GITHUB_BUILDIMAGE_COMPLETE"? Also, I'm trying to change them to be env variables.

actions/entrypoint.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/sh -l
```
cormacpayne: @qianz2 do we need to add this file? Any changes that we're making to our GitHub Action should be done in its repository, [`appservice-build`](https://github.com/Azure/appservice-build)

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -26,6 +28,12 @@ internal abstract class CommandBase
         [Option("--debug", Description = "Print stack traces for exceptions.")]
         public bool DebugMode { get; set; }
 
+        [Option("--github-build-start", Description = "The time when a Github Action started the build.")]
+        public string GitHubBuildStartTime { get; set; }
```
cormacpayne: @qianz2 I'm in the boat that these GitHub Action-specific properties we look for should be set as environment variables during the action -- this will keep us from introducing new parameters each time there's another property of an action we want to capture and use in some fashion, and potentially "cluttering" the set of parameters we have for our CLI.

What do you think? CC: @kichalla @arroyc

actions/entrypoint.sh
qianz2: Gotcha, I'm not sure where to commit my changes at first. It not necessary for oryx to have this file, I'll remove it later :)

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
kichalla: 👍 for environment variables

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
qianz2: Gotcha, I'll think of a better name:)

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
qianz2: I vote for environment variable too

actions/entrypoint.sh
arroyc: IMO environment variable will be better than build parameter, essentially we will be checking the value of the environment variables from oryx to send the calculated value in the telemetry. environment variables or build parameters both works, but imo env variable will help us not to overcrowd oryx build with  parameters that has a limited scope

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -26,6 +28,12 @@ internal abstract class CommandBase
...
+
+        [Option("--github-build-end", Description = "The time when a Github Action completed the build.")]
+        public string GitHubBuildEndTime { get; set; }
+
```
arroyc: Imo environment variable makes much more sense instead of build parameters

actions/entrypoint.sh
qianz2: @arroyc Thanks for the advice. Env variable will save us from having too much CLI parameters. I'll go for env variables.

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -43,6 +51,14 @@ public int OnExecute(CommandLineApplication app, IConsole console)
                 if (envSettings != null && envSettings.GitHubActions)
                 {
                     logger?.LogInformation("The current Oryx command is being run from within a GitHub Action.");
+                    var githubActionBuildStartTime = GitHubBuildStartTime;
+                    var githubActionBuildEndTime = GitHubBuildEndTime;
+                    var buildEventProps = new Dictionary<string, string>()
+                    {
+                        { "githubActionBuildStartTime", githubActionBuildStartTime},
```
dorfire: CamelCase names - github -> gitHub

src/BuildScriptGeneratorCli/CliEnvironmentSettings.cs
```diff
@@ -17,6 +17,8 @@ public class CliEnvironmentSettings
         private const string DisableNodeJsEnvVarName = "DISABLE_NODEJS_BUILD";
         private const string EnableMultiPlatformBuildEnvVarName = "ENABLE_MULTIPLATFORM_BUILD";
         private const string GitHubActionsEnvVarName = "GITHUB_ACTIONS";
+        private const string GitHubActionsBuildContainerStartTimeEnvVarName = "GITHUB_BUILD_CONTAINER_START";
```
cormacpayne: nit: I'm not sure if we would reuse these environment variables in the future for tracking image pull times, but since we're currently only using it for GitHub Actions metrics, I would lean towards prefixing these environment variables with `GITHUB_ACTIONS`.

I would then call the two environment variables
- `GITHUB_ACTIONS_BUILD_START_TIME`
- `GITHUB_ACTIONS_BUILD_END_TIME`

Thoughts? CC: @kichalla @arroyc

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -43,6 +45,30 @@ public int OnExecute(CommandLineApplication app, IConsole console)
                 if (envSettings != null && envSettings.GitHubActions)
                 {
                     logger?.LogInformation("The current Oryx command is being run from within a GitHub Action.");
+
+                    // Format: "2020-02-15T02:51:50.000Z"
+                    var gitHubActionBuildContainerStartTime = envSettings.GitHubActionsBuildContainerStartTime;
+                    var gitHubActionBuildContainerEndTime = envSettings.GitHubActionsBuildContainerEndTime;
+
+                    var dateStart = gitHubActionBuildContainerStartTime.Split('T')[0];
```
cormacpayne: @qianz2 we should be able to avoid doing all of this parsing by using `DateTime.Parse(...)`

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -43,6 +45,30 @@ public int OnExecute(CommandLineApplication app, IConsole console)
...
+                                                  int.Parse(timeStart.Substring(0, 2)), int.Parse(timeStart.Substring(3, 2)), int.Parse(timeStart.Substring(6, 2)));
+                    DateTime date2 = new DateTime(int.Parse(dateEnd.Substring(0, 4)), int.Parse(dateEnd.Substring(5, 2)), int.Parse(dateEnd.Substring(8, 2)),
+                                                  int.Parse(timeEnd.Substring(0, 2)), int.Parse(timeEnd.Substring(3, 2)), int.Parse(timeEnd.Substring(6, 2)));
+                    TimeSpan interval = date2 - date1;
+                    var gitHubActionBuildContainerDurationSeconds = interval.Seconds.ToString();
+                    var buildEventProps = new Dictionary<string, string>()
+                    {
+                        { "gitHubActionBuildContainerStartTime", gitHubActionBuildContainerStartTime },
+                        { "gitHubActionBuildContainerEndTime", gitHubActionBuildContainerEndTime },
+                        { "gitHubActionBuildContainerDurationSeconds", gitHubActionBuildContainerDurationSeconds },
```
cormacpayne: @qianz2 nit: if we're sending the elapsed time for building the action, I don't think we need to also send the start and end times (I can't think of any way that we'd use those pieces of information on top of the elapsed time)

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
qianz2: Excellent suggestion!

src/BuildScriptGeneratorCli/CliEnvironmentSettings.cs
qianz2: I think it's a good idea to have prefix👍

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -43,6 +45,15 @@ public int OnExecute(CommandLineApplication app, IConsole console)
                 if (envSettings != null && envSettings.GitHubActions)
                 {
                     logger?.LogInformation("The current Oryx command is being run from within a GitHub Action.");
+                    DateTime date1 = DateTime.Parse(envSettings.GitHubActionsBuildStartTime);
```
cormacpayne: @qianz2 for safety, I think we should first do a null-check on the two environment variables to make sure that they are set, and then use `DateTime.TryParse` to avoid any exception from being thrown for incorrectly set dates. It would look something like

```cs
DateTime startTime, endTime;
if (DateTime.TryParse(envSettings.GitHubActionsBuildStartTime, out startTime) &&
    DateTime.TryParse(envSettings.GitHubActionsBuildEndTime, out endTime))
{
   // Log results
}
```

src/BuildScriptGeneratorCli/CliEnvironmentSettings.cs
```diff
@@ -17,6 +17,8 @@ public class CliEnvironmentSettings
         private const string DisableNodeJsEnvVarName = "DISABLE_NODEJS_BUILD";
         private const string EnableMultiPlatformBuildEnvVarName = "ENABLE_MULTIPLATFORM_BUILD";
         private const string GitHubActionsEnvVarName = "GITHUB_ACTIONS";
+        private const string GitHubActionsBuildStartTimeEnvVarName = "GITHUB_ACTIONS_BUILD_START_TIME";
+        private const string GitHubActionsBuildEndTimeEnvVarName = "GITHUB_ACTIONS_BUILD_END_TIME";
```
kichalla: Would the following be appropriate?
`GITHUB_ACTIONS_BUILD_START_TIME` => `GITHUB_ACTIONS_BUILD_IMAGE_PULL_START_TIME`
`GITHUB_ACTIONS_BUILD_END_TIME` => `GITHUB_ACTIONS_BUILD_IMAGE_PULL_END_TIME`

qianz2: Thanks! I'll go for these names, they are specific and no ambiguity.

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -43,6 +45,20 @@ public int OnExecute(CommandLineApplication app, IConsole console)
...
+                    if (envSettings.GitHubActionsBuildImagePullStartTime != null
+                        && envSettings.GitHubActionsBuildImagePullEndTime != null
+                        && DateTime.TryParse(envSettings.GitHubActionsBuildImagePullStartTime, out startTime)
+                        && DateTime.TryParse(envSettings.GitHubActionsBuildImagePullEndTime, out endTime))
+                    {
+                        TimeSpan interval = endTime - startTime;
+                        var gitHubActionBuildImagePullDurationSeconds = interval.TotalSeconds.ToString();
+                        var buildEventProps = new Dictionary<string, string>()
+                        {
+                            { "gitHubActionBuildImagePullDurationSeconds", gitHubActionBuildImagePullDurationSeconds },
```
arroyc: spacing?

PR #492, opened 2020-02-19
Publish Oryx binaries as tar.gz file to Github release

kichalla:


arroyc:
seems like in the ci-build download pipeline artifact task has failed … it's may not be your change which caused this, it may be azure devops again

PR #498, opened 2020-02-21
update platform images (php, dotnetcore, composer)

arroyc:
update platform images for php and dotnet core 2.2 along with composer

todo: once base image build is successful, tag name needs to be changed


build/__dotNetCoreSdkVersions.sh
```diff
@@ -2,6 +2,6 @@
 
 DOT_NET_CORE_11_SDK_VERSION='1.1.14'
 DOT_NET_CORE_21_SDK_VERSION='2.1.804'
-DOT_NET_CORE_22_SDK_VERSION='2.2.402'
+DOT_NET_CORE_22_SDK_VERSION='2.2.207'
```
kichalla: I would avoid doing this. .NET Core team has 207 released after 402 which was kind of weird. So just skip this version 2.2. also 2.2 is not supported anymore

images/runtime/node/12.16/base.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM oryx-node-run-base
```
kichalla: We should not need this file. Update the `12`'s dockerfile

images/runtime/node/13.9/base.Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN ARCH= && dpkgArch="$(dpkg --print-architecture)" \
     *) echo "unsupported architecture"; exit 1 ;; \
   esac
 
-ENV NODE_VERSION 12.9.1
```
kichalla: is 12.9 specified by AppService? if so, we should not do this and add a new `13` file

PR #502, opened 2020-02-24
Added a new dockerfile for build image to be used in GitHub actions

kichalla:


images/build/GitHubActions.Dockerfile
```diff
@@ -0,0 +1,165 @@
...
+ARG IMAGES_DIR=/tmp/oryx/images
+ARG BUILD_DIR=/tmp/oryx/build
+ARG SDK_STORAGE_ENV_NAME
+ARG SDK_STORAGE_BASE_URL_VALUE
+# Determine where the image is getting built (DevOps agents or local)
+ARG AGENTBUILD
+
+# Use the exact layer that is present on the GitHub agent machines
+# https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md
+FROM buildpack-deps:stretch@sha256:8bcd320ec29cf67052985f28891586fb853051f69ad0646fc7a49f47d6e3ee1a AS main
```
cormacpayne: @kichalla I'm not sure how often GitHub is updating their runners with a new SHA of this image, so my only concern would be keeping this Dockerfile in sync with their cache, and what it means to our pull times if it becomes out of sync.

images/build/GitHubActions.Dockerfile
```diff
@@ -0,0 +1,165 @@
+# Folders in the image which we use to build this image itself
```
cormacpayne: @kichalla is the goal of this image to improve the pull time in a GitHub Action workflow by basing it off of the same `buildpack-deps` image they're using, or are there additional benefits we're trying to get from it as well?

kichalla: Yes, the goal is improve to the pull time in a GitHub action workflow. This PR uses the same layers of buildpacks-deps that currently the GitHub agents. This image also does not install any platform SDKs (except Yarn and Hugo for now). Platform SDKs would be dyanmically installed(Python for now, node and .NET Core support will be added soon).

images/build/GitHubActions.Dockerfile
kichalla: Yes, that is a cause of concern. We have to work with DevOps team to make sure both of our team's base images are aligned.

build/__variables.sh
```diff
@@ -26,6 +26,7 @@ declare -r ORYXTESTS_SLIM_BUILDIMAGE_DOCKERFILE="$REPO_DIR/tests/images/build/sl
 declare -r RUNTIME_IMAGES_SRC_DIR="$REPO_DIR/images/runtime"
 declare -r BUILD_IMAGES_CLI_DOCKERFILE="$REPO_DIR/images/build/cli.Dockerfile"
 declare -r BUILD_IMAGES_AZ_FUNCS_JAMSTACK_DOCKERFILE="$REPO_DIR/images/build/AzureFunctions.JamStack.Dockerfile"
+declare -r BUILD_IMAGES_GITHUB_ACTIONS_DOCKERFILE="$REPO_DIR/images/build/GitHubActions.Dockerfile"
```
arroyc: nit: I may be thinking too much … but I feel its more like base image for all the future githubaction images .. so can we name it accordingly?

images/build/GitHubActions.Dockerfile
arroyc: nit: use directories instead of folders may be?

images/build/GitHubActions.Dockerfile
```diff
@@ -0,0 +1,165 @@
...
+ENV ${SDK_STORAGE_ENV_NAME} ${SDK_STORAGE_BASE_URL_VALUE}
+
+ARG GIT_COMMIT=unspecified
+ARG BUILD_NUMBER=unspecified
+ARG RELEASE_TAG_NAME=unspecified
+LABEL com.microsoft.oryx.git-commit=${GIT_COMMIT}
+LABEL com.microsoft.oryx.build-number=${BUILD_NUMBER}
+LABEL com.microsoft.oryx.release-tag-name=${RELEASE_TAG_NAME}
+
+ENTRYPOINT [ "benv" ]
```
arroyc: nit: This dockerfile seems to have lot of similarities with azurefunction/slim.dockerfile of oryx ... can we add this as basic image layer to all the images?

tests/Oryx.Integration.Tests/Python/PythonDynamicInstallationTest.cs
```diff
@@ -65,6 +65,50 @@ public async Task CanBuildAndRunPythonApp(string pythonVersion)
                 });
         }
 
+        [Theory]
+        [InlineData("2.7")]
```
arroyc: user constants from constant file instead of hardcoded versions?

build/__variables.sh
kichalla: The thing is this image is strongly tied to the layers of the image that the GitHub agent machines have so we might not want it to be the base image. AppService could use latest version of buildpack-deps having the security patches and need not be strongly tied to the one which GitHub action uses.

images/build/GitHubActions.Dockerfile
kichalla: This is just  a comment :-)

images/build/GitHubActions.Dockerfile
kichalla: Check my earlier comment regarding this.

tests/Oryx.Integration.Tests/Python/PythonDynamicInstallationTest.cs
kichalla: Sure

PR #504, opened 2020-02-25
Append to .npmrc instead of overwriting

dorfire:


dorfire:
> Looks great. I noticed that NodeBuildScriptGenerationTest.cs also has '> ~/.npmrc'. Do we also need to make some changes on that file?

Good catch, updated the tes

PR #509, opened 2020-02-27
change .NET core framework 2.1 to 3.1

qianz2:
Move oryx repo from .NET core 2.1 to 3.1.
In order to reduce command line length and size to pull images

### DIFF
--- a/src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
+++ b/src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
-    <RuntimeFrameworkVersion>2.1</RuntimeFrameworkVersion>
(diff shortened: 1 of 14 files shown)


src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
-    <RuntimeFrameworkVersion>2.1</RuntimeFrameworkVersion>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
kichalla: Can you create new projects in 3.1 and see what new xml elements they generate and have them similarly here? Idea is to avoid having elements which might not make sense any longer.

kichalla: Also we need to place a request with the DDIT team to upgrade the .NET Core SDK to 3.1 version first otherwise your PR build will not succeed.

arroyc:
@qianz2  .. do we still need this PR? if not you can close it

PR #511, opened 2020-03-02
Ignore case when reading blob metadata

kichalla:


src/BuildScriptGenerator/SdkStorageVersionProviderBase.cs
```diff
@@ -7,54 +7,58 @@
...
-                .GetStringAsync($"{sdkStorageBaseUrl}/{platformName}?restype=container&comp=list&include=metadata")
-                .Result;
-            var xdoc = XDocument.Parse(blobList);
+            var blobServiceClient = new BlobServiceClient(serviceUri: new Uri(sdkStorageBaseUrl));
+            var blobContainerClient = blobServiceClient.GetBlobContainerClient(platformName);
+
+            // Example: nodejs-10.1.0.tar.gz is the format of a blob
+            var blobList = blobContainerClient.GetBlobs(BlobTraits.Metadata, prefix: $"{platformName}-");
             var supportedVersions = new List<string>();
-            foreach (var runtimeVersionElement in xdoc.XPathSelectElements(
```
kichalla: There seem to be no straight forward way to find elements by ignoring case, so I chose to use the Azure Blob API to read the blob metadata which is case-INsensitive. Due to this we no longer need HttpClient api.

PR #512, opened 2020-03-02
Added sanity tests to verify dev and prod storage accounts.

kichalla:


platforms/__common.sh
```diff
@@ -39,6 +39,12 @@ shouldBuildSdk() {
 	else
 		# return whatever exit cdoe the following returns
 		blobExists $containerName $blobName
+		exitCode=$?
+		if [ "$exitCode" == 0 ]; then
+			return 1
```
arroyc: should we follow exitcode = 0 no error? why are we changing it here?

kichalla: Look at the method name `shouldBuildSdk`, if the blob exists we do not want to build the sdk again. Earlier it was building again even though it already exists. So fixed it

arroyc: thanks for explaining .. the comment "# return whatever exit cdoe the following returns" was bit confusing .. also nit: code instead of cdoe :)

PR #513, opened 2020-03-03
Remove extra execute signatures

qianz2:
Remove extra execute signatures.

### DIFF
--- a/src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
+++ b/src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
@@ -111,8 +111,6 @@ namespace Microsoft.Oryx.BuildScriptGeneratorCli
 
         internal override int Execute(IServiceProvider serviceProvider, IConsole console)
         {
-            DataReceivedEventHandler stdOutHandler = null;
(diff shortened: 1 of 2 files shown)


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -111,16 +111,8 @@ public static string BuildOperationName(IEnvironment env)
...
-
-        // To enable unit testing
-        internal int Execute(
-            IServiceProvider serviceProvider,
-            IConsole console,
-            DataReceivedEventHandler stdOutHandler,
-            DataReceivedEventHandler stdErrHandler)
-        {
+            DataReceivedEventHandler stdOutHandler = null;
+            DataReceivedEventHandler stdErrHandler = null;
```
dorfire: do we need these two vars? we can just pass nulls to `serviceProvider.GetRequiredService<IScriptExecutor>().ExecuteScript(...)`

qianz2: You are right, I see them only used once inside ExecuteScript(). Will remove them

PR #514, opened 2020-03-03
Created a action workflow to test few samples with our github actions

kichalla:
cc @cormacpayne @gpcastro @qianz2


.github/workflows/sample-apps-with-v2-beta.yml
```diff
@@ -0,0 +1,24 @@
+name: Building sample apps with v2-beta action
+
+on:
+  push:
+    branches:
+      - master
+      - patch/**
+      - actions/**
```
kichalla: I added this branch pattern so that if anyone wants to test a workflow with an action, then they can create a branch in the format `actions/kichalla/testing.v3.action` for example

.github/workflows/sample-apps-with-v2-beta.yml
```diff
@@ -0,0 +1,24 @@
+name: Building sample apps with v2-beta action
```
kichalla: I wanted a way for us to readily look to see the status of our existing github actions against few samples. Initially I wanted to create few sample apps in the actions repo itself but that might increase the pull time of the actions for actual users, so instead creating it here.

kichalla: We should probably also have an end-to-end sample of using Oryx to build and deploy to a sample Azure Web App site. This way end users would have a quick sample to look at it and get a feel of an E2E with Oryx.

.github/workflows/sample-apps-with-v2-beta.yml
```diff
@@ -0,0 +1,24 @@
...
+
+on:
+  push:
+    branches:
+      - master
+      - patch/**
+      - actions/**
+jobs:
```
kichalla: https://github.com/microsoft/Oryx/runs/482915102?check_suite_focus=true

.github/workflows/sample-apps-with-v2-beta.yml
```diff
@@ -0,0 +1,24 @@
+name: Building sample apps with v2-beta action
+
+on:
+  push:
+    branches:
+      - master
+      - patch/**
```
cormacpayne: If we're looking to have a mechanism to test our action, I don't know if we necessarily need to kick off one of these builds each time someone pushes to `master` or a patch branch. Since this workflow is fixed to a specific version of our action, and that action's version is fixed to a specific image tag, wouldn't we essentially be running the same build each time unless a change was made in this workflow to update the version of our action used?

.github/workflows/sample-apps-with-v2-beta.yml
cormacpayne: I don't think we should necessarily bind the name of this file to the version of the action we're using -- if this workflow is used for testing our action against different scenarios, I think the workflow would look something like the following:

- Release a new version of our action
- Open a pull request against `master` from a new `actions/**` branch that updates this workflow to reflect the new version
- Verify that the kicked-off actions build succeeds and merge the pull request

Do you think that there are scenarios where we would want to support multiple versions of the action at once? If so, we could update this workflow to kick off multiple runs per build that would each test a different version of the action.

.github/workflows/sample-apps-with-v2-beta.yml
kichalla: yeah I agree and was also thinking about the same. Things could change in two places, one in the actions and other in the sample apps itself. I wanted a place to see the whole E2E flow and potentially point end users to an example of how things work. Ideally I would have a separate repo for samples which demo our github actions but I would like to avoid doing that for now.
Building on every merge to `master` might not be that bad since it is not as frequent as other private branches.
I was looking into using the `repository_dispatch` even too which will enable us to trigger an action from the actions repo to this repo whenever any changes happen. But this will not handle any changes to the sample app in this repo itself though.
Thoughts?

.github/workflows/sample-apps-with-v2-beta.yml
cormacpayne: That makes sense -- I think we can get away with having a single workflow file that kicks off a single job with multiple runs, each testing a different version of the action that we're interested in. Initially, the file might be large and repetitive since we would be testing the same apps across different action versions, but the GitHub Actions team said they're looking into some form of templating (similar to what we do now for ADO). Having a single workflow will make it easier to track all of the runs for a given branch push (or any additional triggers we may add).

.github/workflows/sample-apps-with-v2-beta.yml
cormacpayne: That's true, we don't push to `master` that often, so we could do that as well, I just wasn't sure what kind of value we would get out of jobs that didn't change the workflow file since the results _should_ be the same (but maybe that's what the jobs check 😊). I think we should be fine with triggering the job on pushes to `master` and `actions/**`.

.github/workflows/v1.yml
```diff
@@ -0,0 +1,25 @@
+on:
```
cormacpayne: I think it would be better to consolidate the two workflows into one, having them executed as separate jobs within the same run, which will reduce the number of runs that appear on our actions page. The workflow would look something like this:

```
on:
  push:
    branches:
      - master
      - actions/**

jobs:
  v1:
    runs-on: ubuntu-latest
    steps:
        ...
  v2:
    runs-on: ubuntu-latest
    steps:
      ...
```

This will run the workflows in parallel jobs and easy to reference within a single run. What do you think?

kichalla: Sure, sounds good to me. I will be closing this PR and will create a new one in the new repo that we discussed in scrum meeting today.

kichalla:
Closing this PR in favor of creating ones in repo: https://github.com/MicrosoftOryx/githubactions-samples

PR #516, opened 2020-03-04
Fix app platform version detection and resolution

kichalla:


tests/BuildScriptGeneratorCli.Tests/BuildpackDetectCommandTest.cs
```diff
@@ -100,6 +100,7 @@ private static IServiceProvider GetServiceProvider(BuildpackDetectCommand cmd)
                 {
                     svcs.Replace(ServiceDescriptor.Singleton(typeof(IEnvironment), env));
                     svcs.AddSingleton<INodeVersionProvider, TestNodeVersionProvider>();
+                    svcs.AddSingleton<IPhpVersionProvider, TestPhpVersionProvider>();
```
arroyc: why?>

arroyc: nit: Can we change the filename BuildpackDetectCommandTest to something else? I was initially confused .. I thought we adding php in buildpack .. so I made that "why" comment

arroyc:
I would also like to see a successful nightly build

kichalla:
> I would also like to see a successful nightly build

Actually I had run the Nightly build but with an older commit : https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3527848

I had made 2 commits after that nightly build, but these 2 commits changes are related to unit tests which have been tested already in the validation run.

images/build/benv.sh
```diff
@@ -24,33 +24,18 @@ matchesName() {
...
 done < <(set | grep -i '^python=')
-while read benvvar; do
-  set -- "$benvvar" "$@"
-done < <(set | grep -i '^python_version=')
 while read benvvar; do
   set -- "$benvvar" "$@"
 done < <(set | grep -i '^node=')
-while read benvvar; do
-  set -- "$benvvar" "$@"
-done < <(set | grep -i '^node_version=')
```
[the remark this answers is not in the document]
kichalla: 2 things: 
1. I want to remove `benv` from supporting this since we have the logic (the spec which i sent in an email last week) within the c# code.
2. Sure we can consider both `node_vesion` and as well as `nodejs_version` in our code.

tests/BuildScriptGeneratorCli.Tests/BuildpackDetectCommandTest.cs
kichalla: I added this because this was failing a test. In this PR I modified the logic as per the spec.

PR #517, opened 2020-03-05
Fixed dotnet core base image break and add Python version 3.8.2

kichalla:


platforms/python/buildPython.sh
```diff
@@ -46,17 +45,12 @@ buildPython() {
 		else
 			dockerFile="$pythonPlatformDir/$dockerFile"
 		fi
-
-		if [ -z "$pipVersion" ]; then
```
kichalla: Made this change since we have come across a requirement where we lock into specific version of PIP for a particular Python version.

PR #518, opened 2020-03-07
Enable building and copying output to nested directory of the source directory

kichalla:


tests/Oryx.Integration.Tests/DotNetCore/DotNetCoreRuntimeVersion31Tests.cs
```diff
@@ -62,5 +62,49 @@ public async Task CanBuildAndRun_NetCore31WebApp()
                     Assert.Contains("Welcome to ASP.NET Core MVC!", data);
                 });
         }
+
+        [Fact]
+        public async Task CanBuildAndRunApp_FromNestedOutputDirectory()
```
arroyc: how about a test where the intermediate directory and output directory both are nested under source directory?
intermediate directory is nested under output directory?

kichalla: > how about a test where the intermediate directory and output directory both are nested under source directory?

We currently do not allow nesting an intermediate directory under the source directory. We throw an `InvalidUsageException`. We do this because this scenario is not common for folks to do(but if you feel strongly about it, probably we can discuss within the team and handle it in a separate PR). But yes we do allow the output directory to be nested under source directory and this PR is about supporting that.

> intermediate directory is nested under output directory?

This is not common scenario for users to do.

PR #521, opened 2020-03-09
Use 'create-script' command for all runtime script generators

kichalla:
Currently only Python and Node runtime script generators have the `create-script` and `version` subcommands, With this PR I made change to even make DotNetCore and PHP runtime script generators to have this subcommand. This is so that AppService does not need to do special logic per platform and can have a consistent experience all across.

### DIFF
--- a/src/startupscriptgenerator/src/common/commandLineFlags.go
+++ b/src/startupscriptgenerator/src/common/commandLineFlags.go
@@ -10,18 +10,26 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 )
 
-func ValidateCommands(versionCommand *flag.FlagSet, scriptCommand *flag.FlagSet, setupEnvCommand *flag.FlagSet) {
+func getCommandNameList(commands []*flag.FlagSet) []string {
+	cmdsList := make([]string, len(commands))
+	for i := 0; i < len(cmdsList); i++ {
+		cmdsList[i] = commands[i].Name()
+	}
+	return cmdsList
+}
+
+func ValidateCommands(commands []*flag.FlagSet) {
 	// Verify that a subcommand has been provided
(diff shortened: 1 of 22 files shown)


src/startupscriptgenerator/src/common/commandLineFlags.go
```diff
@@ -10,34 +10,52 @@ import (
...
-		os.Exit(1)
+	} else {
+		isValidCommand := false
+		for i := 0; i < len(commands); i++ {
+			if suppliedCommandName == commands[i].Name() {
+				isValidCommand = true
+				commands[i].Parse(os.Args[2:])
+				break
+			}
+		}
```
qianz2: This for loop assumes there is only one valid command, do we guarantee such assumption?

kichalla: I did not understand your question. Here we try to find the command which is of the same name as what the end user has provided. If we find it, then we break.

qianz2: Could we avoid for loop and use an existing method to check if the 'commands' array contains 'suppliedCommandName'?

kichalla: `suppliedCommandName` is a string where as the type of the element in `commands` array is not (for example, I am doing `commands[i].Name()`), so I need to go through each element and do a `.Name()` on that command and compare. Are you asking if there is a C# equivalent in Golang for something like this? `commands.FirstOrDefault(cmd => cmd.Name() ==  suppliedCommandName)`...I checked online and seems like it does not have a built-in way to do this.

qianz2: Gotcha, thanks for explaining. The annotations are also helpful for us to understand it better.

PR #528, opened 2020-03-13
fix php-fpm images and corresponding tests

arroyc:


build/buildBuildImageBases.sh
```diff
@@ -52,13 +52,13 @@ case $IMAGE_DIR_TO_BUILD in
...
 		buildImages "python"
 		;;
-	'php')
-		echo "Building PHP base images"
-		echo
-
-		docker build -f $BUILD_IMAGES_DIR/php/prereqs/Dockerfile -t "php-build-prereqs" $REPO_DIR
-		buildImages "php"
-		;;            
+	#'php')
```
kichalla: Can you send out a separate clean up PR?

arroyc: ok .. I will remove from here

PR #529, opened 2020-03-13
run build from setup.py for python apps

arroyc:
This PR is to support following scenarios:
1. if there is a setup.py file in the root level, and there is no requirement.txt, oryx should be able to detect it as python app and build smoothly.
2. if there is setup.py as well as requirement.txt, build from requirement.txt.

Note: we can also run build from both the files if both are present. I would like to know what's your thought on that.

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -1,5 +1,5 @@
 declare -r TS_FMT='[%T%z] '
-declare -r REQS_NOT_FOUND_MSG='Could not find requirements.txt; Not running pip install'
+declare -r REQS_NOT_FOUND_MSG='Could not find setup.py or requirements.txt; Not running pip install'
 echo "Python Version: $python"
 PIP_CACHE_DIR=/usr/local/share/pip-cache
 
@@ -45,6 +45,18 @@ then
 	then
 		exit $pipInstallExitCode
 	fi
+elif [ -e "setup.py" ]
+then
+	echo
+	START_TIME=$SECONDS
+
+	echo "Running python setup.py install..."
+	python setup.py install | ts $TS_FMT
+	pythonBuildExitCode=${PIPESTATUS[0]}
+	if [[ $pythonBuildExitCode != 0 ]]
+	then
+		exit $pythonBuildExitCode
+	fi
 else
 	echo $REQS_NOT_FOUND_MSG
 fi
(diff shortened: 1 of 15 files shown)


src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -68,6 +86,23 @@ then
...
+then
+	echo
+	START_TIME=$SECONDS
+	pip install --upgrade pip
+	ELAPSED_TIME=$(($SECONDS - $START_TIME))
+	echo "Done in $ELAPSED_TIME sec(s)."
+
+	echo "Running python setup.py install..."
+	$python setup.py install --user| ts $TS_FMT
+	cd *.egg-info
```
arroyc: python<version> setup.py install build and installs dependent packages. It installs packages in the userbase (it's needed for build image as in anteras  build env we don't have write permission) and also creates a <projectname>.egg-info folder. Inside that egg-info folder we have the requires.txt. which we do install again using pip (like we do for our normally when we see a requirement.txt in root of the app).

arroyc:
python integration tests are successful .. some random node integration test failed/crashed

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
```diff
@@ -400,6 +400,90 @@ public void GeneratesScript_AndBuilds_WhenExplicitLanguageAndVersion_AreProvided
...
+                result.GetDebugInfo());
+        }
+
+        // This is to test if we can build an app when both the files requirement.txt
+        // and setup.py are provided, we tend to prioritize the root level requirement.txt
+        [Fact]
+        public void GeneratesScript_AndBuilds_With_Both_Files_areProvided()
+        {
+            // Arrange
+            var appName = "flask-app";
```
kichalla: The test is talking about 'both' files being provided but this app only has `requirements.txt` file and not `setup.py`. Did you want to use `flask-setup-py-requirement-txt` app instead?

tests/Oryx.Integration.Tests/Python/PythonEndToEndTests.cs
```diff
@@ -151,6 +151,44 @@ public async Task BuildWithVirtualEnv_RemovesOryxPackagesDir_FromOlderBuild(stri
                 });
         }
 
+        [Theory]
+        [InlineData("3.6")]
```
kichalla: Do we need `2.7` too?

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
arroyc: yes .. good catch … thanks !

PR #533, opened 2020-03-20
Make dynamic install controllable by an environment variable

kichalla:


src/BuildScriptGeneratorCli/Options/BuildScriptGeneratorOptionsSetup.cs
```diff
@@ -0,0 +1,26 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Oryx.BuildScriptGenerator;
+
+namespace Microsoft.Oryx.BuildScriptGeneratorCli.Options
+{
+    public class BuildScriptGeneratorOptionsSetup : IConfigureOptions<BuildScriptGeneratorOptions>
+    {
+        private readonly IEnvironment _environment;
+
+        public BuildScriptGeneratorOptionsSetup(IEnvironment environment)
```
kichalla: FYI: Ideal way is to get `IConfiguration` from DI and use it to get values of keys (for example: `configuration.GetValue("ENABLE_DYNAMIC_INSTALL")`) and then bind to different properties of this options. The advantage of this way is that `configuration.GetValue` call would get the hierarchically resolved value to you i.e command line value over environment variable over file settings

What I have here is a stop-gap solution to prevent this PR to get very big. I will later send out a different PR as it is going to have substantial changes.

src/BuildScriptGeneratorCli/Commands/BuildScriptGeneratorOptionsHelper.cs
```diff
@@ -53,8 +52,6 @@ internal static class BuildScriptGeneratorOptionsHelper
 
             options.ScriptOnly = scriptOnly;
 
-            options.EnableDynamicInstall = enableDynamicInstall;
```
qianz2: Since we removed it from command line options, we may need to remove it from entrypoint of appservice-build as well.

kichalla: Yes, right. Once we get a release build out we can update the tag and make the change to environment variable. Actually I feel we need this environment variable to be the default for the image `github-actions` since this image does not have any sdks at all and so does not make sense to have it optional.

kichalla: In the latest commit, I made installing dynamically 'true' by default for github-actions image

qianz2: Cool, that makes sense for the `github-actions`.

PR #534, opened 2020-03-20
Enable providing a custom NPM run build command

kichalla:


src/BuildScriptGeneratorCli/Options/NodeScriptGeneratorOptionsSetup.cs
```diff
@@ -0,0 +1,27 @@
...
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using Microsoft.Extensions.Options;
+using Microsoft.Oryx.BuildScriptGenerator;
+using Microsoft.Oryx.BuildScriptGenerator.Node;
+
+namespace Microsoft.Oryx.BuildScriptGeneratorCli.Options
+{
+    public class NodeScriptGeneratorOptionsSetup : IConfigureOptions<NodeScriptGeneratorOptions>
```
gpcastro: documentation?

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -90,18 +90,24 @@ echo "Running '{{ PackageInstallCommand }}'..."
 echo
 {{ PackageInstallCommand }}
 
-{{ if NpmRunBuildCommand | IsNotBlank }}
-echo
-echo "Running '{{ NpmRunBuildCommand }}'..."
-echo
-{{ NpmRunBuildCommand }}
-{{ end }}
+{{ if CustomNpmRunBuildCommand | IsNotBlank }}
```
arroyc: so we are assuming (as per design) we will ignore npm run build if custom npm run build command is enabled

kichalla: yes that's right

PR #535, opened 2020-03-23
Removed support for build.env

kichalla:


src/BuildScriptGenerator/BuildScriptGeneratorServiceCollectionExtensions.cs
```diff
@@ -31,7 +31,6 @@ public static IServiceCollection AddBuildScriptGeneratorServices(this IServiceCo
             services.AddSingleton<ISourceRepoProvider, DefaultSourceRepoProvider>();
             services.AddSingleton<ITempDirectoryProvider, DefaulTempDirectoryProvider>();
             services.AddSingleton<IScriptExecutor, DefaultScriptExecutor>();
-            services.AddSingleton<IEnvironmentSettingsProvider, DefaultEnvironmentSettingsProvider>();
```
kichalla: - This feature was implemented during our initial days and was not advertised.
- We left the source code in to use it later if needed.
- I am sending this PR now because I will later be sending a PR where we would be following `Microsoft.Extensions.Configuration` api to get hierarchical configuration from different 'sources' (command line -> environment variables -> files like `build.env` and in future maybe `app.yml`)

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreSampleAppsTest.cs
```diff
@@ -292,7 +292,7 @@ public void Builds_NetCore31App_UsingNetCore31_DotNetSdkVersion()
                 result.GetDebugInfo());
         }
 
-        [Fact]
+        [Fact(Skip = "1086977")]
```
kichalla: I added a new issue for re-enabling these tests once I have the support for getting settings using `IConfiguration` api

src/BuildScriptGenerator/Helpers/PreAndPostBuildCommandHelper.cs
```diff
@@ -4,42 +4,36 @@
...
 using System.IO;
+using Microsoft.Oryx.Common;
 
 namespace Microsoft.Oryx.BuildScriptGenerator
 {
     internal static class PreAndPostBuildCommandHelper
     {
-        public static (string preBuildCommand, string postBuildCommand) GetPreAndPostBuildCommands(
-            ISourceRepo sourceRepo,
-            EnvironmentSettings settings)
```
arroyc: nit: this seem to be a breaking change. do we still be supporting old settings or  from now on users will mention/keep build.env in their repo instead if having app_settings? how does this fit into azure function and the vs online and may be GitHub actions?

kichalla: Take a look at my other comments. They will give an idea of the changes here.

src/BuildScriptGenerator/BuildScriptGeneratorServiceCollectionExtensions.cs
qianz2: So do we also want the same hierarchical config for run-time command line and env variables as well in the future?

kichalla: yes, that's right.

src/BuildScriptGenerator/Helpers/PreAndPostBuildCommandHelper.cs
```diff
@@ -4,42 +4,36 @@
...
-            string postBuildCommand = null;
-
-            if (!string.IsNullOrEmpty(settings.PreBuildScriptPath))
-            {
-                preBuildCommand = $"\"{settings.PreBuildScriptPath}\"";
-            }
-            else if (!string.IsNullOrEmpty(settings.PreBuildCommand))
+            var preBuildCommandOrScriptPath = string.IsNullOrEmpty(preBuildScriptPath)
+                ? preBuildCmd : preBuildScriptPath;
+            if (!string.IsNullOrEmpty(preBuildCommandOrScriptPath))
```
qianz2: Looks good. It's much readable and clear.

qianz2: Probably waiting for a succeed nightly build.

PR #536, opened 2020-03-23
Fix how .NET Core and PHP read configuration settings

kichalla:


src/BuildScriptGeneratorCli/Options/PhpScriptGeneratorOptionsSetup.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    /// <summary>
+    /// Gets hierarchical configuration from IConfiguration api and
+    /// binds the properties on <see cref="PhpScriptGeneratorOptions"/>.
+    /// </summary>
+    public class PhpScriptGeneratorOptionsSetup : IConfigureOptions<PhpScriptGeneratorOptions>
+    {
+        private readonly IEnvironment _environment;
+
+        public PhpScriptGeneratorOptionsSetup(IEnvironment environment)
```
kichalla: I am sending this PR now because I will later be sending a PR where we would be following `Microsoft.Extensions.Configuration` api to get hierarchical configuration from different 'sources' (command line -> environment variables -> files like build.env and in future maybe app.yml). 
So instead of reading from `IEnvironment` directly here, we would be injecting `IConfiguration` and do something like `configuation["PHP_VERSION"]` where this `configuration` will take care of 'resolving' the final value based on the hierarchically set values.

src/BuildScriptGeneratorCli/Options/OptionsServiceCollectionExtensions.cs
```diff
@@ -19,6 +21,10 @@ public static IServiceCollection AddOptionsServices(this IServiceCollection serv
                 ServiceDescriptor.Transient<IConfigureOptions<BuildScriptGeneratorOptions>, BuildScriptGeneratorOptionsSetup>());
             services.TryAddEnumerable(
                 ServiceDescriptor.Transient<IConfigureOptions<NodeScriptGeneratorOptions>, NodeScriptGeneratorOptionsSetup>());
+            services.TryAddEnumerable(
+                ServiceDescriptor.Transient<IConfigureOptions<DotNetCoreScriptGeneratorOptions>, DotNetCoreScriptGeneratorOptionsSetup>());
+            services.TryAddEnumerable(
+                ServiceDescriptor.Transient<IConfigureOptions<PhpScriptGeneratorOptions>, PhpScriptGeneratorOptionsSetup>());
```
arroyc: nit: do we need to add python later? (not in this PR)

kichalla: Yes, I am adding it in another PR.

src/BuildScriptGenerator/Php/PhpScriptGeneratorServiceCollectionExtensions.cs
```diff
@@ -16,7 +15,6 @@ public static IServiceCollection AddPhpScriptGeneratorServices(this IServiceColl
         {
             services.TryAddEnumerable(ServiceDescriptor.Singleton<ILanguageDetector, PhpLanguageDetector>());
             services.TryAddEnumerable(ServiceDescriptor.Singleton<IProgrammingPlatform, PhpPlatform>());
-            services.TryAddEnumerable(ServiceDescriptor.Singleton<IConfigureOptions<PhpScriptGeneratorOptions>, PhpScriptGeneratorOptionsSetup>());
             services.AddSingleton<IPhpVersionProvider, PhpVersionProvider>();
             services.AddScoped<PhpLanguageDetector>();
```
arroyc: nit:this can be singleton too?

src/BuildScriptGeneratorCli/Options/PhpScriptGeneratorOptionsSetup.cs
arroyc: is that other PR needs to go first before this one? or it  does not matter?

tests/BuildScriptGenerator.Tests/Php/PhpLanguageDetectorTest.cs
```diff
@@ -95,15 +93,13 @@ public void Detect_ReturnsVersionFromCliSwitch_EvenIfEnvironmentVariable_AndComp
...
-            var environment = new TestEnvironment();
-            environment.Variables[PhpConstants.PhpRuntimeVersionEnvVarName] = "7.2.5";
             var detector = CreatePhpLanguageDetector(
                 supportedPhpVersions: new[] { "7.3.14", "7.2.5", "5.6.0" },
                 defaultVersion: "7.3.14",
-                environment);
+                new PhpScriptGeneratorOptions { PhpVersion = "7.2.5" });
             var repo = new MemorySourceRepo();
             var version = "5.6.0";
             repo.AddFile("{\"require\":{\"php\":\"" + version + "\"}}", PhpConstants.ComposerFileName);
```
arroyc: nit: maybe its just better to have a sample file instead of having all these string operations multiple times

src/BuildScriptGenerator/Php/PhpScriptGeneratorServiceCollectionExtensions.cs
kichalla: yes, thanks. I am sending a cleanup PR having fixes for this.

src/BuildScriptGeneratorCli/Options/PhpScriptGeneratorOptionsSetup.cs
kichalla: I am not too concerned about the order at this point since `build.env` has not been advertised yet and also my PR would be coming out soon.

tests/BuildScriptGenerator.Tests/Php/PhpLanguageDetectorTest.cs
kichalla: Not really related to this PR's change, so I am not going to fix this. I am resisting the temptation to do changes not related to this PR's intention :-)

PR #537, opened 2020-03-24
Add pre-run script 

qianz2:
Enabling to use a pre-run script to run command before the app is started.
Set the script path as an environment variable; Updated constants.yaml
For dotnet core platform first, then apply to other platforms.


build/constants.yaml
```diff
@@ -3,6 +3,7 @@
     environment-type: ORYX_ENV_TYPE # Currently exported only in the Visual Studio Online build image
     application-insights-instrumentation-key-env-var-name: ORYX_AI_INSTRUMENTATION_KEY
     app-service-app-name-env-var-name: APPSETTING_WEBSITE_SITE_NAME
+    app-service-pre-run-script-file-path-env-var-name: APPSETTING_PRE_RUN_SCRIPT_FILE_PATH
```
kichalla: Since we want this environment variable to be used by user to either provide command(s) or a script path, let's name it something like `PRE_RUN_COMMAND`. Also do not include the prefix `APPSETTING` since it is implicit. The setting `APPSETTING_WEBSITE_SITE_NAME` is from `AppService` actually which we use to log in some metrics.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -42,6 +42,16 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
 	}
 
 	dotnetBinary := "/usr/share/dotnet/dotnet"
+
+	preRunScriptPath := common.GetPathEnvironmentVariable(consts.AppServicePreRunScriptFilePathEnvVarName)
+	if preRunScriptPath != "" && !common.PathExists(dotnetBinary) {
```
kichalla: As mentioned above, this variable could be a list of commands or a relative path to a script file.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -42,6 +42,16 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
+		scriptBuilder.WriteString("echo 'Running the user provided pre-run command within pre-run script... ")
+		preRunScriptFileDir := filepath.Dir(preRunScriptPath)
+		scriptBuilder.WriteString("cd \"" + preRunScriptFileDir + "\"\n")
+		preRunCommand := "bash \"" + preRunScriptPath + "\""
```
kichalla: A user could specify shebang in their script file, so we do not need to explicit mention this?

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -42,6 +42,16 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
+		scriptBuilder.WriteString(preRunCommand + "\n")
```
kichalla: I think this new block should be added after line # 58 (i.e after we install the platform sdk). For example, a user might try to use some platform tools to do something in their pre-run commands.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -42,6 +42,16 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
+	}
+
```
kichalla: Can you add an integration test for this?

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
qianz2: Sounds good. I'm leaning to either a relative path to a script file or a single string command. A list of commands might be redundant here?

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,137 @@
...
+               .AddCommand(
+                $"oryx build {appDir} --platform dotnet --platform-version {runtimeVersion} -o {appOutputDir}")
+               .ToString();
+            var runtimeImageScript = new ShellScriptBuilder()
+                .SetEnvironmentVariable(SettingsKeys.EnableDynamicInstall, true.ToString())
+                .SetEnvironmentVariable(
+                    SdkStorageConstants.SdkStorageBaseUrlKeyName,
+                    SdkStorageConstants.DevSdkStorageBaseUrl)
+                .SetEnvironmentVariable(FilePaths.PreRunCommandOrScriptEnvVarName, "./prerunscript.sh")
+                .AddCommand(
```
qianz2: Wondering how we're going to test for prerunscript.sh file on this integration test? Since there doesn't exist one, where should we create one for testing?

build/constants.yaml
```diff
@@ -98,6 +98,7 @@
     node-global-modules-path: /usr/local/lib/node_modules
     bash: /bin/bash
     benv: /opt/oryx/benv
+    pre-run-command-or-script-env-var-name: PRE_RUN_COMMAND_OR_SCRIPT
```
kichalla: just make it `PRE_RUN_COMMAND` which would be inline with `PRE_BUILD_COMMAND`.

src/startupscriptgenerator/src/common/environmentVariablesHelper.go
```diff
@@ -23,3 +23,11 @@ func GetBooleanEnvironmentVariable(key string) bool {
 	}
 	return result
 }
+
+func GetEnvironmentVariable(key string) string {
+	value, ok := os.LookupEnv(key)
+	if !ok {
+		panic(fmt.Sprintf("'%s' not set\n", key))
```
kichalla: Setting the `PRE_RUN_COMMAND` is optional, so we should not fail running the app if it is not present.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -42,11 +42,25 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
 
 	dotnetBinary := "/usr/share/dotnet/dotnet"
+
 	enableDynamicInstall := common.GetBooleanEnvironmentVariable(consts.EnableDynamicInstallKey)
 	if enableDynamicInstall && !common.PathExists(dotnetBinary) {
 		scriptBuilder.WriteString(fmt.Sprintf("oryx setupEnv -appPath %s\n", appPath))
 	}
 
+	preRunCommandOrScript := common.GetEnvironmentVariable(consts.PreRunCommandOrScriptEnvVarName)
+	if preRunCommandOrScript != "" && !common.PathExists(dotnetBinary) {
```
kichalla: Why is the condition `!common.PathExists(dotnetBinary)` required here?

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -42,11 +42,25 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
+		if common.FileExists(preRunCommandOrScript) {
```
kichalla: Possible values for `preRunCommandOrScript`: 
- `,/installPreqs.sh`
- `apt-get install foo`
- `./installPreqs.sh && apt-get install foo`

To make it less complicated for ourselves, we can avoid parsing the content and checking if a file exists or not. We can tell users that the runtime script is always run from root of the app directory. Let's make sure that we run from the app directory.

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,137 @@
...
+namespace Microsoft.Oryx.Integration.Tests
+{
+    [Trait("category", "dotnetcore")]
+    public class DotNetCorePreRunCommandOrScriptTest : DotNetCoreEndToEndTestsBase
+    {
+        public DotNetCorePreRunCommandOrScriptTest(ITestOutputHelper output, TestTempDirTestFixture testTempDirTestFixture)
+            : base(output, testTempDirTestFixture)
+        {
+        }
+        [Theory]
```
kichalla: nit: space before this line

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,137 @@
+// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using Microsoft.Oryx.BuildScriptGeneratorCli;
+using Microsoft.Oryx.Common;
+using Microsoft.Oryx.Tests.Common;
+using System.IO;
```
kichalla: nit: System directives before all other directives
image

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,137 @@
...
+                _output,
+                new[] { volume },
+                _imageHelper.GetGitHubActionsBuildImage(),
+                "/bin/sh",
+                new[]
+                {
+                    "-c",
+                    buildImageScript
+                },
+                _imageHelper.GetTestRuntimeImage("dotnetcore", "prerun"),
```
kichalla: There is no image with `prerun` tag? It should be `dynamic` here?

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,137 @@
...
+                    SdkStorageConstants.DevSdkStorageBaseUrl)
+               .AddCommand(
+                $"oryx build {appDir} --platform dotnet --platform-version {runtimeVersion} -o {appOutputDir}")
+               .ToString();
+            var runtimeImageScript = new ShellScriptBuilder()
+                .SetEnvironmentVariable(SettingsKeys.EnableDynamicInstall, true.ToString())
+                .SetEnvironmentVariable(
+                    SdkStorageConstants.SdkStorageBaseUrlKeyName,
+                    SdkStorageConstants.DevSdkStorageBaseUrl)
+                .SetEnvironmentVariable(FilePaths.PreRunCommandOrScriptEnvVarName, "echo \"Here has a pre run command.\"")
```
kichalla: How are you testing that this command ran successfully? One way to achieve it here is to create a file in the `appOutputDir` and assert this file's existence at line 77

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,137 @@
...
+    public class DotNetCorePreRunCommandOrScriptTest : DotNetCoreEndToEndTestsBase
+    {
+        public DotNetCorePreRunCommandOrScriptTest(ITestOutputHelper output, TestTempDirTestFixture testTempDirTestFixture)
+            : base(output, testTempDirTestFixture)
+        {
+        }
+        [Theory]
+        [InlineData("2.1", NetCoreApp21WebApp, "Hello World!")]
+        [InlineData("3.1", NetCoreApp31MvcApp, "Welcome to ASP.NET Core MVC!")]
+        public async Task CanBuildAndRun_NetCore31WebApp_UsingPreRunCommand_WithDynamicInstalls(
```
kichalla: `DynamicInstalls` => `DynamicInstall`

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
kichalla: Just before this line you can add a command which creates this file called `prerunscript.sh` in the `appOutputDir`. Also make sure you put some content in the `prerunscript.sh` in such a way that you can Assert that indeed this script ran

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,137 @@
...
+                _output,
+                new[] { volume },
+                _imageHelper.GetGitHubActionsBuildImage(),
+                "/bin/sh",
+                new[]
+                {
+                    "-c",
+                    buildImageScript
+                },
+                _imageHelper.GetTestRuntimeImage("dotnetcore", "prerun"),
```
kichalla: same here...no `prerun` image, but `dynamic` needs to be used.

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,137 @@
...
+                },
+                async (hostPort) =>
+                {
+                    var data = await _httpClient.GetStringAsync($"http://localhost:{hostPort}/");
+                    Assert.Contains(expectedResponseContent, data);
+                });
+        }
+
+        [Theory]
+        [InlineData("2.1", NetCoreApp21WebApp, "Hello World!")]
```
kichalla: For both these tests, let us target to only 1 version. Seems like an overkill to run for both versions given the pivot (scenario) we are testing is not dependent on the version as such.

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,138 @@
...
+               .AddCommand(
+                $"oryx build {appDir} --platform dotnet --platform-version {runtimeVersion} -o {appOutputDir}")
+               .ToString();
+            var runtimeImageScript = new ShellScriptBuilder()
+                .SetEnvironmentVariable(SettingsKeys.EnableDynamicInstall, true.ToString())
+                .SetEnvironmentVariable(
+                    SdkStorageConstants.SdkStorageBaseUrlKeyName,
+                    SdkStorageConstants.DevSdkStorageBaseUrl)
+                .SetEnvironmentVariable(FilePaths.PreRunCommandEnvVarName, $"touch \"{appOutputDir}/test_pre_run.txt\"")
+                .AddCommand(
```
qianz2: use `touch` command to create a txt test file, then test for its existence.

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,138 @@
...
+               .ToString();
+            var runtimeImageScript = new ShellScriptBuilder()
+                .SetEnvironmentVariable(SettingsKeys.EnableDynamicInstall, true.ToString())
+                .SetEnvironmentVariable(
+                    SdkStorageConstants.SdkStorageBaseUrlKeyName,
+                    SdkStorageConstants.DevSdkStorageBaseUrl)
+                .SetEnvironmentVariable(FilePaths.PreRunCommandEnvVarName, "./prerunscript.sh && apt-get install foo")
+                .AddCommand($"touch {appOutputDir}/prerunscript.sh")
+                .AddCommand($"echo \"export PRE_RUN_TEST_SUCCESS=true\" > {appOutputDir}/prerunscript.sh")
+                .AddCommand($"chmod 755 {appOutputDir}/prerunscript.sh")
```
qianz2: Set a env inside script, then see if this env exists, if so then means it ran successfully?

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,138 @@
...
+                async (hostPort) =>
+                {
+                    var data = await _httpClient.GetStringAsync($"http://localhost:{hostPort}/");
+                    Assert.Contains("Welcome to ASP.NET Core MVC!", data);
+                    Assert.IsTrue(File.Exists($"{appOutputDir}/prerunscript.sh"));
+                    Assert.Equals(Environment.GetEnvironmentVariable("PRE_RUN_TEST_SUCCESS"), "true");
+                });
+        }
+    }
+}
```
kichalla: We need tests for other platforms too.

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,138 @@
...
+                {
+                    "-c",
+                    runtimeImageScript
+                },
+                async (hostPort) =>
+                {
+                    var data = await _httpClient.GetStringAsync($"http://localhost:{hostPort}/");
+                    Assert.Contains("Welcome to ASP.NET Core MVC!", data);
+                    Assert.IsTrue(File.Exists($"{appOutputDir}/prerunscript.sh"));
+                    Assert.Equals(Environment.GetEnvironmentVariable("PRE_RUN_TEST_SUCCESS"), "true");
```
kichalla: I am not sure how this will work. Does this work at all? Take a look at my earlier comment to see how to do it.

src/startupscriptgenerator/src/common/environmentVariablesHelper.go
```diff
@@ -23,3 +23,8 @@ func GetBooleanEnvironmentVariable(key string) bool {
 	}
 	return result
 }
+
+func GetEnvironmentVariable(key string) string {
```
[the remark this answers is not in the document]
qianz2: I feel it's a good place to put env related helper function for future use. But it seems not much coding involved so I'll delete it and just use `os.Getenv`.

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,138 @@
...
+            : base(output, testTempDirTestFixture)
+        {
+        }
+
+        [Fact]
+        public async Task CanBuildAndRun_NetCore21WebApp_UsingPreRunCommand_WithDynamicInstall()
+        {
+            // Arrange
+            var runtimeVersion = "2.1";
+            var appName = "NetCoreApp30WebApp";
```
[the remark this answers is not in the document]
qianz2: It's my misunderstanding. It'll be revised.

src/startupscriptgenerator/src/common/preRunScriptHelper.go
```diff
@@ -0,0 +1,20 @@
...
+package common
+
+import (
+	"strings"
+	"common/consts"
+)
+
+// Run pre-run script from root of the app directory
+func SetupPreRunScript(scriptBuilder *strings.Builder) {
+	preRunCommandOrScript := GetEnvironmentVariable(consts.PreRunCommandEnvVarName)
```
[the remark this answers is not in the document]
qianz2: If there has quotes within env's value, then it need to have escape `\` before it. For example: `export TEST_QUOTE="Hello "World""`, we get `Hello World`. And  `export TEST_QUOTE="Hello \"world\""` we get `Hello "world"`.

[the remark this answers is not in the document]
qianz2: So that means users need to add `\` before every quotes themselves for commands to work?

qianz2:
> Is this PR ready to review for all?

Sorry for the confusion, it's still in progress.

src/startupscriptgenerator/src/common/preRunScriptHelper.go
```diff
@@ -0,0 +1,22 @@
...
+	"strings"
+	"common/consts"
+)
+
+// Run pre-run script from root of the app directory
+
+func SetupPreRunScript(scriptBuilder *strings.Builder)  {
+	preRunCommandOrScript := os.Getenv(consts.PreRunCommandEnvVarName)
+	if preRunCommandOrScript != "" {
+		scriptBuilder.WriteString("echo 'Running the provided pre-run command...'")
```
kichalla: You need a `\n` here

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -42,11 +42,14 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
 	}
 
 	dotnetBinary := "/usr/share/dotnet/dotnet"
+
 	enableDynamicInstall := common.GetBooleanEnvironmentVariable(consts.EnableDynamicInstallKey)
 	if enableDynamicInstall && !common.PathExists(dotnetBinary) {
 		scriptBuilder.WriteString(fmt.Sprintf("oryx setupEnv -appPath %s\n", appPath))
 	}
 
+	common.SetupPreRunScript(scriptBuilder)
```
kichalla: `&scriptBuilder`

tests/Oryx.Integration.Tests/Php/PhpPreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,95 @@
...
+            var appName = "twig-example";
+            var hostDir = Path.Combine(_hostSamplesDir, "php", appName);
+            var volume = DockerVolume.CreateMirror(hostDir);
+            var appDir = volume.ContainerDir;
+            var appOutputDir = $"{appDir}/myoutputdir"; 
+            var buildScript = new ShellScriptBuilder()
+               .AddCommand($"oryx build {appDir} --platform php --language-version {phpVersion} -o {appOutputDir}")
+               .ToString();
+            var runScript = new ShellScriptBuilder()
+                .SetEnvironmentVariable(FilePaths.PreRunCommandEnvVarName, "./prerunscript.sh")
```
kichalla: This is fine for now, but as @gpcastro was mentioning, probably installing a PHP extension that we do not include as part of the image would be good scenario to test.

tests/Oryx.Integration.Tests/Python/PythonPreRunScriptTest.cs
```diff
@@ -0,0 +1,118 @@
...
+                .AddCommand($"oryx create-script -appPath {appOutputDir} -bindPort {ContainerPort}")
+                .AddFileExistsCheck($"{appOutputDir}/test_pre_run.txt")
+                .AddCommand(DefaultStartupFilePath)
+                .ToString();
+
+            await EndToEndTestHelper.BuildRunAndAssertAppAsync(
+                appName,
+                _output,
+                new[] { volume },
+                _imageHelper.GetTestSlimBuildImage(),
```
kichalla: use the githubactions image here and in the other test

tests/Oryx.Integration.Tests/Node/NodePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,138 @@
...
+                .AddCommand($"oryx create-script -appPath {appOutputDir} -bindPort {ContainerPort}")
+                .AddFileExistsCheck($"{appOutputDir}/test_pre_run.txt")
+                .AddCommand(DefaultStartupFilePath)
+                .ToString();
+
+            await EndToEndTestHelper.BuildRunAndAssertAppAsync(
+                appName,
+                _output,
+                new[] { volume },
+                _imageHelper.GetTestSlimBuildImage(),
```
kichalla: use githubactions image here and in the next test

kichalla:
Can you kick off a Nighly build too?

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
qianz2: The scriptBuilder within dotnet script generator is already a pointer type.

kichalla: Gotcha thanks

qianz2:
> Make sure nightly build passes before merging this PR

Thanks, will make a nightly build for it.

tests/Oryx.Integration.Tests/Node/NodePreRunCommandOrScriptTest.cs
```diff
@@ -0,0 +1,140 @@
...
+                .ToString();
+            var runScript = new ShellScriptBuilder()
+                .SetEnvironmentVariable(
+                    SdkStorageConstants.SdkStorageBaseUrlKeyName,
+                    SdkStorageConstants.DevSdkStorageBaseUrl)
+                .SetEnvironmentVariable(SettingsKeys.EnableDynamicInstall, true.ToString())
+                .SetEnvironmentVariable(FilePaths.PreRunCommandEnvVarName, "./prerunscript.sh")
+                .AddCommand($"touch {appOutputDir}/prerunscript.sh")
+                .AddFileExistsCheck($"{appOutputDir}/prerunscript.sh")
+                .AddCommand($"echo \"touch test_pre_run.txt\" > {appOutputDir}/prerunscript.sh")
```
arroyc, line 106: can we test a script which is bit more complicated or rather which matches more with a regular bash/shell script? this is more like running a command rather than script to me.

src/startupscriptgenerator/src/common/preRunScriptHelper.go
```diff
@@ -0,0 +1,22 @@
...
+	"common/consts"
+)
+
+// Run pre-run script from root of the app directory
+
+func SetupPreRunScript(scriptBuilder *strings.Builder)  {
+	preRunCommandOrScript := os.Getenv(consts.PreRunCommandEnvVarName)
+	if preRunCommandOrScript != "" {
+		scriptBuilder.WriteString("echo 'Running the provided pre-run command...'\n")
+		scriptBuilder.WriteString(preRunCommandOrScript + "\n")
```
arroyc, line 20: how do we know the path of the script? do we assume absolute path to be provided or do we support relative path as well?

kichalla: It could be error prone for us to parse the contents of this environment variable since user could provide a command, list of commands, mix commands and file paths etc. In my opinion we should add documentation suggesting users to have relative paths to scripts.

arroyc: agreed …. we should have doc related to that .. but at the basic level we could check as well whether the content of the env variable is a file or not a file. what do you think? @kichalla

kichalla: You mean if somebody just has `./prerun-script.sh` or `scripts/prerun-script.sh`, we could resolve them to full paths and run? Sure, we could try doing that but I do not see much value in it to be honest. If we can guarantee and document it that the user that the app's run context is always going to be root of the directory, then it should just work. Are there any other cases?

arroyc: I think, from customer point of view it doesn't add value much. But from tooling side if we know what user was trying to run (it's a command or it's a script file) it will be easier for us to troubleshoot seeing the log. End to end scenario wise I agree there is not much value, but adding a small check won't hurt us.

kichalla: > So if we check the file and error it out at-least user will know from basic context.

Sure I agree this would help the user.

qianz2: @arroyc for now I'm leaning to assume the path to be relative and provide a documentation for users, that'll save us some checks. And yes I'll add a check to see if the environment variable is a file or commands, good for users to see some context.

qianz2: An user case could be like `./prerun-script.sh && touch sample.txt`, we could not simply check if the environment variable is a path or not, the combination may also exist.

arroyc: should we let user to do that? we also have userstartup command to run commands. 
> Enabling to use a pre-run script to run command before the app is started.
Set the script path as an environment variable

As per the description and the test you had, I thought it will always be script path, if t is not then we have a bit different story here. But the use case you mentioned can you please add a relevant test.

kichalla: FYI: this environment variable can allow a command, list of commands, a script path, mix of these etc.

PR #538, opened 2020-03-25
Make CDN endpoint the default for Prod scenarios

kichalla:


build/__sdkStorageConstants.sh
```diff
@@ -3,7 +3,7 @@
 ENABLE_DYNAMIC_INSTALL_KEY='ENABLE_DYNAMIC_INSTALL'
 SDK_STORAGE_BASE_URL_KEY_NAME='ORYX_SDK_STORAGE_BASE_URL'
 DEV_SDK_STORAGE_BASE_URL='https://oryxsdksdev.blob.core.windows.net'
-PROD_SDK_STORAGE_BASE_URL='https://oryxsdks.blob.core.windows.net'
+PROD_SDK_STORAGE_BASE_URL='https://oryxsdk-cdn.azureedge.net'
```
qianz2: Do we need the blob storage url as the backup base url?

kichalla: Since Azure CDN promises availability and reliability, I do not want to optimize for a scenario that might not happen. We can revisit this later based on how things go on.

PR #540, opened 2020-03-25
Fixed services: language detector registrations removed and some cleanup

kichalla:
cc @arroyc


src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
```diff
@@ -63,7 +63,6 @@ public string GenerateDockerfile(DockerfileContext ctx)
                 }
 
                 runImage = ConvertToRuntimeName(platform.Name);
-
```
kichalla: Fixed some code analysis warnings along the way

src/BuildScriptGenerator/Php/PhpLanguageDetector.cs
```diff
@@ -12,7 +12,7 @@
 
 namespace Microsoft.Oryx.BuildScriptGenerator.Php
 {
-    internal class PhpLanguageDetector : ILanguageDetector
+    internal class PhpLanguageDetector : IPlatformDetector
```
qianz2: Maybe change its name to  `PhpPlatformDetector.cs` as well since here we renamed all `Language` to `Platform`. It'll look better if all conforms. Same as other platforms: `NodePlatformDetector.cs`, `PythonPlatformDetector.cs` and `DotnetCorePlatformDetector.cs`.

qianz2: And it will match well with PhpPlatformInstaller.

tests/Oryx.Tests.Common/TestLanguageDetector.cs
```diff
@@ -7,36 +7,36 @@
...
         private readonly string _languageVersion;
 
-        public TestLanguageDetectorUsingLangName(string detectedLanguageName, string detectedLanguageVersion)
+        public TestPlatformDetectorUsingLangName(string detectedPlatformName, string detectedPlatformVersion)
         {
-            _languageName = detectedLanguageName;
-            _languageVersion = detectedLanguageVersion;
+            _languageName = detectedPlatformName;
+            _languageVersion = detectedPlatformVersion;
         }
```
qianz2: The naming of this file doesn't match with its class name within. Change its class name to `TestLanguageDetector` or change file name to `TestLanguageDetectorUsingLangName.cs`.

qianz2: I prefer to change the class name to `TestLanguageDetector`, but it looks like lots of places need to change.

src/BuildScriptGenerator/Php/PhpLanguageDetector.cs
kichalla: Good catch. Will do

tests/Oryx.Tests.Common/TestLanguageDetector.cs
kichalla: We chose to use 'platform' because the word 'language' does not suit everything...for example `python` is a language, but `dotnet` is not but C# is. Sure I will change the file names accordingly.

arroyc:
> cc @arroyc

@kichalla is it possible to limit the size of this PR? :)

kichalla:
> @kichalla is it possible to limit the size of this PR? :)

This is mostly renames and stuff. I do not want to split this into multiple PRs :-). However, good point, I kicked off a nightly build now.

PR #542, opened 2020-03-26
add an autoreload sample app integration test for php

arroyc:
Adding a sample app for php that use autoreload …. add couple of integration test as well
sample app source : https://github.com/frob/greetings
image

### DIFF
--- a/tests/Oryx.Integration.Tests/Php/PhpGreetingsAppTest.cs
+++ b/tests/Oryx.Integration.Tests/Php/PhpGreetingsAppTest.cs
@@ -1,0 +1,61 @@
+﻿// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.Oryx.Common;
+using Microsoft.Oryx.Tests.Common;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Oryx.Integration.Tests
+{
+    [Trait("category", "php")]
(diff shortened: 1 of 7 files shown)


tests/SampleApps/php/greetings/index.php
```diff
@@ -0,0 +1,9 @@
+<?php
```
arroyc: @kichalla  … I added this file for our integration test … had to adjust the require_once path require_once __DIR__ . '/vendor/autoload.php';

tests/SampleApps/php/greetings/composer.json
```diff
@@ -0,0 +1,26 @@
...
+        }
+    ],
+    "minimum-stability": "dev",
+    "require": {
+        "php": ">=5.3.0",
+        "phpunit/php-timer": "^1.0"
+    },
+    "autoload": {
+        "psr-0": {
+            "HelloWorld": "src//"
```
arroyc: changed this from original app …  in original app it was 
`
"autoload": {

        "psr-0": {

            "HelloWorld": "src/"

        },

        "classname": {

          "PHP_Timer": "src/"

        }

    },
`

tests/Oryx.Integration.Tests/Php/PhpGreetingsAppTest.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        private const string ExifImageTypePng = "3";
+
+        public PhpGreetingsAppTest(ITestOutputHelper output, TestTempDirTestFixture fixture)
+            : base(output, fixture)
+        {
+        }
+
+        [Theory]
+        [InlineData("7.4")]
```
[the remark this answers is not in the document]
arroyc: I would keep this .. total time spent for all these set is 3 min. So it's ok I guess

PR #548, opened 2020-04-02
Provide option to make 'build' step mandatory when building node apps

kichalla:


src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -28,6 +28,10 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
         PruneDevDependenciesPropertyKey,
         "When using different intermediate and output folders, only the prod dependencies are copied to the output. " +
         "Options are 'true', blank (same meaning as 'true'), and 'false'. Default is false.")]
+    [BuildProperty(
+        RequireBuildPropertyKey,
+        "Requires either 'npm run build' or 'yarn run build' or custom run build command  to be run. Default is false. " +
+        "If value is not provided, it is assumed to be 'true'.")]
```
arroyc, line 34: nit: im bit confused with this sentence:  "If value is not provided, it is assumed to be 'true'.", shouldn't it be the other way round?

tests/BuildScriptGenerator.Tests/Node/NodePlatformTest.cs
```diff
@@ -160,6 +161,122 @@ public void BuildScript_DoesNotHaveSdkInstallScript_IfDynamicInstallNotEnabled_A
...
         }
 
+        [Fact]
+        public void GeneratedBuildSnippet_DoesNotThrowException_IfPackageJsonHasBuildNode_AndRequireBuildPropertyIsSet()
+        {
+            // Arrange
+            const string packageJson = @"{
+              ""main"": ""server.js"",
+              ""scripts"": {
+                ""build"": ""build-node"",
```
arroyc, line 171: nit: it may sound crazy, but your package.json is simple string. can we send build and build:azure as parameter to this test and make it [theory] instead of two different [fact] uts

tests/BuildScriptGenerator.Tests/Node/NodePlatformTest.cs
```diff
@@ -160,6 +161,122 @@ public void BuildScript_DoesNotHaveSdkInstallScript_IfDynamicInstallNotEnabled_A
...
+            var commonOptions = new BuildScriptGeneratorOptions();
+            var nodePlatform = CreateNodePlatform(
+                commonOptions,
+                new NodeScriptGeneratorOptions(),
+                new NodePlatformInstaller(Options.Create(commonOptions), new TestEnvironment()));
+            var repo = new MemorySourceRepo();
+            repo.AddFile(packageJson, NodeConstants.PackageJsonFileName);
+            var context = CreateContext(repo);
+            context.NodeVersion = "10.10";
+            context.Properties[NodePlatform.RequireBuildPropertyKey] = "true";
```
arroyc, line 268: nit: can we add  some negative tests as well

e.g.
context.Properties[NodePlatform.RequireBuildPropertyKey] = ""
context.Properties[NodePlatform.RequireBuildPropertyKey]="false"

tests/Oryx.BuildImage.Tests/Node/NodeJsSampleAppsOtherTests.cs
```diff
@@ -836,7 +836,7 @@ public void BuildsApp_ByRunningNpmInstall_AndCustomBuildCommand()
             var subDir = Guid.NewGuid();
             var script = new ShellScriptBuilder()
                 .SetEnvironmentVariable(
-                SettingsKeys.CustomNpmRunBuildCommand,
+                SettingsKeys.CustomRunBuildCommand,
                 $"echo > /tmp/foo.txt")
```
arroyc, line 840: is this the only buildimage test where we are testing this feature?

src/BuildScriptGenerator/Node/NodePlatform.cs
kichalla: This meant the scenario: `oryx build /app -p require_build` or `oryx build /app -p require_build=true` are the same. In the first case requiring the build is already part of the name so it is unnecessary to say `true` again. We do the same for prune dev dependencies property.

tests/BuildScriptGenerator.Tests/Node/NodePlatformTest.cs
kichalla: Yeah I had considered it initially but I created separate cases because they are 2 different pivots and also since these are unit tests, we also need not optimize for performance over pivot separation.

tests/BuildScriptGenerator.Tests/Node/NodePlatformTest.cs
kichalla: Sure, sounds good.

tests/Oryx.BuildImage.Tests/Node/NodeJsSampleAppsOtherTests.cs
kichalla: yes, `NodePlatformTest` unit test has good amount of tests for this so I added only test for this. Given how long it takes our builds to run, I feel we should try to write as many unit tests as possible which cover the scenarios. This way tests run fast and we can add lot more variations there and add few integration tests.

kichalla:
gentle ping

PR #550, opened 2020-04-03
add wordpress integration tests for php images

arroyc:
php integration tests for WordPress sample apps

### DIFF
--- a/tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
+++ b/tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
@@ -68,5 +68,128 @@ namespace Microsoft.Oryx.Integration.Tests
                     Assert.Contains("<title>WordPress &rsaquo; Setup Configuration File</title>", data);
                 });
         }
+
+        [Theory]
+        [InlineData("7.4")]
+        [InlineData("7.3")]
+        [InlineData("7.2")]
+        [InlineData("7.0")]
+        [InlineData("5.6")]
+        public async Task CanBuildAndRun_Wordpress_SampleApp(string phpVersion)
+        {
+            // Arrange
+            var appName = "wordpress-example";
+            var hostDir = Path.Combine(_hostSamplesDir, "php", appName);
+            var volume = DockerVolume.CreateMirror(hostDir);
+            var appDir = volume.ContainerDir;
+
+            // build-script to download wordpress cli and build
+            var buildScript = new ShellScriptBuilder()
+                .AddCommand($"cd {appDir}")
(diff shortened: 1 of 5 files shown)


tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
```diff
@@ -68,5 +68,128 @@ public async Task WordPress51(string phpVersion)
...
+                    Assert.Contains("Remember Me", data);
+                    Assert.Contains("Lost your password?", data);
+                    Assert.Contains("Back to localsite", data);
+                });
+        }
+
+        [Theory]
+        [InlineData("7.4")]
+        [InlineData("7.3")]
+        [InlineData("7.2")]
```
qianz2, line 136: Maybe add 7.0, 5.6 version as well?

tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
```diff
@@ -68,5 +68,128 @@ public async Task WordPress51(string phpVersion)
...
+                .AddCommand($"cd {appDir}")
+                .AddFileExistsCheck("oryx-manifest.toml")
+                .AddFileExistsCheck("wp-cli.phar")
+                .AddFileExistsCheck("create_wordpress_db.sh")
+                .AddFileExistsCheck("configure_wordpress.sh")
+                .AddCommand($"chmod +x create_wordpress_db.sh && ./create_wordpress_db.sh")
+                .AddCommand($"chmod +x configure_wordpress.sh && ./configure_wordpress.sh")
+                .AddCommand($"oryx create-script -appPath {appDir} -bindPort {ContainerPort} -output {RunScriptPath}")
+                .AddCommand("mkdir -p /home/site/wwwroot")
+                .AddCommand($"cp -a {appDir}/. /home/site/wwwroot")
```
qianz2, line 168: will `cp -rf` be better instead of `cp -a`?

arroyc: I tried with cp -rf first and it was failing so trying with cp -a, as per doc cp -a seems to be a better version of cp -rf

tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
arroyc: we dont support php fpm images of 7.0 and 5.6. those are only supported in regular php images that has apache

tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
qianz2: gotcha, thanks

kichalla:
Make sure the Nightly passes before checkin

PR #551, opened 2020-04-03
add integration tests for php-fpm images

arroyc:


tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
```diff
@@ -68,5 +68,52 @@ public async Task WordPress51(string phpVersion)
...
+
+        [Theory]
+        [InlineData("7.4")]
+        [InlineData("7.3")]
+        [InlineData("7.2")]
+        public async Task PhpFpmWordPress51(string phpVersion)
+        {
+            // Arrange
+            string hostDir = Path.Combine(_tempRootDir, "wordpress");
+            if (!Directory.Exists(hostDir))
```
kichalla, line 80: There are 2 problems with this and earlier test in this file.
- for all the iterations in **this** test, the same folder is used to build the app. We ideally want unique folder for each variation so that earlier variation's build does not affect the next iteration's build
- This test's folder and earlier test's folder in this file are the same, so we would be having similar issues as mentioned in the above point.

tests/Oryx.Integration.Tests/Php/PhpTwigExampleTest.cs
```diff
@@ -55,5 +55,43 @@ public async Task TwigExample(string phpVersion)
                     Assert.Contains("<h1>Hello World!</h1>", data);
                 });
         }
+
+        [Theory]
+        [InlineData("7.4")]
+        [InlineData("7.3")]
+        [InlineData("7.2")]
```
kichalla, line 62: Curious what is the semantic versioning policy in PHP world? Do minor version upgrades have breaking changes as in Python or they are just feature additions?

tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
arroyc: I will take a look into this, these tests were originally added in regular php-images and I added tests for fpm images in addition.

tests/Oryx.Integration.Tests/Php/PhpTwigExampleTest.cs
arroyc: minor version can cause lot of pains (personal experience), php 7.4 is very different than php 7.3.

arroyc:
not sure why these tests failed here .. can't repro locally 
image

qianz2:
> not sure why these tests failed here .. can't repro locally
Doesn't look like an issue from PR itself, it's like some outdated AntiMalware from pipeline. Maybe rerun the build again?

arroyc:
> not sure why these tests failed here .. can't repro locally
> Doesn't look like an issue from PR itself, it's like some outdated AntiMalware from pipeline. Maybe rerun the build again?

seems like it randomly fails on downloading the cli, I will try keeping the cli as part of the sample app and see if we can minimize this inconsistency

PR #552, opened 2020-04-06
Resolve sdk version based on global.json spec

kichalla:


src/BuildScriptGenerator/DotNetCore/GlobalJsonSdkResolver.cs
```diff
@@ -0,0 +1,371 @@
...
+                allowPrerelease value, the highest installed SDK version is used (equivalent to setting rollForward
+                to latestMajor).
+            */
+            if (globalJson?.Sdk == null)
+            {
+                globalJson = new GlobalJsonModel();
+                globalJson.Sdk = new Sdk
+                {
+                    Version = "0.0.000",
+                    RollForward = LatestMajor,
```
kichalla, line 47: As per the spec, if there is no global.json file, then it should be considered as though there is one and that the roll forward policy is set to latest major, and the spec also says that if no setting is provided by the user for `allowPrerelease` flag, then it is assumed to be `true` by default, so that means prerelease versions must also be considered. 

This is going to be a problem for *our* case because as an end user you might have developed your 2.1/3.1 application locally using a 2.1/3.1 specific non-preview version sdks and it is also very common for people to *not* have global.json files. So based on the global.json spec, we should be considering latest major(including previews) for this app, which means we might try to build a 2.1/3.1 app using a `5.0.100-preview` sdk if that is the most available version in our storage. This might break existing users.

We could follow the following approaches:
1. Ask users to provide a global.json file
2. If there is no global.json file provided by the user, we will try getting the most recent *non-preview* version for that major version. i.e if the runtime version is `3.1`, then we will try getting an sdk version whose major is `3` but rest of the parts like minor, feature and patch would be the latest available one.

I prefer option `2.` above.

src/BuildScriptGenerator/DotNetCore/GlobalJsonSdkResolver.cs
```diff
@@ -0,0 +1,371 @@
...
+                };
```
qianz2, line 48: I'm confused about how this Sdk object created because I didn't see a constructor within the Sdk class above..

src/BuildScriptGenerator/DotNetCore/SdkVersionInfo.cs
```diff
@@ -0,0 +1,139 @@
...
+            var index = sdkVersion.IndexOf("-preview", StringComparison.OrdinalIgnoreCase);
+            if (index >= 0)
+            {
+                isPrerelease = true;
+                previewVersion = sdkVersion.Substring(index + 1, sdkVersion.Length - (index + 1));
+                sdkVersion = sdkVersion.Substring(0, index);
+            }
+
+            // Example: 3.1.102
+            var sdkVersionParts = sdkVersion.Split(".");
```
qianz2, line 52: Any other way to parse string replacing Substring, indexOf and Split method?

kichalla:
> Is it necessary to create some tests for this Sdk resolver and see if it'll successfully cover all cases?

This PR already has tests. I think GitHub is hiding the content. Can you try looking again?

src/BuildScriptGenerator/DotNetCore/SdkVersionInfo.cs
kichalla: In general it is discouraged to use regex, so avoiding it here.

kichalla:
gentle ping

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1217 @@
...
+        {
+            // Arrange
+            var previewVersionInfo = SdkVersionInfo.Parse(previewVersion);
+            var nonPreviewVersionInfo = SdkVersionInfo.Parse(nonPreviewVersion);
+
+            // Act
+            var actual = previewVersionInfo.CompareTo(nonPreviewVersionInfo);
+
+            // Assert
+            Assert.Equal(-1, actual);
```
qianz2, line 97: Let's keep `-1` as input parameter named `expected` as below.

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1217 @@
...
+
+            // Act
+            var actual = globalJsonHelper.GetSatisfyingSdkVersion(globalJson, availableVersions);
+
+            // Assert
+            Assert.Null(actual);
+        }
+
+        [Fact]
+        public void Patch_ReturnsSameVersion_IfPresent_AndDoesNotGetLatestPatchVersion()
```
qianz2, line 273: Maybe we could come up with a naming rules for test class names, like `SomeFeature_**Returns**What_**If**WhatHappens`. In my opinion, this one could be `Patch_ReturnsSameVersion_IfPresentsAndDoesNotReturnLatestPatchVersion`.

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1217 @@
...
+
+            // Act
+            var actual = globalJsonHelper.GetSatisfyingSdkVersion(globalJson, availableVersions);
+
+            // Assert
+            Assert.Equal(expectedVersion, actual);
+        }
+
+        [Fact]
+        public void Patch_ReturnsNull_IfSpecifiedPatchOrLatestPatchVersion_DoesNotExist()
```
qianz2, line 315: nit: `Patch_ReturnsNull_IfSpecifiedPatchOrLatestPatchVersionDoesNotExist`

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1217 @@
...
+
+            // Act
+            var actual = globalJsonHelper.GetSatisfyingSdkVersion(globalJson, availableVersions);
+
+            // Assert
+            Assert.Equal(expectedVersion, actual);
+        }
+
+        [Fact]
+        public void Feature_ReturnsNull_WhenNextHigherFeatureAndLatestPatchLevel_IsNotPresent()
```
qianz2, line 391: nit: `Feature_ReturnsNull_IfNextHigherFeatureAndLatestPatchLevelIsNotPresent`

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1217 @@
...
+
+            // Act
+            var actual = globalJsonHelper.GetSatisfyingSdkVersion(globalJson, availableVersions);
+
+            // Assert
+            Assert.Null(actual);
+        }
+
+        [Fact]
+        public void Feature_ReturnsLatestPatchLevel_ForSameFeatureIfPresent()
```
qianz2, line 335: nit: Feature_ReturnsLatestPatchLevel_IfSameFeatureIsPresent

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1217 @@
...
+
+            // Act
+            var actual = globalJsonHelper.GetSatisfyingSdkVersion(globalJson, availableVersions);
+
+            // Assert
+            Assert.Null(actual);
+        }
+
+        [Fact]
+        public void Minor_ReturnsLatestPatchLevel_IfSpecifiedMinorAndFeature_ArePresent()
```
qianz2, line 418: Same here.

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1217 @@
...
+
+            // Act
+            var actual = globalJsonHelper.GetSatisfyingSdkVersion(globalJson, availableVersions);
+
+            // Assert
+            Assert.Equal(expectedVersion, actual);
+        }
+
+        [Fact]
+        public void LatestPatch_OnlyConsidersPatchWithinSameFeature()
```
qianz2, line 677: Not sure about this one: `LatestPatch_ReturnsSamePatchVersion_IfPatchVersionsWithinSameFeatureAreNotAvailable`?

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1217 @@
...
+
+            // Act
+            var actual = globalJsonHelper.GetSatisfyingSdkVersion(globalJson, availableVersions);
+
+            // Assert
+            Assert.Equal(expectedVersion, actual);
+        }
+
+        [Fact]
+        public void NoRollFoward_UseLatestPatchAsDefault()
```
qianz2, line 778: RollForward_ReturnsLatestPatchAsDefault_IfRollForwardIsNotPresent? Focusing so much on the naming lol..

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
kichalla: Sure, sounds good.

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
kichalla: I think currently this test name follows the format you suggested `TestFeature_ReturnsWhat_IfWhatHappens_AndDoesNotDoOtherThings`

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
kichalla: Nice!

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
kichalla: Nice! 

>  me focusing too much on the naming lol..

Actually this is good since it will help our team to have consistent and good names. Thanks!

src/BuildScriptGenerator/DotNetCore/GlobalJsonSdkResolver.cs
```diff
@@ -0,0 +1,371 @@
...
+                else
+                {
+                    throw new InvalidUsageException(
+                        $"Invalid value {sdkNodeInGlobalJson.AllowPreRelease} for " +
+                        $"'allowPrelease' in global.json. Allowed values are either 'true' or 'false'.");
+                }
+            }
+
+            string resolvedVersion = null;
+            switch (rollForward.ToLower())
```
arroyc, line 97: this is good .. which resolves my previous question.

src/BuildScriptGenerator/DotNetCore/GlobalJsonSdkResolver.cs
```diff
@@ -0,0 +1,371 @@
...
+                    var nextHighestMinorSdk = higherMinorSdks.OrderBy(sdk => sdk).First();
+                    var nextHighestMinorSdks = higherMinorSdks.Where(sdk => sdk.Minor == nextHighestMinorSdk.Minor);
+                    return GetFeature(nextHighestMinorSdks, nextHighestMinorSdk);
+                }
+            }
+
+            return null;
+        }
+
+        private string GetMajor(IEnumerable<SdkVersionInfo> availableSdks, SdkVersionInfo versionToResolve)
```
arroyc, line 245: repeating my question here … 
as per this logic it looks like 
if user is locally using some non-preview version of 2.1.4 (for example)
we will try to get the latest patch for the same, if not we move forward with next higher feature band within 2.1 and so on. so in that case we are good accepting point 2 that you mentioned above.

can you confirm?

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1218 @@
+﻿// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
```
arroyc, line 5: nit: interesting. this file didn't show as GitHub doesn't render large changes. I need to review this one

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1218 @@
...
+                RollForward = GlobalJsonSdkResolver.Patch,
+            };
+            var globalJsonHelper = GetGlobalJsonHelper();
+
+            // Act
+            var actual = globalJsonHelper.GetSatisfyingSdkVersion(globalJson, availableVersions);
+
+            // Assert
+            Assert.Equal(expectedVersion, actual);
+        }
```
arroyc, line 313: so it return 3.1.102?

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1218 @@
...
+
+        [Fact]
+        public void Minor_ReturnsNextMinorAndNextHigherFeatureAndLatestPatchLevel_IfSpecifiedMinorIsNotPresent()
```
arroyc, line 466: yes this is the test i was looking for? is it going to break the user? minor version releases breaking changes as well right?

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
```diff
@@ -0,0 +1,1218 @@
...
+                "3.0.101",
+                "3.2.100",
+                "3.2.101",
+                "3.2.200",
+                "3.2.201",
+            };
+            var globalJson = new GlobalJsonModel();
+            globalJson.Sdk = new Sdk
+            {
+                Version = "3.2.202",
```
arroyc, line 512: imo we should return 3.2.201 instead of null here. what do you think?

qianz2:
> @qianz2 I addressed your comments. Can you take a look again?

The naming now looks good, thanks

src/BuildScriptGenerator/DotNetCore/GlobalJsonSdkResolver.cs
kichalla: > can dotnet core releases have breaking changes in their minor version?

No, .NET Core team would not do breaking changes in minor versions intentionally. Imagine this, currently if there is *no* global.json file, .NET Core cli picks up the *latest major* version, so what we are suggesting here we do is far safer than the default.

src/BuildScriptGenerator/DotNetCore/GlobalJsonSdkResolver.cs
kichalla: Just to make it clear, this PR does not handle the scenario of the points (`1.` and `2.`) that I made above. I just wanted to bring up some points for consideration in a later PR (if you look at this PR it is not integrated with rest of Oryx yet)

> if user is locally using some non-preview version of 2.1.4 (for example)
we will try to get the latest patch for the same, if not we move forward with next higher feature band within 2.1 and so on. so in that case we are good accepting point 2 that you mentioned above.

I did not understand your scenario clearly but if a user did not provide a global.json file, we can decide on either doing `minor` or `latestMinor` based on a runtime version.

kichalla:
> can we have a test where we don't have 3.0 or newer sdk as available sdk and we use rollforward or allowprerelease in global json? I wanted to see how do we handle this sort of errors?

As I mentioned in one of my comments, this is going to be addressed in a later PR. This PR was to just show our implementation of global.json spec.

tests/BuildScriptGenerator.Tests/DotnetCore/GlobalJsonSdkResolverTest.cs
kichalla: Few things to consider here:
- It will be highly unlikely that a specified version would not exist, like we would having a bunch of sdks in our storage.
- .NET Core team would not do breaking changes in minor versions intentionally. As I mentioned in one of my earlier comments, if there is no global.json spec the default is `latest major` so what I am suggesting we do is `>=minor` which is far safer option.

src/BuildScriptGenerator/DotNetCore/GlobalJsonSdkResolver.cs
arroyc: ah ok

> Just to make it clear, this PR does not handle the scenario of the points (1. and 2.) that I made above. I just wanted to bring up some points for consideration in a later PR (if you look at this PR it is not integrated with rest of Oryx yet)
> 
> if user is locally using some non-preview version of 2.1.4 (for example)
> we will try to get the latest patch for the same, if not we move forward with next higher feature band within 2.1 and so on. so in that case we are good accepting point 2 that you mentioned above.
> 
> I did not understand your scenario clearly but if a user did not provide a global.json file, we can decide on either doing minor or latestMinor based on a runtime version.

im fine with current test coverage then

src/BuildScriptGenerator/DotNetCore/GlobalJsonModel.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class GlobalJsonModel
+    {
+        public Sdk Sdk { get; set; }
+    }
+
+    public class Sdk
+    {
+        public string Version { get; set; }
+
+        public string AllowPreRelease { get; set; }
```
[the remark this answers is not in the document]
kichalla: I updated the PR by making this and the next property strongly typed. Can you take a look again?

kichalla:
gentle ping

PR #556, opened 2020-04-09
clean up build image creation

arroyc:


images/build/Dockerfiles/Dockerfile
```diff
@@ -235,12 +211,6 @@ RUN cp ${IMAGES_DIR}/build/dotNetCoreAll.Readme.md /opt/dotnet/Readme.md
 # Copy NodeJs, NPM and Yarn related content
 COPY --from=node-install /opt /opt
 
-# Install PHP pre-reqs
-RUN ${IMAGES_DIR}/build/php/prereqs/installPrereqs.sh
-
-# Oryx depends on the run script generators for most of its
-# `IProgrammingPlatform.GenerateBashRunScript()` implementations
-COPY --from=startupScriptGens /opt/startupcmdgen/ /opt/startupcmdgen/
```
qianz2, line 243: So wondering why we're removing all php prereqs files, we don't need them anymore?

qianz2: Do we want to change build image tests accordingly or it won't affect those tests

arroyc: ans1. yes we may have some failing build images. idea is to find them and make sure they are taken care of by either removing them (don't need anymore) or fixing them.
ans2. so … we are now dynamically installing  php binaries in the build image .. so we don't need to create php build bases anymore .. that's why we can get rid of these files.
.

qianz2: Got it, thanks. Let's make sure all those tests will pass.

images/build/Dockerfiles/Dockerfile
```diff
@@ -235,12 +211,6 @@ RUN cp ${IMAGES_DIR}/build/dotNetCoreAll.Readme.md /opt/dotnet/Readme.md
 # Copy NodeJs, NPM and Yarn related content
 COPY --from=node-install /opt /opt
 
-# Install PHP pre-reqs
-RUN ${IMAGES_DIR}/build/php/prereqs/installPrereqs.sh
```
kichalla, line 239: These are just the pre-reqs which I think are required to be present in the build image. No?

images/build/Dockerfiles/Dockerfile
```diff
@@ -197,30 +197,6 @@ FROM dotnet-install AS buildscriptbuilder
 
 FROM mcr.microsoft.com/oryx/base:build-yarn-cache-${YARN_CACHE_BASE_TAG} AS yarn-cache-base
 
-###
-# Build run script generators (to be used by the `oryx run-script` command)
```
kichalla, line 201: Avoid doing this change in this PR and handle it separately. This way you can delete this content as well as the corresponding command, tests etc. in the same PR. In case we want to revert a commit, we can hopefully.

build/buildBuildImages.sh
```diff
@@ -16,6 +16,7 @@ source $REPO_DIR/build/__phpVersions.sh    # For PHP_BUILD_BASE_TAG
 source $REPO_DIR/build/__nodeVersions.sh   # For YARN_CACHE_BASE_TAG
 source $REPO_DIR/build/__nodeVersions.sh   # For YARN_CACHE_BASE_TAG
 source $REPO_DIR/build/__sdkStorageConstants.sh
+source $REPO_DIR/build/__buildBuildImageFunctions.sh
```
kichalla, line 19: Some of these functions use variables like `BUILD_SIGNED` in them which is set actually later than this line, so they might fail.

arroyc:
this is not needed now

PR #558, opened 2020-04-10
add custom startup command tests and sample python app

qianz2:


tests/Oryx.Integration.Tests/Python/PythonCustomStartUpCommandTest.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        public PythonCustomStartUpCommandTest(ITestOutputHelper output, TestTempDirTestFixture testTempDirTestFixture)
+            : base(output, testTempDirTestFixture)
+        {
+        }
+
+        [Theory]
+        [InlineData("2.7")]
+        [InlineData("3.6")]
+        [InlineData("3.7")]
```
arroyc, line 29: add 3.8 as well?

tests/Oryx.Integration.Tests/Python/PythonCustomStartUpCommandTest.cs
```diff
@@ -0,0 +1,124 @@
...
+            var volume = CreateAppVolume(appName);
+            var appDir = volume.ContainerDir;
+
+            var buildScript = new ShellScriptBuilder()
+                .AddCommand($"oryx build {appDir} --platform {PythonConstants.PlatformName} --platform-version {pythonVersion}")
+                .ToString();
+
+            // Create a custom startup script
+            const string customStartUpScript = "customStartup.sh";
+            File.WriteAllText(Path.Join(volume.MountedHostDir, customStartUpScript),
```
arroyc, line 43: may be its better to add this script as part of the sample app and adjust the filepath here.

tests/Oryx.Integration.Tests/Python/PythonCustomStartUpCommandTest.cs
```diff
@@ -0,0 +1,124 @@
...
+                {
+                    var data = await _httpClient.GetStringAsync($"http://localhost:{hostPort}/");
+                    Assert.Contains("Hello Gunicorn!", data);
+                });
+        }
+
+        [Theory]
+        [InlineData("2.7")]
+        [InlineData("3.6")]
+        [InlineData("3.7")]
```
arroyc, line 80: 3.8?

tests/SampleApps/python/http-server-py/myapp.py
```diff
@@ -0,0 +1,8 @@
+
+def app(environ, start_response):
+    data = b"Hello Gunicorn!\n"
+    start_response("200 OK", [
+        ("Content-Type", "text/plain"),
+        ("Content-Type", str(len(data)))
```
arroyc, line 6: content-length?

tests/Oryx.Integration.Tests/Python/PythonCustomStartUpCommandTest.cs
```diff
@@ -0,0 +1,124 @@
...
+
+            var buildScript = new ShellScriptBuilder()
+                .AddCommand($"oryx build {appDir} --platform {PythonConstants.PlatformName} --platform-version {pythonVersion}")
+                .ToString();
+
+            // Create a custom startup script
+            const string customStartUpScript = "customStartup.sh";
+            File.WriteAllText(Path.Join(volume.MountedHostDir, customStartUpScript),
+                "#!/bin/bash\n" +
+                "pip install gunicorn\n" +
```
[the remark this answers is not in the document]
qianz2: yeah it'll work without this command since the custom startup command test passed without including it so I'll remove it from here.

tests/Oryx.Integration.Tests/Python/PythonCustomStartUpCommandTest.cs
qianz2: Forgot to include 3.8,  I'll add it as well, thanks

kichalla: This got me thinking...the pivot that is being tested is if the user's startup command is being invoked instead of our default command. So for both these test cases here testing with only one version is enough since this decision of whether a user's startup command needs to be used is not dependent on the version.

arroyc: I agree … but I think its better to keep two versions at least .. 2.7 and one of the 3.*

tests/SampleApps/python/http-server-py/myapp.py
qianz2: Good catch, this should be `Content-Length`

tests/Oryx.Integration.Tests/Python/PythonCustomStartUpCommandTest.cs
qianz2: Cool, I'll change the versions to include only 2.7 and 3.8.

PR #564, opened 2020-04-15
Removed support for zipping all output from .NET Core platform

kichalla:
This feature has not been used by AppService so cutting it out since it only adds more work for us.


kichalla:
> How about the ZipAllOutput in buildManifestHelper.go parameter and main.go line 111

Good catch. I will make this change in another PR that I current have opened

PR #565, opened 2020-04-16
Apply multi workers strategy to Gunicorn

qianz2:
Apply multi workers strategy as recommended by Gunicorn docs.


src/startupscriptgenerator/src/python/scriptgenerator.go
```diff
@@ -224,6 +226,13 @@ func (gen *PythonStartupScriptGenerator) buildGunicornCommandForModule(module st
 	// Default to App Service's timeout value (in seconds)
 	args := "--timeout 600 --access-logfile '-' --error-logfile '-'"
 
+	// 2N+1 number of workers is recommended by Gunicorn docs.
```
kichalla, line 229: This is not what we want to do here. You can find more details in the work item's description. 
We had done a similar thing few months ago and then we noticed that it broke some users' apps mainly when they were using in-memory session scenarios.

kichalla: So as per yesterday's meeting, this code is good...we need to introduce a environment variable based flag - which if is true we set the worker count dynamically...Also let's first send out an email out to Jenny and Sanchit to come to an agreement on this.

src/Common/ExtVarNames.cs
```diff
@@ -12,5 +12,6 @@ public static class ExtVarNames
         public const string UserAppInsightsKeyEnv = "APPINSIGHTS_INSTRUMENTATIONKEY";
         public const string UserAppInsightsEnableEnv = "ApplicationInsightsAgent_EXTENSION_VERSION";
         public const string UserAppInsightsConnectionStringEnv = "APPLICATIONINSIGHTS_CONNECTION_STRING";
+        public const string PythonEnableGunicornDynamicWorkerEnvVarName = "PYTHON_ENABLE_GUNICORN_DYNAMIC_WORKER";
```
kichalla, line 15: I am not sure I like this name...how about `PYTHON_ENABLE_GUNICORN_MULTIWORKERS`?

src/startupscriptgenerator/src/python/scriptgenerator.go
```diff
@@ -224,6 +226,15 @@ func (gen *PythonStartupScriptGenerator) buildGunicornCommandForModule(module st
 	// Default to App Service's timeout value (in seconds)
 	args := "--timeout 600 --access-logfile '-' --error-logfile '-'"
 
+	pythonEnableGunicornMultiWorkers := common.GetBooleanEnvironmentVariable(consts.PythonEnableGunicornMultiWorkersEnvVarName)
+	if pythonEnableGunicornMultiWorkers {
+		// 2N+1 number of workers is recommended by Gunicorn docs.
+		// Where N is the number of CPU threads.
+		// One worker will be reading or writing from the socket while the other worker is processing a request.
+		workers := strconv.Itoa((2 * runtime.NumCPU()) + 1)
+		args = appendArgs(args, "--workers="+workers)
```
arroyc, line 235: nit: may be it would be a good practice to check for any errors for strconv.Itoa(). Though it is highly unlikely but I would call appendArgs when workers > 0.

src/startupscriptgenerator/src/common/consts/ext_var_names.go
```diff
@@ -10,3 +10,4 @@ const PhpOriginEnvVarName string = "PHP_ORIGIN"
 const UserAppInsightsKeyEnv string = "APPINSIGHTS_INSTRUMENTATIONKEY"
 const UserAppInsightsEnableEnv string = "ApplicationInsightsAgent_EXTENSION_VERSION"
 const UserAppInsightsConnectionStringEnv string = "APPLICATIONINSIGHTS_CONNECTION_STRING"
+const PythonEnableGunicornMultiWorkersEnvVarName string = "PYTHON_ENABLE_GUNICORN_MULTIWORKERS"
```
arroyc, line 13: is this python_ something appservice team recommended? sounds very weird, but if they are recommending it .. it is what it is

qianz2: They recommend us to add a python prefix on it for future use, and we'll double check with them on next week's meeting.

src/startupscriptgenerator/src/python/scriptgenerator.go
qianz2: Thanks, I think that's a good point, will add an error checking after this part.

PR #569, opened 2020-04-24
Added libgdiplus library to build and .net core runtime images.

kichalla:
For example to enable users to use System.Drawing api in their apps


images/build/Dockerfiles/gitHubActions.Dockerfile
```diff
@@ -45,6 +45,7 @@ RUN apt-get update \
         libssl1.0.2 \
         libstdc++6 \
         zlib1g \
+        libgdiplus \
```
kichalla, line 48: Adding this library to 'build' images is not required for AppService scenarios for example, but this is for VSO users who might want to run an app in the build image.

tests/Oryx.Integration.Tests/DotNetCore/DotNetCoreRuntimeVersion31Tests.cs
```diff
@@ -107,5 +107,50 @@ public async Task CanBuildAndRunApp_FromNestedOutputDirectory()
                     Assert.Contains("Welcome to ASP.NET Core MVC!", data);
                 });
         }
+
+        [Fact]
+        public async Task CanRunAppWhichUsesGDILibrary()
+        {
+            // Arrange
+            var appName = "ImageResizingWebApp";
+            var version = "3.1";
```
kichalla, line 116: I added this integration test only for 3.1 version. Feels like an overkill to have one for every runtime version even though ideally we should have one but probably I can add just runtime image tests checking for presence of `libgdiplus` library

tests/Oryx.RuntimeImage.Tests/DotNetCore/LibraryVerificationTest.cs
```diff
@@ -0,0 +1,49 @@
...
+using Microsoft.Oryx.Tests.Common;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Oryx.RuntimeImage.Tests
+{
+    public class LibraryVerificationTest : TestBase
+    {
+        public LibraryVerificationTest(ITestOutputHelper output) : base(output)
+        {
```
qianz2, line 15: Could we add `[Trait("category", "dotnetcore")]` to specify the platform, and probably add a prefix in front of `LibraryVerificationTest` current test name?

qianz2: Maybe change to `DotNetCoreLibraryVerificationTest` in order to match with other tests under dotnetcore platform..

qianz2:
Since it involved some integration tests, we'll need to pass a nightly build for this PR.

build/__dotNetCoreRunTimeVersions.sh
```diff
@@ -1,6 +1,6 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
-DOT_NET_CORE_RUNTIME_BASE_TAG='20200331.2'
+DOT_NET_CORE_RUNTIME_BASE_TAG='20200424.2'
```
qianz2, line 3: How about other platforms' base tags, the ones in `_nodeVersion.sh`, `_phpVersion.sh` and `_pythonVersion.sh`? I saw they are not updated to most recent ones yet, maybe update in a new PR?..

kichalla: Sure sounds good.

build/constants.yaml
```diff
@@ -30,7 +30,7 @@
       file-name-prefix: __
 - name: dot-net-core-run-time-versions
   constants:
-    dot-net-core-runtime-base-tag: 20200331.2
+    dot-net-core-runtime-base-tag: 20200424.2
```
qianz2, line 33: Same here, other base tags in this `constants.yaml` not updated

kichalla: This is because I added this library only in build and .NET Core runtime images. We typically update _all_ the runtime base tags of platforms unless it is required to do so.

PR #571, opened 2020-04-28
Use template for generating .NET Core build script and also enable it to be part of a multi-platform build.

kichalla:


src/BuildScriptGenerator/BuildScriptGenerator.csproj
```diff
@@ -78,9 +78,14 @@
     <Content Remove="Python\PythonBashBuildSnippet.sh.tpl" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="DotNetCore\DotNetCoreBashBuildSnippet.sh.tpl" />
```
arroyc, line 82: nit: content remove instead of none remove

arroyc:
it would be nice to have a sample app integration test with node and dotnetcore as well

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -145,112 +145,31 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
...
             return new BuildScriptSnippet
             {
-                BashBuildScriptSnippet = scriptBuilder.ToString(),
-                IsFullScript = true,
+                BashBuildScriptSnippet = script,
+                BuildProperties = manifestFileProperties,
+                PlatformInstallationScriptSnippet = installationScriptSnippet,
+
+                // Setting this to false to avoid copying files like '.cs' to the destination
+                CopySourceDirectoryContentToDestinationDirectory = false,
```
qianz2, line 172: Just a little confused why we're setting it to be false while within `BuildScriptSnippet.cs` we let it default be true

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreSampleAppsTest.cs
```diff
@@ -459,7 +459,7 @@ public void Build_Executes_InlinePreAndPostBuildCommands()
         }
 
         [Fact]
-        public void MultiPlatformBuild_IsDisabled()
+        public void CanUseDotNetCoreAsPartOfMultiPlatformBuild()
```
qianz2, line 462: nit: maybe name it starting with `Build_` to conform with other test names.

qianz2:
Please make sure all tests are passed, some of failed tests in `Test Build Images` are related to dotnet core dynamic install tests.

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
arroyc: > Just a little confused why we're setting it to be false while within BuildScriptSnippet.cs we let it default be true

because default is true, means for all platforms we are copying the source directory content, except dotnetcore apps and multiplatform apps where we do dotnet publish/build

kichalla: In case of Node and Python platforms we copy the source files to the target folder and they get interpreted & compiled during run of an app where as in case of .NET Core we compile from source files to get the output and run using them so we do not need the source files (ex: *.cs).

PR #572, opened 2020-04-28
Support node 14 version in Oryx

qianz2:
Support Nodejs 14.0.0;
Support Nodejs LTS version 12.16.3


build/__nodeVersions.sh
```diff
@@ -9,4 +9,5 @@ NODE6_VERSION='6.17.1'
 NODE8_VERSION='8.17.0'
 NODE10_VERSION='10.19.0'
 NODE12_VERSION='12.16.1'
+NODE14_VERSION='14.0.0'
```
kichalla, line 12: Can you add tests for this? You can take a look at existing tests for version `12` to get an idea.

build/constants.yaml
```diff
@@ -116,6 +116,7 @@
     node8-version: 8.17.0
     node10-version: 10.19.0
     node12-version: 12.16.1
```
kichalla, line 118: Seems like there is a new version of 12 (12.16.3). It is your call if you want to update that also as part of this PR.

qianz2: Oh I see, thanks for checking it, will update them together then. It's the latest LTS version so going to replace 12.16.1 with it.

qianz2: Oh a little correction, not replacing with the new version, will add it to the existing versions.

arroyc:
as kiran pointed out add some test for node 14 and this is good to be merged

PR #573, opened 2020-04-29
Added .NET Core 5 preview versions and support for installing preview versions

kichalla:


arroyc:
should we hold on to this for a moment till we get into buster?

kichalla:
> should we hold on to this for a moment till we get into buster?

I do not think we need to wait for that. .NET Core typically is not as tied up to underlying platform as others that we have seen so it is good to go..for example, even 3.1 was targeted for buster but we are still building on stretch but run on buster

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -292,26 +292,7 @@ private void LogScriptIfGiven(string type, string scriptPath)
         /// <returns>The maximum version that satisfies the requested version spec.</returns>
         private string GetMatchingTargetVersion(IProgrammingPlatform platform, string targetVersionSpec)
         {
-            string targetVersion;
```
kichalla, line 295: This piece of logic has been removed because:
1. This duplicates with what each individual platform's detector does.
2. SemanticVersionResolver does not handle each platform's preview format resolution. For example currently it does not handle .NET Core preview version format. I am fixing this issue in .NET Core's platform detector instead of having switch-case conditions within SemanticVersionResolver. Now since this detector handles preview version too, I am removing this code and pushing this responsibility to each individual platform itself.

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -292,26 +292,7 @@ private void LogScriptIfGiven(string type, string scriptPath)
         /// <returns>The maximum version that satisfies the requested version spec.</returns>
         private string GetMatchingTargetVersion(IProgrammingPlatform platform, string targetVersionSpec)
```
kichalla, line 293: This api is used when an explicit version is provided by a user. Since the version provided by an end user can be just major or major.minor or major.minor.patch, this api tries to resolve it. Note that we do a similar thing in each of the platform detectors too...so there the use case is that an explicit platform was not provided but instead a version was detected and that could also be just major or major.minor or major.minor.patch and we need to resolve it.

platforms/dotnet/versionsToBuild.txt
```diff
@@ -1,24 +1,32 @@
...
 2.2.402, 81937DE0874EE837E3B42E36D1CF9E04BD9DEFF6BA60D0162AE7CA9336A78F733E624136D27F559728DF3F681A72A669869BF91D02DB47C5331398C0CFDA9B44,
+
 3.0.100-preview8-013656, 448C740418F0AB43B3A8D9F7CCB532E71E590692D3B64239C3F21D46DF3A46788B5B824E1A10236E5ABE51D4A5143C27B90D08B342A683C96BD9ABEBC2D33017,
 3.0.100-preview9-014004, B4E4580250F172C7B14CBCE96D2DE54D13A0B312ED71F86986CC53A9392491347E83CDEFC2805A23BD652FC2CC0658C2E74FC8C05212C3983C19E2ADBC1FE9A3,
 3.0.100-rc1-014190, 24CFE258849579399B93AC81813BDDDADD1D1C546E3EB5DAE7586263F157A4B3DE3223E6A54B5458ECD103E4C70499CAECC330B57570302E378A2203EADDE671,
 3.0.100, 766da31f9a0bcfbf0f12c91ea68354eb509ac2111879d55b656f19299c6ea1c005d31460dac7c2a4ef82b3edfea30232c82ba301fb52c0ff268d3e3a1b73d8f7,
 3.0.101, 8cd6070c9df1272ac3f2a0e9bd8c1fc2cfef38907fa2e3c81384f005168f45036b5904b4f731cbe37d2b3fe558f560008c481939093b0d08ef0f1778ca97f186,
 3.0.102, 77bc287d9c20630976ac4c0736192ba4899154c9e7cc5b87bc9d94d5d8abafdd832cfe8f385b6ba584c702d9261566109df15ab46b0d62bd218d950d3b47893e,
 3.0.103, 22acd337c1f837c586b9d0e3581feeba828c7d6dc64e4c6c9b24bdc6159c635eb7019c3fb0534edeb4f84971e9c3584c7e3a4d80854cf5664d2792ee8fde189b,
+
```
arroyc, line 21: nit: space

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
arroyc: can you add  some tests for this? ignore if they are already there

> This api is used when an explicit version is provided by a user. Since the version provided by an end user can be just major or major.minor or major.minor.patch, this api tries to resolve it. Note that we do a similar thing in each of the platform detectors too...so there the use case is that an explicit platform was not provided but instead a version was detected and that could also be just major or major.minor or major.minor.patch and we need to resolve it.

qianz2: So there are several tests within `DefaultBuildScriptGeneratorTest` do the exact things for testing but I'm not sure if they covered all the cases.

qianz2: @kichalla could you check if all cases in the scenarios are covered in the tests?

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
```diff
@@ -252,5 +252,63 @@ public void BuildsApplication_IgnoresRuntimeVersionBasedSdkVersion_AndUsesSdkVer
                 },
                 result.GetDebugInfo());
         }
+
+        [Fact]
+        public void BuildsDotNetCore50App_UsingSdkFromGlobalJson()
```
qianz2, line 257: Will it help if we don't name it specifically as dotnetcore50app but instead add it as InlineData for future use?

tests/Oryx.Integration.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
```diff
@@ -64,7 +64,7 @@ public DotNetCoreDynamicInstallTest(ITestOutputHelper output, TestTempDirTestFix
                     "-c",
                     buildImageScript
                 },
-                _imageHelper.GetRuntimeImage("dotnetcore", "dynamic"),
+                _imageHelper.GetRuntimeImage("dotnetcore", runtimeVersion),
```
qianz2, line 67: Wondering why we want to change it here.

tests/Oryx.Integration.Tests/DotNetCore/DotNetCoreRuntimeVersion50Tests.cs
```diff
@@ -0,0 +1,74 @@
...
+        public DotNetCoreRuntimeVersion50Tests(ITestOutputHelper output, TestTempDirTestFixture testTempDirTestFixture)
+            : base(output, testTempDirTestFixture)
+        {
+        }
+
+        [Fact]
+        public async Task CanBuildAndRun_NetCore50MvcApp()
+        {
+            // Arrange
+            var dotnetcoreVersion = "5.0";
```
qianz2, line 28: We could add this new version of MVC app name to the constants of `TestsBase`.

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
kichalla: Sure makes sense. Will update it.

tests/Oryx.Integration.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
kichalla: Good catch. This change should not be required.

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
kichalla: > can you add some tests for this? ignore if they are already there

yes, we already have several tests for these.

> @kichalla could you check if all cases in the scenarios are covered in the tests?

We currently have several tests for these but they are not necessarily as unit tests under `DefaultBuildScriptGeneratorTest`. We have quite a few covered by `SemantingVersioningResolver`, each platform's detector test and integration tests too.

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
kichalla: I added this as part of an existing test's inline data but changed this particular test's name to indicate that we are `testing a preview version from global.json sdk version`

kichalla:
> nit: can you also add a test where an explicit version is mentioned but it's not what we (oryx) support today. How do wew resolve that?
For example, let's say node 14.1.0 is the explicit version and we support 14.0.1 .. what's the use case?
also, lets say node 13.0.5 is the explicit version but we dont support 13 , either node 12 or node 14 .. what will be the experience

Either if a user provides a version explicitly to oryx command or if we detect a version through some file (ex: node engine field in package.json or python's runtime txt file or php's composer file), we depend on SemanticVersionResolver for resolving these versions and we already have tests which handles it:
https://github.com/microsoft/Oryx/blob/master/tests/BuildScriptGenerator.Tests/SemanticVersionResolverTest.cs#L36-L120

PR #575, opened 2020-05-05
update base image tags to have latest openssl

arroyc:
thi is to update openssl for our master .. I will send another  for ant-87 patch


build/__dotNetCoreRunTimeVersions.sh
```diff
@@ -1,6 +1,6 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
-DOT_NET_CORE_RUNTIME_BASE_TAG='20200424.2'
+DOT_NET_CORE_RUNTIME_BASE_TAG='20200505.1'
```
kichalla, line 3: You can avoid making this change since our latest images already have this openssl update...also i have a PR out (.net core 5) which also updates this same image tag...want to avoid collision if it is really not required to do so:

**April 24 image:**

```

C:\oryx [kichalla/add.dotnetcore.5 ≡]> docker run --entrypoint bash mcr.microsoft.com/oryx/base:dotnetcore-3.1-20200424.2 -c "apt-cache policy openssl"
openssl:
  Installed: 1.1.1d-0+deb10u3
  Candidate: 1.1.1d-0+deb10u3
  Version table:
 *** 1.1.1d-0+deb10u3 100
        100 /var/lib/dpkg/status

```

arroyc:
closing this PR as it is not needed anymore based on discussion with @MikeHolman

PR #577, opened 2020-05-05
update dotnet core runtime image bases for openssl [ant-87 patch]

arroyc:
This is for patch release for ant-87 dotnetcore runtime fix for openssl


qianz2:
So what is this branch `patch/20200114.13` for? Should we merge into current master branch instead

arroyc:
> So what is this branch patch/20200114.13 for? Should we merge into current master branch instead

so this branch is made from the commit that was made for build 20200114.3, currently appservice use this build for ant87. We are sending them a patch for this ant 87, so that's the use for this branch. We should not merge it to master branch as we it is already updated.

kichalla:
We do not want this PR anymore right?

kichalla:
Is this PR still required?

PR #578, opened 2020-05-05
Update python docs to include gunicorn support

qianz2:


doc/runtimes/python.md
```diff
@@ -64,13 +66,19 @@ The following process is applied to determine how to start an app.
         * `index.py`
         * `server.py`
 
+### Gunicorn multiple workers suport
```
kichalla, line 69: support

PR #579, opened 2020-05-05
Update pip version to 20.1

qianz2:


build/__pythonVersions.sh
```diff
@@ -1,7 +1,7 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
 PYTHON_BASE_TAG='20200217.4'
-PIP_VERSION='20.0.2'
+PIP_VERSION='20.1'
```
kichalla, line 4: As you know pip is also built from source just like python and included within it but currently our PlatformBinaries pipeline does not build an sdk version if it is already present in the dev storage account. This was an optimization to avoid building sdks if they are really required.
There is a environment variable called `OVERWRITE_EXISTING_SDKS` which you can set when queuing a build in this pipeline and as the name indicates it will cause all sdks to be built again, but since it is going to trigger ALL platforms' sdk builds, I would suggest to introduce platform specific variable (like `OVERWRITE_EXISTING_PYTHON_SDKS`) to override or we could even create separate pipelines for each platform (yeah this is lot of work)

qianz2: Introducing the environment variable like `OVERWRITE_EXISTING_PYTHON_SDKS` sounds plausible for this case, so it will shorten the overall build time and we want to create one for each platform by editing the pipeline script.. And one thing I noticed maybe unrelated to the above is that we have `Oryx-baseImages-PlatformName ` builds for each platform except Python platform, should we create one for Python?..

kichalla: No we do not need base image build for Python.

vsts/pipelines/PlatformBinaries/dotnetcore.yml
```diff
@@ -0,0 +1,53 @@
...
+  - job: DotNetCore
+    timeoutInMinutes: 250
+    pool:
+      name: OryxLinux
+    steps:
+    - template: templates/_platformBinariesTemplate.yml
+      parameters:
+        platformName: 'dotnet'
+
+- stage: Release
```
kichalla, line 16: Let's move this to a separate template since it is a common thing across all platforms here.

kichalla:
Also could you kick off a platform binaries build for each of these platforms?

qianz2:
Capture
Looks like a permission will be needed to add these for pipelines into DevOps.

PR #580, opened 2020-05-06
Create nuget packages for build script generator and cli

kichalla:


src/Detector.Cli/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Microsoft.Oryx.Detector.Cli
```
kichalla, line 3: @qianz2 Just FYI...I added two projects `Detector` and `Detector.Cli` temporarily so that I can make changes at my end without waiting.

vsts/pipelines/templates/_signBinary.yml
```diff
@@ -35,15 +35,75 @@ steps:
...
+  condition: and(succeeded(), eq(variables['setSignTypeVariable.SignType'], 'real'))
+  
+- powershell: |
+    $env:GIT_COMMIT=$(git rev-parse HEAD)
+    $env:BUILD_NUMBER=$env:BUILD_BUILDNUMBER
+    dotnet publish -r linux-x64 -c $(BuildConfiguration) src\BuildScriptGeneratorCLI\BuildScriptGeneratorCli.csproj
+  displayName: 'Publish Oryx BuildScriptGenerator CLI'
   condition: and(succeeded(), eq(variables['setSignTypeVariable.SignType'], 'real'))
   
 - task: VSBuild@1
```
kichalla, line 105: The task related to Nuget package sign will come here

vsts/pipelines/templates/_signBinary.yml
```diff
@@ -10,7 +10,7 @@ steps:
   displayName: 'Set sign binary variable'
   condition: >
    and(in(variables['Build.Reason'], 'Schedule', 'Manual'), eq(variables['Build.DefinitionName'], 'Oryx-CI'), 
-   or(startsWith(variables['Build.SourceBranch'], 'refs/heads/master'), startsWith(variables['Build.SourceBranch'],'refs/heads/patch/' )))
+   or(startsWith(variables['Build.SourceBranch'], 'refs/heads/master'), startsWith(variables['Build.SourceBranch'],'refs/heads/kichalla/' )))
```
qianz2, line 13: Maybe we should keep it as `patch` and avoid using personal branch name?

vsts/pipelines/templates/_releaseJobTemplate.yml
```diff
@@ -78,5 +78,9 @@ jobs:
       repositoryName: Microsoft/Oryx
       tagSource: manual
       tag: '$(RELEASE_TAG_NAME)'
-      assets: $(Build.SourcesDirectory)/artifacts/oryx_linux_x64.tar.gz
+      assets: |
+          $(Build.SourcesDirectory)/artifacts/oryx_linux_x64.tar.gz
+          $(Build.SourcesDirectory)/artifacts/Microsoft.Oryx.Detector.0.2.0.nupkg
+          $(Build.SourcesDirectory)/artifacts/Microsoft.Oryx.Detector.Cli.0.2.0.nupkg
```
qianz2, line 84: So for the detector and detector.cli that later we add on, we need to match these two names here?

vsts/pipelines/templates/_signBinary.yml
kichalla: That was a temporary change to make my branch be built using CI build definition.

vsts/pipelines/templates/_releaseJobTemplate.yml
kichalla: After our discussion few days ago, for now I am creating packages based on existing BuildScriptGenerator and it's CLI

Oryx.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27703.2042
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30011.22
 MinimumVisualStudioVersion = 10.0.40219.1
```
arroyc, line 5: nit: I think for now it's better if we keep min vs version as 16.0.* as well

kichalla:
> nit: how did you test the signing task?

Here you go. This is the build which I ran against a branch where I updated the yml file to even consider my private branch
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3724837&view=logs&j=679d8e30-cde9-5bbf-f15c-33d949595c89&t=0399b1a4-50d1-5f70-dc50-22db6ad5a4f6&l=1

Oryx.sln
kichalla: So this is OK right?

arroyc: I would still like to change the minimum version to 16

PR #582, opened 2020-05-07
[WIP] Add buster based oryx run time images

arroyc:
Currently some of the platforms (node, php, python) has this support, work is in progress for php-fpm, dotnet core is not needed as the support for 3.1 is in buster it self

step1: successful base image build <failing>
step2: update base image tag for corresponding buster based runtime images
step3: update tests and CI for buster based images


.gitignore
```diff
@@ -256,10 +256,14 @@ paket-files/
 
 # Node and .NET Core runtime Dockerfiles, that are auto generated
 images/runtime/node/*/Dockerfile
+images/runtime/node/*/buster.Dockerfile
 images/runtime/dotnetcore/*/Dockerfile
+images/runtime/dotnetcore/*/buster.Dockerfile
```
qianz2, line 261: add one for python?

images/runtime/php-fpm/.gitignore
```diff
@@ -1,2 +1,4 @@
 Dockerfile
-base.Dockerfile
\ No newline at end of file
+buster.Dockerfile
+base.Dockerfile
+base.Buster.Dockerfile
```
qianz2, line 4: Update gitignore files under other platforms' runtime images as well?

images/runtime/commonbase/buster.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM debian:buster-slim
```
kichalla, line 1: >debian:buster-slim [](start = 5, length = 18)

I see that the stretch version of this base image is deriving from `buildpack-deps:stretch-curl`, but here we are not doing the same. Why? is it because of the size?

kichalla:
Is this PR still required?

PR #585, opened 2020-05-12
Fix hugo app build by adding a separate platform for Hugo

kichalla:


src/BuildScriptGenerator/Hugo/HugoBashBuildSnippet.sh.tpl
```diff
@@ -0,0 +1,8 @@
+echo
```
arroyc, line 1: is there other way? For this much, aren't we over doing templating hugo?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -57,11 +57,11 @@ RUN apt-get update \
         gnupg \
         dirmngr \
     && rm -rf /var/lib/apt/lists/*
-ARG HUGO_VERSION=0.59.1
-RUN curl -fsSLO --compressed "https://github.com/gohugoio/hugo/releases/download/v${HUGO_VERSION}/hugo_${HUGO_VERSION}_Linux-64bit.tar.gz" \
+RUN . ${BUILD_DIR}/__hugoConstants.sh \
```
arroyc, line 60: nit: keep line 61 to line  64 in a script and execute the script rather

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -164,12 +164,6 @@ public BuildScriptSnippet GenerateBashBuildScriptSnippet(BuildScriptGeneratorCon
                 configureYarnCache = true;
                 packageInstallerVersionCommand = NodeConstants.YarnVersionCommand;
             }
-            else if (StaticSiteGeneratorHelper.IsHugoApp(ctx.SourceRepo, _environment))
-            {
-                packageManagerCmd = NodeConstants.HugoCommand;
-                packageInstallCommand = NodeConstants.HugoCommand;
-                packageInstallerVersionCommand = NodeConstants.HugoVersionCommand;
```
qianz2, line 171: we could remove these constants from NodeConstants if they're not used anymore. Ignore if they're already removed.

kichalla: Sure, I will handle this clean up in a separate PR.

PR #589, opened 2020-05-13
Revert failed test temporarily and fix errors

qianz2:


src/startupscriptgenerator/src/python/scriptgenerator.go
```diff
@@ -226,15 +224,6 @@ func (gen *PythonStartupScriptGenerator) buildGunicornCommandForModule(module st
 	// Default to App Service's timeout value (in seconds)
 	args := "--timeout 600 --access-logfile '-' --error-logfile '-'"
 
-	pythonEnableGunicornMultiWorkers := common.GetBooleanEnvironmentVariable(consts.PythonEnableGunicornMultiWorkersEnvVarName)
```
kichalla, line 229: Why do you want to revert all of this? You can probably just skip the test? Also does this test fail on your machine too?

qianz2: I'm testing it on my machine now but the VM is running extremely slow and don't know the result yet. So gonna to make it skip the test and then fix the test later.

PR #590, opened 2020-05-13
Refactor detector library and add detect cli command

qianz2:
Work in process:
add test for the defaultPlatformDetector;
add test for detect cli command.


src/BuildScriptGenerator/Contracts/IDetector.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// An abstraction over the detection of all platforms in a given repository.
+    /// </summary>
+    public interface IDetector
+    {
+        /// <summary>
+        /// Returns all platforms detected for the given repository context.
+        /// </summary>
+        /// <param name="ctx">The <see cref="RepositoryContext"/>.</param>
+        /// <returns>A dictionary mapping the detected platforms to the version.</returns>
+        IDictionary<IProgrammingPlatform, string> GetAllDetectedPlatforms(RepositoryContext ctx);
```
kichalla, line 20: Did you rebase your branch on top of master? I ask this because you can then test this api usage experience as an end user would do..for example, have you tried using this api from within a console app (i mean a console app outside Oryx) and how is your experience like...for example, how easy is the api to use, is it exposing lot of things which are not really required to,

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreVersionProvider.cs
```diff
@@ -11,15 +11,15 @@ namespace Microsoft.Oryx.BuildScriptGenerator.DotNetCore
 {
     internal class DotNetCoreVersionProvider : IDotNetCoreVersionProvider
     {
-        private readonly BuildScriptGeneratorOptions _cliOptions;
+        private readonly CommonOptions _cliOptions;
```
qianz2, line 14: Adding a CommonOptions because we want the version providers work for both detectorOptions and ScriptGeneratorOptions. We could rename it if there has a better name.

src/BuildScriptGenerator/Options/DetectorOptions.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the source repo where the application is stored.
+        /// </summary>
+        public string SourceDir { get; set; }
+
+        public string PlatformName { get; set; }
+
+        public string PlatformVersion { get; set; }
+
```
qianz2, line 21: This part probably not needed in a detector, not sure actually..

kichalla:
Is this PR still required?

qianz2:
Will come up with a new PR with only detect CLI and test for it.

PR #595, opened 2020-05-18
Ignore exceptions from standard out line parsers

kichalla:


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -230,7 +230,17 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
 
                 foreach (var processor in stdOutEventLoggers)
                 {
-                    processor.ProcessLine(line);
+                    try
+                    {
+                        processor.ProcessLine(line);
+                    }
+                    catch (Exception ex)
```
kichalla, line 239: I will port this fix to Ant87 branch too.

Following was the exception seen by the end user when using the `appdynamics` native module in their NodeJS app:

```
Unhandled Exception: System.IndexOutOfRangeException: Index was outside the bounds of the array.\n   
at Microsoft.Oryx.BuildScriptGenerator.PipDownloadEventLogger.ProcessLine(String line) in E:\A\_work\2510\s\src\BuildScriptGeneratorCLI\PipDownloadEventLogger.cs:line 33
\n   at Microsoft.Oryx.BuildScriptGeneratorCli.BuildCommand.<>c__DisplayClass24_0.<Execute>b__0(Object sender, DataReceivedEventArgs args) in E:\A\_work\2510\s\src\BuildScriptGeneratorCLI\Commands\BuildCommand.cs:line 254
\n   at System.Diagnostics.Process.OutputReadNotifyUser(String data)\n   at System.Diagnostics.AsyncStreamReader.FlushMessageQueue(Boolean rethrowInNewThread)\n--- End of stack trace from previous location where exception was thrown ---
\n   at System.Diagnostics.AsyncStreamReader.<>c.<FlushMessageQueue>b__17_0(Object edi)\n   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
\n--- End of stack trace from previous location where exception was thrown ---\n   at System.Threading.ThreadPoolWorkQueue.Dispatch()\n/opt/Kudu/KuduConsole/Scripts/starter.sh
```

kichalla:
> adding a test would be nice

Actually I made change to the `webfrontend` app's package.json itself which is used by many of our tests, but what i am seeing is that this package `appdynamics` is not compatible with `6.17.1` version, so I think I have to create a new sample for it unfortunately.

PR #600, opened 2020-05-19
Add new solution for detector library

qianz2:


src/Detector/DetectorLibrary/DetectorLibrary.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
kichalla, line 4: Change this to `netstandard2.0`

src/Detector/DetectorLibrary/DetectorOptions.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// Options to create a detector.
+    /// </summary>
+    public class DetectorOptions
+    {
+        /// <summary>
+        /// Gets or sets the source repo where the application is stored.
+        /// </summary>
+        public string SourceDir { get; set; }
+
+        public string PlatformName { get; set; }
```
kichalla, line 18: Do we need properties from line 18 to 29?

src/Detector/DetectorLibrary/DotNetCore/DotnetCorePlatformDetector.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.Xml.Linq;
+using System.Xml.XPath;
+using Microsoft.Extensions.Logging;
+using Microsoft.Oryx.Detector.Exceptions;
+
+namespace Microsoft.Oryx.Detector.DotNetCore
+{
+    internal class DotNetCorePlatformDetector : IPlatformDetector
+    {
+        private readonly DotNetCoreVersionProvider _versionProvider;
```
kichalla, line 18: These version providers tell the caller what versions are supported by Oryx. In case of detection, since we only 'detect', we need not use this.

src/Detector/DetectorLibrary/DotNetCore/VersionProviders/DotNetCoreOnDiskVersionProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+// --------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Oryx.Detector.DotNetCore
+{
+    public class DotNetCoreOnDiskVersionProvider
```
kichalla, line 13: See my previous comment

src/Detector/DetectorLibrary/Exceptions/UnsupportedPlatformException.cs
```diff
@@ -0,0 +1,18 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+namespace Microsoft.Oryx.Detector.Exceptions
+{
+    /// <summary>
+    /// Supplied platform is not supported.
+    /// </summary>
+    public class UnsupportedPlatformException : InvalidUsageException
```
kichalla, line 11: This exception should not be required. We just have a bunch of detectors which will detect and give us back info.

src/Detector/DetectorLibrary/Exceptions/UnsupportedVersionException.cs
```diff
@@ -0,0 +1,29 @@
...
+// --------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace Microsoft.Oryx.Detector.Exceptions
+{
+    /// <summary>
+    /// Supplied version is not supported.
+    /// </summary>
+    public class UnsupportedVersionException : InvalidUsageException
```
kichalla, line 13: Same here. Not required.

src/Detector/DetectorLibrary/DetectorOptions.cs
qianz2: True we don't need them anymore.

src/Detector/DetectorLibrary/DotNetCore/DotnetCorePlatformDetector.cs
qianz2: imo when we try to get version, it used the version providers to get version info, `detect `is very related to the version providers.

src/Detector/DetectorLibrary/DetectorLibrary.csproj
qianz2: Can I know what's the reason we want to change target framework? So this solution used some extensions from the common project and I added into it. If we changed to the netstandard2.0, it will not be compatible with common project.

kichalla: I forwarded you an email that had sent our team earlier about this. It has details about this one.

src/Common/Extensions/StringExtensions.cs
```diff
@@ -24,16 +24,16 @@ public static bool EqualsIgnoreCase(this string a, string b)
             return string.Equals(a, b, StringComparison.OrdinalIgnoreCase);
         }
 
-        public static string JoinKeyValuePairs(IDictionary<string, string> pairs, char pairSep = ' ', char kvSep = '=')
-        {
-            return string.Join(pairSep, pairs.Select(pair => pair.Key + kvSep + pair.Value));
+        public static string JoinKeyValuePairs(IDictionary<string, string> pairs, char pairSep = ' ', char kvSep = '=')	
+        {
```
kichalla, line 28: nit: remove whitespace? `Ctrl + K + D` short cut to remove all whitespaces from a file

src/Detector/DetectorLibrary/DetectorLibrary.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
kichalla, line 1: How are you testing this library? I am not seeing it being referenced within `BuildScriptGenerator.csproj`?

src/Detector/PlatformDetectorProvider.cs
```diff
@@ -0,0 +1,61 @@
...
+using Microsoft.Oryx.Detector.DotNetCore;
+using Microsoft.Oryx.Detector.Node;
+using Microsoft.Oryx.Detector.Php;
+using Microsoft.Oryx.Detector.Python;
+
+
+namespace Microsoft.Oryx.Detector
+
+{
+    public class PlatformDetectorProvider : IPlatformDetectorProvider
```
kichalla, line 18: You do not need this api since one can register a list of detectors in DI

qianz2: I think it will be easier to extend if we have it, such as adding a hugo detector in the future. We iterate through all detectors, we don't need to add each detectors but call this api in `DefaultPlatformDetector.cs`

kichalla: > we don't need to add each detectors but call this api in DefaultPlatformDetector.cs

Let's say if you write a new detector, we still would have to add it to DI and also over here which feels unnecessary. You can take a look at the following api in current build script generator

```
services.TryAddEnumerable(
                ServiceDescriptor.Singleton<IProgrammingPlatform, NodePlatform>());
```

We can add detectors similar to the above and the `DefaultPlatformDetector` could inject in `IEnumerable<IPlatformDetector>` and loop through them.

qianz2: Got it, that's a good point.. I'll try to use this way instead

src/Detector/DetectorLibrary/DetectorLibrary.csproj
qianz2: I'm considering to keep a common project for common files within build script generator and detector, please let me know if I was wrong. For example, when we make changes within generator `public PlatformDetectorResult Detect(RepositoryContext context)`, it needs to pass in the `RepositoryContext` which both sides are using, ideally we want to call the detector api and pass in the context but the the namespace needs to be converted, one side is `BuildScriptGenerator.RepositoryContext` and the other is `Detector.RepositoryContext`. Imo the best way to keep a common files and they're both using the same RepositoyContext.

qianz2: Also files within Common.Extensions(such as FileExtension.cs, StringExtension.cs) are also duplicated in detector proj because we did't share the project. This kind of blocks me because we are using different target framework.

kichalla: So are you suggesting you want to move files to Common project so that the can be used in both Detector and BuildScriptGenerator project? Sure, sounds good.

qianz2: Yeah I'm thinking to add Common proj as dependency, and Common proj target framework is netcoreapp2.1 which are not compatible with netstandard2.0.. so we still need to move detector proj to netcoreapp2.1.

qianz2:
Included as part of #635

src/BuildScriptGenerator/BuildScriptGenerator.csproj
```diff
@@ -92,6 +92,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Common\Common.csproj" />
+    <ProjectReference Include="..\Detector\Detector.csproj" />
```
kichalla, line 95: revert

images/build/Dockerfiles/buildScriptGenerator.Dockerfile
```diff
@@ -14,6 +14,7 @@ COPY src/BuildScriptGenerator src/BuildScriptGenerator
 COPY src/BuildScriptGeneratorCli src/BuildScriptGeneratorCli
 COPY src/Common src/Common
 COPY src/CommonFiles src/CommonFiles
+COPY src/Detector src/Detector
```
kichalla, line 17: revert

src/BuildScriptGenerator/BuildScriptGeneratorServiceCollectionExtensions.cs
```diff
@@ -9,6 +9,7 @@
 using System.Net.Http.Headers;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Oryx.Detector;
```
kichalla, line 12: revert

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatformDetector.cs
```diff
@@ -11,6 +11,7 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Oryx.BuildScriptGenerator.Exceptions;
+using Microsoft.Oryx.Detector.DotNetCore;
```
kichalla: revert

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatformDetector.cs
```diff
@@ -8,6 +8,9 @@
 using System.Xml.Linq;
 using System.Xml.XPath;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
```
kichalla, line 11: revert

tests/Detector.Tests/Detector.Tests.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
kichalla, line 3: To make sure that your tests are running in CI, for now, you can update the file called `build\testBuildScriptGenerator.sh` with the following content:

```
testProjectName="Detector.Tests"
dotnet test \
    --blame \
    --test-adapter-path:. \
    --logger:"xunit;LogFilePath=$ARTIFACTS_DIR\testResults\\$testProjectName.xml" \
    -c $BUILD_CONFIGURATION
```

src/Detector/DetectorContext.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// Context to create a detector.
+    /// </summary>
+    public class DetectorContext
+    {
+        public ISourceRepo SourceRepo { get; set; }
+
+        /// <summary>
+        /// Gets or sets specific properties for the generated script.
+        /// </summary>
+        public IDictionary<string, string> Properties { get; set; }
```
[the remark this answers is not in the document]
qianz2: This is used in the projectFileProvider related implementation, so it'll need to be included.

src/Detector/DetectorOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Gets or sets the relative path to the project to be built.
+        /// </summary>
+        public string Project { get; set; }
+
+        /// <summary>
+        /// Gets or sets the arguments to be passed into the detector.
+        /// </summary>
+        public string[] PassThruArguments { get; set; }
```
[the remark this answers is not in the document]
qianz2: Sure, will remove it

src/Detector/Detector.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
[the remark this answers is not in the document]
qianz2: As we discussed earlier, this library will use netcoreapp2.1 for two reasons. It's using Common project `Mircorosft.Oryx.Common.Extension` from it so if we change to netstandard then they won't be compatible. And later when we integrate library in generator, we'll also have other common files such as LocalRepo or RepositoryContext which they both are using.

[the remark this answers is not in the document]
qianz2: We could do it if we move Common project to be netstandard2.0 first

src/Detector/DetectorContext.cs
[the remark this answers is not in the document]
kichalla: I see, seems like the ExplicitProjectFileProvider.cs is using it to get the explicitly provided project file. However I would just delete that code..i.e lines 59-65 in that file (it should have never been there actually...i should have deleted it once my hierarchical config was added few weeks ago).

src/Detector/Detector.csproj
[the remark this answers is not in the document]
kichalla: I see. We can leave it as netcoreapp2.1 for now then. Can revisit later.

src/Detector/DetectorOptions.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// Output metatdata information of the detected platforms in JSON format.
+        /// </summary>
+        public bool OutputJson { get; set; }
+
+        /// <summary>
+        /// Gets or sets the relative path to the project to be built.
+        /// </summary>
+        public string Project { get; set; }
```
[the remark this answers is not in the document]
qianz2: Other properties in DetectorOptions are not used actually,  so I will just delete them here

src/Detector/DotNetCore/ProjectFileProviders/ExplicitProjectFileProvider.cs
```diff
@@ -33,9 +33,8 @@ public string GetRelativePathToProjectFile(DetectorContext context)
             if (string.IsNullOrEmpty(projectPath))
             {
                 _logger.LogDebug(
-                    "No request to build a particular project file explicitly either using the " +
-                    $"PROJECT environment variable or the " +
-                    $"'{DotNetCoreConstants.ProjectBuildPropertyKey}' build property");
+                    "No request to build a particular project file explicitly using the " +
```
kichalla, line 36: 👍

PR #608, opened 2020-05-21
Create a new Detector solution, pipeline and nuget packages

kichalla:


qianz2:
We may stop triggering oryx-validation-github build when we push code into detector solution.

build/__detectorNugetPackagesVersions.sh
```diff
@@ -0,0 +1,5 @@
+# This file was auto-generated from 'constants.yaml'. Changes may be overridden.
+
+VERSION_PREFIX='1.0.0'
+VERSION_SUFFIX="pre-$BUILD_BUILDNUMBER"
```
gpcastro, line 4: why the `pre-` prefix?

build/constants.yaml
```diff
@@ -174,3 +174,30 @@
     - type: shell
       directory: build
       file-name-prefix: __
+- name: detector-nuget-packages-versions
```
gpcastro, line 177: why not have this just once, but an array of different outputs?

build/testDetector.sh
```diff
@@ -0,0 +1,29 @@
...
+# Load all variables
+source $REPO_DIR/build/__variables.sh
+
+echo
+echo "Building and running tests..."
+testProjectName="Detector.Tests"
+cd "$TESTS_SRC_DIR/$testProjectName"
+dotnet test \
+    --test-adapter-path:. \
+    --logger:"xunit;LogFilePath=$ARTIFACTS_DIR\testResults\\$testProjectName.xml" \
```
gpcastro, line 20: how about we error out if `ARTIFACTS_DIR` is blank? We should probably document that this is intended for build agents only

src/Detector/Class1.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+
+namespace Detector
```
gpcastro, line 3: remove?

tests/Detector.Tests/UnitTest1.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+using Xunit;
+
+namespace Detector.Tests
+{
+    public class UnitTest1
+    {
+        [Fact]
+        public void Test1()
```
gpcastro, line 9: remove?

vsts/pipelines/ci-detector.yml
```diff
@@ -0,0 +1,94 @@
...
+    - job: ValidatePipeline
+      displayName: Validate pipeline
+      pool:
+        name: OryxLinux
+      variables:
+        skipComponentGovernanceDetection: true
+      steps:
+      - script: |
+          set -x
+          if [ "$(Build.Reason)" != "Manual" ]; then
```
gpcastro, line 20: why only manual?

vsts/pipelines/ci-detector.yml
```diff
@@ -0,0 +1,94 @@
...
+            echo
+            echo "Invalid configuration."
+            echo "Build should be Manual"
+            exit 1
+          fi
+          if [ "$(Build.DefinitionName)" != "Oryx-Detector-CI" ]; then
```
gpcastro, line 26: why do we need this check?

build/__detectorNugetPackagesVersions.sh
kichalla: Let's say we make a release of a nuget package today with a version `1.0.0`. Let's say we would then update this version to `1.0.1` or later, we want to release them as `prerelease` packages and after our customers have tried and tested them, we can then make a release by remove the `prerelease` prefix

build/constants.yaml
kichalla: Yeah I had tried it, but the problem was with how the 'values' of these keys are being interpreted by the scripts. For example, I want to use `pre-$(BUILD_BUILDNUMBER)` for msbuild scenario, but since this content also gets generated for `bash` file, it throws an error saying that it is invalid to use `$(BUILD_BUILDNUMBER)`, so I had to split it into multiple sections.

build/testDetector.sh
kichalla: Sure, good catch

src/Detector/Class1.cs
kichalla: Sure can remove it. I just left it like this hoping Stella will update it's contents anyway.

tests/Detector.Tests/UnitTest1.cs
kichalla: Sure

vsts/pipelines/ci-detector.yml
kichalla: My thought was if we want to keep releasing nuget packages if the source code hasn't changed at all. For example, this could happen in case of `scheduled` automated builds. We can kick off a CI build if we know that we made changes to Detector api. Also the process of upgrading a pre-release package to released package should be done consciously and manually.

vsts/pipelines/ci-detector.yml
kichalla: (I followed the pattern that we have in existing Oryx-CI). I think this was a gate added to make sure that only this build definition can make release builds which also ensures that we only release signed bits.

build/tools/SharedCodeGenerator/Outputs/MSBuildOutput.cs
```diff
@@ -0,0 +1,52 @@
...
+            var name = _collection.Name.Camelize();
+            name = char.ToLowerInvariant(name[0]) + name.Substring(1);
+            return Path.Combine(_directory, _fileNamePrefix + name + ".props");
+        }
+
+        public string GetContent()
+        {
+            StringBuilder body = new StringBuilder();
+            var autoGeneratedMessage = Program.BuildAutogenDisclaimer(_collection.SourcePath);
+            body.AppendLine($"<!-- {autoGeneratedMessage} -->"); // Can't use AppendLine becuase it appends \r\n
```
arroyc, line 37: i feel like this method is more like setcontent

kichalla: I did not get what you mean. This is a method on stringbuilder

src/BuildScriptGenerator/Hugo/DetectorNugetPackagesVersions.cs
```diff
@@ -0,0 +1,11 @@
+// This file was auto-generated from 'constants.yaml'. Changes may be overridden.
+
+namespace Microsoft.Oryx.BuildScriptGenerator.Hugo
+{
+    public static class DetectorNugetPackagesVersions
+    {
+        public const string VersionPrefix = "1.0.0";
+        public const string IsPrerelease = "true";
+        public const string VersionSuffix = "pre-$(BUILD_BUILDNUMBER)";
```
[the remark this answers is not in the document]
kichalla: Take a look at my response to similar question by @gpcastro

vsts/pipelines/templates/_buildTemplate.yml
```diff
@@ -30,7 +30,7 @@ steps:
 - task: ShellScript@2
   displayName: 'Build Oryx.sln'
   inputs:
-    scriptPath: ./build/buildSln.sh
+    scriptPath: ./build/buildSln.sh Oryx.sln
```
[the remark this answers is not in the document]
kichalla: I did not understand what you mean here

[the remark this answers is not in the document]
arroyc: are you passing oryx.sln as an argument to the buildsln script?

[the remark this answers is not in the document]
kichalla: yes, I am using the same script to build either Oryx.sln or Detector.sln file

build/tools/SharedCodeGenerator/Outputs/MSBuildOutput.cs
kichalla: This method is part of `IOutputFile` but irrespective of that this api is only for getting the content which is finally going to be written to a file that is generated at the call site. So `GetContent` feels correct to me.

vsts/pipelines/templates/_buildTemplate.yml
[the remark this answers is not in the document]
kichalla: Good catch. Thanks!

vsts/scripts/createNuGetPackages.ps1
```diff
@@ -0,0 +1,32 @@
...
+# Delete any existing directory and zip file. Could have been from an earlier build.
+DeleteItem "$detectorName"
+DeleteItem "$detectorName.zip"
+Rename-Item -Path "$detectorName.$VERSION.nupkg" -NewName "$detectorName.zip"
+Expand-Archive -Path "$detectorName.zip" -DestinationPath "$detectorName"
+DeleteItem "$detectorName.zip"
+
+Copy-Item `
+    -Path "$repoRoot\src\Detector\bin\Release\netstandard2.0\$detectorName.dll" `
+    -Destination "$detectorName\lib\netstandard2.0\$detectorName.dll" `
```
qianz2, line 27: netcoreapp2.1 to match with current changes?

kichalla: good catch. Will fix it.

src/Detector/Detector.csproj
```diff
@@ -2,12 +2,38 @@
...
-    <AssemblyVersion>0.2</AssemblyVersion>
+    <IsPackable>true</IsPackable>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+    <Authors>Microsoft</Authors>
+    <Company>Microsoft</Company>
+    <RepositoryUrl>https://github.com/microsoft/oryx</RepositoryUrl>
+    <RepositoryType>git</RepositoryType>
+  </PropertyGroup>
+  
+  <Import Project="$(MSBuildThisFileDirectory)\..\..\build\__detectorNugetPackagesVersions.props" />
```
qianz2, line 24: nit: a sub folder `build\detector` to keep all related detector build files

build/testBuildScriptGenerator.sh
```diff
@@ -29,14 +29,6 @@ dotnet test \
     --logger:"xunit;LogFilePath=$ARTIFACTS_DIR\testResults\\$testProjectName.xml" \
     -c $BUILD_CONFIGURATION
 
-testProjectName="Detector.Tests"
-cd "$TESTS_SRC_DIR/$testProjectName"
-dotnet test \
-    --blame \
-    --test-adapter-path:. \
-    --logger:"xunit;LogFilePath=$ARTIFACTS_DIR\testResults\\$testProjectName.xml" \
-    -c $BUILD_CONFIGURATION
```
qianz2, line 38: A quick question, this will be included after we integrated library?

kichalla: Ideally we want Detector and Detector.tests projects to be built, tested and released separately (you can think of them as though they are part of a different repo). So build script generator will be referencing Detector via nuget package reference.
To summarize, once we integrate Detector library, we will no longer require this script as it is being tested as part of `testDetector.sh` file.

kichalla:
> i also see signproj files, but we need to add an workitem to do signing via detector ci or oryx ci .

It is already part of the `signBinaryDetector.yml` in this PR.

PR #613, opened 2020-05-22
Enable installing hugo dynamically

kichalla:


build/__hugoConstants.sh
```diff
@@ -1,4 +1,7 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
 VERSION='0.71.0'
-NAME='Hugo'
+PLATFORM_NAME='hugo'
+INSTALLED_HUGO_VERSIONS_DIR='/opt/hugo'
+INSTALLATION_URL_FORMAT='https://github.com/gohugoio/hugo/releases/download/v#VERSION#/#TAR_FILE#'
```
kichalla, line 6: For now, I have the dynamic install feature to install from this location.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -8,7 +8,7 @@ ARG AGENTBUILD
 # NOTE: This imge is NOT based on 'githubrunners-buildpackdeps-stretch' because AzFunctions
 # team wanted a consistent experience at their end and do not want to see latency that might
 # be caused due to when the GitHub runners' layers and Oryx image layers go out of sync.
-FROM debian:stretch AS main
+FROM githubrunners-buildpackdeps-stretch AS main
```
kichalla, line 11: Changing this after discussing with jam stack team. They are seeing reports where users are unable to use certain libraries which have native dependencies and they are not present in `debian:stretch`

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -96,7 +96,8 @@ ARG SDK_STORAGE_ENV_NAME
 ARG SDK_STORAGE_BASE_URL_VALUE
 WORKDIR /
 
-ENV PATH="$PATH:/opt/oryx:/opt/nodejs/lts/bin:/opt/yarn/stable/bin:/opt/hugo"
+ENV ORYX_PATHS="/opt/oryx:/opt/nodejs/lts/bin:/opt/yarn/stable/bin:/opt/hugo/lts"
```
kichalla, line 99: Adding an extra environment variable which contains all paths related to Oryx. Came across a scenario from jam stack team where their container's path was being overwritten causing Oryx executables not being found (since path was modified). This new environment variable will allow jam stack team to always make sure (in their entrypoint script) oryx paths are present

images/build/Dockerfiles/vso.Dockerfile
```diff
@@ -8,7 +8,7 @@ ARG RELEASE_TAG_NAME=unspecified
 
 RUN rm -rf /tmp/oryx
 
-ENV PATH="$ORIGINAL_PATH:/opt/oryx:/opt/nodejs/lts/bin:/opt/dotnet/sdks/lts:/opt/python/latest/bin:/opt/yarn/stable/bin:/opt/hugo"
```
kichalla, line 11: Removes duplication

tests/Oryx.BuildImage.Tests/Hugo/HugoDynamicInstallationTest.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            // Arrange
+            var version = "0.59.1";
+            var installationDir = $"{BuildScriptGenerator.Constants.TemporaryInstallationDirectoryRoot}/hugo/{version}";
+            var appName = SampleAppName;
+            var volume = CreateSampleAppVolume(appName);
+            var appDir = volume.ContainerDir;
+            var appOutputDir = "/tmp/app-output";
+            var script = new ShellScriptBuilder()
+                .SetEnvironmentVariable("HUGO_VERSION", version)
```
kichalla, line 47: Users can set this environment variable in their github action to dynamically install the version they want

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
arroyc: it would be good to document and list out those libraries ... we will surely need this for buster images

tests/Oryx.BuildImage.Tests/Hugo/HugoDynamicInstallationTest.cs
arroyc: so we default to 0.59.1, but user can override it?

tests/Oryx.BuildImage.Tests/Hugo/HugoDynamicInstallationTest.cs
```diff
@@ -0,0 +1,125 @@
...
+                return data;
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(ImageNameData))]
+        public void GeneratesScript_AndBuilds(string imageName)
+        {
+            // Arrange
+            var version = "0.59.1";
```
arroyc, line 40: what is this version for? can you add a prefix to version so that it's easy to understand

tests/Oryx.BuildImage.Tests/Hugo/HugoDynamicInstallationTest.cs
```diff
@@ -0,0 +1,125 @@
...
+                        result.StdOut);
+                },
+                result.GetDebugInfo());
+        }
+
+        [Fact]
+        public void DynamicInstall_ReInstallsSdk_IfSentinelFileIsNotPresent()
+        {
+            // Arrange
+            var version = "0.70.0"; //NOTE: use the full version so that we know the install directory path
```
arroyc, line 78: nit: same, please add a prefix to this version

kichalla:
> nit: add a test with multiple values of hugo version being used may be?

I cannot think of a scenario for this. Let's get some feedback and react on it.

PR #618, opened 2020-05-29
Install all tools before generating build script for them

kichalla:


src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -162,13 +183,23 @@ internal class DefaultBuildScriptGenerator : IBuildScriptGenerator
 
         private IList<BuildScriptSnippet> GetBuildSnippets(
             BuildScriptGeneratorContext context,
+            IEnumerable<PlatformDetectorResult> detectionResults,
+            bool runDetection,
```
qianz2, line 187: Here does `runDetection` mean detect from all platforms? If so, seems the logic here is opposite..

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -162,13 +183,23 @@ internal class DefaultBuildScriptGenerator : IBuildScriptGenerator
...
-            var platformsToUse = GetCompatiblePlatforms(context);
+            IDictionary<IProgrammingPlatform, string> platformsToUse;
+            if (runDetection)
+            {
+                platformsToUse = _platformDetector.GetCompatiblePlatforms(context);
+            }
+            else
+            {
+                platformsToUse = _platformDetector.GetCompatiblePlatforms(context, detectionResults);
+            }
```
qianz2, line 202: Should we reverse the if/else content?

src/BuildScriptGenerator/DefaultCompatiblePlatformDetector.cs
```diff
@@ -34,13 +36,34 @@ public class DefaultCompatiblePlatformDetector : ICompatiblePlatformDetector
...
+            return GetCompatiblePlatforms(ctx, detectionResults: null, runDetection: true);
+        }
+
+        /// <inheritdoc/>
+        public IDictionary<IProgrammingPlatform, string> GetCompatiblePlatforms(
+            RepositoryContext ctx,
+            IEnumerable<PlatformDetectorResult> detectionResults)
+        {
+            return GetCompatiblePlatforms(ctx, detectionResults, runDetection: false);
+        }
```
qianz2, line 49: I see here makes sense.. please ignore previous comment. The name may be a little confusing

kichalla: Thanks @qianz2 .Yeah basically I want to avoid running the detection again if it was already run earlier. Ideally I want to refactor this logic to fix the layering, but since that is going to take more time and might conflict with what you are currently doing, this is a temporary solution for now.

PR #622, opened 2020-06-03
Refactor code to fix how versions are resolved

kichalla:


src/BuildScriptGenerator/DotNetCore/DotnetCoreScriptGeneratorOptions.cs
```diff
@@ -16,5 +16,7 @@ public class DotNetCoreScriptGeneratorOptions
         /// Gets or sets the MSBuild configuration that needs to be used when doing 'dotnet build'.
         /// </summary>
         public string MSBuildConfiguration { get; set; }
+
+        public string DotNetCoreRuntimeVersion { get; set; }
```
qianz2, line 20: Wondering why we are moving this `ResolvedPlatformRuntimeVersion` from Context into Option?

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -83,34 +83,56 @@ public IEnumerable<string> SupportedVersions
...
+                detectionResult = new PlatformDetectorResult
+                {
+                    Platform = DotNetCoreConstants.PlatformName,
+                    PlatformVersion = explicitVersion,
+                };
+            }
+            else
+            {
+                detectionResult = _detector.Detect(context);
+            }
```
qianz2, line 98: From this logic, we check the explicit version from here so that means we'll remove `ResolvedPlatformRuntimeVersion` from detectorContext. I'll accomodate this change as well, so we don't want to keep the DetectorContext in the library?

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -261,49 +306,91 @@ private string GetBuildConfiguration()
             buildProperties[DotNetCoreManifestFilePropertyKeys.StartupDllFileName] = startupDllFileName;
         }
 
-        public string GetInstallerScriptSnippet(BuildScriptGeneratorContext context)
+        public string ResolveVersion(string versionToResolve)
+        {
+            var resolvedVersion = GetVersionUsingHierarchicalRules(versionToResolve);
+            resolvedVersion = GetMaxSatisfyingVersionAndVerify(resolvedVersion);
+            return resolvedVersion;
+        }
```
qianz2, line 314: I made an IPlatformVersionResolver interface for keeping version resolving logics like this, and then use it from each Platform, we could discuss more about it!

src/BuildScriptGeneratorCli/Commands/PrepareEnvironmentCommand.cs
```diff
@@ -0,0 +1,350 @@
...
+            int exitCode;
+            using (var timedEvent = logger.LogTimedEvent("EnvSetupCommand"))
+            {
+                var context = BuildScriptGenerator.CreateContext(serviceProvider, operationId: null);
+
+                IEnumerable<PlatformDetectorResult> detectedPlatforms = null;
+                if (SkipDetection)
+                {
+                    console.WriteLine(
+                        $"Skipping platform detection since '{SkipDetectionTemplate}' switch was used...");
```
qianz2, line 78: What does SkipDetectionTemplate do? What kind of scenarios will a user use this feature?

src/BuildScriptGeneratorCli/Commands/PrepareEnvironmentCommand.cs
```diff
@@ -0,0 +1,350 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Oryx.BuildScriptGenerator;
+using Microsoft.Oryx.BuildScriptGeneratorCli.Options;
+using Microsoft.Oryx.Common;
+
+namespace Microsoft.Oryx.BuildScriptGeneratorCli
+{
+    [Command(Name, Description = "Sets up environment by detecting and installing platforms.")]
+    internal class PrepareEnvironmentCommand : CommandBase
```
qianz2, line 24: The name maybe a little ambiguous, we could think of a clear name for this command?

src/BuildScriptGenerator/DotNetCore/DotnetCoreScriptGeneratorOptions.cs
kichalla: We have been doing this before my change:
1. Run detect on all platforms by calling the `Detect` on `IProgammingPlatform`
2. Once we get the detection result, we invoke the `SetVersion` on `IProgrammingPlatform` to enable the platform which detected to set the version on context. We did this way because we wanted the platform to set the particular, ex: `ResolvedDotNetCoreVersion`, `ResolvedPythonVersion` property which the platform knows about
4. We then call the platform's `GenetateBashScriptSnippet` to generate script, which in turn used the version set on the on the context to generate script.

Based on the above flow, you should notice that this flow is complicated and could be simplified. With this PR we will be passing in the detected result directly to the method being called, which is `GenetateBashScriptSnippet`

src/BuildScriptGeneratorCli/Commands/PrepareEnvironmentCommand.cs
kichalla: Since the command shortcut is called `prep` (which means prepare), I wanted to keep it in align with it.

src/BuildScriptGeneratorCli/Commands/PrepareEnvironmentCommand.cs
kichalla: For example, AppService folks wanted to install the 'popular' sdks upfront in an image to boost build performance of apps. Since they might install it upfront (ex: calling `RUN oryx prep --skip-detection --platforms-and-versions dotnet=3.1.300,python=3.7.7, node=12.16.3` in their Dockerfile), they need not have an app to do it and skip detection serves this purpose

src/BuildScriptGenerator/DotNetCore/DotnetCoreScriptGeneratorOptions.cs
qianz2: Makes sense.. that's good way to simplify the flow.

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
qianz2: The `DetectorContext ` inherits `RepositoryContext ` and I moved `RepositoryContext `into Common proj, so both library and build script generator are pointing to the same `RepositoryContext `. Because when we use it from build script generator, we want to pass in the `RepositoryContext ` parameter to detectorlibrary directly, otherwise we need to do conversion. This is the current way to handle this part.

src/BuildScriptGeneratorCli/Commands/PrepareEnvironmentCommand.cs
qianz2: Got it, that sounds more like an `oryx install-sdk` command to me.

qianz2: Since it's part of preparation process so it makes sense.

images/build/benv.sh
```diff
@@ -130,7 +130,11 @@ benv-resolve() {
   local value=$(echo $1 | sed 's/^.*=//')
 
   # Resolve node versions
-  if matchesName "node" "$name" || matchesName "node_version" "$name" && [ "${value::1}" != "/" ]; then
+  if matchesName "nodejs" "$name" \
```
arroyc, line 133: why bringing back nodejs? :)

qianz2:
Should we merge this PR after my refactor PR? we're dealing with same part differently so lots of conflicts here.. I'm unable to merge my PR.

PR #624, opened 2020-06-04
Set benv with all the detected platforms

kichalla:


src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -5,27 +5,22 @@
...
     {
         private readonly ILogger<ProbeAndFindProjectFileProvider> _logger;
-        private readonly IStandardOutputWriter _writer;
 
         // Since this service is registered as a singleton, we can cache the lookup of project file.
         private bool _probedForProjectFile;
         private string _projectFileRelativePath;
 
-        public ProbeAndFindProjectFileProvider(ILogger<ProbeAndFindProjectFileProvider> logger, IStandardOutputWriter writer)
+        public ProbeAndFindProjectFileProvider(ILogger<ProbeAndFindProjectFileProvider> logger)
```
qianz2, line 21: Why do we not keep the output writer to alert users in this part

kichalla: This is because we have received feedback from jamstack team about their users complaining about this message which it seem is confusing for them (as jamstack only supports node + some static tools based apps). Also we support `.fsproj` apart from `.csproj` files, so instead of adding more confusing logs to existing issue, I am adding a general message like `Detecting platforms...` so that it is is applicable to all scenarios.

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
```diff
@@ -330,5 +330,40 @@ public void BuildsApplication_UsingPreviewVersionOfSdk()
...
+            // Arrange
+            var appName = "dotnetreact";
+            var hostDir = Path.Combine(_hostSamplesDir, "multilanguage", appName);
+            var volume = DockerVolume.CreateMirror(hostDir);
+            var appDir = volume.ContainerDir;
+            var appOutputDir = $"{appDir}/myoutputdir";
+            var buildScript = new ShellScriptBuilder()
+                .AddBuildCommand(
+                $"{appDir} -o {appOutputDir} --platform {DotNetCoreConstants.PlatformName} --platform-version 2.1")
+                .AddFileExistsCheck($"{appOutputDir}/dotnetreact.dll")
```
arroyc, line 346: nit: as discussed add node_modules check

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
```diff
@@ -330,5 +330,40 @@ public void BuildsApplication_UsingPreviewVersionOfSdk()
...
+                CommandToExecuteOnRun = "/bin/bash",
+                CommandArguments = new[] { "-c", buildScript }
+            });
+
+            // Assert
+            RunAsserts(
+                () =>
+                {
+                    Assert.True(result.IsSuccess);
+                    Assert.Contains(@"npm install", result.StdOut);
```
arroyc, line 364: nit: as discussed, add an asset for dotnet build?

PR #634, opened 2020-06-10
keep track of acr-tasks that runs on schedule for cleaning up container registry repositories

arroyc:
image

### DIFF
--- a/acr-tasks/purge-prod-acr-tasks.sh
+++ b/acr-tasks/purge-prod-acr-tasks.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+# --------------------------------------------------------------------------------------------
+
+# Runs every saturday at 22:00 hrs UTC and deletes images older than 180 days 
+
+set -e
+
+PURGE_CMD="acr purge  --filter 'public/oryx/build:.*'  --ago 180d --untagged"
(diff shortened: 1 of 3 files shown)


acr-tasks/purge-prod-acr-tasks.sh
```diff
@@ -0,0 +1,39 @@
...
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+# --------------------------------------------------------------------------------------------
+
+# Runs every saturday at 22:00 hrs UTC and deletes images older than 180 days 
+
+set -e
+
+PURGE_CMD="acr purge  --filter 'public/oryx/build:.*'  --ago 180d --untagged"
```
kichalla, line 11: what does `--untagged` mean? also you could make the `ago` value into a variable so that it is easy to change

kichalla: Also why do we need 3 scripts. Can we just use one script and parameterize it?

arroyc: @kichalla 
The --untagged parameter indicates to delete manifests in addition to tags
yes we can keep it in one script only .. but again this script we only run once .. we do not need to run it .. its for documentation purpose
we actually dont need to run these scripts .. this is one time script only  to create the ACR tasks ... ACR tasks are created already .. they will run on schedule .. I wanted to keep these scripts for our reference to create future tasks  (if needed)

kichalla: I see, got it.

PR #639, opened 2020-06-16
Fixed platforms command to show list of available versions in a dynamic install enabled environment

kichalla:


src/BuildScriptGeneratorCli/Commands/PlatformsCommand.cs
```diff
@@ -112,6 +114,31 @@ private IList<string> SortVersions(IEnumerable<string> versions)
             return result.Select(v => v.ToString()).ToList();
         }
 
+        internal override IServiceProvider GetServiceProvider(IConsole console)
+        {
+            // NOTE: Order of the following is important. So a command line provided value has higher precedence
+            // than the value provided in a configuration file of the repo.
+            var config = new ConfigurationBuilder()
+                .AddEnvironmentVariables()
```
arroyc, line 122: confusion: are environmentvariables are provided in commandline?

kichalla: No they are normal environment variables.

PR #640, opened 2020-06-17
fix a stylecop and override error during buildscriptgenerator build

arroyc:
this fix a stylecop error during build

### DIFF
--- a/src/BuildScriptGeneratorCli/Commands/PlatformsCommand.cs
+++ b/src/BuildScriptGeneratorCli/Commands/PlatformsCommand.cs
@@ -66,6 +66,31 @@ namespace Microsoft.Oryx.BuildScriptGeneratorCli
             return ProcessConstants.ExitSuccess;
         }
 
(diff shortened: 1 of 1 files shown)


src/BuildScriptGeneratorCli/Commands/PlatformsCommand.cs
```diff
@@ -114,7 +114,7 @@ private IList<string> SortVersions(IEnumerable<string> versions)
             return result.Select(v => v.ToString()).ToList();
         }
 
-        internal override IServiceProvider GetServiceProvider(IConsole console)
+        internal IServiceProvider GetServiceProvider(IConsole console)
```
kichalla, line 117: We need to override the `TryGetServiceProvider` method here. Could you change it?

arroyc: sure .. thanks

arroyc:
@kichalla .. please take a look at the updated PR .. if looks good .. please merge it to master

PR #644, opened 2020-06-18
update oryx to dotnetcore 3.1 & add dotnetcore task to pipelines

arroyc:
This PR is to verify migrating smoothly to dotnetcore 3.1..
There are lot of dependency issues in buster with DotNetCore 2.1. DotNetCore 2.1 doesn't support libss1.1  that comes with buster. Some other compatibility issues are there as well.

This PR will enable us to avoid the compatibility issue as well as it will enable the oryx blessed buster based build images (the PR in draft)

### DIFF
--- a/src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
+++ b/src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
@@ -23,14 +23,14 @@
   <ItemGroup>
     <PackageReference Include="McMaster.Extensions.CommandLineUtils" Version="2.3.4" />
     <PackageReference Include="Microsoft.ApplicationInsights.NLogTarget" Version="2.10.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="2.2.4" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Ini" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="2.2.0" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="2.2.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.CommandLine" Version="3.1.5" />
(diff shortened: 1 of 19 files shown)


src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
-    <RuntimeFrameworkVersion>2.1</RuntimeFrameworkVersion>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
kichalla, line 4: Could you create a 'new' project in VS or dotnet cli and copy paste most of the content from there to these .csproj files...and also we have package references to 2.2 which need to be moved to 3.1 as well

vsts/pipelines/templates/_signBinary.yml
```diff
@@ -41,7 +41,12 @@ steps:
     dotnet publish -r linux-x64 -c Release src\BuildScriptGeneratorCLI\BuildScriptGeneratorCli.csproj
   displayName: 'dotnet publish and after setting git_commit and build_number as env variable'
   condition: and(succeeded(), eq(variables['setSignTypeVariable.SignType'], 'real'))
-  
+
+- task: UseDotNet@2
+  displayName: 'Use .NET Core sdk 3.1.x'
```
kichalla, line 46: this appears a bit late step for this task (before line 37 or 31 would be better)

arroyc: good catch .. we need to do it before dotnet publish

src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
arroyc: yepp .. i will do the clean up that includes the corresponding package references in dotnetcore 3.1, wanted to quickly test my theory that moving to 3.1 will work .. so i will do the changes now .. only 1 test failed .. but it will work I guess

vsts/scripts/removeDockerArtifacts.sh
```diff
@@ -27,5 +27,5 @@ echo
 if [ -d "$mountedDirs" ]; then
     echo
     echo "Cleaning up files created by test containers ..."
-    docker run -v $mountedDirs:/tempDirs oryxdevmcr.azurecr.io/public/oryx/build /bin/bash -c "rm -rf /tempDirs/* && ls /tempDirs"
+    docker run -v $mountedDirs:/tempDirs alpine /bin/sh -c "rm -rf /tempDirs/* && ls /tempDirs"
```
arroyc, line 30: i found we are downloading big and bulky build image to clean up docker artifacts when we are doing clean up in runtime image job, so I changed this to alpine .. its just 5 mb ... so even though we have to pull it it should still be faster

PR #647, opened 2020-06-22
Fix for CI build break: Removed nuget restore step

kichalla:


vsts/pipelines/templates/_signBinary.yml
```diff
@@ -28,12 +28,6 @@ steps:
     signType: '$(setSignTypeVariable.SignType)'
   condition: and(succeeded(), eq(variables['setSignTypeVariable.SignType'], 'real'))
 
-- task: NuGetCommand@2
```
kichalla, line 31: This step is not required since dotnet CLI does the restore for us automatically.

arroyc: @kichalla ... I tried some other things as well .. I saw the build broke too .. we also need to change singing agents .. VSEng-MicroBuildVS2019 .. also i recommend ... rename the branch from patch so that we can test the signing

PR #648, opened 2020-06-23
Detector: Make only few types public and the rest internal

kichalla:


src/Detector/AssemblyInfo.cs
```diff
@@ -0,0 +1,8 @@
+﻿// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.Oryx.Detector.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c96528184668742ee5bc6a1f8b4447fdc42d482d7c339989f050d7858ea7277df2f86c7ad0b8afa733987baaf6adb2858b170995d03ba3ad612bbd0b5a389e1f6392cc5ad158f726d018d9aa75c362e0350da1c6f92b75ca449591be97fc08e68c576d95aff6cfc9e58a4653e4b6e87a1d83e3060f98a6214eacfd62a45cc8bf")]
```
kichalla, line 8: To allow internal types to be accessible via tests

src/Detector/DotNetCore/DotnetCoreConstants.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Oryx.Detector.DotNetCore
 {
-    public static class DotNetCoreConstants
+    internal static class DotNetCoreConstants
```
kichalla, line 8: Throughout this PR I have made types `internal` which I felt are not required to be exposed to clients as they are internal details.

src/Common/Extensions/StringExtensions.cs
```diff
@@ -24,7 +24,10 @@ public static bool EqualsIgnoreCase(this string a, string b)
             return string.Equals(a, b, StringComparison.OrdinalIgnoreCase);
         }
 
-        public static string JoinKeyValuePairs(IDictionary<string, string> pairs, char pairSep = ' ', char kvSep = '=')
+        public static string JoinKeyValuePairs(
+            IDictionary<string, string> pairs,
+            string pairSep = " ",
```
kichalla, line 29: The change is from using `' '` to `"  "`...In this PR I made a change to Detector project where it now adds couple of files `FileExtensions.cs` and `StringExtensions.cs` as 'links' from the `Common` project. This way we need not duplicate code.
However there are some caveats with this approach. For example, since the files here get build either in 'netstandard2.0' or 'netcoreapp3.1' projects, we need to have code which only works in both projects. Here using `" "` works because it is supported in 'netstandard2.0' and since 'netcoreapp3.1' supports 'netstandard2.0', things just work.

Summary is to stick to API which is supported in 'netstandard2.0' to avoid errors

kichalla: Scratch the above comment since I noticed a problem in doing so..Since Detector is being referenced in BuildScriptGenerator too, there was a conflict since the same type name existed in both the assemblies. So I instead split the projects into multiple ones.

qianz2: Just to summarize the new changes here to make sure, so now we have split into `Oryx.Common` and `BuildScriptGenerator.Common`. `BuildScriptGenerator.Common` and `Detector ` are referencing `Oryx.Common`. `BuildScriptGenerator `and `Test.Common` are referencing `BuildScriptGenerator.Common`.

qianz2: BTW, in the VS, it doesn't show as` Oryx.Common` but as `Common`, and in the folder I can see it..so maybe need to configure csproj

kichalla: Good catch. I assigned AssemblyName to Microsoft.Oryx.Common now. Thanks.

PR #654, opened 2020-06-24
Enable dynamic install for Python preview version

qianz2:
- Fix python preview version not able to be dynamic installed since semanticVersion could not parse preview format.
- Remove 3.9 preview from full build image.
- Add dynamic installation test for preview versions.

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonPlatform.cs
+++ b/src/BuildScriptGenerator/Python/PythonPlatform.cs
@@ -434,22 +434,21 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Python
 
         private string GetMaxSatisfyingVersionAndVerify(string version)
         {
-            var versionInfo = _versionProvider.GetVersionInfo();
-            var versionMap = versionInfo.SupportedVersions;
+            var supportedVersions = SupportedVersions;
 
             // Since our semantic versioning library does not work with Python preview version format, here
             // we do some trivial way of finding the latest version which matches a given runtime version.
             // Preview version of sdks have alphabet letter in the version name. Such as '3.8.0b3', '3.9.0b1',etc.
-            var nonPreviewRuntimeVersions = versionMap.Where(v => !v.Any(c => char.IsLetter(c)));
+            var nonPreviewRuntimeVersions = supportedVersions.Where(v => !v.Any(c => char.IsLetter(c)));
             var maxSatisfyingVersion = SemanticVersionResolver.GetMaxSatisfyingVersion(
                 version,
                 nonPreviewRuntimeVersions);
-                
+
             // Check if a preview version is available
             if (string.IsNullOrEmpty(maxSatisfyingVersion))
             {
                 // Preview versions: '3.8.0b3', '3.9.0b1', etc
-                var previewRuntimeVersions = versionMap
+                var previewRuntimeVersions = supportedVersions
                     .Where(v => v.Any(c => char.IsLetter(c)))
                     .Where(v => v.StartsWith(version))
                     .OrderByDescending(v => v);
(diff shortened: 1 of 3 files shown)


src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -435,9 +435,29 @@ private string GetVirtualEnvironmentName(BuildScriptGeneratorContext context)
         private string GetMaxSatisfyingVersionAndVerify(string version)
         {
             var versionInfo = _versionProvider.GetVersionInfo();
+            var versionMap = versionInfo.SupportedVersions;
```
kichalla, line 438: Did you see your tests passing? i was suspecting this will not work since this property will _not_ have the preview versions in it since the parsing of the version `3.9.0b1` for example throws.

kichalla: Never mind. I remember now

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -435,9 +435,29 @@ private string GetVirtualEnvironmentName(BuildScriptGeneratorContext context)
...
             var maxSatisfyingVersion = SemanticVersionResolver.GetMaxSatisfyingVersion(
                 version,
-                versionInfo.SupportedVersions);
+                nonPreviewRuntimeVersions);
+                
+            // Check if a preview version is available
+            if (string.IsNullOrEmpty(maxSatisfyingVersion))
+            {
+                // Preview versions: '3.8.0b3', '3.9.0b1', etc
+                var previewRuntimeVersions = versionMap
```
kichalla, line 452: rename 'versionMap' to `supportedVersions` since it is not really a dictionary as in .net core scenario.

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -435,9 +435,29 @@ private string GetVirtualEnvironmentName(BuildScriptGeneratorContext context)
...
+                    .Where(v => v.Any(c => char.IsLetter(c)))
+                    .Where(v => v.StartsWith(version))
+                    .OrderByDescending(v => v);
```
kichalla, line 455: We need to fix the experience of `oryx platforms` too as that command's code path does not go through these lines of code. Could you handle it in the work item related to making sure preview version is supported is all platforms?

qianz2: sure, will have a separate PR for it

tests/Oryx.BuildImage.Tests/Python/PythonDynamicInstallationTest.cs
```diff
@@ -76,6 +76,50 @@ public void GeneratesScript_AndBuilds(string imageName)
                 result.GetDebugInfo());
         }
 
+        [Theory]
+        [InlineData("3.8.0b3")]
```
arroyc, line 80: remove 3.8.0b3 may be? and replace it with 3.8.3?

tests/Oryx.BuildImage.Tests/Python/PythonDynamicInstallationTest.cs
```diff
@@ -21,24 +21,20 @@ public PythonDynamicInstallationTest(ITestOutputHelper output) : base(output)
         {
         }
 
-        public static TheoryData<string> ImageNameData
+        public static IEnumerable<object[]> ImageNameData()
```
kichalla, line 24: In general strongly type objects are preferred over `object`...in this case you could have used `TheoryData<string, string>`

tests/Oryx.BuildImage.Tests/Python/PythonDynamicInstallationTest.cs
arroyc: @qianz2 3.8.3 here?

qianz2: I added it onto the other test, since this test is only testing for preview versions.

PR #656, opened 2020-06-25
Fix Python app detection

kichalla:


src/BuildScriptGenerator/Python/PythonPlatformDetector.cs
```diff
@@ -29,46 +29,52 @@ internal class PythonPlatformDetector : IPlatformDetector
...
-                    $"Oryx will try to build from '{PythonConstants.SetupDotPyFileName}'that exists in source repo");
+                Platform = PythonConstants.PlatformName,
+                PlatformVersion = versionFromRuntimeFile,
+            };
+        }
+
+        private bool IsPythonApp(ISourceRepo sourceRepo)
+        {
+            if (sourceRepo.FileExists(PythonConstants.RequirementsFileName))
+            {
```
arroyc, line 52: if we have an app where req.txt doesnt exist but setup.py exists .. we  are not detecting it as python app

tests/BuildScriptGenerator.Tests/Python/PythonPlatformDetectorTest.cs
```diff
@@ -39,10 +41,9 @@ public void Detect_ReturnsNull_WhenSourceDirectoryIsEmpty()
         }
 
         [Fact]
-        public void Detect_ReutrnsNull_WhenRequirementsFileDoesNotExist()
+        public void Detect_ReutrnsResult_WhenRequirementsFileDoesNotExist_ButDotPyFilesExist()
```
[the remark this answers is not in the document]
arroyc: let me find an example for you !

tests/BuildScriptGenerator.Tests/Python/PythonPlatformDetectorTest.cs
```diff
@@ -119,11 +124,14 @@ public void Detect_ReutrnsNull_WhenRuntimeTextFileExists_ButDoesNotTextInExpecte
             Assert.Null(result);
         }
 
-        [Fact]
-        public void Detect_ReturnsVersionFromRuntimeTextFile_IfOptionsValueIsNotPresent()
+        [Theory]
+        [InlineData("3")]
+        [InlineData("3.7")]
+        [InlineData("3.7.5")]
+        [InlineData("3.7.5b01")]
```
arroyc, line 131: do we need tests for both 3.7.5 and 3.7.5b01?

src/BuildScriptGenerator/Python/PythonPlatformDetector.cs
```diff
@@ -29,46 +29,52 @@ internal class PythonPlatformDetector : IPlatformDetector
...
-                _logger.LogInformation($"'{PythonConstants.SetupDotPyFileName} doesn't exist in source repo.' " +
-                    $"Oryx will try to build from '{PythonConstants.RequirementsFileName}'that exists in source repo");
+                _logger.LogInformation(
+                    $"Cound not find {PythonConstants.RequirementsFileName} at the root of the repo.");
             }
 
-            // This detects if a runtime.txt file exists if that is a python file
-            var versionFromRuntimeFile = DetectPythonVersionFromRuntimeFile(context.SourceRepo);
-            if (string.IsNullOrEmpty(versionFromRuntimeFile))
+            var files = sourceRepo.EnumerateFiles(PythonConstants.PythonFileNamePattern, searchSubDirectories: true);
```
arroyc, line 62: so there is no setup.py in the root ... and no req.txt in the root .. but there is app.py or something .. which successfully detects it as python app ... but I think build will fail .. can we have a test on this? even a fail test just to confirm the scenario

kichalla: Good point. Will add a test.

tests/BuildScriptGenerator.Tests/Python/PythonPlatformDetectorTest.cs
kichalla: This is just to test that our parsing of runtime text file works for all valid python versions..for example runtime.txt version format is like `python-3.5.5`

PR #658, opened 2020-06-26
Detect Blazor Wasm and write function platform to manifest

arroyc:


src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -83,6 +88,11 @@ public string GetRelativePathToProjectFile(RepositoryContext context)
                 projectFile = GetProject(azureFunctionsProjects);
             }
 
+            if (projectFile == null)
```
kichalla, line 91: Just realized we have a bug here. We need to ideally pass the combination (AddRange) of all these projects to see if there is more than 1 project that we found to build...or let's say we can look for a setting called 'ORYX_APP_TYPE' which can give us a hint on what type of app we are trying to build..if the app type is either functions or staticwebapps, then we only consider those apps

src/BuildScriptGenerator/DotNetCore/DotnetCoreConstants.cs
```diff
@@ -39,6 +39,12 @@ public static class DotNetCoreConstants
         public const string ProjectBuildPropertyKey = "project";
         public const string ProjectBuildPropertyKeyDocumentation = "Relative path of the project file to build.";
 
+        public const string AzureBlazorWasmTargetFrameWorkXPathExpression =
+            "/Project/PropertyGroup/TargetFramework";
```
kichalla, line 43: we should already have a constant for this above

src/BuildScriptGenerator/ManifestFilePropertyKeys.cs
```diff
@@ -20,5 +20,7 @@ public static class ManifestFilePropertyKeys
         internal const string PythonVersion = nameof(PythonVersion);
 
         internal const string HugoVersion = nameof(HugoVersion);
+
+        internal const string AzFunctionPlatform = "az-function-platform";
```
kichalla, line 24: We can just call it `PlatformName` and this need not be specific to azure functions

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProjectFileHelpers.cs
```diff
@@ -65,6 +70,27 @@ public static bool IsAzureFunctionsProject(XDocument projectFileDoc)
             return HasPackageReference(projectFileDoc, DotNetCoreConstants.AzureFunctionsPackageReference);
         }
 
+        public static bool IsAzureBlazorWebAssemblyProject(XDocument projectFileDoc)
```
kichalla, line 73: can you add tests for this

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -145,6 +145,11 @@ public PlatformDetectorResult Detect(RepositoryContext context)
                 return null;
             }
 
+            if (ProjectFileHelpers.IsAzureFunctionsProject(context.SourceRepo, projectFile))
+            {
+                manifestFileProperties[ManifestFilePropertyKeys.AzFunctionPlatform] = DotNetCoreConstants.PlatformName;
```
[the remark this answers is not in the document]
arroyc: I updated this

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
arroyc: I added some logic to deal with this..  please take a look

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProjectFileHelpers.cs
arroyc: sure .. I added a basic buildimage test .. but i will add a test for this as well

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -139,12 +139,20 @@ public PlatformDetectorResult Detect(RepositoryContext context)
                 manifestFileProperties[ManifestFilePropertyKeys.DotNetCoreSdkVersion] = globalJsonSdkVersion;
             }
 
+            var oryxAppTypeEnvironmentVar = Environment.GetEnvironmentVariable("Oryx_App_Type");
```
kichalla, line 142: We do not want to read environment variables directly this way because users should be able to set this via `build.env` file or environment variables...look how we do for `PROJECT` setting key for example...basically this should be accessible via options api

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreSampleAppsTest.cs
```diff
@@ -579,9 +579,52 @@ public void Builds_AzureFunctionsProject()
             var appDir = volume.ContainerDir;
             var appOutputDir = "/tmp/AzureFunctionsHttpTriggerApp-output";
             var script = new ShellScriptBuilder()
+                .SetEnvironmentVariable("Oryx_App_type", "Functions")
```
kichalla, line 582: Can you also add unit tests for the new wasm project detection in `ProjectFileHelpersTest`?

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
kichalla: Also let us validate the supplied value for app type and maybe write a warning saying it is not supported and list the supported app types.

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -75,14 +77,33 @@ public string GetRelativePathToProjectFile(RepositoryContext context)
...
 
-            projectFile = GetProject(webAppProjects);
-            if (projectFile == null)
+            // Assumption: some env variable "Oryx_App_type" will be passed to oryx
+            // which will indicate if the app is an azure function type or static type app
+            var oryxAppTypeEnvironmentVar = Environment.GetEnvironmentVariable("Oryx_App_Type");
+            if (projectFile == null && !string.IsNullOrEmpty(oryxAppTypeEnvironmentVar)
+                && oryxAppTypeEnvironmentVar.ToLower().Contains("functions"))
             {
                 projectFile = GetProject(azureFunctionsProjects);
```
kichalla, line 92: Having an AppType should not be mandatory for example to build functions projects...only if there are more than 1 projects that were found we use the app type as a hint to filter down the list..so this logic does not look correct to me.

src/BuildScriptGenerator/Hugo/HugoPlatform.cs
```diff
@@ -72,8 +72,15 @@ public PlatformDetectorResult Detect(RepositoryContext context)
             PlatformDetectorResult detectorResult)
         {
             var manifestFileProperties = new Dictionary<string, string>();
+            manifestFileProperties[ManifestFilePropertyKeys.PlatformName] = detectorResult.Platform;
             manifestFileProperties[ManifestFilePropertyKeys.HugoVersion] = detectorResult.PlatformVersion;
 
+            var oryxAppTypeEnvironmentVar = Environment.GetEnvironmentVariable("Oryx_App_Type");
```
kichalla, line 78: instead of each platform doing this should this go in the basebashbuildproperties itself..since this applies to all platforms?

src/BuildScriptGenerator/ManifestFilePropertyKeys.cs
```diff
@@ -20,5 +20,9 @@ public static class ManifestFilePropertyKeys
         internal const string PythonVersion = nameof(PythonVersion);
 
         internal const string HugoVersion = nameof(HugoVersion);
+
+        internal const string PlatformName = nameof(PlatformName);
+
+        internal const string OryxAppType = nameof(OryxAppType);
```
kichalla, line 26: nit: remove `Oryx` prefix since this is oryx manifest file...it made sense for environment variable but is not required here...

arroyc: ok

src/BuildScriptGenerator/Hugo/HugoPlatform.cs
arroyc: yeah I was just about to do this ... thanks

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreSampleAppsTest.cs
arroyc: sure

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
arroyc: I have updated the logic .. have some questions though .. lets discuss tomorrow

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -63,26 +64,85 @@ public string GetRelativePathToProjectFile(RepositoryContext context)
...
+            bool webAppProjectExists = false;
+            bool blazorWasmProjectExists = false;
+
             foreach (var file in projectFiles)
             {
                 allProjects.Add(file);
                 if (ProjectFileHelpers.IsAspNetCoreWebApplicationProject(sourceRepo, file))
                 {
-                    webAppProjects.Add(file);
+                    if (ProjectFileHelpers.IsAzureBlazorWebAssemblyProject(sourceRepo, file))
```
kichalla, line 78: hmm, this looks incorrect. Currently wasm project supports `netstandard2.0` and we look for `netcoreapp` in case of webapplication project. So if this code is nested under webapplication project, it would not be executed. Am i missing soemthing? can you can write unit tests for this scenario?

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -276,6 +276,35 @@ private void LogScriptIfGiven(string type, string scriptPath)
...
+                && context.Properties.TryGetValue(Constants.OryxAppType, out oryxAppType))
+            {
+                if (!string.IsNullOrEmpty(oryxAppType)
+                    || !string.IsNullOrWhiteSpace(oryxAppType))
+                {
+                    _logger.LogDebug($"Context Property Key {Constants.OryxAppType} with value {oryxAppType} is written into manifest");
+                    buildProperties[Constants.OryxAppType] = oryxAppType;
+                }
+            }
+            else if (!string.IsNullOrEmpty(Environment.GetEnvironmentVariable(Constants.OryxAppType))
```
kichalla, line 300: If you recall we already have a system to deal with hierarchical config and that is through .NET Core's Configuration API, so we want to stick to it and also the above code does not handle content coming from `build.env`. You can take a look at line 410 in `BuildCommand.cs` on how we do this for `RequiredOsPackages`...so basically the value set from `-p` overrides value set via environment variable which in turn overrides value set in `build.env`.

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
arroyc: yeah ... i wanted to discuss this .. i think we need to add more logic to current implementation of IsAspNetCoreWebApplicationProject which returns true if it IsOfSdkProjectType. Both blazor wasm and web apps have Microsoft.Net.Sdk.Web in their csproj

kichalla: Ok, a project can be a wasm one only if it has `Microsoft.Net.Sdk.Web` and `netstandard` prefix?

kichalla: adding to my previous comment: and `Microsoft.AspNetCore.Components.WebAssembly`

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
kichalla: `build.env` is our own file...so i am not sure if you will find an example of it...sure, let us discuss about it

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -276,6 +276,19 @@ private void LogScriptIfGiven(string type, string scriptPath)
             var copySourceDirectoryContentToDestinationDirectory = buildScriptSnippets.All(
                 snippet => snippet.CopySourceDirectoryContentToDestinationDirectory);
 
+            string oryxAppType = string.Empty;
```
kichalla, line 279: We do not need this block of code since you are getting app type from Options

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -139,12 +139,21 @@ public PlatformDetectorResult Detect(RepositoryContext context)
                 manifestFileProperties[ManifestFilePropertyKeys.DotNetCoreSdkVersion] = globalJsonSdkVersion;
             }
 
+            string oryxAppTypeValue = null;
```
kichalla, line 142: This block of code is not needed since you are getting app type from Options.

src/BuildScriptGenerator/Options/BuildScriptGeneratorOptions.cs
```diff
@@ -54,5 +54,7 @@ public class BuildScriptGeneratorOptions
         public string PostBuildScriptPath { get; set; }
 
         public string PostBuildCommand { get; set; }
+
+        public string OryxAppType { get; set; }
```
kichalla, line 58: nit: remove prefix `Oryx`

src/BuildScriptGeneratorCli/Commands/BuildCommandBase.cs
```diff
@@ -33,6 +33,12 @@ internal abstract class BuildCommandBase : CommandBase
             Description = "Comma-separated list of operating system packages that will be installed (using apt-get) before building the application.")]
         public string OsRequirements { get; set; }
 
+        [Option(
+            OptionTemplates.AppType,
+            CommandOptionType.SingleValue,
+            Description = "Application type in a multi-app repo, for example a repo that has functions api and blazor-wasm client app.")]
```
kichalla, line 39: `Type of application that the source directory has, for example: 'functions' or 'static-sites' etc.`

tests/BuildScriptGenerator.Tests/DotnetCore/ProbeAndFindProjectFileProviderTest.cs
```diff
@@ -17,6 +17,7 @@ public ProbeAndFindProjectFileProviderTest(TestTempDirTestFixture testFixture) :
         {
         }
 
+        protected BuildScriptGeneratorOptions commonOptions = new BuildScriptGeneratorOptions();
```
kichalla, line 20: looks weird. move this to each individual test that is using it.

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -63,26 +69,82 @@ public string GetRelativePathToProjectFile(RepositoryContext context)
...
+            // Assumption: some build option "--apptype" will be passed to oryx
+            // which will indicate if the app is an azure function type or static type app
+
+            // If there are multiple projects, we will look for --appty to detect corresponding project.
+            // for example azurefunction and blazor both projects can reside
+            // at the same repo, so more than 2 csprojs will be found. Now we will
+            // look for --apptype value to determine which project needs to be built
+            string oryxAppTypeEnv = string.Empty;
+
+            if (_options != null
```
kichalla, line 107: You can simplify this to `if (!string.IsNullorEmpty(_options.Value.OryxAppType))` because `_options` cannot be null since DI ensures we get an instance

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -63,26 +69,82 @@ public string GetRelativePathToProjectFile(RepositoryContext context)
...
+                else
+                {
+                    _logger.LogDebug($"Invalid value {oryxAppTypeEnv} for env:{Constants.OryxAppType}. Currently, supported values are 'functions', 'blazor-wasm', 'static-sites'");
+                }
+            }
+            else
+            {
+                // If multiple project exists, webapp gets priority if appType is not set
+                // orderwise webapp then blazor-wasm and then functions
+                if (projectFile == null && webAppProjectExists)
```
kichalla, line 133: I would change this logic like following:
1. Merge webAppProjects + azureBlazorWasmProjects + azureFunctionsProjects into one list
2. GetProject(merged list), that will throw ambiguity exception if we detected multiple projects

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -63,26 +69,82 @@ public string GetRelativePathToProjectFile(RepositoryContext context)
 
             var webAppProjects = new List<string>();
             var azureFunctionsProjects = new List<string>();
+            var azureBlazorWasmProjects = new List<string>();
             var allProjects = new List<string>();
+            bool functionProjectExists = false;
```
kichalla, line 74: These feel unnecessary...you can use `azureFunctionsProjects.Count`

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -63,26 +69,82 @@ public string GetRelativePathToProjectFile(RepositoryContext context)
...
+                }
+                else if (blazorWasmProjectExists
+                    && (oryxAppTypeEnv.ToLower().Contains("static-sites")
+                    || oryxAppTypeEnv.ToLower().Contains("blazor-wasm")))
+                {
+                    projectFile = GetProject(azureBlazorWasmProjects);
+                }
+                else
+                {
+                    _logger.LogDebug($"Invalid value {oryxAppTypeEnv} for env:{Constants.OryxAppType}. Currently, supported values are 'functions', 'blazor-wasm', 'static-sites'");
```
kichalla, line 126: This is too late for this kind of validation to happen. You need to do this in `BuildCommnd.IsValidInput`

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -63,26 +69,82 @@ public string GetRelativePathToProjectFile(RepositoryContext context)
...
+                oryxAppTypeEnv = _options.Value.OryxAppType;
+                _logger.LogInformation($"ProbeAndProjectFileProvider: {Constants.OryxAppType} is set to {oryxAppTypeEnv}");
+
+                if (functionProjectExists && oryxAppTypeEnv.ToLower().Contains("functions"))
+                {
+                    projectFile = GetProject(azureFunctionsProjects);
+                }
+                else if (blazorWasmProjectExists
+                    && (oryxAppTypeEnv.ToLower().Contains("static-sites")
+                    || oryxAppTypeEnv.ToLower().Contains("blazor-wasm")))
```
kichalla, line 120: I thought we should need only 2 app types for now: 'webapp' and 'static-sites'...webapp is for AppService's scenarios and 'static-sites' is for static webapps.

When 'static-sties' is used, the idea was to identity a maximum of 2 projects...one functions and one blazor-wasm and once if we find them we return those files and these files can then be used to publish them to different folders.

Let us chat about this today. We might want to handle this in a separate PR. I want us to at least get the part which Mitch is looking for now.

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
arroyc: imo its better to default to webapp when no value is provided to apptype. but let's discuss it

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
arroyc: yeah I know .. I was testing something and kept it here ..I will remove it from here

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -276,6 +276,17 @@ private void LogScriptIfGiven(string type, string scriptPath)
             var copySourceDirectoryContentToDestinationDirectory = buildScriptSnippets.All(
                 snippet => snippet.CopySourceDirectoryContentToDestinationDirectory);
 
+            if (_cliOptions != null
```
kichalla, line 279: revert the changes here...as it is not required

src/BuildScriptGenerator/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -63,26 +69,77 @@ public string GetRelativePathToProjectFile(RepositoryContext context)
...
+                    projectFile = GetProject(azureFunctionsProjects);
+                }
+                else if (blazorWasmProjectExists
+                    && _options.AppType.ToLower().Contains("static-sites"))
+                {
+                    projectFile = GetProject(azureBlazorWasmProjects);
+                }
+                else
+                {
+                    _logger.LogDebug($"Invalid value {_options.AppType} for env:{Constants.AppType}. Currently, supported values are 'functions', 'blazor-wasm', 'static-sites'");
```
kichalla, line 121: `env:{Constants.AppType}` => `{SettingsKeys.AppType}`...also the supported values can be put in a constants file which can referenced here and as well as in BuildCommand.IsValidInput

tests/BuildScriptGenerator.Tests/DotnetCore/ProbeAndFindProjectFileProviderTest.cs
```diff
@@ -88,7 +88,75 @@ public void GetRelativePathToProjectFile_ReturnsAzureFunctionsApp_OnlyWhenNoWebA
...
+
+            // Act
+            var actual = provider.GetRelativePathToProjectFile(context);
+
+            // Assert
+            Assert.Equal(expectedRelativePath, actual);
+        }
+
+        [Fact]
+        public void GetRelativePathToProjectFile_ReturnsAzureBlazorWasmApp_OnlyWhenOryxAppType_IsSetTo_BlazorWasm()
```
kichalla, line 133: `IsSetTo_BlazorWasm` => IsSetTo_StaticSites

tests/BuildScriptGenerator.Tests/DotnetCore/ProjectFileProviderTestBase.cs
```diff
@@ -37,6 +37,27 @@ public abstract class ProjectFileProviderTestBase : IClassFixture<TestTempDirTes
...
         ";
 
+        protected const string AzureBlazorWasmClientProjectFile = @"
+        <Project Sdk=""Microsoft.NET.Sdk.Web"">
+          <PropertyGroup>
+            <TargetFramework>netstandard2.1</TargetFramework>
+            <RazorLangVersion>3.0</RazorLangVersion>
+          </PropertyGroup>
+          <ItemGroup>
+            <PackageReference Include=""Microsoft.AspNetCore.Components.WebAssembly"" Version=""3.2.0-rc1.20223.4"" />
```
kichalla, line 47: Could you add one test case where this package reference is not there and only netstandard2.1 is present..should not be detected as wasm project

tests/BuildScriptGeneratorCli.Tests/BuildCommandTest.cs
```diff
@@ -121,6 +121,56 @@ public void IsValidInput_IsTrue_EvenIfDestinationDirExists_AndIsEmpty()
             Assert.Empty(testConsole.StdError);
         }
 
+        [Theory]
+        [InlineData("functions")]
+        [InlineData("static-sites")]
+        [InlineData("")]
+        [InlineData(null)]
+        public void IsValidInput_IsTrue_WhenApptype_Has_Valid_Input(string input)
```
kichalla, line 129: 👍

tests/BuildScriptGeneratorCli.Tests/BuildCommandTest.cs
```diff
@@ -121,6 +121,56 @@ public void IsValidInput_IsTrue_EvenIfDestinationDirExists_AndIsEmpty()
             Assert.Empty(testConsole.StdError);
         }
 
+        [Theory]
+        [InlineData("functions")]
```
kichalla, line 125: Could you add variations with different case too?

tests/BuildScriptGenerator.Tests/DotnetCore/ProjectFileProviderTestBase.cs
arroyc: ok

PR #659, opened 2020-06-26
Added hugo to detector library and updated contract of IDetector

kichalla:


tests/Detector.Tests/StaticSiteGeneratorHelperTest.cs
```diff
@@ -0,0 +1,106 @@
...
+
+using System;
+using System.IO;
+using Microsoft.Oryx.Detector.Hugo;
+using Microsoft.Oryx.Tests.Common;
+using Xunit;
+
+namespace Microsoft.Oryx.Detector.Tests
+{
+    public class StaticSiteGeneratorHelperTest : IClassFixture<TestTempDirTestFixture>
```
qianz2, line 14: Could you also add HugoDetectorTest

src/Detector/Hugo/StaticSiteGeneratorHelper.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// <param name="sourceRepo">Source repo for the application.</param>
+        /// <param name="environment">Environment abstraction.</param>
+        /// <returns>True if the app is a Hugo app, false otherwise.</returns>
+        public static bool IsHugoApp(ISourceRepo sourceRepo, HugoDetectorOptions options)
+        {
+            // Check for Hugo environment variables
+            if (options.HasHugoEnvironmentVariables)
+            {
+                return true;
+            }
```
qianz2, line 52: imo, this HugoDetectorOptions  will be within BuildScriptGenerator instead of within detector library

qianz2: Like as part of version resolving of Hugo platform.

src/Detector/Hugo/HugoConstants.cs
```diff
@@ -0,0 +1,18 @@
+// This file was auto-generated from 'constants.yaml'. Changes may be overridden.
+
+namespace Microsoft.Oryx.Detector.Hugo
+{
+    public static class HugoConstants
+    {
+        public const string Version = "0.71.0";
+        public const string PlatformName = "hugo";
+        public const string InstalledHugoVersionsDir = "/opt/hugo";
+        public const string InstallationUrlFormat = "https://github.com/gohugoio/hugo/releases/download/v#VERSION#/#TAR_FILE#";
```
qianz2, line 10: we can exclude some constants that are not used

src/Detector/DefaultPlatformDetector.cs
```diff
@@ -26,15 +26,18 @@ public class DefaultPlatformDetector : IDetector
             _detectorOptions = detectorOptions;
         }
 
-        public IDictionary<PlatformName, string> GetAllDetectedPlatforms(DetectorContext ctx)
+        public IDictionary<PlatformName, PlatformDetectorResult> GetAllDetectedPlatforms(DetectorContext context)
```
qianz2, line 29: PlatformDetectorResult already has Platform name as property, so how about make it return a list of `PlatformDetectorResult `

kichalla: Sure, yeah have been thinking of making that change to keep it consistent with what we already have in BuildScriptGenerator.

src/Detector/Hugo/StaticSiteGeneratorHelper.cs
kichalla: Hmm, why do you think HugoDetector should not have it's own options? It is possible for a detector could have it's own settings. That said I do not like that we say that an app is a Hugo app if just this environment variable exists. We should ideally remove it.

tests/Detector.Tests/StaticSiteGeneratorHelperTest.cs
kichalla: Since HugoDetector is just passing on the responsibility to `StaticSiteGeneratorHelper`, I added test for this type...I ideally want to remove this `StaticSiteGeneratorHelper` and keep all logic in HugoDetector itself.

src/Detector/Hugo/HugoConstants.cs
kichalla: I want to handle this in a separate PR. I created a new PR to update our tool which generates constants to look for multiple constants files. Once that is in, I want to create a separate Detector specific constants file.

src/Oryx.Common/Common.csproj
```diff
@@ -13,4 +13,11 @@
   </PropertyGroup>
 
   <Import Project="$(MSBuildThisFileDirectory)\..\CommonFiles\AssemblyVersion.proj" />
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.5" />
+    <PackageReference Include="Nett" Version="0.13.0" />
+    <PackageReference Include="Newtonsoft.Json" Version="11.0.2" />
+    <PackageReference Include="YamlDotNet" Version="8.0.0" />
+  </ItemGroup>
```
qianz2, line 22: Are these packages used by common proj

kichalla: I put these in Common because we still have Hugo related stuff in BuildScriptGenerator. I need them in common so that newly moved Hugo related stuff in Detector library can use these too.

PR #660, opened 2020-06-27
Added option to limit the scope of the generated C# constants file and some enhancements.

kichalla:


build/tools/SharedCodeGenerator/Outputs/CSharpConstants.cs.tpl
```diff
@@ -2,7 +2,7 @@
 
 namespace {{ Namespace }}
 {
-    public static class {{ Name }}
+    {{ Scope }} static class {{ Name }}
```
kichalla, line 5: For example in case of detector we want to limit the types which are exposed as public

build/tools/SharedCodeGenerator/Outputs/CSharpOutput.cs
```diff
@@ -45,13 +48,21 @@ public string GetPath()
 
         public string GetContent()
         {
+            var scope = "public";
```
kichalla, line 51: Tried keeping default as internal but that is breaking existing code and will cause even more changes in this PR.

build/generateConstants.sh
```diff
@@ -9,4 +9,14 @@ set -e
 declare -r REPO_DIR=$( cd $( dirname "$0" ) && cd .. && pwd )
 
 cd $REPO_DIR
-dotnet run --no-launch-profile --project $REPO_DIR/build/tools/SharedCodeGenerator/SharedCodeGenerator.csproj -- $REPO_DIR/build/constants.yaml $REPO_DIR
+
+# Get comma separated list of all constants files
+constantsFiles=$(find "$REPO_DIR/build" -type f -name '*constants.yaml' -o -name '*constants.yml')
```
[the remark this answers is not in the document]
arroyc: nit: not sure why is this required .. lets discuss this offline

build/tools/SharedCodeGenerator/Outputs/CSharpOutput.cs
arroyc: so all constants are going to be public now .. if we need to change some to internal or protected we need to manually change them after the constants are generated?

PR #665, opened 2020-07-01
Add output to file option to platforms command

qianz2:


src/BuildScriptGeneratorCli/Commands/PlatformsCommand.cs
```diff
@@ -26,6 +27,13 @@ internal class PlatformsCommand : CommandBase
         [Option("--json", Description = "Output the supported platform data in JSON format.")]
         public bool OutputJson { get; set; }
 
+        [Option(
+            "-i|--output",
```
kichalla, line 31: The output to a file was for the `build` command and not platforms command

qianz2:
> what is the scenario? why is this needed?

The app service want to add output to a file option to `Build `command, but I wrongly added it to the `platforms `command, will come up with another PR for that.

PR #668, opened 2020-07-02
Integrate Detector library into BuildScriptGenerator

kichalla:


src/Detector/DotNetCore/DotNetCoreServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+using Microsoft.Oryx.Detector.DotNetCore;
+
+namespace Microsoft.Oryx.Detector
+{
+    internal static class DotNetCoreServiceCollectionExtensions
+    {
+        public static IServiceCollection AddDotNetCoreServices(this IServiceCollection services)
+        {
+            services.AddSingleton<DotNetCoreDetector>();
+            // Factory to make sure same detector instance is returned when same implementation type is resolved via
```
kichalla, line 18: FYI..I added unit tests for verifying this.

src/Detector/Contracts/IPlatformDetector.cs
```diff
@@ -17,11 +17,5 @@ public interface IPlatformDetector
         /// <returns>An instance of <see cref="PlatformDetectorResult"/> if detection was
         /// successful, <c>null</c> otherwise</returns>
         PlatformDetectorResult Detect(DetectorContext context);
-
-        /// <summary>
-        /// Platform that the detector is used for.
-        /// </summary>
-        /// <returns>An instance of <see cref="Detector.PlatformName"/></returns>
-        PlatformName PlatformName { get; }
```
kichalla, line 25: @qianz2 since we are no longer returning a dictionary from the getalldetectedplatforms method, i removed this

src/Detector/PlatformDetectorServiceCollection.cs
```diff
@@ -21,41 +15,12 @@ public static IServiceCollection AddPlatformDetectorServices(this IServiceCollec
...
-            // Hugo
-            services.AddSingleton<HugoDetector>();
-            services.TryAddEnumerable(
-                ServiceDescriptor.Singleton<IPlatformDetector, HugoDetector>());
+            services
+                .AddDotNetCoreServices()
+                .AddNodeServices()
+                .AddPythonServices()
+                .AddPhpServices()
+                .AddHugoServices();
```
qianz2, line 23: nit: maybe better to name them as `AddHugoDetectorServices()` and so on

tests/Detector.Tests/DotnetCoreDetectorTest.cs
```diff
@@ -12,42 +12,6 @@ namespace Microsoft.Oryx.Detector.Tests.DotNetCore
 {
     public class DotNetCoreDetectorTest
```
qianz2, line 13: seems that we already have this test file within the `DotNetCore `folder

qianz2: Their content are different so probably keep this one and remove the other one

tests/Oryx.Tests.Common/TestPlatformDetector.cs
```diff
@@ -54,7 +56,9 @@ public class TestPlatformDetectorSimpleMatch : IPlatformDetector
 
         public bool DetectInvoked { get; private set; }
 
-        public PlatformDetectorResult Detect(RepositoryContext context)
+        public string PlatformName => _platformName;
```
qianz2, line 59: we could remove this and the one above since interface no longer contains it

tests/Detector.Tests/DotNetCore/RootDirectoryProjectFileProviderTest.cs
```diff
@@ -4,11 +4,10 @@
 // --------------------------------------------------------------------------------------------
 
 using System.IO;
-using Microsoft.Oryx.Detector.DotNetCore;
 using Microsoft.Oryx.Tests.Common;
 using Xunit;
 
-namespace Microsoft.Oryx.Detector.Tests.DotNetCore
+namespace Microsoft.Oryx.Detector.DotNetCore
```
qianz2, line 10: still want to keep it as Microsoft.Oryx.Detector.Test.DotNetCore, same as other test files' namespace

tests/Oryx.Tests.Common/TestProgrammingPlatform.cs
```diff
@@ -75,7 +79,7 @@ public IEnumerable<string> GetDirectoriesToExcludeFromCopyToBuildOutputDir(Build
             return Array.Empty<string>();
         }
 
-        public bool IsCleanRepo(ISourceRepo repo)
+        public bool IsCleanRepo(BuildScriptGenerator.ISourceRepo repo)
```
qianz2, line 82: nit: should we move `ISourceRepo `and `LocalSourceRepo `into `common` proj so that they're pointing to the same one?

src/Detector/Php/PhpPlatformDetectorResult.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.Oryx.Detector.Php
+{
+    public class PhpPlatformDetectorResult : PlatformDetectorResult
+    {
+        public bool ComposerFileExists { get; set; }
+
+        public bool ComposerLockFileExists { get; set; }
+
+        public IDictionary<string, string> Dependencies { get; set; }
```
qianz2, line 18: It's a good idea to add this dependencies so that users can get information through it, seems that it didn't have implementation, maybe we could add its implementation in the future PR

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -112,6 +114,14 @@ public PlatformDetectorResult Detect(RepositoryContext context)
             BuildScriptGeneratorContext context,
             PlatformDetectorResult detectorResult)
         {
+            var dotNetCorePlatformDetectorResult = detectorResult as DotNetCorePlatformDetectorResult;
+            if (dotNetCorePlatformDetectorResult == null)
+            {
+                throw new ArgumentException(
+                    $"Expected '{nameof(detectorResult)}' argument to be of type " +
+                    $"'{typeof(DotNetCorePlatformDetectorResult)}' but got '{detectorResult.GetType()}'.");
```
qianz2, line 122: Do we want to directly throw `ArgumentException ` or use more user friendly exceptions like `InvalidException`

kichalla: `InvalidUsageException` is only to expose details to end users. The exception scenario here is something that would not be triggered by an end user since this area of code is used by our own types (ex: DefaultBuildScriptGenerator).

src/Detector/Php/PhpPlatformDetectorResult.cs
kichalla: Good catch. I actually should have reverted this because initially I wanted to add 'Dependencies' here but later I realized that if we should include very 'build' specific stuff as part of detection logic itself. I know this is not an easy answer since I am sure there would be arguments for and against it. That said we need to work on designing these sub-types' structure also well since they are going to be part of the contract. For now, I will revert this change since it is not really being used.

src/Detector/Python/PythonDetector.cs
```diff
@@ -24,61 +23,52 @@ public PythonDetector(ILogger<PythonDetector> logger)
...
-                    $"Detected '{PythonConstants.RequirementsFileName}'that exists in source repo");
-            }
 
-            // This detects if a runtime.txt file exists if that is a python file
+            var isPythonApp = IsPythonApp(sourceRepo);
+
+            // This detects if a runtime.txt file exists and if that is a python file
             var versionFromRuntimeFile = DetectPythonVersionFromRuntimeFile(context.SourceRepo);
-            if (string.IsNullOrEmpty(versionFromRuntimeFile))
+            if (!isPythonApp && string.IsNullOrEmpty(versionFromRuntimeFile))
```
qianz2, line 31: Just want to double check if this logic should be: it's not a python app **or** nothing found in runtime file, then return null?

src/Detector/Php/PhpPlatformDetectorResult.cs
qianz2: I see, let's revert it and take some time to think about it and design it in the future.

src/Detector/Python/PythonDetector.cs
kichalla: No, the logic is fine...for example, you can imagine this line as following instead:
```
if (!isPythonApp)
{
    if (string.IsNullOrEmpty(versionFromRuntimeFile))
    {
        return null;
    }
}

```

tests/BuildScriptGenerator.Tests/DotnetCore/DotNetCorePlatformTest.cs
```diff
@@ -19,82 +19,41 @@ public class DotNetCorePlatformTest
...
-            var platform = CreatePlatform(projectFile: projectFile, defaultVersion: "2.2.7");
-
-            var exception = Assert.Throws<UnsupportedVersionException>(
-                () => platform.Detect(context));
-            Assert.Equal(
-                $"Platform 'dotnet' version '0.0' is unsupported. " +
-                "Supported versions: 2.2.7",
-                exception.Message);
-        }
+            var detector = CreateDetector(detectedVersion: "0.0");
```
qianz2, line 22: Looks like some tests are removed here, please make sure they're covered either in detector tests or here in the generator tests

tests/BuildScriptGenerator.Tests/Python/PythonPlatformTests.cs
```diff
@@ -235,67 +228,19 @@ public void Detect_ReturnsVersionFromOptions_EvenIfRuntimeTextFileHasVersion()
             Assert.Equal(expectedVersion, result.PlatformVersion);
         }
 
-        [Fact]
-        public void Detect_ReturnsResult_WhenOnlyMajorAndMinorVersion_AreSpecifiedInRuntimeTxtFile()
```
qianz2, line 239: same here

tests/BuildScriptGenerator.Tests/DotnetCore/DotNetCorePlatformTest.cs
kichalla: Sure, will do. But to give some context what the change I made here: Earlier this test used to depend on the detector (present in BuildScriptGenerator) to parse the `netcoreapp0.0` and then throw the exception, but now since Detector library took the responsibility of detecting the version, this test is only supplying the `detectedVersion` directly and is expecting the exception to be thrown.

tests/BuildScriptGenerator.Tests/DotnetCore/DotNetCorePlatformTest.cs
```diff
@@ -19,82 +19,41 @@ public class DotNetCorePlatformTest
...
+            var context = CreateContext();
+            var platform = CreatePlatform(
+                detector,
+                defaultVersion: "2.1.1",
+                supportedVersions: new Dictionary<string, string> { { "2.1.1", "3.1.3" } });
 
-        [Fact]
-        public void Detect_ThrowsUnsupportedException_WhenNoVersionFoundReturnsMaximumSatisfyingVersion()
```
qianz2, line 45: We still want to keep this one imo

tests/BuildScriptGenerator.Tests/Python/PythonPlatformTests.cs
kichalla: Sure, in general I agree with the concern regarding lost coverage. In this test instead of now parsing the runtime.txt file and then resolving the version, we are assuming that the PythonDetector gave us a detected version and we are getting the max satisfying version. And for this we have a test in this PR called `Detect_ReturnsResult_WhenOnlyPartialVersionIsProvidedByDetector`

tests/BuildScriptGenerator.Tests/DotnetCore/DotNetCorePlatformTest.cs
kichalla: I removed this test case because it is already covered by the test above this one.

PR #671, opened 2020-07-07
Make dynamic install root directory configurable

kichalla:


kichalla:
> I have some scenarios (edge cases), please take a look if we need to add tests for those

I do not see your comments. You missed submitting them?

tests/Oryx.BuildImage.Tests/PrepareEnvironmentCommandTest.cs
```diff
@@ -195,5 +195,42 @@ public void SkipsDetectionAndInstallsPlatformsFromFile()
...
+                $"{customDynamicInstallRootDir}/{NodeConstants.PlatformName}/{nodeVersion}")
+                .AddCommand($"source benv node={nodeVersion}")
+                .AddCommand("nodePath=$(which node)")
+                .AddCommand("echo \"Node path is: $nodePath\"")
+                .ToString();
+
+            // Act
+            var result = _dockerCli.Run(new DockerRunArguments
+            {
+                ImageId = _imageHelper.GetGitHubActionsBuildImage(),
```
arroyc, line 222: yeah .. so I would like to add a test for dynamic install custom directory for full build image as well
1. where custom root dir is /opt/node or opt/nodejs with an existing node (e.g. 10)
2. with node 14 as it is been removed

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
```diff
@@ -367,5 +367,70 @@ public void BuildsAppAfterInstallingAllRequiredPlatforms()
                 },
                 result.GetDebugInfo());
         }
+
+        [Fact]
+        public void BuildsApplication_ByDynamicallyInstallingSDKs_IntoCustomDynamicInstallationDir()
+        {
```
arroyc, line 373: can we add one test for python and php as well?

src/BuildScriptGeneratorCli/Commands/OptionTemplates.cs
```diff
@@ -20,5 +20,6 @@ internal static class OptionTemplates
         public const string Property = "-p|--property <key-value>";
         public const string ManifestDir = "--manifest-dir <directory-path>";
         public const string AppType = "--apptype <application-type>";
+        public const string DynamicInstallRootDir = "--dynamic-install-root-dir <dir-path>";
```
arroyc, line 23: DynamicInstallSdkDir instead DynamicInstallRootDir? IMO keeping root in the name may be confusing

kichalla: I wanted to mention it explicitly because there are some cases where users might want to install it in a way which benv and oryx does not understand, for example installing at /usr/bin. We cannot support it currently since we expect certain structure for our platforms.

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
kichalla: Sure, will do

tests/Oryx.BuildImage.Tests/PrepareEnvironmentCommandTest.cs
kichalla: Good scenario. Will add it.

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -424,8 +424,21 @@ internal override IServiceProvider TryGetServiceProvider(IConsole console)
             SetPlatformVersion(PlatformName, PlatformVersion);
 
             commandLineConfigSource.Set(SettingsKeys.CreatePackage, ShouldPackage.ToString());
-            commandLineConfigSource.Set(SettingsKeys.RequiredOsPackages, OsRequirements);
-            commandLineConfigSource.Set(SettingsKeys.AppType, AppType);
+
+            if (!string.IsNullOrEmpty(OsRequirements))
+            {
+                commandLineConfigSource.Set(SettingsKeys.RequiredOsPackages, OsRequirements);
```
gpcastro, line 430: nit: add a `SetIfNotEmpty` to avoid repeating all these ifs?

kichalla: Sure, will do

images/build/benv.sh
```diff
@@ -95,8 +128,9 @@ benv-showSupportedVersionsErrorInfo() {
 benv-getPlatformDir() {
```
gpcastro, line 128: add some documentation for these functions?

PR #672, opened 2020-07-08
Make kitchensink, vso image base image to be same as github runners's base digest

kichalla:


images/build/Dockerfiles/ltsVersions.Dockerfile
```diff
@@ -7,7 +7,7 @@ ARG SDK_STORAGE_BASE_URL_VALUE
 # Determine where the image is getting built (DevOps agents or local)
 ARG AGENTBUILD
 
-FROM buildpack-deps:stretch AS main
+FROM githubrunners-buildpackdeps-stretch AS main
```
kichalla, line 10: This is so that the base layers are shared between the full build image and github-actions image in case of AppService scenario (since they are going to adopt dynamic install iteratively)

PR #674, opened 2020-07-09
Add detect cli command

qianz2:


src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+    [Command(Name, Description = "Detect all platforms and versions in the given app source directory.")]
+    internal class DetectCommand : CommandBase
+    {
+        public const string Name = "detect";
+
+        [Argument(0, Description = "The source directory. If no value is provided, the current directory is used.")]
+        [DirectoryExists]
+        public string SourceDir { get; set; }
+
+        [Option("--json", Description = "Output the detected platform data in JSON format.")]
```
gpcastro, line 30: what is the format if this is not provided? We should probably use --output=json instead, so we can add more formats later and prevent conflicts like both `--json` and say `--table` (or any other format) be passed at the same time

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+                    {
+                        console.WriteLine("Detection result:");
+                        console.WriteLine(ListFormatResult(detectedPlatforms));
+                    }
+
+                    return ProcessConstants.ExitSuccess;
+                }
+            }
+
+            return ProcessConstants.ExitFailure;
```
gpcastro, line 64: if we don't detect anything we fail? Why not just return nothing?

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+            var logger = loggerFactory.CreateLogger<DetectCommand>();
+            var sourceRepo = new LocalSourceRepo(SourceDir, loggerFactory);
+            var ctx = new DetectorContext
+            {
+                SourceRepo = sourceRepo,
+            };
+
+            var detector = serviceProvider.GetRequiredService<IDetector>();
+            var detectedPlatforms = detector.GetAllDetectedPlatforms(ctx);
+            using (var timedEvent = logger.LogTimedEvent("DetectCommand"))
```
gpcastro, line 45: consider using it without braces (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/using-statement)

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                if (detectedPlatforms != null && detectedPlatforms.Any())
+                {
+                    if (OutputJson)
+                    {
+                        console.WriteLine("Detection result in Json format:");
```
gpcastro, line 51: I'd say we shouldn't print anything other than the JSON itself so that whoever is calling it can just pipe the command output to a json parser

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+                SourceRepo = sourceRepo,
+            };
+
+            var detector = serviceProvider.GetRequiredService<IDetector>();
+            var detectedPlatforms = detector.GetAllDetectedPlatforms(ctx);
+            using (var timedEvent = logger.LogTimedEvent("DetectCommand"))
+            {
+                if (detectedPlatforms != null && detectedPlatforms.Any())
+                {
+                    if (OutputJson)
```
gpcastro, line 49: if nothing is detected, we could return an empty JSON array or object to facilitate the parsing by the caller

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+                logger?.LogError("Could not find the source directory.");
+                console.WriteErrorLine($"Could not find the source directory: '{SourceDir}'");
+
+                return false;
+            }
+
+            return true;
+        }
+
+        private string ListFormatResult(IEnumerable<PlatformDetectorResult> detectedPlatforms)
```
gpcastro, line 83: instead of adding it to string builds, parse it into a string, to then print it, how about call it `PrintTableResult` for example and just write directly to the console?

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
qianz2: If the format is not provided, then by default it'll simply print the detection results as a list to console. Users could choose only one of the format, so it won't be cases that we having both formats passed in at the same time? yeah agree, we could do this --output=json instead of having them separately, will change it

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
qianz2: Yeah good point, will remove this line.

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
qianz2: If nothing detected, we'll print out some warning/logging for users instead, will change it.

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+            var loggerFactory = serviceProvider.GetRequiredService<ILoggerFactory>();
+            var logger = loggerFactory.CreateLogger<DetectCommand>();
+            var sourceRepo = new LocalSourceRepo(SourceDir, loggerFactory);
+            var ctx = new DetectorContext
+            {
+                SourceRepo = sourceRepo,
+            };
+
+            var detector = serviceProvider.GetRequiredService<IDetector>();
+            var detectedPlatforms = detector.GetAllDetectedPlatforms(ctx);
```
kiranongithub, line 44: `var detectedResults` or `var detectedPlatformResults`

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+            using (var timedEvent = logger.LogTimedEvent("DetectCommand"))
+            {
+                if (detectedPlatforms != null && detectedPlatforms.Any())
+                {
+                    if (OutputJson)
+                    {
+                        console.WriteLine("Detection result in Json format:");
+                        console.WriteLine(JsonFormatResult(detectedPlatforms));
```
kiranongithub, line 52: Is this json format indented? it would be nice to be in indented format

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+                defs.AddDefinition("Platform", detectedPlatform.Platform);
+                defs.AddDefinition("Version", detectedPlatform.PlatformVersion ?? "N/A");
+
+                result.AppendLine(defs.ToString());
+            }
+
+            return result.ToString();
+        }
+
+        private string JsonFormatResult(IEnumerable<PlatformDetectorResult> detectedPlatforms)
```
kiranongithub, line 100: change the variable name here too

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            var detectionResult = new Dictionary<string, string>
+            {
+                ["App_Path"] = SourceDir,
+            };
+            var platformData = new List<Dictionary<string, string>>();
+            foreach (var detectedPlatform in detectedPlatforms)
+            {
+                Dictionary<string, string> data = new Dictionary<string, string>
```
kiranongithub, line 109: Since a detector result could have more information than just the name and version, we should not be doing this. To make it simpler we could just serialize the list of platform detector result itself. In my opinion things like `App_Path` are not interested to be put into the output.

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,123 @@
...
+                {
+                    ["Platform_Name"] = detectedPlatform.Platform,
+                    ["Platform_Version"] = detectedPlatform.PlatformVersion,
+                };
+                platformData.Add(data);
+            }
+
+            var jsonPlatformData = JsonConvert.SerializeObject(platformData, Formatting.Indented);
+            detectionResult["Platform_Data"] = jsonPlatformData;
+            var json = JsonConvert.SerializeObject(detectionResult, Formatting.Indented);
```
kiranongithub, line 119: 👍 i see the output is being indented.

tests/BuildScriptGeneratorCli.Tests/DetectCommandTest.cs
```diff
@@ -0,0 +1,255 @@
...
+
+            // Assert
+            Assert.False(isValidInput);
+            Assert.Contains(
+                $"Could not find the source directory",
+                testConsole.StdError);
+        }
+
+        [Fact]
+        public void Execute_OutputsNodePlatformAndNullVersion_WhenJsonFileIsEmpty()
```
kiranongithub, line 71: This name does not sound right...here the version is not detected..similarly fix other test names too

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
qianz2: Good catch, I forgot that the fact that detector results could have more properties, will fix this

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+        }
+
+        private void PrintJsonResult(IEnumerable<PlatformDetectorResult> detectedPlatformResults, IConsole console)
+        {
+            if (detectedPlatformResults == null || !detectedPlatformResults.Any())
+            {
+                var platformData = new List<Dictionary<string, string>>();
+                Dictionary<string, string> data = new Dictionary<string, string>
+                {
+                    ["Platform"] = "Not Detected",
```
kichalla, line 121: I think we do not need to print these if nothing is detected. We can just print an `{}` indicating there were no platforms detected.

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+                else
+                {
+                    PrintTableResult(detectedPlatformResults, console);
+                }
+
+                return ProcessConstants.ExitSuccess;
+            }
+        }
+
+        internal override bool IsValidInput(IServiceProvider serviceProvider, IConsole console)
```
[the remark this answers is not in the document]
qianz2: I think we validated the outputs within DetectCommandTest, are you saying validate the console outputs or else?

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
qianz2: I see, we could do this in JSON format. And if it's table format, we still want to output some messages like 'Not Detected' imo.

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -0,0 +1,137 @@
...
+    internal class DetectCommand : CommandBase
+    {
+        public const string Name = "detect";
+
+        [Argument(0, Description = "The source directory. If no value is provided, the current directory is used.")]
+        [DirectoryExists]
+        public string SourceDir { get; set; }
+
+        [Option(
+            "--output",
```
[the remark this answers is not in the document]
qianz2: sorry didn't notice these comments, will come up with another PR for addressing these

qianz2:
> a simple test?

yes it has some unit tests attached.

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
kichalla: Sure that sounds fine to me.

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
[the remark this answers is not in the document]
kichalla: I meant if a user types in `oryx detect -o xml`, we should validate and show an error message since we do not support that format

PR #676, opened 2020-07-09
Added tests for Detector nuget package

kichalla:


build/testDetector.sh
```diff
@@ -11,15 +11,42 @@ declare -r REPO_DIR=$( cd $( dirname "$0" ) && cd .. && pwd )
 # Load all variables
 source $REPO_DIR/build/__variables.sh
 
+# A hint to Detector.csproj file to generate the nuget package in a specific way so that
```
kichalla, line 14: Will explain in meeting as to why we are doing this

src/Detector/Detector.csproj
```diff
@@ -35,18 +35,26 @@
     <PackageProjectUrl>https://github.com/microsoft/oryx</PackageProjectUrl>
     <PackageTags>Oryx Detector OryxDetector</PackageTags>
     <Description>A library to detect tools and frameworks used in an application.</Description>
+    <Version>1.0.0-pre-</Version>
```
gpcastro, line 38: can we just have latest here?

tests/Oryx.Tests.Common/TestTempDirTestFixture.cs
```diff
@@ -6,7 +6,7 @@
 using System;
 using System.IO;
 
-namespace Microsoft.Oryx.Tests.Common
+namespace Microsoft.Oryx.Detector.Tests
```
gpcastro, line 9: why the rename?

kichalla: This was by mistake. I will revert it actually.

src/Detector/Detector.csproj
kichalla: Prerelease packages on NuGet.org typically start with `pre`, so I am using that convention here.

PR #684, opened 2020-07-14
Add validate output format in detect cli

qianz2:


src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -118,8 +134,8 @@ private void PrintJsonResult(IEnumerable<PlatformDetectorResult> detectedPlatfor
                 var platformData = new List<Dictionary<string, string>>();
                 Dictionary<string, string> data = new Dictionary<string, string>
                 {
-                    ["Platform"] = "Not Detected",
-                    ["PlatformVersion"] = "Not Detected",
+                    ["Platform"] = "{}",
```
kichalla, line 137: you do not need to even use the serializer in this case, you can just print out `{}` or if you want indented output you can just new up the dictionary without setting these keys and values and when the serializer does it's job it should print `{ }`...does it not do that?

qianz2: I didn't try that actually, yeah will try it and see if I can just make it as empty

PR #685, opened 2020-07-14
Clean up some duplicates from BuildScriptGenerator

qianz2:


src/BuildScriptGenerator/Hugo/StaticSiteGeneratorHelper.cs
```diff
@@ -112,11 +113,15 @@ public static bool IsHugoApp(ISourceRepo sourceRepo, IEnvironment environment)
             foreach (var path in yamlFilePaths)
             {
                 var yamlNode = ParserHelper.ParseYamlFile(sourceRepo, path);
-                if (yamlNode.Children.Keys
-                    .Select(k => k.ToString())
-                    .Any(k => HugoConfigurationVariables.Contains(k, StringComparer.OrdinalIgnoreCase)))
+                var yamlMappingNode = yamlNode as YamlMappingNode;
+                if (yamlMappingNode != null)
```
qianz2, line 117: If it is not a yamlMappingNode but a yamlSequenceNode, we probably still want to look into the sequence and get document for each item?

src/BuildScriptGenerator/Helpers/ParserHelper.cs
```diff
@@ -29,18 +29,18 @@ public static TomlTable ParseTomlFile(ISourceRepo sourceRepo, string filePath)
...
-        /// Parse a .yaml file into a YamlMappingNode from the YamlDotNet library.
+        /// Parse a .yaml file into a YamlNode from the YamlDotNet library.
         /// See https://github.com/aaubry/YamlDotNet for more information.
         /// </summary>
         /// <param name="sourceRepo">Source repo for the application.</param>
         /// <param name="filePath">The path to the .yaml file.</param>
-        /// <returns>A YamlMappingNode object containing information about the .yaml file.</returns>
-        public static YamlMappingNode ParseYamlFile(ISourceRepo sourceRepo, string filePath)
+        /// <returns>A YamlNode object containing information about the .yaml file.</returns>
+        public static YamlNode ParseYamlFile(ISourceRepo sourceRepo, string filePath)
```
kichalla, line 38: Ah can you check if this helper is indeed used in BuildScriptGenerator since i copied it to Detector too. If it is not used we can delete this file.

qianz2: I found this is no longer used so deleted it, thanks for reminding, and also removed projectFileProviders related from generator since we already have that in detector.

PR #686, opened 2020-07-14
update global node module path

arroyc:
still working on a test which can will output something in standard output randomly before we run our run script

### DIFF
--- a/src/startupscriptgenerator/src/node/scriptgenerator.go
+++ b/src/startupscriptgenerator/src/node/scriptgenerator.go
@@ -74,7 +74,8 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 
 	scriptBuilder.WriteString("\n# Enter the source directory to make sure the script runs where the user expects\n")
 	scriptBuilder.WriteString("cd \"" + gen.SourcePath + "\"\n\n")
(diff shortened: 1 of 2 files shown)


src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -74,7 +74,8 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 
 	scriptBuilder.WriteString("\n# Enter the source directory to make sure the script runs where the user expects\n")
 	scriptBuilder.WriteString("cd \"" + gen.SourcePath + "\"\n\n")
-	scriptBuilder.WriteString("export NODE_PATH=$(npm root --quiet -g):$NODE_PATH\n")
+	globalNodeModulesDir := consts.NodeGlobalModulesPath
```
kichalla, line 77: you missed running the `generateConstants.sh` file? Also should we have a test which runs on all runtime images...basically in the test you would run `$(npm root --quiet -g)` to confirm if the global modules is the expected one. I suspect if they would really change it so probably it is an overkill. Ignore it.

arroyc: i saw we already have this constant, so did not run the generateconstants.sh ... i will add a test as you suggested

PR #688, opened 2020-07-15
Provide option to restrict detection of apps only to the root directory

kichalla:


arroyc:
Can we also have a simple integration test with an app similar to mitch's example, so that it will be easier to follow up in case any changes need to done later?

kichalla:
> Can we also have a simple integration test with an app similar to mitch's example, so that it will be easier to follow up in case any changes need to done later?

@arroyc I updated the PR based on what we all discussed. Take a look again if you want to.

PR #689, opened 2020-07-16
Update .NET Core versions

kichalla:


arroyc:
can you also kick off a nightly build? I will also kick off one for  my pr

PR #690, opened 2020-07-16
update node 10, 12 and python 3.8

arroyc:
I updated the node and python in build image, also we need to update npm, our npm is very old


platforms/nodejs/versionsToBuild.txt
```diff
@@ -36,10 +37,11 @@
 12.14.1
 12.16.0
 12.16.1
+12.18.2
```
kichalla, line 40: Remove this

platforms/nodejs/versionsToBuild.txt
```diff
@@ -27,6 +27,7 @@
 10.16.3
 10.18.0
 10.18.1
+10.21.0
```
kichalla, line 30: Remove this

qianz2:
will be great to add them to the existing tests

PR #693, opened 2020-07-17
Enable setting to be applied through command line

kichalla:


src/Detector/Python/PythonDetector.cs
```diff
@@ -66,6 +66,11 @@ private bool IsPythonApp(ISourceRepo sourceRepo)
             }
 
             var searchSubDirectories = !_options.DisableRecursiveLookUp;
+            if (!searchSubDirectories)
+            {
+                _logger.LogDebug("Skipping search for files in sub-directories as it has been disabled.");
```
qianz2, line 71: add same logging for hugo, node, and php?

kichalla: They currently do not do a recursive look up so did not need to add it there

PR #694, opened 2020-07-17
Enable choosing the image that should be built

kichalla:


build/buildBuildImages.sh
```diff
@@ -48,6 +48,34 @@ fi
 storageArgs="--build-arg SDK_STORAGE_ENV_NAME=$SDK_STORAGE_BASE_URL_KEY_NAME"
 storageArgs="$storageArgs --build-arg SDK_STORAGE_BASE_URL_VALUE=$PROD_SDK_CDN_STORAGE_BASE_URL"
 
+# https://medium.com/@Drew_Stokes/bash-argument-parsing-54f3b81a6a8f
+PARAMS=""
+while (( "$#" )); do
+  case "$1" in
+    -t|--type)
```
kichalla, line 55: I do not think we will use this switch in CI. As of now it is meant to make local development easy

kichalla:
> can you trigger a nightly build

Sure, sounds good

PR #695, opened 2020-07-18
Make detector result immutable and enable to get property info list

qianz2:


src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -91,8 +91,7 @@ public PlatformDetectorResult Detect(RepositoryContext context)
             }
 
             var version = ResolveVersion(detectionResult.PlatformVersion);
-            detectionResult.PlatformVersion = version;
-            return detectionResult;
+            return new PlatformDetectorResult(detectionResult.Platform, version);
```
kichalla, line 94: I was hoping for this work item to get triaged before we start to implement. For my perspective having this immutable is not of a big concern. The more richer we make these detector results, this constructor pattern will not be friendly.

src/Detector/DotNetCore/DotNetCorePlatformDetectorResult.cs
```diff
@@ -12,11 +14,38 @@ namespace Microsoft.Oryx.Detector.DotNetCore
...
+            : base(platform, platformVersion)
+        {
+            this.ProjectFile = projectFile;
+        }
+
+        /// <summary>
+        /// Gets pairs of property names and values from the DotNetCorePlatformDetectorResult.
+        /// </summary>
+        /// <returns></returns>
+        public Dictionary<string, string> GetPlatformDetectorResultPropertyInfos()
```
gpcastro, line 40: we should return `IReadOnlyDictionary` instead to make it immutable. Also, why can't we return an object here instead of a dictionary?

kichalla: +1 on what @gpcastro said. We do not need this method and should instead have plain properties.

PR #697, opened 2020-07-20
Handle InvalidProjectFileException from BuildScriptGenerator

qianz2:


src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -80,19 +80,27 @@ public IEnumerable<string> SupportedVersions
...
 
-            if (detectionResult == null)
+                var version = ResolveVersion(detectionResult.PlatformVersion);
+                detectionResult.PlatformVersion = version;
+                return detectionResult;
+            }
+            catch (InvalidProjectFileException e)
             {
-                return null;
+                _logger.LogError(e, "InvalidProjectFileException caught: Either could not find the .NET Core project file" +
```
kichalla, line 100: `LogError` should have an overload which takes in an exception, so for the message part you can just say like 'error occurred while detecting for .net core application(s)'.

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -80,19 +80,27 @@ public IEnumerable<string> SupportedVersions
...
+                    "or found ambiguity in selecting a project to build.");
+                throw new InvalidUsageException($"Invalid project file exception caught from IDotNetCorePlatformDetector. ");
```
kichalla, line 102: Note that the message that we put in `InvalidUsageException` should be something which can be acted upon by an end user...You can just assign the message of the `InvalidProjectFileException` here.

PR #698, opened 2020-07-24
Enable detectors to output directories and frameworks

qianz2:


src/Detector/Contracts/IDetector.cs
```diff
@@ -16,7 +16,7 @@ public interface IDetector
         /// Returns all platforms detected for the given repository context.
         /// </summary>
         /// <param name="context">The <see cref="DetectorContext"/>.</param>
-        /// <returns>A dictionary mapping the detected platforms to the version.</returns>
+        /// <returns>A list of platform detection results that contains detected platforms, versions and directories.</returns>
```
kichalla, line 19: As we keep adding richer metadata this comment might not be a place to put all that information.

src/Detector/Hugo/HugoDetector.cs
```diff
@@ -42,22 +42,24 @@ public HugoDetector(ILogger<HugoDetector> logger)
...
         public PlatformDetectorResult Detect(DetectorContext context)
         {
-            var isHugoApp = IsHugoApp(context.SourceRepo);
+            var isHugoApp = IsHugoApp(context.SourceRepo, out string directory);
             if (isHugoApp)
             {
                 return new PlatformDetectorResult
                 {
                     Platform = HugoConstants.PlatformName,
+                    Directory = directory,
```
kichalla, line 51: The name `Directory` does not really tell much. Can it be called `AppDirectory`?

src/Detector/Constants.cs
```diff
@@ -9,5 +9,6 @@ internal static class Constants
     {
         public const string FunctionApplications = "functions";
         public const string StaticSiteApplications = "static-sites";
+        public const string RelativeRootDirectory = "./";
```
kichalla, line 12: Does it need to be like this? can it be like "src/webapp" instead?

src/Detector/Hugo/HugoDetector.cs
```diff
@@ -100,6 +102,7 @@ private bool IsHugoApp(ISourceRepo sourceRepo)
                 {
                     if (IsHugoTomlFile(sourceRepo, tomlFile))
                     {
+                        directory = RelativeDirectoryHelper.GetRelativeDirectoryToRoot(tomlFile, sourceRepo.RootPath);
```
kichalla, line 105: What would be the app directory in this case? It should still be the parent directory of the `config` directory.

src/Detector/Constants.cs
kichalla: Actually take a look at how `System.IO.Path.GetRelativePath` behaves here...I think it would return `.`

src/Detector/Hugo/HugoDetector.cs
qianz2: sure, will change it to `AppDirectory`

src/Detector/Constants.cs
qianz2: Noticed for .NET standard 2.0, it doesn't have get relative path method. For the relative path, my understanding is, if a root repo path is `C:/apps` and absolute app directory is `C:/apps/src/webapp`, then it will output `./src/webapp`. And windows and linux expression are different, windows is like `.\src\webapps`, so we'll just nomalize it to be `./src/webapp`.

src/Detector/Hugo/HugoDetector.cs
qianz2: I see, made changes to it

src/Detector/Node/NodeDetector.cs
```diff
@@ -151,5 +149,57 @@ private dynamic ReadJsonObjectFromFile(ISourceRepo sourceRepo, string fileName)
...
         }
+
+        private IEnumerable<FrameworkInfo> DetectFrameworkInfos(DetectorContext context)
+        {
+            var detectedFrameworkResult = new List<FrameworkInfo>();
+            var packageJson = GetPackageJsonObject(context.SourceRepo, _logger);
+            if (packageJson?.devDependencies != null)
+            {
+                var devDependencyObject = packageJson?.devDependencies;
+                foreach (var KeyWordToName in NodeConstants.DevDependencyFrameworkKeyWordToName)
```
kichalla, line 160: nit: KeyWordToName => keyWordToName

src/Detector/Node/NodeDetector.cs
```diff
@@ -151,5 +149,57 @@ private dynamic ReadJsonObjectFromFile(ISourceRepo sourceRepo, string fileName)
...
+                {
+                    var keyword = KeyWordToName.Key;
+                    if (devDependencyObject[keyword] != null)
```
kichalla, line 163: Would this check throw if the key is not present? Also can you check if this does a case-INsensitive way of lookup?

src/Detector/Node/NodePlatformDetectorResult.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.Oryx.Detector.Node
+{
+    /// <summary>
+    /// Represents a <see cref="PlatformDetectorResult"/> returned by the <see cref="NodeDetector"/> and
+    /// contains additional information related to the detected applications.
+    /// </summary>
+    public class NodePlatformDetectorResult : PlatformDetectorResult
+    {
+
```
kichalla, line 16: nit: remove extra space

src/Detector/Node/NodeDetector.cs
qianz2: It will throw exception if no key found, added the catch exception here. And I checked it's case insensitive by default.

PR #701, opened 2020-07-28
Fix checkers to avoid throwing exceptions unnecessarily and some clean-up

kichalla:


src/BuildScriptGenerator/Node/NodeDependencyChecker.cs
```diff
@@ -30,10 +30,11 @@ public NodeDependencyChecker(ILogger<NodeDependencyChecker> logger)
 
         public IEnumerable<ICheckerMessage> CheckSourceRepo(ISourceRepo repo)
         {
-            dynamic packageJson = NodePlatform.GetPackageJsonObject(repo, null);
+            dynamic packageJson = NodePlatform.GetPackageJsonObject(repo, _logger);
```
kichalla, line 33: This was causing exceptions to be thrown because `GetPackageJsonObject` tries writing to a log.

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -402,6 +404,11 @@ public string ResolveVersion(string versionToResolve)
         /// <returns>Package json Object.</returns>
         internal static dynamic GetPackageJsonObject(ISourceRepo sourceRepo, ILogger logger)
         {
+            if (!sourceRepo.FileExists(NodeConstants.PackageJsonFileName))
```
kichalla, line 407: Noticed quite a few filenotfound exception for this file, so adding this check to avoid that.

src/BuildScriptGeneratorCli/Commands/BuildCommandBase.cs
```diff
@@ -36,7 +36,7 @@ internal abstract class BuildCommandBase : CommandBase
         [Option(
             OptionTemplates.AppType,
             CommandOptionType.SingleValue,
-            Description = "Type of application that the source directory has, for example: 'functions' or 'static-sites' etc.")]
+            Description = "Type of application thatheck the source directory has, for example: 'functions' or 'static-sites' etc.")]
```
qianz2, line 39: nit: space

PR #705, opened 2020-07-30
update appinsights version to 1.8.3

arroyc:
update appinsights to 1.8.3


arroyc:
need to update the node base image tag

PR #706, opened 2020-07-30
patch to fix appinsights issue in ant-88 by updating appinsights 

arroyc:


arroyc:
need to update the node base image once it's done

arroyc:
closing for #707

PR #708, opened 2020-07-31
Reduce build images layers

kichalla:


build/buildBuildImages.sh
```diff
@@ -36,17 +29,9 @@ else
...
-	buildMetadataArgs="$buildMetadataArgs --build-arg BUILD_NUMBER=$BUILD_NUMBER"
-	buildMetadataArgs="$buildMetadataArgs --build-arg RELEASE_TAG_NAME=$RELEASE_TAG_NAME"
-	echo "Build metadata args: $buildMetadataArgs"
-fi
-
-storageArgs="--build-arg SDK_STORAGE_ENV_NAME=$SDK_STORAGE_BASE_URL_KEY_NAME"
-storageArgs="$storageArgs --build-arg SDK_STORAGE_BASE_URL_VALUE=$PROD_SDK_CDN_STORAGE_BASE_URL"
+# NOTE: We are using only one label here and put all information in it 
+# in order to limit the number of layers that are created
+labelContent="git_commit=$GIT_COMMIT, build_number=$BUILD_NUMBER, release_tag_name=$RELEASE_TAG_NAME"
```
kichalla, line 34: A hack to combine 3 labels into 1. Since we want capture the state and apply to an image, it does not really matter for us if that is put as 3 labels or just 1 label as long as we can understand it.

build/buildBuildImages.sh
```diff
@@ -171,120 +87,190 @@ function createImageNameWithReleaseTag() {
...
+		.
+}
+
+function buildTemporaryFilesImage() {
+	buildGitHubRunnersBaseImage
+	
+	# Create the following image so that it's contents can be copied to the rest of the images below
+	echo
+	echo "-------------Creating temporary files image-------------------"
+	docker build -t tempfiles \
```
kichalla, line 122: An image used to store docker build context (ex: images, build directories which contain scripts etc.) which is used in building images. This also helped in reducing the image layers by avoiding us to pass in `ARG` statements which was causing layers to be added.

build/buildBuildImages.sh
```diff
@@ -171,120 +87,190 @@ function createImageNameWithReleaseTag() {
...
 function buildBuildScriptGeneratorImage() {
+	buildTemporaryFilesImage
+
 	# Create the following image so that it's contents can be copied to the rest of the images below
 	echo
 	echo "-------------Creating build script generator image-------------------"
 	docker build -t buildscriptgenerator \
 		--build-arg AGENTBUILD=$BUILD_SIGNED \
-		$buildMetadataArgs \
+		--build-arg GIT_COMMIT=$GIT_COMMIT \
```
kichalla, line 135: Need not worry about these args as the image being built here is an intermediate image only.

build/buildBuildImages.sh
```diff
@@ -171,120 +87,190 @@ function createImageNameWithReleaseTag() {
...
+	# Do not write this image tag to the artifacts file as we do not intend to push it
+	testImageTag="$ORYXTESTS_BUILDIMAGE_REPO:lts-versions"
+	docker build -t $testImageTag -f "$ORYXTESTS_LTS_VERSIONS_BUILDIMAGE_DOCKERFILE" .
 }
 
 function buildFullImage() {
 	buildLtsVersionsImage
 
+	# Pull and tag the image with the name that this image's Dockerfile expects
+	yarnImage="mcr.microsoft.com/oryx/base:build-yarn-cache-$YARN_CACHE_BASE_TAG"
```
kichalla, line 227: One more hack (or workaround) to avoid the ARG usage and hence layer

build/buildBuildImages.sh
```diff
@@ -171,120 +87,190 @@ function createImageNameWithReleaseTag() {
...
+
 	echo
 	echo "-------------Creating full build image-------------------"
-	buildDockerImage "$BUILD_IMAGES_DOCKERFILE" \
-					"$ACR_BUILD_IMAGES_REPO" \
-					"$ORYXTESTS_BUILDIMAGE_DOCKERFILE" \
-					"$ORYXTESTS_BUILDIMAGE_REPO" \
-					"$DEVBOX_BUILD_IMAGES_REPO"
+	builtImageTag="$ACR_BUILD_IMAGES_REPO"
+	# NOTE: do not pass in label as it is inherited from base image
```
kichalla, line 236: note this

images/build/Dockerfiles/Dockerfile
```diff
@@ -1,40 +1,38 @@
...
+# End build run script generators
+###
 
 # Install .NET Core
-FROM main AS dotnet-install
-ARG BUILD_DIR
-ARG IMAGES_DIR
-RUN apt-get update \
+FROM main AS pre-final
+COPY --from=tempfiles /tmp/oryx/ /opt/tmp
```
kichalla, line 29: Instead of multiple intermediate images (like dotnet-install, node-install), we have only 1 now and we try to copy all info to `/opt` folder because this allows us to use only 1 COPY instruction (hence only 1 layer) when copying from this intermediate stage to final stage. There is perf cost here but unavoidable in my opinion.

images/build/Dockerfiles/Dockerfile
```diff
@@ -44,7 +42,7 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 # Check https://www.microsoft.com/net/platform/support-policy for support policy of .NET Core versions
-
+ENV NUGET_PACKAGES=/opt/nuget
```
kichalla, line 45: Target everything to go under `/opt` folder in this intermediate stage and in final image we copy back from `/opt/nuget` folder to `/var/nuget` folder and also set the environment variable appropriately.

images/build/Dockerfiles/gitHubActions.Dockerfile
```diff
@@ -87,63 +63,50 @@ RUN set -ex \
...
-# Bake Application Insights key from pipeline variable into final image
-ARG AI_KEY
-ENV ORYX_AI_INSTRUMENTATION_KEY=${AI_KEY}
-
-ENV ENABLE_DYNAMIC_INSTALL=true
-ENV ${SDK_STORAGE_ENV_NAME} ${SDK_STORAGE_BASE_URL_VALUE}
+    && rm -rf /var/lib/apt/lists/* \
+    && rm -rf $tmpDir
+
+# Docker has an issue with variable expansion when all are used in a single ENV command.
```
kichalla, line 93: note this

arroyc:
what is the size now? did  layer reduction help reducing image size as well?

images/build/Dockerfiles/gitHubActions.Dockerfile
```diff
@@ -87,63 +63,50 @@ RUN set -ex \
...
-ARG BUILD_NUMBER=unspecified
-ARG RELEASE_TAG_NAME=unspecified
-LABEL com.microsoft.oryx.git-commit=${GIT_COMMIT}
-LABEL com.microsoft.oryx.build-number=${BUILD_NUMBER}
-LABEL com.microsoft.oryx.release-tag-name=${RELEASE_TAG_NAME}
+ENV LANG="C.UTF-8" \
+    ORIGINAL_PATH="$PATH" \
+    PATH="$ORYX_PATHS:$PATH" \
+    NUGET_XMLDOC_MODE="skip" \
+	DOTNET_SKIP_FIRST_TIME_EXPERIENCE="1" \
```
arroyc, line 106: spaces

images/build/Dockerfiles/temporaryFiles.Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM githubrunners-buildpackdeps-stretch
+ARG IMAGES_DIR=/tmp/oryx/images
+ARG BUILD_DIR=/tmp/oryx/build
+RUN mkdir -p ${IMAGES_DIR} \
+    && mkdir -p ${BUILD_DIR}
+COPY images ${IMAGES_DIR}
+COPY build ${BUILD_DIR}
+RUN find ${IMAGES_DIR} -type f -iname "*.sh" -exec chmod +x {} \; \
+    && find ${BUILD_DIR} -type f -iname "*.sh" -exec chmod +x {} \;
```
arroyc, line 9: can we rename this to buildcontext.dockerfiles or something like that?

images/build/Dockerfiles/ltsVersions.Dockerfile
```diff
@@ -171,98 +146,84 @@ RUN set -ex \
...
-LABEL com.microsoft.oryx.git-commit=${GIT_COMMIT}
-LABEL com.microsoft.oryx.build-number=${BUILD_NUMBER}
-LABEL com.microsoft.oryx.release-tag-name=${RELEASE_TAG_NAME}
 
-ENV ${SDK_STORAGE_ENV_NAME} ${SDK_STORAGE_BASE_URL_VALUE}
+ENV LANG="C.UTF-8" \
+    ORIGINAL_PATH="$PATH" \
+    PATH="$ORYX_PATHS:$PATH" \
+    NUGET_XMLDOC_MODE="skip" \
+	DOTNET_SKIP_FIRST_TIME_EXPERIENCE="1" \
```
arroyc, line 223: spacing

images/build/Dockerfiles/gitHubActions.Dockerfile
```diff
@@ -87,63 +63,50 @@ RUN set -ex \
...
-RUN ln -s /opt/buildscriptgen/GenerateBuildScript /opt/oryx/oryx
+ARG AI_KEY
 
-# Install PHP pre-reqs
-RUN ${IMAGES_DIR}/build/php/prereqs/installPrereqs.sh
-# NOTE: do not include the following lines in prereq installation script as
-# doing so is causing different version of libargon library being installed
-# causing php-composer to fail
-RUN apt-get update \
+COPY --from=pre-final /opt /opt
```
arroyc, line 69: nit: rename pre-final

images/build/Dockerfiles/ltsVersions.Dockerfile
```diff
@@ -77,9 +55,9 @@ ENV DOTNET_RUNNING_IN_CONTAINER=true \
     DOTNET_USE_POLLING_FILE_WATCHER=true \
 	NUGET_XMLDOC_MODE=skip \
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE=1 \
-	NUGET_PACKAGES=/var/nuget
+	NUGET_PACKAGES=/opt/nuget
```
arroyc, line 58: space

images/build/Dockerfiles/ltsVersions.Dockerfile
```diff
@@ -77,9 +55,9 @@ ENV DOTNET_RUNNING_IN_CONTAINER=true \
...
 
-RUN mkdir /var/nuget
+RUN mkdir /opt/nuget
```
arroyc, line 60: you can attach this mkdir command to a previous or post run statements as well

images/build/Dockerfiles/ltsVersions.Dockerfile
```diff
@@ -171,98 +146,84 @@ RUN set -ex \
...
 
-FROM main AS python
-ARG BUILD_DIR
-ARG IMAGES_DIR
-# It's not clear whether these are needed at runtime...
-RUN apt-get update \
+FROM main AS final
+ARG AI_KEY
+ARG SDK_STORAGE_BASE_URL_VALUE
+COPY --from=pre-final /opt /opt
```
arroyc, line 152: nit: rename pre-final

images/build/Dockerfiles/gitHubActions.Dockerfile
kichalla: Any suggestions?

kichalla:
> what is the size now? did layer reduction help reducing image size as well?

It is the same. No, I do not see any reduction (which is expected I would say)

build/buildBuildImages.sh
arroyc: nit: instead of tempfiles rename it something more meaningful like buildcontext etc.

images/build/Dockerfiles/ltsVersions.Dockerfile
kichalla: I am not worried about this since it is an intermediate stage

images/build/Dockerfiles/gitHubActions.Dockerfile
kichalla: Take a look at the new name in the latest commit that I pushed. I just named it `intermediate`

images/build/Dockerfiles/temporaryFiles.Dockerfile
kichalla: In my latest commit I avoided creating the name with `buildcontext` since there is actually a concept of docker build context and i did not want to confuse the name with that, so i created a different name `supportFilesForBuildingBuildImages.Dockerfile`

images/build/Dockerfiles/gitHubActions.Dockerfile
arroyc: seems like we are preparing opt folder in final in this layer .. so can it be assemble-opt-dir-for-final .... will that be too long?

kichalla: Hmm, I do not think that is necessary

PR #709, opened 2020-07-31
Remove retired platform version from image and enable dynamic install

qianz2:


qianz2:
> I would suggest to wait till monday .. by that time @kichalla 's docker layer pr would be merged and that would be a lot of merge conflict to this PR.

Sure, will leave it for now.

images/build/Dockerfiles/Dockerfile
```diff
@@ -103,9 +89,6 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 RUN cd ${IMAGES_DIR} \ 
  && . ${BUILD_DIR}/__nodeVersions.sh \
- && ./installPlatform.sh nodejs 4.4.7 \
- && ./installPlatform.sh nodejs 4.5.0 \
- && ./installPlatform.sh nodejs 4.8.0 \
  && ./installPlatform.sh nodejs 6.2.2 \
```
kichalla, line 92: I thought we are removing version 6 too? @arroyc?

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreSampleAppsTest.cs
```diff
@@ -29,94 +29,14 @@ public DotNetCoreSampleAppsTest(ITestOutputHelper output) : base(output)
 
         private readonly string SdkVersionMessageFormat = "Using .NET Core SDK Version: {0}";
 
-        [Fact]
-        public void Builds_NetCore10App_UsingNetCore11_DotNetSdkVersion()
```
kichalla, line 33: We need to have coverage for this version in our platform so can you check if there is already a test for this version? else we can change this test to use the github-actions image. Similarly can you check for other tests too?

images/build/Dockerfiles/Dockerfile
```diff
@@ -63,7 +59,6 @@ RUN set -ex \
 RUN set -ex \
  && sdksDir=/opt/dotnet/sdks \
  && cd $sdksDir \
- && ln -s 1.1 1 \
  && ln -s 3.0 3
```
kichalla, line 62: we do not need 2.2 and 3.0 also in my opinion

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreSampleAppsTest.cs
qianz2: Sure, will check the coverage for this version.

kichalla: I just found that dynamic install is currently failing for 1.0 and 1.1 versions in github-actions and lts-versions image. I think it is because these two images are missing pre-requisite system packages for those versions to work.
For example, you can add the following highlighted system packages:

image

PR #712, opened 2020-08-05
Fixed tagging of build images

kichalla:


build/buildRunTimeImages.sh
```diff
@@ -115,6 +115,11 @@ for dockerFile in $dockerFiles; do
         $labels \
         .
 
+    echo
+    echo "'$localImageTagName' image history:"
+    echo docker history $localImageTagName
```
kichalla, line 120: I added this because i am noticing even some runtime images' layers are a lot, so we might need to do some work over there too

build/buildBuildImages.sh
```diff
@@ -146,7 +146,7 @@ function buildGitHubActionsImage() {
 	
 	echo
 	echo "-------------Creating build image for GitHub Actions-------------------"
-	builtImageName="$ACR_BUILD_GITHUB_ACTIONS_IMAGE_NAME"
+	local builtImageName="$ACR_BUILD_GITHUB_ACTIONS_IMAGE_NAME"
```
qianz2, line 149: wondering what `local` here used for

vsts/pipelines/validation.yml
```diff
@@ -17,7 +17,7 @@ jobs:
   displayName: Build and Test Build Images
   pool:
     name: OryxLinux
-  timeoutInMinutes: 300
+  timeoutInMinutes: 480
```
qianz2, line 20: should we leave it as `300`.

build/buildBuildImages.sh
qianz2: oh never mind, got it

kichalla: This is so that scope of this variable is limited to this function only

vsts/pipelines/validation.yml
kichalla: I increased it because the building and testing of images is now taking a long time due to recent changes related to reducing layers of build images and also conda related tests. If we do not increase the timeout devops is killing off the job.

PR #714, opened 2020-08-10
Upgrade packages whenever runtime images are built and upgrade npm cli

kichalla:


images/build/Dockerfiles/gitHubRunners.BuildPackDepsStretch.Dockerfile
```diff
@@ -1,4 +1,4 @@
 # The digest here must match with what is present in the following location
 # so that downloading of the image is very fast in GitHub workflows
 # https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md
-FROM buildpack-deps:stretch@sha256:23c6d55a4d5ad7c9476638196920d67a97702433b5b48465496af6ab3214e7e4
+FROM buildpack-deps:stretch@sha256:d95db8b9293c71d0f9b6a12d96d1ace65af81fd6535e5cb07078df85b3147a76
```
arroyc, line 4: again?

arroyc: should we just create another githubrunner image that always points to latest  buildpackdepsstretch so that it will ensure all security patches ... we can keep this one for github actions specifically

kichalla: > again?

If they are building their VMs regularly then yeah this will keep happening.

> should we just create another githubrunner image that always points to latest buildpackdepsstretch so that it will ensure all security patches ... we can keep this one for github actions specifically

Yeah, let me think about it.

images/runtime/node/installDependencies.sh
```diff
@@ -11,36 +11,40 @@ set -ex
...
+    npm install -g pm2@3.5.1 -loglevel silent
 else
     echo "Skipping PM2 installation..."
 fi
 
 # Application-Insights is supported as an option for telemetry when running the app,
 # so we need to make sure it is available in our images.
 # Updated to 1.8.3 that doesn't emit json payload in stdout which is causing issues to customers in ant-88
-npm install -g applicationinsights@1.8.3
+npm install -g applicationinsights@1.8.3 -loglevel silent
```
qianz2, line 50: could you move the appinsights version and pm2 version as constants

kichalla: Sure, sounds good

PR #725, opened 2020-08-15
Made fixes to improve build image build perf

kichalla:


images/build/Dockerfiles/Dockerfile
```diff
@@ -166,11 +166,19 @@ RUN set -ex \
...
  && ln -s 5.6.0 /opt/npm/5.6 \
  && ln -s 5.6 /opt/npm/5
+RUN set -ex \
+    && mkdir -p /output \
+    && cd /opt \
+    && tar -zcf /output/opt.tar.gz .
 
 FROM main AS final
-COPY --from=intermediate /opt /opt
+COPY --from=intermediate /output /opt
```
arroyc, line 175: so this will copy the yan-cache, but nuget contents are still in /var/nuget .. can you please confirm /var/nuget is also getting copied in /output?

images/build/yarn-cache/cacheNodePackages.sh
```diff
@@ -203,8 +203,11 @@ for pkg in \
...
-chmod -R 777 $YARN_CACHE_DIR
-
 echo "Caching done. Total cache size: $(du -sh $YARN_CACHE_DIR | cut -f -1)"
-rm -fr $TEMP_APP_DIR
\ No newline at end of file
+rm -fr $TEMP_APP_DIR
+
+mkdir -p /output
+cd $YARN_CACHE_DIR
+tar -zcf /output/yarncache.tar.gz .
```
arroyc, line 211: here we are compressing whole directory which is $YARN_CACHE_DIR ... is it also happen to be /output?

images/build/Dockerfiles/Dockerfile
arroyc: nvm .. i  saw it's only need in ltsversions image which is there .. so ignore this 👍

images/build/yarn-cache/cacheNodePackages.sh
kichalla: basically i am establishing a pattern where either this yarn-cache or the 'intermediate' stage has an folder called 'output' where we can find the stuff to be copy from.

PR #726, opened 2020-08-15
README update

dbalabka:


qianz2:
Thanks @dbalabka , we updated it in #783

PR #728, opened 2020-08-18
Revert build image perf fixes

kichalla:


images/build/Dockerfiles/Dockerfile
```diff
@@ -166,19 +166,11 @@ RUN set -ex \
...
-RUN set -ex \
-    && mkdir -p /output \
-    && cd /opt \
-    && tar -zcf /output/opt.tar.gz .
 
 FROM main AS final
-COPY --from=intermediate /output /opt
+COPY --from=intermediate /opt /opt
 # https://github.com/docker-library/python/issues/147
-RUN cd /opt \
```
kichalla, line 177: I am reverting my earlier check-in related to improving build perf because the build image size increased by 1.2 GB. The reason behind this increase is because docker layers are immutable. For example, the copy command in line 171 copies the tar balls to the final stage but these tar balls form an immutable layer and do not actually go away even though we try to remove in this RUN statement causing this additional image size.

PR #739, opened 2020-08-20
security update for php

arroyc:
need to update the php base image tags ...
some tests needs to be updated as well


kichalla:
Let us watch out for any huge changes to shared size of these runtime images after your changes.

arroyc:
> Let us watch out for any huge changes to shared size of these runtime images after your changes.

ok

PR #742, opened 2020-08-21
Add ruby runtime images

qianz2:


build/buildPlatformBinaries.sh
```diff
@@ -23,6 +23,9 @@ case $PLATFORM_TO_BUILD in
 		;;
 	'nodejs')
 		"$platformsDir/nodejs/getNode.sh"
-		;;            
+		;; 
+	'ruby')
```
kichalla, line 27: Make sure to create a container for 'ruby' before running the platform-binary-build for it. BTW, create a new pipeline for ruby in `vsts/platformbinaries` folder and I can create the build definition for it.

images/build/ruby/2.6/Dockerfile
```diff
@@ -0,0 +1,3 @@
+# Build ruby 2.6
+FROM ruby-build-prereqs AS ruby26-build
```
kichalla, line 2: we might not need these 'build' related dockerfiles since AppService might be interested in installing them dynamically. We could however install it in vso image but anyway these docker files should not be required. Also did you check if our build image should also have any pre-reqs installed for ruby?

images/build/ruby/2.7/Dockerfile
```diff
@@ -0,0 +1,3 @@
+# Build ruby 2.7
+FROM ruby-build-prereqs AS ruby27-build
```
kichalla, line 2: For runtime dockerfiles for ruby, you can check with Sanchit to get their existing Ruby related dockerfiles.

qianz2:
restart to work on this PR and will open a new one once worked out.

PR #746, opened 2020-08-24
Add Ruby platform pipeline

qianz2:


build/__rubyVersions.sh
```diff
@@ -0,0 +1,7 @@
+# This file was auto-generated from 'constants.yaml'. Changes may be overridden.
+
+GEM_VERSION='3.1.4'
```
kichalla, line 3: Make sure to add a sanity over here: `tests\Oryx.Integration.Tests\StorageAccountSanityTestBase.cs`

PR #751, opened 2020-08-28
Added support for detecting Java applications

kichalla:


build/__rubyVersions.sh
```diff
@@ -4,4 +4,4 @@ GEM_VERSION='3.1.4'
 RUBY27_VERSION='2.7.1'
 RUBY27_TAR_SHA256='b224f9844646cc92765df8288a46838511c1cec5b550d8874bd4686a904fcee7'
 RUBY26_VERSION='2.6.6'
-RUBY26_TAR_SHA256='5db187882b7ac34016cd48d7032e197f07e4968f406b0690e20193b9b424841f'
\ No newline at end of file
+RUBY26_TAR_SHA256='5db187882b7ac34016cd48d7032e197f07e4968f406b0690e20193b9b424841f'
```
arroyc, line 7: why is this change?

kichalla: Probably happened during rebase but if you notice the change is only addition of a new line

src/BuildScriptGenerator/Java/JavaPlatform.cs
```diff
@@ -0,0 +1,417 @@
...
+            // Write the platform name and version to the manifest file
+            manifestFileProperties[ManifestFilePropertyKeys.JavaVersion] = detectorResult.PlatformVersion;
+
+            string command = string.Empty;
+
+            if (javaPlatformDetectorResult.UsesMavenWrapperTool)
+            {
+                if (_commonOptions.ShouldPackage)
+                {
+                    command = "./mvnw clean package";
```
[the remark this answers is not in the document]
kichalla: Sure 👍

src/BuildScriptGenerator/Java/JavaPlatform.cs
```diff
@@ -0,0 +1,417 @@
...
+
+                // Maven spits out lot of information related to downloading of packages which is too verbose.
+                // Since the --quiet option is too quiet, we are trying to use a new switch below to just mute the
+                // messages related to transfer progress of these downloads.
+                // https://maven.apache.org/docs/3.6.1/release-notes.html#user-visible-changes
+                var minVersionHavingLessVerbositySupport = new SemVer.Version("3.6.1");
+                var currentMavenVersion = new SemVer.Version(javaPlatformDetectorResult.MavenVersion);
+                if (currentMavenVersion.CompareTo(minVersionHavingLessVerbositySupport) >= 0)
+                {
+                    command = $"{command} --no-transfer-progress";
```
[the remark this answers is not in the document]
kichalla: Please check my comments above

src/BuildScriptGenerator/Java/JavaPlatform.cs
```diff
@@ -0,0 +1,417 @@
...
+    /// </summary>
+    internal class JavaPlatform : IProgrammingPlatform
+    {
+        /// <summary>
+        /// The tar-gz option for Java modules.
+        /// </summary>
+        private readonly BuildScriptGeneratorOptions _commonOptions;
+        private readonly JavaScriptGeneratorOptions _javaScriptGeneratorOptions;
+        private readonly IJavaVersionProvider _javaVersionProvider;
+        private readonly IMavenVersionProvider _mavenVersionProvider;
```
arroyc, line 28: nit: java can be built using gradle or ant etc as well. Should we consider them here as well? so instead of a IMavenVersionProvider and MavenInstaller should we use a wrapper IJavaBuilderVersionProvider and JavaBuildToolInstaller?

src/BuildScriptGenerator/Java/JavaPlatform.cs
```diff
@@ -0,0 +1,417 @@
...
+        private readonly ILogger<JavaPlatform> _logger;
+        private readonly IJavaPlatformDetector _detector;
+        private readonly JavaPlatformInstaller _javaPlatformInstaller;
+        private readonly MavenInstaller _mavenInstaller;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="JavaPlatform"/> class.
+        /// </summary>
+        /// <param name="commonOptions">The <see cref="BuildScriptGeneratorOptions"/>.</param>
+        /// <param name="javaScriptGeneratorOptions">The options for JavaScriptGenerator.</param>
```
arroyc, line 38: nit: ScriptGeneratorOptionsJava?

src/BuildScriptGenerator/Java/JavaPlatform.cs
```diff
@@ -0,0 +1,417 @@
...
+            }
+            else if (javaPlatformDetectorResult.UsesMaven)
+            {
+                if (_commonOptions.ShouldPackage)
+                {
+                    command = "mvn clean package";
+                }
+                else
+                {
+                    command = "mvn clean compile";
```
arroyc, line 135: nit: move these commands to constants

arroyc (changes requested):

src/BuildScriptGenerator/BuildScriptGeneratorServiceCollectionExtensions.cs
```diff
@@ -25,7 +25,8 @@ public static IServiceCollection AddBuildScriptGeneratorServices(this IServiceCo
                 .AddHugoScriptGeneratorServices()
                 .AddPythonScriptGeneratorServices()
                 .AddDotNetCoreScriptGeneratorServices()
-                .AddPhpScriptGeneratorServices();
+                .AddPhpScriptGeneratorServices()
+                .AddJavaScriptGeneratorServices();
```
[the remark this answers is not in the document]
kichalla: `AddScriptGeneratorServicesForJava`..yeah that sounds good to me, Sure we can make changes to other extensions in a different PR

src/BuildScriptGenerator/Java/JavaPlatform.cs
kichalla: Regarding supporting different build systems: I did not add Ant as it did not look like a popular way of doing things currently and many seem to be using Maven. Regarding Gradle, yeah I think we should add support for it but seems like it is mostly used for Android based apps (which is also a valid scenario for codespaces since your could have a github repo for that those kind of apps and want to 'develop' in codespaces environment). I can create a separate PR for adding support for it though.

Regarding wrapper, I do not feel too strong about it. So far we have been having separate installers and providers for our tools and we could do the same for Gradle too. what do you think?

arroyc (approved):
as discussed, lets have an workitem scoping support for gradle vs ant etc  in future .. and for the time being we are good with going forward with maven only, also please add the commands in the constants

PR #756, opened 2020-09-03
Support Ruby in BuildScriptGenerator

qianz2:


arroyc (changes requested):

arroyc:
as discusse: log an workitem to research later where we need to decide if we fail the build when gem file is not available

src/Detector/Ruby/RubyPlatformDetectorResult.cs
```diff
@@ -0,0 +1,11 @@
+﻿// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+namespace Microsoft.Oryx.Detector.Ruby
+{
+    public class RubyPlatformDetectorResult : PlatformDetectorResult
+    {
+    }
```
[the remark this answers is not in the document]
qianz2: There has no Ruby specific detection result we currently want to detect, I think we should add a GemfileExists properties here. Will change here.

src/BuildScriptGenerator/Ruby/RubyBashBuildSnippet.sh.tpl
```diff
@@ -0,0 +1,17 @@
...
+echo "Using Gem version:"
+gem --version
+echo
+echo
+echo "Using Ruby version:"
+ruby --version
+echo
+
+{{ if UseBundlerToInstallDependencies }}
+echo "Running 'gem install bundler'..."
```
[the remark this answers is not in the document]
qianz2: Yes, ruby installation will come with gem if ruby version is newer than 1.*, since we have 2.* so we're good. A Gem is a form of package and if we found a Gemfile, then use bundler to install all dependencies inside the Gemfile. I Will think of the case when there has no Gemfile included in the app.

src/BuildScriptGenerator/Ruby/RubyPlatform.cs
```diff
@@ -0,0 +1,295 @@
...
+        {
+            if (_rubyInstaller.IsVersionAlreadyInstalled(rubyVersion))
+            {
+                _logger.LogDebug(
+                   "Ruby version {version} is already installed. So skipping installing it again.",
+                   rubyVersion);
+            }
+            else
+            {
+                _logger.LogDebug(
```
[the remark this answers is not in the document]
qianz2: If ruby is not installed, then call the `GetInstallerScriptSnippet` to generate installation script so here `LogDebug` will be good I think.

images/build/Dockerfiles/vso.Dockerfile
```diff
@@ -30,5 +30,16 @@ RUN buildDir="/opt/tmp/build" \
     && condaDir="/opt/oryx/conda" \
     && mkdir -p "$condaDir" \
     && cd $imagesDir/build/python/conda \
-    && cp -rf * "$condaDir" \
+    && cp -rf * "$condaDir"
+
+# Install Ruby
+ENV RUBY_BIN="opt/ruby/lts/bin"
```
kichalla, line 36: Since we want to reduce the number of layers that are part of a docker image, these instructions need to be modified. Please take a look at my Java PR for idea on how to make changes in `vso.Dockerfile`

src/BuildScriptGenerator/Ruby/RubyManifestFilePropertyKeys.cs
```diff
@@ -0,0 +1,14 @@
+// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+namespace Microsoft.Oryx.BuildScriptGenerator.Ruby
+{
+    internal static class RubyManifestFilePropertyKeys
+    {
+        internal const string RubyVersion = "rubyversion";
```
kichalla, line 10: nit: nameof(RubyVersion), nameof(GemVersion)

src/BuildScriptGenerator/Ruby/RubyManifestFilePropertyKeys.cs
```diff
@@ -0,0 +1,14 @@
...
+
+        internal const string GemVersion = "gemversion";
```
kichalla, line 12: I would remove this since it comes as part of Ruby? and also make sure to remove it from `RubyScriptGeneratorOptions.cs` file

tests/Oryx.BuildImage.Tests/Ruby/RubySampleAppsTest.cs
```diff
@@ -0,0 +1,129 @@
...
+                .SetEnvironmentVariable(
+                    SdkStorageConstants.SdkStorageBaseUrlKeyName,
+                    SdkStorageConstants.DevSdkStorageBaseUrl)
+                .AddBuildCommand($"{appDir} -o {appOutputDir}")
+                .ToString();
+
+            // Act
+            var result = _dockerCli.Run(new DockerRunArguments
+            {
+                ImageId = _imageHelper.GetVsoBuildImage(),
```
kichalla, line 42: Could you add one test which uses GitHubActions image to test the dynamic install too?

kichalla (approved):

images/build/Dockerfiles/vso.Dockerfile
qianz2: Sure, I made changes to it

arroyc (approved):

arroyc:
please trigger a  successful nightly build before merging the PR to master

PR #786, opened 2020-09-21
Updated .NET Core versions

kichalla:
Updated .NET Core 5.0 versions to latest


src/BuildScriptGenerator/Ruby/RubyVersions.cs
```diff
@@ -9,5 +9,7 @@ public static class RubyVersions
         public const string Ruby27TarSha256 = "b224f9844646cc92765df8288a46838511c1cec5b550d8874bd4686a904fcee7";
         public const string Ruby26Version = "2.6.6";
         public const string Ruby26TarSha256 = "5db187882b7ac34016cd48d7032e197f07e4968f406b0690e20193b9b424841f";
+        public const string Ruby25Version = "2.5.8";
```
arroyc, line 12: why this change?

arroyc (commented):
we need to change the base image builds for dotnet core 5 as well. Can we trigger that

src/BuildScriptGenerator/Ruby/RubyVersions.cs
kichalla: This change is because of running `generateConstants.sh` script (cc @qianz2)

kichalla:
> we need to change the base image builds for dotnet core 5 as well. Can we trigger that

Yes, I updated the PR now. Thanks.

arroyc (approved):

PR #798, opened 2020-09-28
Support building Node Monorepos Apps

qianz2:
Support building Monorepos using two types of tools: lerna and lage.


src/BuildScriptGenerator/Node/NodeConstants.cs
```diff
@@ -18,6 +19,15 @@ public static class NodeConstants
         public const string YarnStartCommand = "yarn run start";
         public const string YarnCommand = "yarn";
         public const string HugoCommand = "hugo";
+        public const string LernaCommand = "lerna";
+        public const string LernaVersionCommand = "echo Using Npm version: && lerna --version";
+        public const string LernaInitCommand = "lerna init";
+        public const string LernaBootstrapCommand = "lerna bootstrap";
+        public const string InstallLernaCommandNpm = "npm install --global lerna";
```
kichalla, line 26: Does this have to be `global`?

src/BuildScriptGenerator/Node/NodeConstants.cs
```diff
@@ -18,6 +19,15 @@ public static class NodeConstants
...
+        public const string InstallLernaCommandYarn = "npm install --global lerna --no-package-lock";
```
kichalla, line 27: What is the difference between `InstallLernaCommandNpm` and `InstallLernaCommandYarn`? I do not see `yarn` being used in the `Yarn` command

tests/SampleApps/nodejs/monorepo-lerna-yarn/lerna.json
```diff
@@ -0,0 +1,8 @@
+{
+  "npmClient": "yarn",
```
kichalla, line 2: Does `lerna` take care of choosing either `npm` or `yarn`?

kichalla (changes requested):

src/BuildScriptGenerator/Node/NodeConstants.cs
qianz2: In its documentation, this is the command to install lerna tool. I didn't try installing it without `global`, I'll try it and see. Just wondering anything not good with `global`?

tests/SampleApps/nodejs/monorepo-lerna-yarn/lerna.json
qianz2: The default is client is npm, Lerna can choose which one by having this line in `lerna.json` file.

src/BuildScriptGenerator/Node/NodeConstants.cs
qianz2: Good point, its basically same, I was thinking not adding `package-lock.json` if it's yarn, so added it, but it seems not necessary since it will not create a brand new one if not exist one previously.

arroyc (changes requested):
couple of questions:
1. using npm command over yarn in learninstall command? what happens if there is a yarn.lock file and also user wants to use lerna?
2. Current Pr we have npmrunbuild command and npmrunbuildazurecommand prior to lerna ... so can you create a test to show which one will run?
3. Do we need to install Lerna/Lage in build image? 
4. Is there any version related restriction? for example for node 14 or newer it works with certain npm version .. is this also applicable to lerna and lage?
5. Not sure if this is valid use case ... what happens when there is bot lerna and lage command ... do we  prefer to choose one over the other?

qianz2:
> couple of questions:
> 
> 1. using npm command over yarn in learninstall command? what happens if there is a yarn.lock file and also user wants to use lerna?
> 2. Current Pr we have npmrunbuild command and npmrunbuildazurecommand prior to lerna ... so can you create a test to show which one will run?
> 3. Do we need to install Lerna/Lage in build image?
> 4. Is there any version related restriction? for example for node 14 or newer it works with certain npm version .. is this also applicable to lerna and lage?
> 5. Not sure if this is valid use case ... what happens when there is bot lerna and lage command ... do we  prefer to choose one over the other?

Here are some of my understandings:
1: The way to install lerna is only through npm, using this command: `npm install --global lerna`. as I tried, running `yarn install` could not install lerna even it's included in `package.json` devdepencies.  If there's a `yarn.lock` file, and  a line `"npmClient": "yarn"` in `lerna.json` file, that means lerna choose yarn to be its client to run commands. But we still need to run `npm install --global lerna` to have lerna in the first place.
2. Sure I will add a test for it. `npmrunbuild ` command is basically from "script" ---> "build" line in `package.json` file, so if users already put "lerna run build" line there, it will be good, if not put there, we'll to run `LernaRunBuildCommand` which is same.
3. If we install it in build image, it will increase image size? I'm not sure actually,. as long as we have npm tool included in the image we should be able to install lerna while generating build script? The current approach is Installing lerna/lage as needed.
4. I didn't find any version restriction related to npm and lerna.
5. Using both package management tool is not valid to us and it's unlikely for users to do so. If it's the case, we can throw an exception? There has no preference from users' perspectives I guess.

qianz2:
Here is the full node bash build script after adding lerna commands for reference:
----------------------------------------------------------------------------
    echo
    echo "Using Node version:"
    node --version
    echo
    echo Using Npm version: && npm --version
    
    
    zippedModulesFileName=
    allModulesDirName=".oryx_all_node_modules"
    prodModulesDirName=".oryx_prod_node_modules"
    PruneDevDependencies=false
    HasProdDependencies=false
    HasDevDependencies=false
    
    # if node modules exist separately for dev & prod (like from an earlier build),
    # rename the folders back appropriately for the current build
    if [ -d "$allModulesDirName" ]
    then
    	echo
    	echo "Found existing folder '$SOURCE_DIR/$allModulesDirName'."
    	echo "Copying modules from '$SOURCE_DIR/$allModulesDirName' to '$SOURCE_DIR/node_modules'..."
    	cd "$SOURCE_DIR"
    	mkdir -p node_modules
    	rsync -rcE --links "$allModulesDirName/" node_modules
    fi
    
    if [ "$PruneDevDependencies" == "true" ] && [ "$HasProdDependencies" == "true" ]
    then
    	# Delete existing prod modules folder so that we do not publish
    	# any unused modules to final destination directory.
    	if [ -d "$prodModulesDirName" ]; then
    		echo "Found existing '$SOURCE_DIR/$prodModulesDirName'. Deleting it..."
    		rm -rf "$prodModulesDirName"
    	fi
    
    	mkdir -p "$prodModulesDirName"
    	cd "$prodModulesDirName"
    
    	if [ -f "$SOURCE_DIR/package.json" ]; then
    		cp -f "$SOURCE_DIR/package.json" .
    	fi
    
    	if [ -f "$SOURCE_DIR/package-lock.json" ]; then
    		cp -f "$SOURCE_DIR/package-lock.json" .
    	fi
    	
    	if [ -f "$SOURCE_DIR/.npmrc" ]; then
    		cp -f "$SOURCE_DIR/.npmrc" .
    	fi
    
    	if [ -f "$SOURCE_DIR/yarn.lock" ]; then
    		cp -f "$SOURCE_DIR/yarn.lock" .
    	fi
    
    	if [ -f "$SOURCE_DIR/.yarnrc" ]; then
    		cp -f "$SOURCE_DIR/.yarnrc" .
    	fi
    
    	echo
    	echo "Installing production dependencies in '$SOURCE_DIR/$prodModulesDirName'..."
    	echo
    	echo "Running 'npm install --unsafe-perm --production'..."
    	echo
    	npm install --unsafe-perm --production
    
    	if [ -d "node_modules" ]; then
    		echo
    		echo "Copying production dependencies from '$SOURCE_DIR/$prodModulesDirName' to '$SOURCE_DIR/node_modules'..."
    		START_TIME=$SECONDS
    		rsync -rcE --links "node_modules/" "$SOURCE_DIR/node_modules"
    	fi
    fi
    
    cd "$SOURCE_DIR"
    
    echo
    echo "Running 'npm install --unsafe-perm'..."
    echo
    npm install --unsafe-perm
    
    echo
    echo "Using npm to install Lerna:"
	npm install --global lerna
	echo
	echo
	echo "Using Lerna version:"
	lerna --version
	echo
	echo
	echo "Running 'lerna init & lerna bootstrap':"
	lerna init
	lerna bootstrap
	echo
	echo
	echo "Running 'lerna run build'..."
	echo
	lerna run build
    
    if [ "$PruneDevDependencies" == "true" ] && [ "$HasDevDependencies" == "true" ]
    then
    	if [ -d "node_modules" ]; then
    		echo
    		echo "Copy '$SOURCE_DIR/node_modules' with all dependencies to '$SOURCE_DIR/$allModulesDirName'..."
    		rsync -rcE --links "node_modules/" "$allModulesDirName" --delete
    	fi
    
    	if [ "$HasProdDependencies" == "true" ] && [ -d "$prodModulesDirName/node_modules/" ]; then
    		echo
    		echo "Copying production dependencies from '$SOURCE_DIR/$prodModulesDirName/node_modules' to '$SOURCE_DIR/node_modules'..."
    		rsync -rcE --links "$prodModulesDirName/node_modules/" node_modules --delete
    	else
    		rm -rf "node_modules/"
    	fi
    fi

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -181,7 +230,9 @@ public PlatformDetectorResult Detect(RepositoryContext context)
             var productionOnlyPackageInstallCommand = string.Format(
                 NodeConstants.ProductionOnlyPackageInstallCommandTemplate, packageInstallCommand);
 
-            if (string.IsNullOrEmpty(_nodeScriptGeneratorOptions.CustomRunBuildCommand))
+            if (string.IsNullOrEmpty(_nodeScriptGeneratorOptions.CustomRunBuildCommand) &&
```
qianz2, line 233: Here are priorities of which build command we should generate in the build script: `customRunBuildCommand ` > lerna 
 or lage commands > build/build:azure under script in `package.json` commands.  So I modifies some logics here hope it makes sense to you @kichalla @arroyc.

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -163,6 +182,36 @@ public PlatformDetectorResult Detect(RepositoryContext context)
...
+                }
+            }
+
+            // If a 'lage.config.js' file exits, run build using lage specifc commands.
+            if (nodePlatformDetectorResult.HasLageConfigJSFile)
+            {
+                runBuildLageCommand = ctx.SourceRepo.FileExists(NodeConstants.YarnLockFileName) ?
+                        NodeConstants.YarnRunLageBuildCommand : NodeConstants.NpmRunLageBuildCommand;
+            }
+
```
arroyc, line 213: line 185 to line 214  (except 171 to 183) can be moved up as else if condition at line 170

tests/BuildScriptGenerator.Tests/Node/NodePlatformTest.cs
```diff
@@ -125,6 +125,42 @@ public void GeneratedBuildSnippet_HasNpmRunBuildAzureCommand()
             Assert.Contains("npm run build:azure", buildScriptSnippet.BashBuildScriptSnippet);
         }
 
+        [Fact]
+        public void GeneratedBuildSnippet_HasLernaRunBuildCommand_IfLernaJsonFileExists()
```
arroyc, line 129: add couple of more tests .. 
1. lerna and lage are both there and we are throwing exception
2. lerna and customrunbuild command is there, customrunbuild should be executed
3. lernarunbuild, customrunbuild and lage command all are there, we should throw exception
4. npmrunbuild and lernarunbuild are there and we should be running lernarunbuild

src/BuildScriptGenerator/Node/NodeConstants.cs
arroyc: @qianz2 we may not have permission to install it globally in appservice build image .. you can try this locally with oryxtests/build image and see if you are having any permission issue or anything as such while building a lerna app

src/BuildScriptGenerator/Node/NodePlatform.cs
qianz2: I put them at here because we want to use lerna chosen npm client not from what detect. for example, if `yarn.lock` found but lerna have `npmClient: npm` line in `lerna.json` we still choose npm. It will be after the file check so it's able to override.

tests/BuildScriptGenerator.Tests/Node/NodePlatformTest.cs
qianz2: Sure, sounds good

images/build/Dockerfiles/Dockerfile
```diff
@@ -107,7 +107,8 @@ RUN set -ex \
  && ln -s 5.3.0 /opt/npm/5.3 \
  && ln -s 5.4.2 /opt/npm/5.4 \
  && ln -s 5.6.0 /opt/npm/5.6 \
- && ln -s 5.6 /opt/npm/5
+ && ln -s 5.6 /opt/npm/5 \
```
arroyc, line 110: can you rebase on master?? I don't think these npm symlinks exist anymore

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -93,23 +93,48 @@ fi
...
+	echo "Using Lerna version:"
+	lerna --version
+	echo
+	echo
+	echo "Running '{{ LernaInitCommand }} & {{ LernaBootstrapCommand }}':"
+	{{ LernaInitCommand }}
+	{{ LernaBootstrapCommand }}
+	echo
+	echo
+	echo "Running '{{ LernaRunBuildCommand }}'..."
```
arroyc, line 115: I found this article and was reading more about lerna ...
https://www.jannikbuschke.de/blog/monorepo-with-lerna-react-and-typescript/

> {
>   "name": "root",
>   "private": true,
>   "devDependencies": {
>     "lerna": "3.13.1"
>   },
>   "scripts": {
>     "bootstrap": "lerna bootstrap --hoist",
>     "start": "lerna run start --stream",
>     "tsc": "lerna run tsc --stream",
>     "watch": "lerna run watch --stream"
>   }
> }
> 
> 

Can we have a test on an app which has a package.json like this to see if this buildscript is good enough for that?

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -93,23 +93,48 @@ fi
...
+	echo
+	{{ LernaRunBuildCommand }}
+{{ else if LageRunBuildCommand | IsNotBlank }}
+	echo
+	echo "Running ' {{ InstallLageCommand }} ':"
```
arroyc, line 120: same comment for lage

arroyc (changes requested):

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
arroyc: try to build/test with this https://github.com/jannikbuschke/lerna-react-typescript-sample

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -93,23 +93,48 @@ fi
 
 cd "$SOURCE_DIR"
 
-echo
-echo "Running '{{ PackageInstallCommand }}'..."
-echo
-{{ PackageInstallCommand }}
-
 {{ if CustomRunBuildCommand | IsNotBlank }}
```
kichalla, line 96: Move this `CustomRunBuildCommand` into the `else` section as having this separate section feels repetitive.
For reference, before your changes in this PR, this is how it was looking like:

image

images/build/Dockerfiles/Dockerfile
```diff
@@ -77,6 +77,7 @@ RUN set -ex \
     && ln -s 10.14.2 10.14 \
     && ln -s 10.21.0 10.21 \
     && ln -s 12.18.2 12.18 \
+    && npm install -g lerna \
```
kichalla, line 80: Which image tag is Terrapin using? I ask because depending on the tag they are using, this might not be present in some tags which results in breaking the buld.

kichalla (changes requested):

images/build/Dockerfiles/Dockerfile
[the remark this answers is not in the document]
arroyc: as far as i know terrapin use full build image ... but i dont know which tag (date) they are currently using in prod

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
qianz2: I agree it looks repetitive because we added this `packageInstallCommand `at multiple places, however since `CustomRunBuildCommand `has priority over other commands, we'll need to put it in the first `if` section.

images/build/Dockerfiles/Dockerfile
qianz2: Terrapin is using lts-versions image, so we'll test lerna in oryx/build:lts-versions image instead of full image now. And for the github-actions image, we don't have node/npm installed in image itself, I'm thinking add at least one node/npm version to it for testing, but that will increase image size/layers.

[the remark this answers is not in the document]
kichalla: Is there a way for users to specify which version of `lerna` to be used? If yes, then having it installed globally will not work

kichalla: > I'm thinking add at least one node/npm version to it for testing, but that will increase image size/layers.

We were discussing about updating the Node path to point to the root folder of the repo's node_modules folder. Have you checked to see if this helps in `lerna` being available to sub project's builds?

qianz2: Users could specify version of lerna and we could install it accordingly but still need to be global, locally didn't work. And here is the path info /opt/nodejs/12.18.3/bin/lerna -> /opt/nodejs/12.18.3/lib/node_modules/lerna/cli.js. I think it's already point to node_modules.
```
root@d0101a29df7f:/# npm install -g lerna@3.22.1
npm WARN deprecated resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
npm WARN deprecated urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
npm WARN deprecated request@2.88.2: request has been deprecated, see https://github.com/request/request/issues/3142
npm WARN deprecated har-validator@5.1.5: this library is no longer supported
npm WARN deprecated mkdirp-promise@5.0.1: This package is broken and no longer maintained. 'mkdirp' itself supports promises now, please switch to that.
**/opt/nodejs/12.18.3/bin/lerna -> /opt/nodejs/12.18.3/lib/node_modules/lerna/cli.js**
+ lerna@3.22.1
added 749 packages from 387 contributors in 20.946s
```

src/Detector/Node/NodeDetector.cs
```diff
@@ -208,5 +224,36 @@ private IEnumerable<FrameworkInfo> DetectFrameworkInfos(DetectorContext context)
...
+                return npmClientName;
+            }
+
+            try
+            {
+                dynamic lernaJson = ReadJsonObjectFromFile(context.SourceRepo, NodeConstants.LernaJsonFileName);
+                if (lernaJson?.npmClient != null)
+                {
+                    npmClientName = lernaJson["npmClient"].Value as string;
+                } else
```
kichalla, line 242: nit: separate line

tests/Oryx.BuildImage.Tests/Node/NodeJsSampleAppsTests.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.IO;
+using Microsoft.Oryx.BuildScriptGenerator.Common;
+using Microsoft.Oryx.BuildScriptGeneratorCli;
+using Microsoft.Oryx.Tests.Common;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Oryx.BuildImage.Tests
+{
+    public class NodeJsSampleAppsTest : SampleAppsTestBase
```
kichalla, line 16: Interesting. Isn't there already a file and class with this name? Also I would rename this file and class to indicate these are mono repo related tests

tests/Oryx.BuildImage.Tests/Node/NodeJsSampleAppsTests.cs
```diff
@@ -0,0 +1,92 @@
...
+                Volumes = new List<DockerVolume> { volume },
+                CommandToExecuteOnRun = "/bin/bash",
+                CommandArguments = new[] { "-c", script }
+            });
+
+            // Assert
+            RunAsserts(
+                () =>
+                {
+                    Assert.True(result.IsSuccess);
```
kichalla, line 53: Do we want to add more assertions here? Like if there are multiple sub-projects, then we want to make sure that modules in each of these sub projects were indeed present after build?

tests/Oryx.BuildImage.Tests/Node/NodeJsSampleAppsTests.cs
```diff
@@ -0,0 +1,92 @@
...
```
kichalla, line 87: Similarly here

tests/Oryx.BuildImage.Tests/Node/NodeJsSampleAppsTests.cs
```diff
@@ -0,0 +1,92 @@
...
+                },
+                result.GetDebugInfo());
+        }
+    }
```
kichalla, line 91: We should add tests for `lage` related samples too?

kichalla (approved):
I am approving now but please take a look at my new comments...mainly about tests...

arroyc (approved):

PR #802, opened 2020-09-30
add an env variable to classify different build images 

arroyc:
This is to verify pulling and tagging is correctly done during integration


tests/Oryx.Integration.Tests/Database/PulledBuildImageTypeTest.cs
```diff
@@ -0,0 +1,69 @@
...
+// --------------------------------------------------------------------------------------------
+
+using System;
+using Microsoft.Oryx.Tests.Common;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Oryx.Integration.Tests
+{
+    [Trait("category", "db")]
```
arroyc, line 13: @kichalla .. i added this as a db category .. it doesn't matter in general .. because all the categories will be pulling and tagging in the same way .. so if db category test fails it ultimately reflects pulling and tagging is wrong for other categories as well.

images/build/Dockerfiles/Dockerfile
```diff
@@ -25,8 +25,12 @@ RUN ./build.sh python     /opt/startupcmdgen/python
 ###
 
 FROM main AS final
+ARG ORYX_BUILDIMAGE_TYPE
 COPY --from=yarn-cache-base /usr/local/share/yarn-cache /usr/local/share/yarn-cache
 COPY --from=startupScriptGens /opt/startupcmdgen/ /opt/startupcmdgen/
+
+ENV ORYX_BUILDIMAGE_TYPE="${ORYX_BUILDIMAGE_TYPE}"
```
kichalla, line 32: These new changes would add more layers. Instead we can create labels using `--label`

kichalla (changes requested):

images/build/Dockerfiles/Dockerfile
arroyc: as discussed moved it to single env "hardcoded"

arroyc (commented):
@kichalla ... locally ran the tests and they are passing ..

image

qianz2 (approved):
Looks good, please make sure the nighly build has passed.

images/build/Dockerfiles/Dockerfile
```diff
@@ -27,6 +28,9 @@ RUN ./build.sh python     /opt/startupcmdgen/python
 FROM main AS final
 COPY --from=yarn-cache-base /usr/local/share/yarn-cache /usr/local/share/yarn-cache
 COPY --from=startupScriptGens /opt/startupcmdgen/ /opt/startupcmdgen/
+
+ENV ORYX_BUILDIMAGE_TYPE="full"
```
kichalla, line 32: This will add a new layer

kichalla (approved):

images/build/Dockerfiles/Dockerfile
```diff
@@ -12,9 +12,10 @@ COPY src/startupscriptgenerator/src .
 ARG GIT_COMMIT=unspecified
 ARG BUILD_NUMBER=unspecified
 ARG RELEASE_TAG_NAME=unspecified
-ENV GIT_COMMIT=${GIT_COMMIT}
-ENV BUILD_NUMBER=${BUILD_NUMBER}
-ENV RELEASE_TAG_NAME=${RELEASE_TAG_NAME}
+ENV GIT_COMMIT=${GIT_COMMIT} \
+    BUILD_NUMBER=${BUILD_NUMBER} \
+    RELEASE_TAG_NAME=${RELEASE_TAG_NAME} \
+    ORYX_BUILDIMAGE_TYPE="full"
```
arroyc, line 18: this is where it should stay @kichalla

images/build/Dockerfiles/Dockerfile
arroyc: this will be removed

PR #803, opened 2020-09-30
Cleaned up npm related references, split tests, .NET Core blazor related detection

kichalla:
Cleaned up npm related references, split tests, .NET Core blazor related detection


images/build/benv.sh
```diff
@@ -199,24 +195,6 @@ benv-resolve() {
     return 0
   fi
 
-  # Resolve npm versions
```
qianz2, line 202: So why we don't need npm benv anymore, could you check if its corresponding tests are also removed?

tests/Oryx.BuildImage.Tests/Python/CondaTests.cs
```diff
@@ -0,0 +1,241 @@
...
+            var appName = "requirements";
+            var volume = CreateCondaSampleAppVolume(appName);
+            var appDir = volume.ContainerDir;
+            var appOutputDir = "/tmp/app-output";
+            var script = new ShellScriptBuilder()
+                .AddBuildCommand($"{appDir} -o {appOutputDir}")
+                .AddDirectoryExistsCheck($"{appOutputDir}/venv")
+                // Following command makes sure that this package 'matplotlib' is present
+                .AddCommand("source /opt/conda/etc/profile.d/conda.sh")
+                .AddCommand($"conda activate {appOutputDir}/venv")
```
qianz2, line 37: Curious are these commands already included in the image or we required to run these commands every time we build a conda app

images/build/benv.sh
kichalla: This is because npm comes along with Node so we do not need to maintain this list. So if a user is expecting to use a particular npm version, they should just use corresponding Node version. This makes things simpler for us too.  Yes the tests were removed earlier but we missed cleaning up all this stuff.

tests/Oryx.BuildImage.Tests/Python/CondaTests.cs
kichalla: These commands are already part of the generated script too. Here we added this to make sure that a particular package is indeed installed within the virtual environment.

arroyc: so the nightly build/tests and validations tests have successed so that means corresponding tests were already removed ...

images/build/benv.sh
arroyc: the way we set are npm is different now .. for each node_version we check for supporting npm version and set the npm version there so we dont use benv to set npm any more

arroyc (approved):

PR #812, opened 2020-10-08
Updated Python version to 3.9.0

kichalla:


qianz2 (approved):
Could you also add 3.9.0 into `PlatformsCommandTest `under `ListsPlatformsAndVersionsAvailableForDynamicInstall` to make sure it will be in the list, thanks!

kichalla:
> Could you also add 3.9.0 into PlatformsCommandTest under ListsPlatformsAndVersionsAvailableForDynamicInstall to make sure it will be in the list, thanks!

Sure, sounds good...but I think we can the test more robust by making sure that whatever versions are build/acquired (for example, from `versionsToBuild.txt`) are listed. One less thing to maintain...

PR #813, opened 2020-10-09
Updated platform versions

kichalla:


arroyc (approved):

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
```diff
@@ -1220,9 +1220,11 @@ public void Build_ExecutesPreAndPostBuildScripts_WithinBenvContext(string versio
                 () =>
                 {
                     Assert.True(result.IsSuccess);
-                    Assert.Matches($"Pre-build script: /opt/python/{version}/bin/python{version}", result.StdOut);
+                    var semVer = new SemVer.Version(version);
+                    var virtualEnvSuffix = $"{semVer.Major}.{semVer.Minor}";
+                    Assert.Matches($"Pre-build script: /opt/python/{version}/bin/python{virtualEnvSuffix}", result.StdOut);
```
qianz2, line 1225: This test is failing on validation build for my PR as well, seems its path `/opt/python/3.9.0/bin/python3.9` should become `/tmp/oryx/platforms/python/3.9.0/bin/python3.9` for 3.9.0 since it only support dynamic install?

kichalla: Yes, exactly...so to avoid this I am using LTS versions of these versions which are expected to be present under `/opt` path since these are always installed in `lts-versions` image.

vsts/pipelines/templates/_buildimageBasesJobTemplate.yml
```diff
@@ -11,8 +11,6 @@ jobs:
   displayName: ${{ parameters.displayName }}
   timeoutInMinutes: 250
   steps:
-  - template: _cleanImageCacheTemplate.yml
```
qianz2, line 14: Should we still keep this?

arroyc: no :) this needs to be gone

qianz2 (approved):

Issue #801, opened 2020-09-29
Oryx error for Hugo with valid config.toml


joshdcar:
### Bug Report

NOTE: This issue was originally reported at https://github.com/Azure/static-web-apps/issues/160 but was advised it should be moved here. 

- At what date and time did you most recently experience the problem?

9/29/2020

- Where did you experience the problem? E.g. Azure Web Apps, Azure Functions, Azure Container Registry, or offline use.

Azure Static Web Apps

- If your repo is publicly available please share its URL:

https://github.com/joshdcar/azure-static-web-app-hugo-build-fail

- What happened?

I recently changed the theme and the corresponding config.toml on an Azure Static Web App configured for Hugo. The build had previously worked. The build started throwing an "Oops.. An unexpected error has occurred" following the update.  The Hugo site built and ran fine locally and the config.toml was valid. 

- What did you expect or want to happen?

If the config.toml is valid Oryx build should not fail.  Worse case the build should provide a more useful error details to help diagnose the issue. 

- How can we reproduce it?

I discovered the root cause of this issue to be "section" ordering in the config.toml. The configuration itself is valid for Hugo (runs\builds fine locally) and the toml is of course valid but will fail in Oryx.

Oryx throw an error after "Detecting platforms..." when a "sub-section" is defined before the "root" section.  

    [params.search]
     enable = true

    [params]
     author = "my name"

However the following works successfully:

    [params]
     author = "my name" 

    [params.search]
     enable = true

After playing with around a dozen themes I found this pattern to be fairly prevalent.  I agree that logically [params] should appear before [params.search] but in general, if the configuration is valid for Hugo it should IMO work in Oryx \ Azure Static Web Apps as well.  I couldn't figure out how to get some more diagnostic details from Oryx. 

I set up a repo that you can use to reproduce this issue at https://github.com/joshdcar/azure-static-web-app-hugo-build-fail . The repo has the same details in the readme and it's also wired up to an azure static web app where I've "broken" the build on purpose.  If you need additional permissions to the repo or details on the static web app instance please let me know. 

- Do you have log output? Please include between the backticks:

![image](https://user-images.githubusercontent.com/24570240/94476867-12b1c480-019f-11eb-8343-826b9b4d373f.png)

qianz2:
Hi @joshdcar ,
[...]

PR #814, opened 2020-10-09
Fix cli generic error messages

qianz2:
Fix issue #801


src/Detector/Hugo/HugoDetector.cs
```diff
@@ -156,17 +156,7 @@ private bool IsHugoTomlFile(ISourceRepo sourceRepo, params string[] subPaths)
...
-            {
-                tomlTable = ParserHelper.ParseTomlFile(sourceRepo, relativeFilePath);
-            }
-            catch (FailedToParseFileException ex)
-            {
-                _logger.LogError(ex, $"An error occurred when trying to parse file '{relativeFilePath}'.");
-                return false;
-            }
-
+            tomlTable = ParserHelper.ParseTomlFile(sourceRepo, relativeFilePath);
```
qianz2, line 159: In CLI `CommandBase`, we only output specific messages from `InvalidUsageException` so we don't want to early catch `FailedToParseFileException ` in detector here.

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -104,10 +105,23 @@ public IEnumerable<string> SupportedVersions
...
+            try
             {
-                SourceRepo = new Detector.LocalSourceRepo(context.SourceRepo.RootPath),
-            });
+                detectionResult = _detector.Detect(new DetectorContext
+                {
+                    SourceRepo = new Detector.LocalSourceRepo(context.SourceRepo.RootPath),
+                });  
+            }
+            catch (FailedToParseFileException ex)
```
kichalla, line 117: Shouldn't we need this guard for all other platforms too since any detector could throw this exception? That said I wonder if we should have a generic exception that a Detector throws and that we can catch in all platforms? I say this because from a platform's perspective this is an internal detail of the detector and there is a possibility of other kind of exceptions that could be thrown from a detector which also could be interesting for a user to react (ex: fix my toml file)

kichalla: We should add unit tests for each platform to see if we handling this properly. for example, mocking the `IPythonPlatformDetector` to throw and see how `PythonPlatform` reacts

kichalla (changes requested):

src/BuildScriptGenerator/Python/PythonPlatform.cs
qianz2: I agree we should have a generic exception for detector only and will work on adding more tests, thanks!

src/Detector/Hugo/HugoDetector.cs
arroyc: so where are we placing the generalexception in terms of order? can we add few tests where we can validate this ... like exception1 gets caught before exception2 and exception2 gets caught before exception3 and so on

arroyc (changes requested):

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
```diff
@@ -1170,7 +1170,7 @@ public void Build_ExecutesPreAndPostBuildScripts_WithinBenvContext(string versio
             // Act
             var result = _dockerCli.Run(new DockerRunArguments
             {
-                ImageId = Settings.BuildImageName,
+                ImageId = Settings.LtsVersionsBuildImageName,
```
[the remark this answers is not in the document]
qianz2: It's not for this PR specifically, it's for fixing a test our validation build failed.

src/Detector/Hugo/HugoDetector.cs
qianz2: Now we're having Detector level exceptions and Platform level exceptions, we can validate each of them in DetectorTests and PlatformTests. The order will be like, a detector level exception throws and we catch it in a platform and throws a platform level exception there.

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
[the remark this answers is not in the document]
qianz2: will remove this and simply rebase from master.

qianz2:
As our discussion, it may cause multiple platforms detector failure, so close this one.

PR #815, opened 2020-10-13
Update python packaging to support  building universal wheel for python

arroyc:
- [X ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as \#\<issue\_number\>.

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -103,7 +103,6 @@ fi
 	echo $APP_PACKAGES_PATH > $SITE_PACKAGES_PATH"/oryx.pth"
 {{ end }}
 
-packageWheelType={{ PythonPackageWheelProperty }}
 
 {{ if RunPythonPackageCommand }}
 	echo
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -103,12 +103,25 @@ fi
 	echo $APP_PACKAGES_PATH > $SITE_PACKAGES_PATH"/oryx.pth"
 {{ end }}
 
+packageWheelType={{ PythonPackageWheelProperty }}
+
 {{ if RunPythonPackageCommand }}
 	echo
 	echo "Running python packaging commands ...."
 	echo
-	echo "Creating python package wheel ...."
-	$python setup.py sdist --formats=gztar,zip,tar bdist_wheel --universal
+	echo "Determining python package wheel ...."
+
+	{{ if PythonPackageWheelProperty }}
+		echo "Creating universal package wheel ...."
+	{{ end }}
+
+	if [ -z "{{ PythonPackageWheelProperty }}" ]
+	then 
+		echo "Creating non universal package wheel ...."
+		$python setup.py sdist --formats=gztar,zip,tar bdist_wheel
+	else		
+		$python setup.py sdist --formats=gztar,zip,tar bdist_wheel --universal
+	fi
 	echo "Now creating python package egg ...."
 	$python setup.py bdist_egg
 	echo
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippetProperties.cs
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippetProperties.cs
@@ -18,7 +18,8 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Python
             bool enableCollectStatic,
             string compressVirtualEnvCommand,
             string compressedVirtualEnvFileName,
-            bool runPythonPackageCommand)
+            bool runPythonPackageCommand,
+            string pythonPackageWheelProperty = null)
         {
             VirtualEnvironmentName = virtualEnvironmentName;
             VirtualEnvironmentModule = virtualEnvironmentModule;
@@ -28,6 +29,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Python
             CompressVirtualEnvCommand = compressVirtualEnvCommand;
             CompressedVirtualEnvFileName = compressedVirtualEnvFileName;
             RunPythonPackageCommand = runPythonPackageCommand;
+            PythonPackageWheelProperty = pythonPackageWheelProperty;
         }
 
         public string VirtualEnvironmentName { get; set; }
@@ -48,5 +50,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Python
         public string CompressedVirtualEnvFileName { get; set; }
 
         public bool RunPythonPackageCommand { get; set;}
+
+        public string PythonPackageWheelProperty { get; set; }
     }
 }
--- a/src/BuildScriptGenerator/Python/PythonPlatform.cs
+++ b/src/BuildScriptGenerator/Python/PythonPlatform.cs
@@ -175,6 +175,12 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Python
             {
                 throw new InvalidUsageException($"Option '{PythonPackageWheelPropertyKey}' can only have 'universal' as value.");
             }
+            else if (isPythonPackageCommandEnabled &&
+                !string.IsNullOrWhiteSpace(pythonPackageWheelType) &&
+                string.Equals(pythonPackageWheelType.ToLower(), "universal"))
+            {
+                manifestFileProperties[PythonManifestFilePropertyKeys.PackageWheel] = pythonPackageWheelType;
+            }
 
             if (!string.IsNullOrWhiteSpace(packageDir) && !string.IsNullOrWhiteSpace(virtualEnvName))
             {
--- a/src/BuildScriptGenerator/Python/PythonPlatform.cs
+++ b/src/BuildScriptGenerator/Python/PythonPlatform.cs
@@ -165,20 +165,18 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Python
 
             if (!isPythonPackageCommandEnabled && !string.IsNullOrWhiteSpace(pythonPackageWheelType))
             {
-                throw new InvalidUsageException($"Option '{PythonPackageWheelPropertyKey}' cann't exist" +
+                throw new InvalidUsageException($"Option '{PythonPackageWheelPropertyKey}' can't exist" +
                     $"without package command being enabled. Please provide --package along with wheel type");
             }
 
             if (isPythonPackageCommandEnabled &&
-                !string.IsNullOrWhiteSpace(pythonPackageWheelType) &&
-                !string.Equals(pythonPackageWheelType.ToLower(), "universal"))
+                !string.IsNullOrWhiteSpace(pythonPackageWheelType))
             {
-                throw new InvalidUsageException($"Option '{PythonPackageWheelPropertyKey}' can only have 'universal' as value.");
-            }
-            else if (isPythonPackageCommandEnabled &&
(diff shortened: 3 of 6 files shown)


vsts/pipelines/templates/_buildimageBasesJobTemplate.yml
```diff
@@ -11,8 +11,6 @@ jobs:
   displayName: ${{ parameters.displayName }}
   timeoutInMinutes: 250
   steps:
-  - template: _cleanImageCacheTemplate.yml
-
```
arroyc, line 15: @kichalla i removed it, because I saw your Pr got closed and this needs to be removed from master

tests/Oryx.BuildImage.Tests/Python/PythonPackageTest.cs
```diff
@@ -103,7 +103,7 @@ public PythonPackageTest(ITestOutputHelper output) : base(output)
                 .ToString();
 
             // Act
-            var image = Settings.BuildImageWithRootAccess;
+            var image = Settings.LtsVersionsBuildImageWithRootAccess;
```
arroyc, line 106: terrapin use lts-versions image for packaging so tests should refer to lts-versions images

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
```diff
@@ -1124,8 +1124,8 @@ public void Django_CollectStaticFailure_DoesNotFailBuild()
         }
 
         [Theory]
-        [InlineData("3")]
-        [InlineData("2")]
+        [InlineData("3.8.5")]
+        [InlineData("2.7.18")]
         public void Build_ExecutesPreAndPostBuildScripts_WithinBenvContext(string version)
```
arroyc, line 1129: this change is only to get rid of test failures in master .. not related to python packaging

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -103,12 +103,25 @@ fi
...
 {{ if RunPythonPackageCommand }}
 	echo
 	echo "Running python packaging commands ...."
 	echo
-	echo "Creating python package wheel ...."
-	$python setup.py sdist --formats=gztar,zip,tar bdist_wheel --universal
+	echo "Determining python package wheel ...."
+
+	{{ if PythonPackageWheelProperty }}
+		echo "Creating universal package wheel ...."
```
kichalla, line 115: Move just before line 123?

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -31,6 +31,12 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Python
     [BuildProperty(
         TargetPackageDirectoryPropertyKey,
         "If provided, packages will be downloaded to the given directory instead of to a virtual environment.")]
+    [BuildProperty(
+        PythonPackageWheelPropertyKey,
+        "If provided, package wheels will be built with universal flag. For example," +
```
kichalla, line 36: nit: fix the content after "For example"

kichalla: Also do we need a 'value' for this property? since it is either universal wheel or not, we can just have "-p build_universa_wheel"?

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -144,6 +161,26 @@ public PlatformDetectorResult Detect(RepositoryContext context)
             var packageDir = GetPackageDirectory(context);
             var virtualEnvName = GetVirtualEnvironmentName(context);
             var isPythonPackageCommandEnabled = _commonOptions.ShouldPackage;
+            var pythonPackageWheelType = GetPythonPackageWheelType(context);
+
+            if (!isPythonPackageCommandEnabled && !string.IsNullOrWhiteSpace(pythonPackageWheelType))
```
kichalla, line 166: 👍

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -144,6 +161,26 @@ public PlatformDetectorResult Detect(RepositoryContext context)
...
+            {
+                throw new InvalidUsageException($"Option '{PythonPackageWheelPropertyKey}' cann't exist" +
+                    $"without package command being enabled. Please provide --package along with wheel type");
+            }
+
+            if (isPythonPackageCommandEnabled &&
```
kichalla, line 172: Check my earlier comment, since it is either universal or not, we can simplify this

kichalla (changes requested):

src/BuildScriptGenerator/Python/PythonPlatform.cs
arroyc: @kichalla ... imo value gives better flexibility .. in case python come up with some new  option 'foo' in future

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
arroyc: actually if I keep it like this it's easier to write a test and assert on the content.

kichalla (approved):

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
kichalla: Just FYI...I merged my PR...you can now rebase on it.

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -144,6 +161,26 @@ public PlatformDetectorResult Detect(RepositoryContext context)
             var packageDir = GetPackageDirectory(context);
             var virtualEnvName = GetVirtualEnvironmentName(context);
             var isPythonPackageCommandEnabled = _commonOptions.ShouldPackage;
+            var pythonPackageWheelType = GetPythonPackageWheelType(context);
+
+            if (!isPythonPackageCommandEnabled && !string.IsNullOrWhiteSpace(pythonPackageWheelType))
+            {
+                throw new InvalidUsageException($"Option '{PythonPackageWheelPropertyKey}' cann't exist" +
```
cormacpayne, line 168: nit: typo -- `cann't` --> `can't`

src/BuildScriptGenerator/Python/PythonPlatform.cs
cormacpayne: I had the same thought as @kichalla 

@arroyc -- in this case, I was thinking we could add an array of "supported package types", or something of the like, to this class, and reference that here in this description and in the validation checks later in the `GenerateBashBuildScriptSnippet` method.

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -144,6 +161,26 @@ public PlatformDetectorResult Detect(RepositoryContext context)
...
+                    $"without package command being enabled. Please provide --package along with wheel type");
+            }
+
+            if (isPythonPackageCommandEnabled &&
+                !string.IsNullOrWhiteSpace(pythonPackageWheelType) &&
+                !string.Equals(pythonPackageWheelType.ToLower(), "universal"))
```
cormacpayne, line 174: If we went with the array model, this is where we would validate that the given type was in our list.

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -144,6 +161,26 @@ public PlatformDetectorResult Detect(RepositoryContext context)
...
+            {
+                throw new InvalidUsageException($"Option '{PythonPackageWheelPropertyKey}' can only have 'universal' as value.");
+            }
+            else if (isPythonPackageCommandEnabled &&
+                !string.IsNullOrWhiteSpace(pythonPackageWheelType) &&
+                string.Equals(pythonPackageWheelType.ToLower(), "universal"))
```
cormacpayne, line 180: nit: I would consolidate this `else if` with the `if` above, just to clean it up a little:

```cs
if (isPythonPackageCommandEnabled && !string.IsNullOrEmptyWhiteSpace(pythonPackageWheelType))
{
    if (!string.Equals(pythonPackageWheelType.ToLower(), "universal")
    {
        throw new InvalidUsageException($"Option '{PythonPackageWheelPropertyKey}' can only have 'universal' as value.'");
    }
    
    manifestFileProperties[PythonManifestFilePropertyKeys.PackageWheel] = pythonPackageWheelType;
}
```

Also, the logic for the nested `if` will need to change if you do decide to go with the array of supported types 😄

src/BuildScriptGenerator/Python/PythonBashBuildSnippetProperties.cs
```diff
@@ -18,7 +18,8 @@ public class PythonBashBuildSnippetProperties
             bool enableCollectStatic,
             string compressVirtualEnvCommand,
             string compressedVirtualEnvFileName,
-            bool runPythonPackageCommand)
+            bool runPythonPackageCommand,
+            string pythonPackageWheelProperty = null)
```
cormacpayne, line 22: nit: is the `= null` part necessary?

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -103,12 +103,25 @@ fi
 	echo $APP_PACKAGES_PATH > $SITE_PACKAGES_PATH"/oryx.pth"
 {{ end }}
 
+packageWheelType={{ PythonPackageWheelProperty }}
```
cormacpayne, line 106: It doesn't look like `packageWheelType` is used -- am I missing something? ☺️

src/BuildScriptGenerator/Python/PythonBashBuildSnippetProperties.cs
arroyc: in that way we are saying it's an optional argument .. so if it doesn't get provided it will be passed as null .. so in most of the case it will be null ... because packaging will be used by terrapin mostly ..

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
arroyc: no it was a debug piece .. will remove it

src/BuildScriptGenerator/Python/PythonPlatform.cs
arroyc: im not sure what is arraymodel you are referring here

src/BuildScriptGenerator/Python/PythonPlatform.cs
arroyc: same comment about arraymodel

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -57,6 +68,12 @@ internal class PythonPlatform : IProgrammingPlatform
         /// The tar-gz option.
         /// </summary>
         internal const string TarGzOption = "tar-gz";
+
+        /// <summary>
+        /// The Universal Wheel option.
+        /// </summary>
+        internal const string UniversalWheel = "universal";
```
qianz2, line 75: nit: we can move these constants into `PythonConstants.cs`

src/BuildScriptGenerator/Python/PythonPlatform.cs
arroyc: as discussed, at this moment  we would go with non-array model, when the time comes to support new value we can add the array model. Similar thing what we do for zipoption and taroption ... we could have simply used one option and have an array with 2 elements zip and tar ... so im keeping that as it is for now

PR #816, opened 2020-10-13
Updated .NET Core versions and updated apptype to support webapps

kichalla:


src/Detector/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
```diff
@@ -57,72 +57,85 @@ public string GetRelativePathToProjectFile(DetectorContext context)
...
+                    {
+                        return null;
+                    }
+
+                    projectFile = GetProject(blazorWasmProjects);
+                }
+                else if (appType.Contains(Constants.WebApplications))
                 {
-                    projectFile = GetProject(azureBlazorWasmProjects);
+                    if (webAppProjects.Count() == 0)
```
arroyc, line 112: why are we checking the count? so are you saying there is a possibility that apptype will be wrongly set ? like there is no webappproject but apptype is set to webapp

arroyc (approved):
I left some comments/questions/ lets discuss this ... other wise it looks good

src/Detector/DotNetCore/ProjectFileProviders/ProbeAndFindProjectFileProvider.cs
kichalla: Exactly, if the users are saying to look for a particular apptype we are looking only for those and if there are none we should not be returning other app type based projects

PR #820, opened 2020-10-21
Add an option to enable build node monorepo apps

qianz2:


src/BuildScriptGenerator/Node/NodeScriptGeneratorOptions.cs
```diff
@@ -18,5 +18,7 @@ public class NodeScriptGeneratorOptions
         public string NpmRegistryUrl { get; set; }
 
         public string NodeVersion { get; set; }
+
+        public bool EnableNodeMonorepoBuild { get; set; }
```
kichalla, line 22: repo => Repo ?

src/BuildScriptGeneratorCli/Options/NodeScriptGeneratorOptionsSetup.cs
```diff
@@ -25,6 +25,7 @@ public void Configure(NodeScriptGeneratorOptions options)
             options.CustomRunBuildCommand = GetStringValue(SettingsKeys.CustomRunBuildCommand);
             options.PruneDevDependencies = GetBooleanValue(SettingsKeys.PruneDevDependencies);
             options.NpmRegistryUrl = GetStringValue(SettingsKeys.NpmRegistryUrl);
+            options.EnableNodeMonorepoBuild = GetBooleanValue(SettingsKeys.EnableNodeMonorepoBuild);
```
kichalla, line 28: Since this is `false` by default, shouldn't you update your `lerna` and `lage` tests to make this `true` for those to pass?

kichalla (changes requested):

src/BuildScriptGeneratorCli/Options/NodeScriptGeneratorOptionsSetup.cs
kichalla: Also could you please run a Nightly build?

src/BuildScriptGenerator/Node/NodeScriptGeneratorOptions.cs
qianz2: _monorepo_ is a word, so i think keep it as same will be good.

kichalla (approved):

Issue #752, opened 2020-08-31
Oryx fails to detect platforms in Hugo blog


alexhudson:
### Bug Report

I created a new Hugo-based blog this morning, and am trying to deploy it via Azure Static Apps preview.

During the CI/CD, I get this response:

```
App Directory Location: '/' was found.
[WARNING] Api Directory Location: 'api' could not be found. Azure Functions will not be created.
Starting to build app with Oryx
Azure Static Web Apps utilizes Oryx to build both static applications and Azure Functions. You can find more details on Oryx here: https://github.com/microsoft/Oryx
Oryx will build app with the following custom override command: hugo
---Oryx build logs---


Operation performed by Microsoft Oryx, https://github.com/Microsoft/Oryx
You can report issues at https://github.com/Microsoft/Oryx/issues

Oryx Version: 0.2.20200719.1, Commit: 80603ef82312ab8b6d950149f332638c5707de74, ReleaseTagName: 20200719.1

Build Operation ID: |AzuAF2Uuhv8=.8a66635f_
Repository Commit : efa54c9c1e4179fd62231c47c8852989db398022

Detecting platforms...
Error: Oops... An unexpected error has occurred.
```

There seems to be an issue with the platform detection, but I don't seem to be able to enable Oryx debug mode to get a better backtrace, and there doesn't appear to be an obvious workaround. I'm not using Oryx to do the build - I have these steps setup:

```
      - name: Build
        run: hugo
      - name: Build And Deploy
        id: builddeploy
        uses: Azure/static-web-apps-deploy@v0.0.1-preview
        with:
          azure_static_web_apps_api_token: ${{ secrets.AZURE_STATIC_WEB_APPS_API_TOKEN }}
          repo_token: ${{ secrets.GITHUB_TOKEN }} # Used for Github integrations (i.e. PR comments)
          action: "upload"
          ###### Repository/Build Configurations - These values can be configured to match you app requirements. ######
          # For more information regarding Static Web App workflow configurations, please visit: https://aka.ms/swaworkflowconfig
          app_build_command: hugo
          app_location: "/" # App source code path
          api_location: "api" # Api source code path - optional
          app_artifact_location: "public" # Built app content directory - optional
          ###### End of Repository/Build Configurations ######
```

... and the build process is working fine; I just need Oryx to do the upload step. Even overriding the app build command, it's still doing the platform detection and I get the same error.

kichalla:
[...]

PR #821, opened 2020-10-22
Update toml file parser for Hugo apps

qianz2:
#801 #752

### DIFF
--- a/src/Detector/ParserHelper.cs
+++ b/src/Detector/ParserHelper.cs
@@ -33,13 +33,7 @@ namespace Microsoft.Oryx.Detector
             try
             {
                 // Gets a syntax tree of the TOML text
-                var doc = Toml.Parse(tomlContent); // returns a DocumentSyntax
-                // Check for parsing errors with doc.HasErrors and doc.Diagnostics
-                // doc.HasErrors => throws an exception
-                if (doc.HasErrors)
-                {
-                    throw new Exception("TOML file content has parsing errors.");
-                }
+                var doc = Toml.Parse(tomlContent);
                 // Gets a runtime representation of the syntax tree
                 var table = doc.ToModel();
                 return table;
--- a/src/Detector/ParserHelper.cs
+++ b/src/Detector/ParserHelper.cs
@@ -7,8 +7,9 @@ using System;
 using System.IO;
 using Microsoft.Oryx.Detector.Exceptions;
 using Microsoft.Oryx.Detector.Resources;
-using Nett;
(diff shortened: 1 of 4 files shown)


src/Detector/ParserHelper.cs
```diff
@@ -31,7 +32,17 @@ public static TomlTable ParseTomlFile(ISourceRepo sourceRepo, string filePath)
 
             try
             {
-                return Toml.ReadString(tomlContent);
+                // Gets a syntax tree of the TOML text
+                var doc = Toml.Parse(tomlContent); // returns a DocumentSyntax
```
kichalla, line 36: Can you kick off a nightly build too?

src/Detector/ParserHelper.cs
```diff
@@ -31,7 +32,17 @@ public static TomlTable ParseTomlFile(ISourceRepo sourceRepo, string filePath)
...
+                // Check for parsing errors with doc.HasErrors and doc.Diagnostics
+                // doc.HasErrors => throws an exception
+                if (doc.HasErrors)
```
kichalla, line 39: Is there a way to get list of errors that this parser gives back? if yes, are those errors captured in an exception? if yes we want to throw that in a new FailedToParseFileException because we want to know what is the actual error that the parser faced and not the following generic message because it does not give us an idea of what the problem is with the file.

src/Detector/ParserHelper.cs
```diff
@@ -19,8 +20,8 @@ namespace Microsoft.Oryx.Detector
     internal static class ParserHelper
     {
         /// <summary>
-        /// Parse a .toml file into a TomlTable from the Nett library.
-        /// See https://github.com/paiden/Nett for more information.
+        /// Parse a .toml file into a TomlTable from the Tomlyn library.
+        /// See https://github.com/xoofx/Tomlyn for more information.
```
kichalla, line 24: Is this just a new version of `Nett` tool?

kichalla (changes requested):

src/Detector/ParserHelper.cs
qianz2: No, this is a new tool which has similar functions as `Nett`

src/Detector/ParserHelper.cs
qianz2: I see, yeah it has a diagnostics property which contains detail error information, I didn't try it yet, I'll try it and if it works, then pass the details into `FailedToParseFileException `.

kichalla (approved):

PR #830, opened 2020-10-29
Unlock rogue coverage reports

dorfire:
these fail the Checkout step in DevOps if left with wrong permissions.

### DIFF
--- a/vsts/pipelines/templates/_agentCleanUpJobTemplate.yml
+++ b/vsts/pipelines/templates/_agentCleanUpJobTemplate.yml
@@ -13,8 +13,8 @@ jobs:
 
   steps:
 
-  - script: echo `pwd`; ls -la; find . -type d -name coverage.cobertura.xml -print -exec chmod +w {} \;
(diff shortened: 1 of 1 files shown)


arroyc:
I dont think we should add any job or steps to unlock specific files in build agents ..

vsts/pipelines/templates/_agentCleanUpJobTemplate.yml
```diff
@@ -2,12 +2,19 @@ parameters:
...
   displayName: ${{ parameters.agentName }}
   pool:
     name: OryxLinux
     demands:
     - agent.name -equals ${{ parameters.agentName }}
   timeoutInMinutes: 30
+
   steps:
+
+  - script: echo Cleaning `pwd`; find . -type d -name coverage.cobertura.xml -print -exec chmod +w {} \;
```
arroyc, line 16: what is this xml file?? how does it get generated?imo if we really need to clean it .. we should clean it where it gets generated ..

arroyc (changes requested):

PR #834, opened 2020-11-04
Include artifact feed prerequisites in the image

qianz2:


images/build/Dockerfiles/ltsVersions.Dockerfile
```diff
@@ -138,6 +138,9 @@ RUN set -ex \
     # Upgrade system python
     && pip install --upgrade cython \
     && pip3 install --upgrade cython \
+    # https://github.com/microsoft/artifacts-keyring
+    # provides auth for publishing or consuming Python packages to or from Azure Artifacts feeds.
+    && pip install twine keyring artifacts-keyring \
```
arroyc, line 143: should it be pip3?

arroyc (approved):

images/build/Dockerfiles/ltsVersions.Dockerfile
qianz2: I think it's `pip` as the docs specify to use this command.

arroyc: Also add a test for build image  to make sure this is also accessible from full build image

PR #841, opened 2020-11-09
Run npm prepublish script

dorfire:


src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -148,6 +148,7 @@ echo
 echo "Running custom packaging scripts that might exist..."
 echo
 npm run package || true
+npm run prepublishOnly || true
```
kichalla, line 151: I am confused what does `npm run package` do in context of `prepublishOnly`
image

kichalla (approved):

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
arroyc: According to the official doc in npm life cycle npm publish comes before npm run package.  

> If you need to perform operations on your package before it is used, in a way that is not dependent on the operating system or architecture of the target system, use a prepublish script. This includes tasks such as:
> 
> Compiling CoffeeScript source code into JavaScript.
> Creating minified versions of JavaScript source code.
> Fetching remote resources that your package will use.

Based on this my assumption is terrapin needs some preparation in terms of configuring env or converting non-javascript source-code into javascript etc. 
Whatever is the use case .. IMO it's always best to write few test-cases which clarifies the use-case or at-least leave some comments for documentation purpose.

dorfire: These are two similar but distinct steps:

First, Oryx attempts to build a package. The build process may include custom steps; _Oryx tries to support as many of those as possible_ - e.g. if lerna is detected, it is invoked; if a `build:azure` script is detected, it is run; etc.

Then, when the `--package` flag is passed, Oryx attempts to also **package** the results of the build.
Here, just like in the build stage, we should try to support common custom steps.
The `package` script is one example.
This PR adds another one I noticed - since Oryx **does not run `npm publish`**, the `prepublishOnly` script isn't run.
Running it explicitly will result in more "correct" builds, compared to what people publish on the public registry.

Issue #823, opened 2020-10-22
Use npm ci instead of npm install


jpoiron:
### Feature Request

Hi, I was wondering if it's possible to use `npm ci` instead of `npm install` when building node.js apps. 

It looks like I could possibly use a [custom build command](https://docs.microsoft.com/en-us/azure/static-web-apps/github-actions-workflow#custom-build-commands) such as `npm ci && npm run build` but if I'm understanding correctly, `npm install` will still always be run automatically first and it might update my dependencies and package-lock.json on me. My goal is to have predictable builds and I'm worried that some sneaky issues could get into other environments if npm install is used.

Thanks!

qianz2:
Hi @jpoiron, thanks for providing us feedbacks. Your understanding is correct, currently we always run `npm install` command before any custom build command. We'll analyze and discuss within team about this feature request, and get back to you. Thanks!

HarryGifford:
@qianz2 I'm not the original person, but it would be useful for my use case to be able to override the `npm install` command. I'm using [PNPM](https://pnpm.js.org/) for a project instead of NPM due to some improved features such as workspaces and would like to be able to use it to build Azure static web apps. Running `npm install` causes the build to fail, since NPM doesn't know where to find the workspace packages (since they're not published.)

qianz2:
Close this as we have added an environment variable `CUSTOM_BUILD_COMMAND`, which enables to override npm install commands. It's in the latest release. Please check it out.

jpoiron:
That's great! I'll check it out soon. If there are any docs you can point me to that would be great. But it sounds like to accomplish what I want I just need to add `CUSTOM_BUILD_COMMAND` to the `env` section of the Azure/static-web-apps-deploy GitHub action and set it to `npm ci && npm run build`. Does that sound right? 

Also, do I need to bump the version of the GitHub action? Currently I have it set to`Azure/static-web-apps-deploy@v0.0.1-preview` and I don't see any other versions within the GitHub actions UI.

PR #844, opened 2020-11-10
Support custom build commands or script for Nodejs apps

qianz2:

Issue #823


src/BuildScriptGenerator/Node/NodeScriptGeneratorOptions.cs
```diff
@@ -13,6 +13,12 @@ public class NodeScriptGeneratorOptions
         /// </summary>
         public string CustomRunBuildCommand { get; set; }
 
+        /// <summary>
+        /// Gets or sets custom build command or script that will run without 'npm install' in the generated
+        /// build script.
+        /// </summary>
```
arroyc, line 19: I'm bit confused .. what is the difference between a customrunbuildcommand and a custombuildcommand

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -93,7 +93,12 @@ fi
 
 cd "$SOURCE_DIR"
 
-{{ if CustomRunBuildCommand | IsNotBlank }}
+{{ if CustomBuildCommand | IsNotBlank }}
+	echo
+	echo "Running '{{ CustomBuildCommand }}'..."
+	echo
+	{{ CustomBuildCommand }}
+{{ else if CustomRunBuildCommand | IsNotBlank }}
```
arroyc, line 101: so you are prioritizing custombuildcommand over customrunbuildcommand .... Can you site an example/use case?

arroyc: I think we need to understand it better in terms of what we are looking to do here and what npm already offers .. may be we  can leverage npm to do this what we are trying to do here.
Let's discuss this over standup ...

arroyc (changes requested):

arroyc:
so let's say we have this in the package.json ... 

> "scripts": {
>   "lint": "jshint **",
>   "build:css": "stylus assets/styles/main.styl > dist/main.css",
>   "build:js": "browserify assets/scripts/main.js > dist/main.js",
>   "build": "npm run build:css && npm run build:js",
>   "prebuild:js": "npm run lint"
> }

https://docs.npmjs.com/cli/v6/using-npm/scripts

I want to see how the customrunbuild or custombuild command helps us compared to just running npm install what we are doing today.

src/BuildScriptGenerator/Node/NodeScriptGeneratorOptions.cs
qianz2: So the difference is basically custombuildcommand will override any PackageInstallCommand Oryx provided(either could be 'npm install' or 'yarn install')

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
qianz2: https://docs.npmjs.com/cli/v6/commands/npm-ci
Currently I have two examples, one is usage of npm ci command, which allows to install dependencies in automated environment(more strict and helps catch inconsistencies). The other one is an app that containing a `Makefile`, which has its own build command, like `make build`.

arroyc (approved):

PR #845, opened 2020-11-11
Move installation script to a separate file

kichalla:

### DIFF
--- a/images/build/Dockerfiles/Dockerfile
+++ b/images/build/Dockerfiles/Dockerfile
@@ -47,21 +47,24 @@ RUN set -ex \
     && ln -s $DOT_NET_CORE_22_SDK_VERSION 2.2 \
     && cd $imagesDir \ 
     && . $buildDir/__nodeVersions.sh \
-    && $imagesDir/installPlatform.sh nodejs $NODE8_VERSION \
-    && $imagesDir/installPlatform.sh nodejs 8.0.0 \
-    && $imagesDir/installPlatform.sh nodejs 8.1.4 \
-    && $imagesDir/installPlatform.sh nodejs 8.2.1 \
-    && $imagesDir/installPlatform.sh nodejs 8.8.1 \
-    && $imagesDir/installPlatform.sh nodejs 8.9.4 \
-    && $imagesDir/installPlatform.sh nodejs 8.11.4 \
-    && $imagesDir/installPlatform.sh nodejs 8.12.0 \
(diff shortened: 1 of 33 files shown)


images/build/Dockerfiles/ltsVersions.Dockerfile
```diff
@@ -171,6 +174,8 @@ RUN set -ex \
         libonig-dev \
     && rm -rf /var/lib/apt/lists/* \
     && cp -f $imagesDir/build/benv.sh /opt/oryx/benv \
+    && cp -f $imagesDir/installPlatform.sh /opt/oryx/installPlatform \
```
qianz2, line 177: So installPlatform will act like a command same as benv in the container? Are we able to run command like 'installPlatform -p python -v 3.9' inside the image?

kichalla: `installPlatform` will be used to install platform versions either when building the image in our CI or even when dynamically installing platforms during customer app's build time too. Once dynamic install is approved for runtime images, we can use this script to install dynamically in runtime images too

arroyc:
this concept is scalable which is good, only thing is we may break things  here and there because this is mostly changes in the bash script  ... I would trigger a nightly build to make sure eveything is ok

images/build/Dockerfiles/Dockerfile
```diff
@@ -47,21 +47,24 @@ RUN set -ex \
...
-    && $imagesDir/installPlatform.sh nodejs 8.12.0 \
-    && $imagesDir/installPlatform.sh nodejs 8.15.1 \
-    && $imagesDir/installPlatform.sh nodejs 9.4.0 \
-    && $imagesDir/installPlatform.sh nodejs 10.1.0 \
-    && $imagesDir/installPlatform.sh nodejs 10.10.0 \
-    && $imagesDir/installPlatform.sh nodejs 10.14.2 \
-    && $imagesDir/installPlatform.sh nodejs 10.21.0 \
-    && $imagesDir/installPlatform.sh nodejs 12.18.2 \
+    && . $buildDir/__sdkStorageConstants.sh \
+    && export DYNAMIC_INSTALL_ROOT_DIR="/opt" \
```
arroyc, line 51: just curious .. aren't we using /tmp for this location? why choosing /opt?

kichalla: This is for baked in platforms in the image which we install under `/opt`

PR #856, opened 2020-11-17
Build ubuntu sdks for all oryx platforms

arroyc:


platforms/php/versionsToBuild_11172020.txt
```diff
@@ -0,0 +1,15 @@
+# version, sha, gpg keys, dockerfile
```
kichalla, line 1: I did not get what is this file for vs. the earlier file

platforms/python/prereqs/build.sh
```diff
@@ -9,6 +9,16 @@ set -ex
 wget https://www.python.org/ftp/python/${PYTHON_VERSION%%[a-z]*}/Python-$PYTHON_VERSION.tar.xz -O /python.tar.xz
 wget https://www.python.org/ftp/python/${PYTHON_VERSION%%[a-z]*}/Python-$PYTHON_VERSION.tar.xz.asc -O /python.tar.xz.asc
 
+debianFlavor=$DEBIAN_FLAVOR
+pythonSdkFileName=""
+
+if [ -z "$debianFlavor" ]; then
```
kichalla, line 15: In this block should `$version` actually be `$PYTHON_VERSION`?

platforms/python/versionsToBuild_11172020.txt
```diff
@@ -0,0 +1,38 @@
+# version, gpg keys, dockerfile
```
kichalla, line 1: Same here. Why is the file suffixed with `_11172020`?

kichalla (approved):

vsts/pipelines/PlatformBinaries/php.yml
```diff
@@ -4,14 +4,25 @@ variables:
 stages:
 - stage: Build
   jobs:
-  - job: Php
+  - job: Php_Stretch_SDK
```
kichalla, line 7: Seems like you might need to fix in few other places for this to work as expected.

Below link will give you the list of blobs in the 'python' container:
https://oryxsdksdev.blob.core.windows.net/python?restype=container&comp=list&include=metadata

I am highlighting the missing details in the new uploaded tarball when compared to existing tarballs

image

platforms/python/versionsToBuild_11172020.txt
arroyc: I only wanted to build current versions of python ... to avoid issues with not finding unsupported old versions of python. So I kept it as backup. Anyway we have corresponding stretch sdks already available, until unless a patch is required we wont need them

vsts/pipelines/PlatformBinaries/php.yml
arroyc: Thanks for pointing that out .. I will rectify it .. good thing is I got a successful build after some tries :) .. this

platforms/php/versionsToBuild_11172020.txt
arroyc: yeah ... so this file I created so that I can build latest sdks of python and php to limit the scope (i.e. only what i need), earlier it was trying to build all the versions and some of them (read unsupported old version e.g. php 5, php 7, python 2 etc) was throwing error, so I got rid of them by creating a new version of this file. From now onward we will probably not be needing to build old sdks (which is already present in storage account) of stretch and only be adding future versions. So, we may not be breaking anything. But I still kept an copy  in case we need to refer old sdks etc.

build/buildPlatformBinaries.sh
```diff
@@ -9,29 +9,30 @@ set -e
...
-		"$platformsDir/python/buildPython.sh"
+		"$platformsDir/python/buildPython.sh" $DEBIAN_FLAVOR_TO_BUILD
 		;;
 	'php')
-		"$platformsDir/php/buildPhp.sh"
+		"$platformsDir/php/buildPhp.sh" $DEBIAN_FLAVOR_TO_BUILD
 		;;
 	'nodejs')
-		"$platformsDir/nodejs/getNode.sh"
+		"$platformsDir/nodejs/getNode.sh" $DEBIAN_FLAVOR_TO_BUILD
```
qianz2, line 26: Is this PR for building ubuntu sdks for all platforms, If so can we change the PR name

platforms/dotnet/versionsToBuild.txt
```diff
@@ -1,54 +1,9 @@
 # version, sha, download url
-1.1.14, E36AA4B278B936E1E2989B5E556F4EC3AEF2CB1EBB5ABD1BD6011D511873FC3E72101D035CF4DAF1563C7A3C43C193863D38CBF24288D375DD143AB42AFAE524, https://dotnetcli.blob.core.windows.net/dotnet/Sdk/1.1.14/dotnet-dev-debian.9-x64.1.1.14.tar.gz
 
-2.1.801, 5781C6E254270803D9BB2C29B1555FB9DB258BF99BA2EFA2B457CA28226C70D6305C9B7C141CBF49ADE4915E136FAE941727ADFC60373E4822C52BF0E92C451B,
```
qianz2, line 4: wondering why delete these versions?

platforms/nodejs/Dockerfile
```diff
@@ -1,6 +1,10 @@
-FROM buildpack-deps:stretch
+ARG DEBIAN_FLAVOR
+FROM buildpack-deps:${DEBIAN_FLAVOR}
+ARG DEBIAN_FLAVOR
```
qianz2, line 3: duplicate arguments?

platforms/nodejs/Dockerfile
```diff
@@ -1,6 +1,10 @@
...
+ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
+        make \
```
qianz2, line 7: I think we already have 'make' tool in nodejs images while I tested, we can still add here to make sure.

platforms/nodejs/Dockerfile
arroyc: no .. we would need one after the from as well ... first arg is to be used in the from statement second one will be used for env variable

platforms/dotnet/versionsToBuild.txt
arroyc: as discussed we wont be needing them in future .. so do not want to build them for buster and ubuntu versions .. we already have them in stretch

platforms/nodejs/Dockerfile
arroyc: this is ubuntu image .. the image you tested comes from a different image base .. so it was throwing error as it didn't have make .. i had to install it

PR #858, opened 2020-11-18
Added tests to verify dynamic install of runtimes in runtime images

kichalla:


src/startupscriptgenerator/src/common/setupEnvironment.go
```diff
@@ -29,6 +30,14 @@ func SetupEnv(script string) {
 }
 
 func GetSetupScript(platformName string, version string, installationDir string) string {
+	dowloadSentinelFileName := ".oryx-sentinel"
+	sentinelFilePath := path.Join(installationDir, dowloadSentinelFileName)
```
qianz2, line 34: wondering what is sentinel file used for runtime image?

tests/Oryx.Integration.Tests/Python/PythonDynamicInstallationTest.cs
```diff
@@ -129,7 +131,7 @@ public async Task CanBuildAndRunPythonApp_UsingScriptCommandAndSetEnvSwitch()
                .ToString();
             var runScript = new ShellScriptBuilder()
                 .AddDefaultTestEnvironmentVariables()
-                .SetEnvironmentVariable(SettingsKeys.EnableDynamicInstall, true.ToString())
+                .AddCommand($"oryx setupEnv -appPath {appOutputDir}")
```
qianz2, line 134: remove this line since you already setup in the dynamic runtime image?

src/startupscriptgenerator/src/common/setupEnvironment.go
kichalla: It is basically a marker file to indicate that the earlier download of the platform runtime was indeed successful. If you notice the generated script for the file download, we create this file only after the download has completed. This is just a gate to handle scenarios where users might stop the download of the runtime midway resulting in an inconsistent state of the runtime.

tests/Oryx.Integration.Tests/Python/PythonDynamicInstallationTest.cs
kichalla: Yes, that is right.

arroyc (approved):
can you trigger a  nightly build before merging to master?

qianz2 (approved):

PR #867, opened 2020-12-02
add buster platform binaries

arroyc:

### DIFF
--- a/vsts/pipelines/PlatformBinaries/dotnetcore.yml
+++ b/vsts/pipelines/PlatformBinaries/dotnetcore.yml
@@ -4,6 +4,16 @@ variables:
 stages:
 - stage: Build
   jobs:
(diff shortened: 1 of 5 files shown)


vsts/pipelines/PlatformBinaries/dotnetcore.yml
```diff
@@ -13,7 +23,7 @@ stages:
       parameters:
         platformName: 'dotnet'
         debianFlavor: 'stretch'
-
+
```
qianz2, line 26: nit: space

qianz2 (commented):
Could you trigger the platform binary pipelines for all platforms? Just want to make sure the buster based sdks are actually in the storage account, you could add tests to verify it in `StorageAccountSanityTestBase` under integration tests.

qianz2 (approved):

PR #871, opened 2020-12-03
fix security build failures

arroyc:
as mentioned in the title

### DIFF
--- a/vsts/pipelines/templates/_securityChecks.yml
+++ b/vsts/pipelines/templates/_securityChecks.yml
@@ -35,7 +35,14 @@ steps:
     BinSkim: true
     CredScan: true
 
-- task: securedevelopmentteam.vss-secure-development-tools.build-task-antimalware.AntiMalware@3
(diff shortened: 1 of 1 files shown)


qianz2 (approved):

vsts/pipelines/templates/_securityChecks.yml
```diff
@@ -35,7 +35,14 @@ steps:
...
-- task: securedevelopmentteam.vss-secure-development-tools.build-task-antimalware.AntiMalware@3
+- task: AntiMalware@3
   displayName: 'Run Antimalware Scan'
   inputs:
-    EnableServices: true
\ No newline at end of file
+    InputType: 'Basic'
+    ScanType: 'CustomScan'
+    EnableServices: true
+    TreatSignatureUpdateFailureAs: 'Warning'
```
arroyc, line 44: i suspect this is why it was failing .. I created a new task in my  own test build definition and it seems these parameters came along ... so changed our template

PR #876, opened 2020-12-03
change db tests to test buster buildimage against buster runtime python

arroyc:
as mentioned in the title


qianz2 (approved):
Looks good, waiting for a success nightly build

PR #888, opened 2020-12-14
Add dotnet monitor tool to dotnet 3.1 and 5.0 runtime images

qianz2:


arroyc (approved):

arroyc:
how much size did it increase?

qianz2:
> how much size did it increase?

it's now 728mb compared to 717mb, so it increases 11MB

Issue #885, opened 2020-12-13
Github azure action with Lerna: lerna: command not found


pkmelee337:
### Bug Report

**At what date and time did you most recently experience the problem?**
13-12-2020 20:40UTC

**- Where did you experience the problem?**
Github actions on the action Azure/static-web-apps-deploy@v0.0.1-preview

**What happened?**
I'm trying to build my lerna repository, with an react-scripts web project and an api, but I can't get it to work. I looked for examples, but the only example I was able to find was [this one](https://github.com/microsoft/Oryx/tree/6eb32856331b221ffe4f5d0c32a175afa97a2573/tests/SampleApps/nodejs/monorepo-lerna-npm), but my repository was already looking like this in the basics. The main problem I'm struggling with is that whatever options I put at app_location and/or output_location I'm always gettings the error `/tmp/BuildScriptGenerator/45f30815be874370b2dcb47c5308aaaa/build.sh: line 223: lerna: command not found`.

I think this problem is easily reproduced by creating a simple lerna repository, but I'm not sure because I'm working from azure and github actions.

My gtihub actions is looking as followed:
```
name: Azure Static Web Apps CI/CD

on:
  push:
    branches:
      - main
  pull_request:
    types: [opened, synchronize, reopened, closed]
    branches:
      - main

jobs:
  build_and_deploy_job:
    if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.event.action != 'closed')
    runs-on: ubuntu-latest
    name: Build and Deploy Job
    steps:
      - uses: actions/checkout@v2
        with:
          submodules: true
      - uses: actions/setup-node@v1
        with:
          node-version: 12.x
          registry-url: https://registry.npmjs.org/

      # - name: Install
      #   run: yarn install

      # - name: Build and test
      #   run: |
      #     npx lerna run build
      #     # npx lerna run test testing not yet implemented

      - name: Build And Deploy
        id: builddeploy
        uses: Azure/static-web-apps-deploy@v0.0.1-preview
        with:
          azure_static_web_apps_api_token: ${{ secrets.AZURE_STATIC_WEB_APPS_API_TOKEN_WITTY_MEADOW_060764403 }}
          repo_token: ${{ secrets.GITHUB_TOKEN }} # Used for Github integrations (i.e. PR comments)
          action: "upload"
          ###### Repository/Build Configurations - These values can be configured to match you app requirements. ######
[...]

PR #894, opened 2020-12-17
Add lerna tool to Jamstack build image

qianz2:


qianz2:
For issue #885

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -4,11 +4,17 @@ ARG DEBIAN_FLAVOR
...
 ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
     ORYX_BUILDIMAGE_TYPE="jamstack" \
-    PATH="/usr/local/go/bin:$PATH"
+    PATH="/opt/nodejs/lts/bin:/usr/local/go/bin:$PATH"
 
 COPY --from=support-files-image-for-build /tmp/oryx/ /tmp
 RUN oryx prep --skip-detection --platforms-and-versions nodejs=12 \
     && echo "jamstack" > /opt/oryx/.imagetype \
+    && . /tmp/build/__nodeVersions.sh \
+    && /tmp/images/installPlatform.sh nodejs $NODE14_VERSION \
```
arroyc, line 13: why do we need this? jamstack image should be minimal ... node 14 is available for dynamic install anyways

arroyc (changes requested):

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
qianz2: It doesn't have NPM which needed to install lerna, to have NPM it needs to install one node version on disk inside the image, so I installed node 14, but curious if there has any other ways..

arroyc: In that case i am just curious if we can solve this by using oryx prep command to install node-14 as  part of the github-action and then the user can use some sort of pre build script to install lerna .... I mean this solution of course works ... but it will add some space .. so i would confirm this with jamstack team (mitchell) if he is fine with this extra space ..

qianz2: I see, I want to try this way locally to see if it works.. if it's working then we can discuss with jamstack team with it

qianz2:
As discussed, we prefer to keep this image minimal for Jamstack team, and found the workaround to fix the issue, so close this PR.

PR #898, opened 2021-01-05
Stop pushing to old ACRs

arroyc:
as mentioned in the title

### DIFF
--- a/vsts/pipelines/templates/_releaseBaseImagesStepTemplate.yml
+++ b/vsts/pipelines/templates/_releaseBaseImagesStepTemplate.yml
@@ -30,41 +30,11 @@ steps:
     TargetFolder: '$(Build.ArtifactStagingDirectory)'
 
 - task: ms-devlabs.utilitytasks.task-Shellpp.Shell++@0
-  displayName: 'Pull and create release tags for staging ACR'
-  inputs:
-    type: FilePath
-    scriptPath: ./vsts/scripts/tagBaseImagesForRelease.sh
-    args: '${{ parameters.baseImageName }} ${{ parameters.acrProdName }}'
-
-- task: ms-devlabs.utilitytasks.task-Shellpp.Shell++@0
   displayName: 'Pull and create release tags for PME staging ACR'
   inputs:
     type: FilePath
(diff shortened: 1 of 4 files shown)


vsts/pipelines/templates/_releaseBaseImagesStepTemplate.yml
```diff
@@ -29,43 +29,13 @@ steps:
...
 
-- task: ms-devlabs.utilitytasks.task-Shellpp.Shell++@0
-  displayName: 'Pull and create release tags for staging ACR'
-  inputs:
-    type: FilePath
-    scriptPath: ./vsts/scripts/tagBaseImagesForRelease.sh
-    args: '${{ parameters.baseImageName }} ${{ parameters.acrProdName }}'
-
 - task: ms-devlabs.utilitytasks.task-Shellpp.Shell++@0
   displayName: 'Pull and create release tags for PME staging ACR'
```
qianz2, line 33: so are we now only push to pme ACRs and will remove all non-pme ACRs?

qianz2: please make sure those build/runtime images under old ACRs will not be used in the future

qianz2 (commented):
Could you also trigger a nightly build afterwards, since it involved in some script changes. btw, looks that the baseImages-ruby pipelines automated triggered, I'll try to make it stop :)

vsts/pipelines/templates/_releaseBaseImagesStepTemplate.yml
arroyc: we dont use them anymore ... it was just sitting there for no reason

arroyc:
> Could you also trigger a nightly build afterwards, since it involved in some script changes. btw, looks that the baseImages-ruby pipelines automated triggered, I'll try to make it stop :)

i can surely trigger .. but these scripts won't do much in nightly build though .. as they are only used during ci build (pulling and tagging etc). I can trigger base image builds tough ..

qianz2 (approved):

PR #903, opened 2021-01-15
Update dotnetcore 5 with security patch

arroyc:
as title says


src/BuildScriptGenerator/Ruby/RubyVersions.cs
```diff
@@ -0,0 +1,15 @@
+// This file was auto-generated from 'constants.yaml'. Changes may be overridden.
+
+namespace Microsoft.Oryx.BuildScriptGenerator.Ruby
```
qianz2, line 3: could you revert this change

src/startupscriptgenerator/src/common/consts/runtime_dynamic_installation_constants.go
```diff
@@ -5,5 +5,4 @@ package consts
 const NodeInstallationDir string = "/tmp/oryx/nodejs"
 const DotNetCoreInstallationDir string = "/tmp/oryx/dotnet"
 const PythonInstallationRootDir string = "/opt/python"
-const RubyInstallationRootDir string = "/opt/ruby"
```
qianz2, line 8: same here

qianz2 (approved):

PR #908, opened 2021-01-20
Support Php 8.0 and update other versions

qianz2:
- Enable dynamic install php 8.0.1 sdks
- Add php 8.0 buster lts version build image and runtime image
- Update php and php-fpm 7.3, 7.4 to latest stable versions


images/build/Dockerfiles/Dockerfile
```diff
@@ -103,6 +103,7 @@ RUN set -ex \
     && $imagesDir/installPlatform.sh php $PHP70_VERSION \
     && $imagesDir/installPlatform.sh php $PHP72_VERSION \
     && $imagesDir/installPlatform.sh php $PHP74_VERSION \
+    && $imagesDir/installPlatform.sh php $PHP80_VERSION \
```
arroyc, line 106: we need to ask app service team if we  need to bake php 8 in the image or we should make it available only for dynamic install? ... also php_8 is buster based i believe .. so  even if we keep it in build image we may have problem running it in buster runtime

arroyc (changes requested):
can you add some tests?

images/build/Dockerfiles/Dockerfile
qianz2: As our discussion today, we'll make php 8 dynamic install and add it to the buster based build image.

images/build/php/prereqs/build.sh
```diff
@@ -83,6 +83,11 @@ else
 	apt-get install -y --no-install-recommends libargon2-dev
 fi
 
+if [ $PHP_MAJOR == '8' ]; then
+	apt-get update
+	apt-get install -y --no-install-recommends libonig-dev
```
arroyc, line 88: i see libonig-dev is already there for php 7.4+ .. so can you get rid of this extra if and use it at line 64?

images/build/php/prereqs/build.sh
```diff
@@ -120,6 +125,10 @@ if [ $PHP_MAJOR == '7' ] && [ $PHP_MINOR != '0' ]; then
 	fi
 fi
 
+if [ $PHP_MAJOR == '8' ]; then
+	versionConfigureArgs='--with-password-argon2 --with-sodium=shared --with-pear'
```
arroyc, line 129: same ... use it with line 119

platforms/php/prereqs/build.sh
```diff
@@ -106,6 +106,11 @@ else
 	DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends libargon2-dev 
 fi
 
+if [ $PHP_MAJOR == '8' ]; then
+	apt-get update
+	apt-get install -y --no-install-recommends libonig-dev
```
arroyc, line 111: why do we need it here? use in line 87

platforms/php/prereqs/build.sh
```diff
@@ -138,7 +143,12 @@ if [ $PHP_MAJOR == '7' ] && [ $PHP_MINOR != '0' ]; then
 	versionConfigureArgs='--with-password-argon2 --with-sodium=shared'
 	if [ $PHP_MINOR == '4' ]; then
 	versionConfigureArgs='--with-password-argon2 --with-sodium=shared --with-pear'
+	fi
 fi
+# in PHP 7.4+, the pecl/pear installers are officially deprecated (requiring an explicit "--with-pear") and will be removed in PHP 8+; 
+	# see also https://github.com/docker-library/php/issues/846#issuecomment-505638494
+if [ $PHP_MAJOR == '8' ]; then
+	versionConfigureArgs='--with-password-argon2 --with-sodium=shared --with-pear'
```
arroyc, line 151: same comment

arroyc (changes requested):
you should add php 8 in lts version buster build image .. and add couple of build image tests ... its ok to add the php 8 runtime later though !

images/build/Dockerfiles/vso.focal.Dockerfile
```diff
@@ -47,7 +47,8 @@ RUN LANG="C.UTF-8" \
     && add-apt-repository universe \
     && apt-get install -y --no-install-recommends python2 \
     && rm -rf /var/lib/apt/lists/* \
-    && curl https://bootstrap.pypa.io/get-pip.py --output get-pip.py \
+    # The file has been moved over here
```
arroyc, line 50: nit:  get-pip.py has been moved from 'https://bootstrap.pypa.io/get-pip.py' to  'https://bootstrap.pypa.io/2.7/get-pip.py'

images/build/php/prereqs/build.sh
qianz2: This line is necessary for 8.0 as well, I tried without it and it will fail because this library was not included.

arroyc (approved):
please make sure you have a successful nightly build before merging this change

PR #910, opened 2021-01-21
remove publish sdk job as part of CI-job 

arroyc:
We already have a seperate pipeline for this, it will be better to keep this separated as azcopy is bit flaky and we dont want to block our release for this job.

I have scheduled the publishsdk to prod as a scheduled pipeline (like we have our nightly builds/CI build) every Monday morning after CI build

### DIFF
--- a/vsts/pipelines/templates/_releaseJobTemplate.yml
+++ b/vsts/pipelines/templates/_releaseJobTemplate.yml
@@ -29,34 +29,6 @@ jobs:
       
   - template: _releaseStepTemplate.yml
 
-- job: Release_PlatformSDKs
-  displayName: Publish SDKs from dev to prod storage account
-  dependsOn: 
-    - Release_BuildImage
-    - Release_RuntimeImages
-  pool:
-    name: OryxLinux
-  variables:
-    skipComponentGovernanceDetection: true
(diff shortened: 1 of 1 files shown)


vsts/pipelines/templates/_releaseJobTemplate.yml
```diff
@@ -29,34 +29,6 @@ jobs:
       
   - template: _releaseStepTemplate.yml
 
-- job: Release_PlatformSDKs
-  displayName: Publish SDKs from dev to prod storage account
```
qianz2, line 33: thanks for updating it, will the pipeline `Oryx-PublishSdkToProd` be automatically triggered each time after a CI build or it's scheduled every Monday

qianz2 (approved):

qianz2:
Capture
just to double check, the scheduled builds are as the screenshot

vsts/pipelines/templates/_releaseJobTemplate.yml
arroyc: yes @qianz2  thats the plan

arroyc:
> Capture
> 
> just to double check, the scheduled builds are as the screenshot

yes

PR #914, opened 2021-01-26
Upgrade npm to 7

qianz2:


qianz2:
https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/
'_Despite massive internal changes, there are no new modules failing with npm 7 vs npm 6 when run against 120+ highly relied upon modules._' Looks that npm 7 is pretty stable and reliable. Since we install npm with node SDKs together (not separately, and I wonder why), we can apply it to node LTS version first.

PR #916, opened 2021-01-28
Resolve ca-certificates issue for .NET 5

qianz2:
A mitigation for .NET 5

### DIFF
--- a/images/runtime/dotnetcore/5.0/base.buster.Dockerfile
+++ b/images/runtime/dotnetcore/5.0/base.buster.Dockerfile
@@ -29,6 +29,14 @@ RUN apt-get update \
     && apt-get upgrade --assume-yes \
     && rm -rf /var/lib/apt/lists/*
 
+# Install ca-certificates from bullseye repository: https://github.com/NuGet/Announcements/issues/49
+RUN echo "deb http://deb.debian.org/debian bullseye main" >> /etc/apt/sources.list \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+         ca-certificates \
+    && rm -rf /var/lib/apt/lists/* \
+    && sed -i '$ d' /etc/apt/sources.list
+
 # Configure web servers to bind to port 80 when present
 ENV ASPNETCORE_URLS=http://+:80 \
     # Enable detection of running in a container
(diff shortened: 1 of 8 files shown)


arroyc (changes requested):
i see this is only changed in the buster images .. but i believe the current issues are in stretch images ... can you please add this to stretch based images too?

qianz2:
> i see this is only changed in the buster images .. but i believe the current issues are in stretch images ... can you please add this to stretch based images too?

https://github.com/NuGet/Announcements/issues/49#issuecomment-768766265
From the announcement, it occurred in Debian 10 buster, so stretch images are good.

arroyc:
> > i see this is only changed in the buster images .. but i believe the current issues are in stretch images ... can you please add this to stretch based images too?
> 
> [NuGet/Announcements#49 (comment)](https://github.com/NuGet/Announcements/issues/49#issuecomment-768766265)
> From the announcement, it occurred in Debian 10 buster, so stretch images are good.

as discussed we need more information on this from mitch's team .. they use  our stretch image but possibly they download a buster sdk .. so we may need to use this certificate in the stretch images as well

qianz2:
> > > i see this is only changed in the buster images .. but i believe the current issues are in stretch images ... can you please add this to stretch based images too?
> > 
> > 
> > [NuGet/Announcements#49 (comment)](https://github.com/NuGet/Announcements/issues/49#issuecomment-768766265)
> > From the announcement, it occurred in Debian 10 buster, so stretch images are good.
> 
> as discussed we need more information on this from mitch's team .. they use our stretch image but possibly they download a buster sdk .. so we may need to use this certificate in the stretch images as well

Yes, thanks for reminding, we should update stretch images too

images/runtime/dotnetcore/5.0/base.buster.Dockerfile
```diff
@@ -57,3 +55,11 @@ RUN set -ex \
     && rm aspnetcore.tar.gz \
     && dotnet-sos install \
     && rm -rf ${BUILD_DIR}
+
+# Install ca-certificates from bullseye repository: https://github.com/NuGet/Announcements/issues/49
```
arroyc, line 59: because of this we also need to create new sets of runtime images for dotnet core 5.0. You need to create new base image tag and then change the corresponding base image tag for dotnet core runtime images

arroyc (approved):
run a succesful nightly build after changing the dotnet core runtime base image tag  and before merging it to master

PR #925, opened 2021-02-04
remove extra _ from JAVA and RUBY build commands

manekinekko:
Hi, 

I wasn't sure about this but it seems that there is an extra `_` in `DISABLE_JAVA__BUILD` and `DISABLE_RUBY__BUILD`. Was that a mistake? if so, then this PR should fix it. If that was intentional, please ignore my PR.

Cheers.


arroyc:
no it was there for a purpose ...

arroyc (changes requested):
thanks for your PR .. but we do need those parameters .. so you can close this

PR #926, opened 2021-02-09
Support Jekyll app builds for static web apps

qianz2:

### DIFF
--- a/src/Detector/Ruby/RubyPlatformDetectorResult.cs
+++ b/src/Detector/Ruby/RubyPlatformDetectorResult.cs
@@ -19,5 +19,10 @@ namespace Microsoft.Oryx.Detector.Ruby
         /// Gets or sets the value if a bundler version specified in 'Gemfile.lock' in the repo.
         /// </summary>
         public string BundlerVersion { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value if a bundler version specified in 'Gemfile.lock' in the repo.
+        /// </summary>
+        public bool ConfigYmlFileExists { get; set; }
     }
 }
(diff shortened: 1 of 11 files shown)


arroyc (commented):
We basically want to cover following things ...
1. jekill app with app_type variable passed as static-sites
2. jekill app with app_type variable not passed
3. ruby app as app_type variable passed as static-sites
4. ruby app as app_type variable not passed

among these 4 there will be 2 negative tests I believe, we need to ensure we (oryx) and static site team agree to these, that's why we need at-least these 4 tests

src/Detector/Ruby/RubyPlatformDetectorResult.cs
```diff
@@ -19,5 +19,10 @@ public class RubyPlatformDetectorResult : PlatformDetectorResult
         /// Gets or sets the value if a bundler version specified in 'Gemfile.lock' in the repo.
         /// </summary>
         public string BundlerVersion { get; set; }
+
+        /// <summary>
+        /// Gets or sets the value if a bundler version specified in 'Gemfile.lock' in the repo.
```
gpcastro, line 24: Is this correct?

gpcastro (approved):

arroyc (approved):

arroyc:
its best to merge after we produce tag for ant-93

arroyc (changes requested):
Stella lets discuss the test scenarios with anthony tomorrow .. and then we can finalize that here

src/Detector/Ruby/RubyPlatformDetectorResult.cs
qianz2: It's not, thanks for catching it, will update

arroyc (approved):

PR #930, opened 2021-02-16
replace oryxlinux pool with official ubuntu build agents

arroyc:
as title

### DIFF
--- a/vsts/pipelines/PlatformBinaries/dotnetcore.yml
+++ b/vsts/pipelines/PlatformBinaries/dotnetcore.yml
@@ -7,7 +7,7 @@ stages:
   - job: DotNetCore_Buster
     timeoutInMinutes: 250
     pool:
-      name: OryxLinux
+      vmImage: 'ubuntu-latest'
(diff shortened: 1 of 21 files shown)


vsts/pipelines/PlatformBinaries/dotnetcore.yml
```diff
@@ -7,7 +7,7 @@ stages:
   - job: DotNetCore_Buster
     timeoutInMinutes: 250
     pool:
-      name: OryxLinux
+      vmImage: 'ubuntu-latest'
```
qianz2, line 10: What was the reason that we want to replace this pool image name?

gpcastro (approved):

vsts/pipelines/PlatformBinaries/dotnetcore.yml
arroyc: we dont want to manage this private buildagent pool "oyxlinux" any more for better and secured supply chain

PR #932, opened 2021-02-16
security updates for dotnet core images

arroyc:
as title ...


build/__dotNetCoreRunTimeVersions.sh
```diff
@@ -1,24 +1,24 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
-DOT_NET_CORE_RUNTIME_BASE_TAG='20210201.3'
+DOT_NET_CORE_RUNTIME_BASE_TAG='20210210.1'
```
qianz2, line 3: I think it'll be updated to today's tag `20210216.2`

platforms/dotnet/versionsToBuild.txt
```diff
@@ -4,12 +4,18 @@
 
 2.1.812, 09fb6fe64434aa83084c1cd17d632648d4be3146a9a6c0e8e50f574748adcac8c1dadb0c49b3e909252223166fedef97173552331052fce4051b216c9f8497d8,
 
+2.1.813, fbf4bbe5ff88bd12ef9da097839e9b426ce6d03b5ebff4a32f8cbc8b082b1086d5b6daacbe6c38e581d606ca2e4d7da98bac86ed608ecbee35616f9b9adbfa53,
+
 3.1.404, 94d8eca3b4e2e6c36135794330ab196c621aee8392c2545a19a991222e804027f300d8efd152e9e4893c4c610d6be8eef195e30e6f6675285755df1ea49d3605,
 
 3.1.405, 924ec0ab3f126d340ef37fe90263a91f31218995716d1ad5a817bdc6ef71e4d8e87a91edeeb785f5dff3912cc08fe87615718986bb5540ff23e9edf2302e38dd,
```
gpcastro, line 11: do we need to keep the older versions here?

gpcastro (approved):

build/__dotNetCoreRunTimeVersions.sh
arroyc: yes I will update it again

platforms/dotnet/versionsToBuild.txt
arroyc: theoretically we don't, we never cleaned this up. I can add a small work item for this

qianz2 (approved):

PR #933, opened 2021-02-16
update number of gunicorn workers defaulted to 1

arroyc:
as title


arroyc:
closing after discussing the usage of having a capability to override gunicorn workers number

Issue #895, opened 2020-12-22
Node and Yarn 2 errors


ericis:
### Bug Report

- At what date and time did you most recently experience the problem?

2020-12-22T15:22:17.052272Z

- Where did you experience the problem? E.g. Azure Web Apps, Azure Functions, Azure Container Registry, or offline use.

Azure Web Apps

- If your repo is publicly available please share its URL:

https://github.com/ourchitecture/panda/tree/main/src/tutorials/node/azure

- What happened?

Deployment with `az webapp up` failed with the error `Usage Error: Couldn't find a configuration settings named "cache-folder"`.

- How can we reproduce it?

Clone the repo and follow the steps.

- Do you have log output? Please include between the backticks:

**az webapp up logs**

```
Webapp 'app-our-tutorials-node-hello' already exists. The command will deploy contents to the existing app.
Creating zip with contents of dir /app ...
Getting scm site credentials for zip deployment
Starting zip deployment. This operation can take a while to complete ...
Deployment endpoint responded with status code 202
Configuring default logging for the app, if not already enabled
ValidationError: Zip deployment failed. {'id': '69d35dc5aad045a998647c8fc9af7021', 'status': 3, 'status_text': '', 'author_email': 'N/A', 'author': 'N/A', 'deployer': 'Push-Deployer', 'message': 'Created via a push deployment', 'progress': '', 'received_time': '2020-12-22T17:47:48.1521144Z', 'start_time': '2020-12-22T17:47:49.4054755Z', 'end_time': '2020-12-22T17:48:21.0986005Z', 'last_success_end_time': None, 'complete': True, 'active': False, 'is_temp': False, 'is_readonly': True, 'url': 'https://app-our-tutorials-node-hello.scm.azurewebsites.net/api/deployments/latest', 'log_url': 'https://app-our-tutorials-node-hello.scm.azurewebsites.net/api/deployments/latest/log', 'site_name': 'app-our-tutorials-node-hello'}. Please run the command az webapp log deployment show -n app-our-tutorials-node-hello -g rg-our-tutorials
```

**Deployment logs**

```
[
  {
    "log_time": "2020-12-22T15:22:17.052272Z",
    "id": "",
    "message": "Command: oryx build /tmp/zipdeploy/extracted -o /home/site/wwwroot --platform nodejs --platform-version 12 -i /tmp/8d8a68d5dd76127 -p compress_node_modules=tar-gz --log-file /tmp/build-debug.log ",
    "type": 0,
[...]

PR #942, opened 2021-02-25
Support yarn 2

qianz2:
https://github.com/microsoft/Oryx/issues/895
https://github.com/Azure/static-web-apps/issues/287

### DIFF
--- a/src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
@@ -15,11 +15,12 @@ echo
 # Yarn config is per user, and since the build might run with a non-root account, we make sure
 # the yarn cache is set on every build.
 YARN_CACHE_DIR=/usr/local/share/yarn-cache
+yarnCacheFolderName={{ YarnCacheFolderName }}
 if [ -d $YARN_CACHE_DIR ]
 then
 	echo
     echo "Configuring Yarn cache folder..."
-    yarn config set cache-folder $YARN_CACHE_DIR
+    yarn config set $yarnCacheFolderName $YARN_CACHE_DIR
 fi
 {{ end }}
 
@@ -73,6 +74,10 @@ then
 
 	if [ -f "$SOURCE_DIR/.yarnrc" ]; then
 		cp -f "$SOURCE_DIR/.yarnrc" .
+	fi
+
+	if [ -f "$SOURCE_DIR/.yarnrc.yml" ]; then
+		cp -f "$SOURCE_DIR/.yarnrc.yml" .
 	fi
 
 	echo
--- a/src/BuildScriptGenerator/Node/NodeConstants.cs
+++ b/src/BuildScriptGenerator/Node/NodeConstants.cs
@@ -34,6 +34,9 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
         public const string YarnVersionCommand = "echo Using Yarn version: && yarn --version";
         public const string HugoVersionCommand = "echo Using Hugo version: && hugo version";
         public const string YarnPackageInstallCommand = "yarn install --prefer-offline";
+        public const string Yarn2PackageInstallCommand = "yarn plugin import workspace-tools && yarn workspaces focus --all";
+        public const string Yarn1ConfigFolderName = "cache-folder";
+        public const string Yarn2ConfigFolderName = "cacheFolder";
         public const string ProductionOnlyPackageInstallCommandTemplate = "{0} --production";
         public const string PkgMgrRunBuildCommandTemplate = "{0} run build";
         public const string PkgMgrRunBuildAzureCommandTemplate = "{0} run build:azure";
--- a/src/BuildScriptGenerator/Node/NodePlatform.cs
+++ b/src/BuildScriptGenerator/Node/NodePlatform.cs
@@ -184,10 +184,10 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
                 packageManagerCmd = NodeConstants.YarnCommand;
                 configureYarnCache = true;
                 packageInstallerVersionCommand = NodeConstants.YarnVersionCommand;
-                // yarn.lock file is valid YAML in Yarn 2+ and yarnrc.yml file replaces yarnrc in Yarn 2+.
+                // yarn.lock file is valid YAML in Yarn 2+ and .yarnrc.yml file replaces .yarnrc in Yarn 2+.
                 // Applying yarn 2 cache folder name and package install command.
-                if (nodePlatformDetectorResult.IsYarnLockFileValidYamlFormat
-                    || nodePlatformDetectorResult.HasYarnrcYmlFile) {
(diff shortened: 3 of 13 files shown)


src/Detector/Node/NodeDetector.cs
```diff
@@ -122,9 +135,26 @@ public PlatformDetectorResult Detect(DetectorContext context)
...
             };
         }
 
+        private bool IsYarnLockFileYamlFile(ISourceRepo sourceRepo, string filePath)
+        {
+            var yamlContent = sourceRepo.ReadFile(filePath);
+            var yamlStream = new YamlStream();
+            try
+            {
+                yamlStream.Load(new StringReader(yamlContent));
```
gpcastro, line 149: should we do any check to make sure it is indeed a yarn 2 file?

gpcastro (approved):

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -15,11 +15,12 @@ echo
 # Yarn config is per user, and since the build might run with a non-root account, we make sure
 # the yarn cache is set on every build.
 YARN_CACHE_DIR=/usr/local/share/yarn-cache
+yarnCacheFolderName={{ YarnCacheFolderName }}
 if [ -d $YARN_CACHE_DIR ]
 then
 	echo
     echo "Configuring Yarn cache folder..."
-    yarn config set cache-folder $YARN_CACHE_DIR
+    yarn config set $yarnCacheFolderName $YARN_CACHE_DIR
```
arroyc, line 23: why is this change? are we passing the cache folder name to the script now?

src/BuildScriptGenerator/Node/NodeConstants.cs
```diff
@@ -34,6 +34,9 @@ public static class NodeConstants
         public const string YarnVersionCommand = "echo Using Yarn version: && yarn --version";
         public const string HugoVersionCommand = "echo Using Hugo version: && hugo version";
         public const string YarnPackageInstallCommand = "yarn install --prefer-offline";
+        public const string Yarn2PackageInstallCommand = "yarn plugin import workspace-tools && yarn workspaces focus --all";
+        public const string Yarn1ConfigFolderName = "cache-folder";
+        public const string Yarn2ConfigFolderName = "cacheFolder";
```
arroyc, line 39: is this someting yarn2 has recommened?

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -181,9 +182,18 @@ public PlatformDetectorResult Detect(RepositoryContext context)
             if (ctx.SourceRepo.FileExists(NodeConstants.YarnLockFileName))
             {
                 packageManagerCmd = NodeConstants.YarnCommand;
-                packageInstallCommand = NodeConstants.YarnPackageInstallCommand;
                 configureYarnCache = true;
                 packageInstallerVersionCommand = NodeConstants.YarnVersionCommand;
+                // yarn.lock file is valid YAML in Yarn 2+ and yarnrc.yml file replaces yarnrc in Yarn 2+.
```
arroyc, line 187: so a repo that doesnt have .yarnrc.yml but has yaml formatted yarn.lock will be considered as yarn2 supported ...

do we commit .yarnrc.yml in the repo too? if not then we may not need to check for this

arroyc (changes requested):
I would also like to have a tests to verify the yamlfileformat checker is working fine. So a test that can assert true if a correct yaml file is passed as input and assert false when a non-yaml file is passed as input

src/BuildScriptGenerator/Node/NodePlatform.cs
qianz2: yarn.lock file can be parsed as valid YAML format since yarn 2. The .yarnc.yml file does not have to be in the repo, it's a configuration file for yarn, sometimes it may not be in the repo for yarn2. So I added both as indicators.

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
qianz2: This is simply the CLI command difference between yarn 1 and yarn 2, so we need to detect and pass in different command parameter here.

src/BuildScriptGenerator/Node/NodeConstants.cs
qianz2: https://next.yarnpkg.com/getting-started/migration Yes, on the bottom CLI command section, it migrated yarn install command into this command.

src/Detector/Node/NodeDetector.cs
qianz2: It's used to parse yarn.lock file, if it's a valid YAML then it'll not throw any exceptions. I added tests for this part to make sure it works good.

src/BuildScriptGenerator/Node/NodeConstants.cs
arroyc: @qianz2  not about yarn install ... looking at yarn cache config folder name. looks like yarn1 is cache-folder while yarn2 is cacheFolder. I'm curious why is that?

arroyc (approved):
thanks for adding two tests where detector returning true for yarn2, can you add 1 negative test where it returns false for yarn2 checks?

Itrel:
Is there a way for me to utilize the Oryx build from this branch to test deployment of Azure static websites using yarn 2? 

What changes are needed to my GitHub actions in order to consume the Oryx build from this branch?

Is there a way I can clone Oryx and run it all locally?

Would it be a useful test case for you for me to attempt this?

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -181,9 +182,18 @@ public PlatformDetectorResult Detect(RepositoryContext context)
             if (ctx.SourceRepo.FileExists(NodeConstants.YarnLockFileName))
             {
                 packageManagerCmd = NodeConstants.YarnCommand;
-                packageInstallCommand = NodeConstants.YarnPackageInstallCommand;
                 configureYarnCache = true;
                 packageInstallerVersionCommand = NodeConstants.YarnVersionCommand;
+                // yarn.lock file is valid YAML in Yarn 2+ and .yarnrc.yml file replaces .yarnrc in Yarn 2+.
+                // Applying yarn 2 cache folder name and package install command.
+                if (nodePlatformDetectorResult.HasYarnrcYmlFile
+                    || nodePlatformDetectorResult.IsYarnLockFileValidYamlFormat) {
```
merceyz, line 190: An alternative to this check is to test if the content of `yarn.lock` matches this regex
```regexp
/^__metadata:$/m
```

src/BuildScriptGenerator/Node/NodeConstants.cs
merceyz: Since you're passing the `--all` flag but not the `--production` flag as well it's the same as if you were to run `yarn install`

> looks like yarn1 is cache-folder while yarn2 is cacheFolder. I'm curious why is that?

We made the settings camelCase in v2

PR #949, opened 2021-03-05
Move validation builds to official ado agents

arroyc:
We will have multiple PRs  to get rid of oryxlinux pool .... first phase was base image builds, detector builds and platform binaries builds. We already have done that, this is our second phase. Next will be nightly builds and at the end we will move ci build.

We are taking phase wise approach as we dont want to block ci/cd for this

### DIFF
--- a/vsts/pipelines/templates/_buildTemplate.yml
+++ b/vsts/pipelines/templates/_buildTemplate.yml
@@ -149,7 +149,7 @@ steps:
 
 - script: |
     docker images && docker system prune -fa && docker images && echo
-  displayName: 'clean docker unknown layers and containers'
+  displayName: 'clean docker layers and containers'
   
 - task: CopyFiles@2
   displayName: 'Copy source projects output to artifacts folder'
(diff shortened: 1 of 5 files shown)


vsts/pipelines/templates/_buildTemplate.yml
```diff
@@ -29,6 +29,14 @@ steps:
     startsWith(variables['Build.SourceBranch'],'refs/heads/patch/'),
     startsWith(variables['Build.SourceBranch'],'refs/heads/exp/')))
 
+- script: |
+    sudo rm -rf /usr/share/dotnet
+    sudo rm -rf /opt/ghc
+    sudo rm -rf "/usr/local/share/boost"
+    sudo rm -rf "$AGENT_TOOLSDIRECTORY"
+    docker images && docker system prune -fa && df -h && echo
+  displayName: 'clean docker images'
```
qianz2, line 38: Will it be faster if not doing it. Sometimes if it's cached, then build build images step will be faster?

qianz2: Such as triggering to re-run failed job directly from the pipeline, I don't know if that's an example

arroyc: unfortunately we cant ... first we dont know which agent we are gonna get .. these are auto managed by ado team .. so they recycle these agents .. secondly there is space limitation .. only 10-12 gb available ..we are sacrificing build speed here over build security @qianz2

qianz2: I see, thanks for clarifying, then I agree we should clean docker images.

qianz2 (approved):

PR #964, opened 2021-03-19
update oryx generated conda-env yml location

arroyc:
this is to redirect the oryx generated conda env yml to non-source location so that user doesn't have to change their git settings (adding it to gitignore or manually delete it)


src/BuildScriptGenerator/Python/JupyterNotebookBashBuildSnippet.sh.tpl
```diff
@@ -6,7 +6,7 @@ conda info
 {{ if EnvironmentYmlFile | IsNotBlank }}
 envFile="{{ EnvironmentYmlFile }}"
 {{ else }}
-envFile="oryx.environment.yml"
+envFile="/opt/oryx/conda/oryx.environment.yml"
```
arroyc, line 9: we were creating this file in user's source repo location which needs to be avoided

qianz2 (approved):

PR #965, opened 2021-03-20
retire vso debian stretch based image

arroyc:
stretch based debian vso image is deprecated and moved to vso-focal, so we should be removing this image and corresponding tests from oryx pipeline.


qianz2 (approved):
Could you trigger a nightly build, just want to make sure integration tests will all pass

PR #970, opened 2021-04-01
Add yarn2 test for vso image and sample app

qianz2:


arroyc (changes requested):
I'm bit confused ... title says add yarn2 test for vso image .. i dont see any test for vso image .. tests are for ltsversion image ... where is the sample app? Did you add a new sample app?

I thought yarn2 support is for all images .. how are you enabling it only for vso image?

qianz2:
> I'm bit confused ... title says add yarn2 test for vso image .. i dont see any test for vso image .. tests are for ltsversion image ... where is the sample app? Did you add a new sample app?
> 
> I thought yarn2 support is for all images .. how are you enabling it only for vso image?

This support is for all images, I titled it because I referred it from a work item we had. Vso image is the place most customers consume this feature. So I have tests here using either ltsversion image or vso image, the results are same. 
Yes, I added the sample app if you can scroll down to the end of files, there has a new submodules which contains the sample yarn2 app.

arroyc:
> > I'm bit confused ... title says add yarn2 test for vso image .. i dont see any test for vso image .. tests are for ltsversion image ... where is the sample app? Did you add a new sample app?
> > I thought yarn2 support is for all images .. how are you enabling it only for vso image?
> 
> This support is for all images, I titled it because I referred it from a work item we had. Vso image is the place most customers consume this feature. So I have tests here using either ltsversion image or vso image, the results are same.
> Yes, I added the sample app, if you can scroll down to the end of files, there has a new submodule which contains the sample yarn2 app.

alright .. it was not there earlier when I reviewed it ... i think you just pushed some changes and I can see 363  files .. so i will take a look at it.

arroyc (commented):
can you delete the contents from ./yarn/cache?? Do we need them?

qianz2:
> can you delete the contents from ./yarn/cache?? Do we need them?

Maybe we can, let me try this locally, if it works fine, I'll remove them

arroyc (approved):

PR #988, opened 2021-04-14
add pm2 back for node runtime images

arroyc:


images/runtime/node/installDependencies.sh
```diff
@@ -36,8 +36,8 @@ if [[ $nodeVersionMajor -ge 10  ]]; then
     echo "Version of npm after upgrade: $currentNpmVersion"
 fi
 
-# Do NOT install PM2 from Node 14 onwards
-if [[ $nodeVersionMajor -lt 14 ]]; then
+# Do NOT install PM2 for Node versions 8 older than 8
```
cormacpayne, line 39: nit: I was a little confused by this comment due to the typo -- I would reword it to be something like:

_Do NOT install PM2 for Node versions below 8 (older than 8)_

tests/Oryx.RuntimeImage.Tests/Node/NodeRuntimeImageContainsRequiredPrograms.cs
```diff
@@ -56,7 +56,7 @@ public void NodeImage_DoesNotContain_PM2(string nodeTag)
             });
 
             // Assert
-            RunAsserts(() => Assert.False(result.IsSuccess), result.GetDebugInfo());
+            RunAsserts(() => Assert.True(result.IsSuccess), result.GetDebugInfo());
```
cormacpayne, line 59: nit: rename the test function to `NodeImage_DoesContain_PM2`

cormacpayne (approved):

wubzz:
Sorry for digging through old PRs but is there any publically available context relating to how PM2 now works in Oryx or how to "force" use it? I see it's been removed, re-added, and modified sveral times throughout 2021 especially in regards to Node 14. Can't find any documentation about this..

<details>
Oryx Version: 0.2.20210826.1, Commit: f8651349d0c78259bb199593b526450568c2f94a, ReleaseTagName: 20210826.1

####

Hosting Azure App Services (Linux) currently in Node version 12. These have always been started with PM2. Changing to version 14 no longer uses pm2, instead it starts the applications using `node server.js`. Switching back to 12 also goes back to using PM2.

The Microsoft Azure documentation states that so long as there is a set of specific files, PM2 will be used to run the application:

https://docs.microsoft.com/en-us/azure/app-service/configure-language-nodejs?pivots=platform-linux#run-with-pm2

The project consists of `server.js` AND `ecosystem.config.js`, and does not contain any npm-run script. Despite this, PM2 still is not used when running Node version 14 - which in turn contradicts the documentation.

With Node 12 leaving maintenance around the corner, I intended to upgrade all apps to version 14.

The same issue goes for Node 16.
</details>

PR #1000, opened 2021-04-29
Use Npm ci command when package lock file is present

qianz2:


arroyc (changes requested):
we should not add this changes directly .. lets discuss this within team/design meeting when rebecca is back as well .. directly adding this change will break other partner teams, we should wrap it inside some  flag or env settings

PR #1008, opened 2021-05-11
Support custom build command for Jekyll apps

qianz2:


tests/BuildScriptGenerator.Tests/Ruby/RubyPlatformTests.cs
```diff
@@ -198,6 +198,42 @@ public void GeneratedBuildSnippet_DoesNotContainsJekyllCommand_WhenOnlyGemFileEx
             Assert.Contains("bundle install", buildScriptSnippet.BashBuildScriptSnippet);
         }
 
+        [Fact]
+        public void GeneratedBuildSnippet_CustomBuildCommandWillExecute_WhenOtherCommandsAlsoExist()
```
arroyc, line 202: can you add some buildimage tests for vso-focal and jamstack image to make sure we are covering the end to end scenarios here?

src/BuildScriptGenerator/Ruby/RubyScriptGeneratorOptions.cs
```diff
@@ -8,5 +8,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Ruby
     public class RubyScriptGeneratorOptions
     {
         public string RubyVersion { get; set; }
+
+        public string CustomBuildCommand { get; set; }
```
arroyc, line 12: How does it work with oryx build command

arroyc (commented):
Add some description such as what is the use case here? how this custom build command will be used currently with oryx build command etc

tests/Oryx.BuildImage.Tests/Ruby/RubySampleAppsTest.cs
```diff
@@ -119,5 +119,38 @@ public void Builds_JekyllStaticWebApp_When_Apptype_Is_SetAs_StaticSiteApplicatio
...
+        public void Builds_JekyllStaticWebApp_UsingCustomBuildCommand(string buildImage)
+        {
+            // Arrange
+            var appName = "Jekyll-app";
+            var volume = CreateSampleAppVolume(appName);
+            var appDir = volume.ContainerDir;
+            var appOutputDir = "/tmp/app-output";
+            var script = new ShellScriptBuilder()
+                .AddDefaultTestEnvironmentVariables()
+                .SetEnvironmentVariable("CUSTOM_BUILD_COMMAND", "touch example.txt")
```
arroyc, line 135: ah .. ok .. so it's not really a property or argument that we pass in oryx command line .. it's just and environment variable

src/BuildScriptGenerator/Ruby/RubyBashBuildSnippetProperties.cs
```diff
@@ -17,5 +17,7 @@ internal class RubyBashBuildSnippetProperties
         public bool GemfileExists { get; set; }
 
         public bool ConfigYmlFileExists { get; set; }
+
+        public string CustomBuildCommand { get; set; }
```
arroyc, line 21: is it an environment variable or a property that will be passed to Oryx? I'm bit confused now ... according to the test it's an environment variable .. so do we really need it as buildsnippet properties?

qianz2: It is passed in as env variable, and then as options (Hierarchy). And then we need to checking this property for generating the buildSnippet so we need to include it into the `RubyBashBuildSnippetProperties.cs` file.

qianz2:
> > We have `CUSTOM_BUILD_COMMAND ` env applying to Node apps for now. Customers from Jekyll static web apps also need this feature to use their custom commands, so we're going to add it into Ruby Bash Build Snippet.
> 
> Can we reuse the custom_build_command which we already have? think about this if something comes for php or python tomorrow are we going to add mosre custom build commands to each platform?

We couldn't add it into baseBashBuildSnippet.cs as we discussed offline, but I'll try to see if we can move this property/option to the base one.

arroyc:
> > > We have `CUSTOM_BUILD_COMMAND ` env applying to Node apps for now. Customers from Jekyll static web apps also need this feature to use their custom commands, so we're going to add it into Ruby Bash Build Snippet.
> > 
> > 
> > Can we reuse the custom_build_command which we already have? think about this if something comes for php or python tomorrow are we going to add mosre custom build commands to each platform?
> 
> We couldn't add it into baseBashBuildSnippet.cs as we discussed offline, but I'll try to see if we can move this property/option to the base one.

no need ... i think its better the way it is now

arroyc (approved):

PR #1012, opened 2021-05-14
Add npm version test and remove node 10, 9, 8 from extensive testing, support and maintenance

arroyc:


tests/Oryx.RuntimeImage.Tests/Node/NodeLegacyVersionsRuntimeImageTests.cs
```diff
@@ -91,6 +97,33 @@ public void NodeVersionMatchesImageName(string nodeTag, string nodeVersion)
                 result.GetDebugInfo());
         }
 
+        [Theory]
+        [InlineData("10")]
+        [InlineData("10.1")]
+        [InlineData("10.10")]
+        [InlineData("10.12")]
+        [InlineData("10.14")]
+        public void Node10ImageContains_Correct_NPM_Version(string imageTag)
```
qianz2, line 106: Do we still need this if we don't support node 10 anymore?

images/runtime/node/installDependencies.sh
```diff
@@ -30,7 +30,7 @@ echo "Version of npm: $currentNpmVersion"
 # Upgrade npm to the latest available version
 if [[ $nodeVersionMajor -ge 10  ]]; then
     echo "Upgrading npm..."
-    npm install npm@6.14.10 -g -loglevel silent
+    npm install npm@$NPM_VERSION -g -loglevel silent
```
qianz2, line 33: If node version is greater than 10, they'll all have the same npm version in the image? 
What if we want to keep some latest node version have npm 7 instead?

tests/Oryx.RuntimeImage.Tests/Node/NodeLegacyVersionsRuntimeImageTests.cs
```diff
@@ -68,6 +68,12 @@ public void NodeImage_Contains_VersionAndCommit_Information(string version)
         [InlineData("6.9", "6.9.5")]
         [InlineData("6.10", "6.10.3")]
         [InlineData("6.11", "6.11.5")]
+        [InlineData("8", NodeVersions.Node8Version)]
+        [InlineData("8.2", "8.2.1")]
+        [InlineData("8.11", "8.11.4")]
+        [InlineData("9.4", "9.4.0")]
+        [InlineData("10", NodeVersions.Node10Version)]
+        [InlineData("10.10", "10.10.0")]
```
qianz2, line 76: same here, look that they're no longer supported.

qianz2:
Could you check telemetry usage to make sure we don't have any existing customers using node 10?

images/runtime/node/installDependencies.sh
arroyc: we dont have any npm 7 requirement yet ..  when we start supporting npm 7 in a feature some of these  codes needs to be revisited for sure. This is to make sure we have coverage for what we have now

tests/Oryx.RuntimeImage.Tests/Node/NodeLegacyVersionsRuntimeImageTests.cs
arroyc: actually the title of the pr is misleading .. we dont test extensively on these versions any more .. just a  basic level of end to end testing

arroyc:
> Could you check telemetry usage to make sure we don't have any existing customers using node 10?

I referred to your last ran query which only showed 12 and 14 in use and also 10 is got officially retired from maintenance cycle of node official team  on april 30th. We are still releasing them,  we are just stopping extensive testing on these versions for the time being

tests/Oryx.RuntimeImage.Tests/Node/NodeLegacyVersionsRuntimeImageTests.cs
arroyc: we still release 10, so technically we do need to test them to some extent, but we wont do extensive testing any more

qianz2 (approved):

Issue #1032, opened 2021-06-05
Jamstack image - include .NET Core 3.1 SDK in base image


anthonychu:
### Feature Request

In order to build .NET isolated function apps for Static Web Apps, we need .NET Core 3.1 _and_ the version of .NET specified in `<TargetFramework>` to be installed.

After discussion with @arroyc, the best way to support this might be to always include .NET Core 3.1 in the Jamstack image. Can we see how much the image size increases if we do this?

qianz2:
After installing .NET 3.1 SDKs into Jamstack image, it increased by 336M. The total size of image is 2.3G. If it's ok, we will proceed to add .NET Core 3.1 SDK on it.

qianz2:
@anthonychu, we've recently supported it, so closing this issue now.

PR #1034, opened 2021-06-08
Include .NET 3 SDK to Jamstack build image

qianz2:
https://github.com/microsoft/Oryx/issues/1032


arroyc (commented):
I would add a test where we check that this image has 3.1 .. we have couple of such tests where we check what sdk versions we have in the image

arroyc (approved):

PR #1056, opened 2021-07-13
Add GIS libraries for GeoDjango

gpcastro:
Adding libraries required for GeoDjango; updating a sample app to validate that the libraries work when building an app the enables it.


arroyc (approved):

gme5k:
Is there a schedule for when this 14 July version or newer of Oryx will run on the Azure App Service?

PR #1067, opened 2021-07-29
remove plugin import command from yarn2 install

arroyc:

### DIFF
--- a/src/BuildScriptGenerator/Node/NodeConstants.cs
+++ b/src/BuildScriptGenerator/Node/NodeConstants.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
         public const string YarnVersionCommand = "echo Using Yarn version: && yarn --version";
         public const string HugoVersionCommand = "echo Using Hugo version: && hugo version";
         public const string YarnPackageInstallCommand = "yarn install --prefer-offline";
-        public const string Yarn2PackageInstallCommand = "yarn workspaces focus --all";
+        public const string Yarn2PackageInstallCommand = "yarn workspaces focus --all || || yarn install";
         public const string ProductionOnlyPackageInstallCommandTemplate = "{0} --production";
         public const string PkgMgrRunBuildCommandTemplate = "{0} run build";
         public const string PkgMgrRunBuildAzureCommandTemplate = "{0} run build:azure";
--- a/src/BuildScriptGenerator/Node/NodeConstants.cs
+++ b/src/BuildScriptGenerator/Node/NodeConstants.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
         public const string YarnVersionCommand = "echo Using Yarn version: && yarn --version";
         public const string HugoVersionCommand = "echo Using Hugo version: && hugo version";
         public const string YarnPackageInstallCommand = "yarn install --prefer-offline";
-        public const string Yarn2PackageInstallCommand = "yarn plugin import workspace-tools && yarn workspaces focus --all";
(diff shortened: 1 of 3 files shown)


tests/Oryx.BuildImage.Tests/Node/NodeJsSampleAppsOtherTests.cs
```diff
@@ -976,6 +976,9 @@ public void  CanBuildAndRunNodeApp_UsingYarn2ForBuild()
             var appOutputDir = "/tmp/nextjs-yarn2-example";
             var appDir = volume.ContainerDir;
             var script = new ShellScriptBuilder()
+                .AddCommand($"yarn set version berry")
```
arroyc, line 979: we needed to set the workspace one time, this should  be user's responsibility, so adding these commands in test prior to oryx build command.
Assumption is User will run these commands in his/her local env before checking it in (including workspace) in git

src/BuildScriptGenerator/Node/NodeConstants.cs
```diff
@@ -34,7 +34,7 @@ public static class NodeConstants
         public const string YarnVersionCommand = "echo Using Yarn version: && yarn --version";
         public const string HugoVersionCommand = "echo Using Hugo version: && hugo version";
         public const string YarnPackageInstallCommand = "yarn install --prefer-offline";
-        public const string Yarn2PackageInstallCommand = "yarn plugin import workspace-tools && yarn workspaces focus --all";
+        public const string Yarn2PackageInstallCommand = "yarn workspaces focus --all";
```
iarna, line 37: Doesn't this mean that users without workspaces will fail? You can call yarn install if the workspace focus subcommand is unavailable with:

```
yarn workspaces focus --all || yarn install
```

arroyc: yeah  good catch .. i will add it ...

iarna (approved):

src/BuildScriptGenerator/Node/NodeConstants.cs
```diff
@@ -34,7 +34,7 @@ public static class NodeConstants
         public const string YarnVersionCommand = "echo Using Yarn version: && yarn --version";
         public const string HugoVersionCommand = "echo Using Hugo version: && hugo version";
         public const string YarnPackageInstallCommand = "yarn install --prefer-offline";
-        public const string Yarn2PackageInstallCommand = "yarn plugin import workspace-tools && yarn workspaces focus --all";
+        public const string Yarn2PackageInstallCommand = "yarn workspaces focus --all || yarn install";
```
qianz2, line 37: `yarn plugin import workspace-tools` I think this is the plugin required for running `yarn workspaces focus --all`, if we remove it we may not able to run the later command.

arroyc: @qianz2  ... so basicaly you are kind of true but the whole scenario/workflow is little different ... here is a good reference https://github.com/yarnpkg/berry/issues/3199

so based on this, I think "yarn plugin import workspace-tools" is usually gets executed only once, it should not run every time .. user should run it once and check the plugin in in their repo and from next time onwards it should be only yarn workspace focus --all (as the workspace is already been set). So assumption is this command is basically user's responsibility and oryx should not do it.

If the workspace is not set yarn workspace focus --all will fail but then we will run yarn install command for yarn2.

arroyc: @qianz2  you can look at the test that I added to get more insight ..  let me know if you have some questions

qianz2: I see, then we will let users run this plugin command, thanks for the explain. And we can include this into oryx docs.

qianz2 (approved):

PR #1078, opened 2021-08-06
Change .NET OutputType to return raw value

anthonychu:
With the current logic, .NET apps that are not Azure Functions apps will still output OutputTypes of "isolated" or "in-process". It'd be better for Oryx to write the OutputType from the csproj without transforming the value. Static Web Apps can use the OutputType to determine function app type.

In addition, OutputType is optional in a .csproj. It defaults to "Library".


arroyc (approved):
because of branch policy I will create a branch from this and check it in.

arroyc:
closing on behalf of #1079

PR #1089, opened 2021-08-20
update for benv support 

arroyc:
for internal review only

### DIFF
--- a/images/build/Dockerfiles/Dockerfile
+++ b/images/build/Dockerfiles/Dockerfile
@@ -21,7 +21,6 @@ RUN ./build.sh dotnetcore /opt/startupcmdgen/dotnet
 RUN ./build.sh node       /opt/startupcmdgen/nodejs
 RUN ./build.sh php        /opt/startupcmdgen/php
 RUN ./build.sh python     /opt/startupcmdgen/python
-RUN ./build.sh golang     /opt/startupcmdgen/golang
 
 ###
 # End build run script generators
--- a/images/build/Dockerfiles/Dockerfile
+++ b/images/build/Dockerfiles/Dockerfile
@@ -83,12 +83,6 @@ RUN set -ex \
     && yarnCacheFolder="/usr/local/share/yarn-cache" \
     && mkdir -p $yarnCacheFolder \
     && chmod 777 $yarnCacheFolder \
-    
-    # Install Golang SDKs
-    && . $buildDir/__golangVersions.sh \
-    && $imagesDir/installPlatform.sh golang $GOLANG_VERSION \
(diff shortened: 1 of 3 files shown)


images/build/Dockerfiles/Dockerfile
```diff
@@ -21,7 +21,6 @@ RUN ./build.sh dotnetcore /opt/startupcmdgen/dotnet
 RUN ./build.sh node       /opt/startupcmdgen/nodejs
 RUN ./build.sh php        /opt/startupcmdgen/php
 RUN ./build.sh python     /opt/startupcmdgen/python
-RUN ./build.sh golang     /opt/startupcmdgen/golang
```
arroyc, line 24: this one was causing problem ... not sure why this was added ... this is for the runtime scriptgenerator ..we are only focused about buildscript generator for golang

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -27,7 +27,8 @@ RUN oryx prep --skip-detection --platforms-and-versions nodejs=12 \
     && . /tmp/build/__goVersions.sh \
     && downloadedFileName="go${GO_VERSION}.linux-amd64.tar.gz" \
     && curl -SLsO https://golang.org/dl/$downloadedFileName \
-    && tar -C /usr/local -xzf $downloadedFileName \
+    && mkdir -p /usr/local \
+    && tar -xzf $downloadedFileName -C /usr/local \
```
arroyc, line 31: this tar command was initially failing so I added a mkdir command before it to make sure the directory exists where we untar the components

images/build/Dockerfiles/Dockerfile
william-msft: Ah got it! Looks like it was from my very initial commits while I was referencing Ruby. Thank you for catching this! 👍

images/build/Dockerfiles/Dockerfile
```diff
@@ -84,12 +83,6 @@ RUN set -ex \
     && yarnCacheFolder="/usr/local/share/yarn-cache" \
     && mkdir -p $yarnCacheFolder \
     && chmod 777 $yarnCacheFolder \
-    
-    # Install Golang SDKs
-    && . $buildDir/__golangVersions.sh \
-    && $imagesDir/installPlatform.sh golang $GOLANG_VERSION \
-    && ln -sfn $GOLANG_VERSION 1.16 \
-
```
arroyc, line 92: we dont need it here as we build this image from lts-version layer so we already have support for golang by now

PR #1096, opened 2021-09-01
Enable PGO and LTO on all Python images

tonybaloney:
Python is being compiled from source without the recommended PGO and LTO optimisations.

This change will give a 10-20% performance gain on all Python execution.
Updates all supported Python 3.x versions to the latest stable bug fix releases

3.6.12 -> 3.6.13
3.7.9 -> 3.7.11
3.8.6 -> 3.8.12
3.9.0 -> 3.9.7

Changes Python 3 to use virtualenv


arroyc:
@tonybaloney  thanks will take care of it soon

arroyc (dismissed):

tonybaloney:
@arroyc  has this already been upstreamed?

PR #1097, opened 2021-09-02
<DO NOT MERGE - In Progress> add support for oryx build end point

arroyc:
- [ ] Tests are included and/or updated for code changes. To do


src/BuildServer/BuildServerResponse.cs
```diff
@@ -0,0 +1,96 @@
...
+// --------------------------------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.Oryx.BuildServer
+{
+    enum BuildState
+    {
+        Success,
+        Running,
```
william-msft, line 13: Nit: Maybe we should consider to rename to `Building`. Running sounds like it like the built app is running to me.

src/BuildServer/BuildServerRequests.cs
```diff
@@ -0,0 +1,39 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Oryx.BuildServer
+{
+    public class BuildServerRequests
+    {
```
william-msft, line 11: Would be nice to have a documentation link to a JSON sample

src/BuildServer/BuildServerResponse.cs
arroyc: I like Building .. good one

src/BuildServer/BuildServerRequests.cs
arroyc: of course we should .. good point

PR #1107, opened 2021-09-17
Draft: Williamhe/1357976 add ruby run build command

william-msft:


src/BuildScriptGenerator/Ruby/RubyBashBuildSnippet.sh.tpl
```diff
@@ -19,29 +19,34 @@ echo
...
-        {{ if GemfileExists }}
-        echo "Running 'bundle install'..."
-        echo
-        bundle install
+if [ ! -z "$RUN_BUILD_COMMAND" ]
+then
+	echo 'Running $RUN_BUILD_COMMAND'
+	${RUN_BUILD_COMMAND}
+else
+    {{ if CustomBuildCommand | IsNotBlank }}
```
cormacpayne, line 27: Do you have an example of a `CustomBuildCommand` that you can provide? Would it include the `RUN_BUILD_COMMAND` and that's why we don't want them in the same flow? Just making sure I understand the purposes of the different variables.

src/BuildScriptGenerator/Ruby/RubyPlatform.cs
```diff
@@ -114,6 +114,13 @@ public PlatformDetectorResult Detect(RepositoryContext context)
 
             var buildProperties = new Dictionary<string, string>();
 
+            string runBuildCommand = Environment.GetEnvironmentVariable("RUN_BUILD_COMMAND");
```
cormacpayne, line 117: The `RUN_BUILD_COMMAND` string is used in the Python PR (and I believe other PRs) -- maybe we should put this string in a global file that they can all use?

william-msft:
Closing until we re-visit

PR #1108, opened 2021-09-17
Draft: Williamhe/1357976 add python run build command

william-msft:


src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -162,6 +162,13 @@ public PlatformDetectorResult Detect(RepositoryContext context)
 
             var manifestFileProperties = new Dictionary<string, string>();
 
+            string runBuildCommand = Environment.GetEnvironmentVariable("RUN_BUILD_COMMAND");
+            if (!String.IsNullOrEmpty(runBuildCommand))
+            {
+                _logger.LogWarning("RUN_BUILD_COMMAND environment variable detected");
```
arroyc, line 168: nit: logdebug  would be better IMO

arroyc (changes requested):
rebase to main and then rerun validation build .. because you are using flask  app which broke our pipeline after the dependencies got updated. we want to make sure this test is running fine before we merge them to main

src/BuildScriptGenerator/Python/PythonPlatform.cs
william-msft: Agree, updated

doc/configuration.md
```diff
@@ -48,6 +48,7 @@ PYTHON\_VERSION              | Specify which Python version the app is using
 DISABLE\_PYTHON\_BUILD       | Do not apply Python build even if repo indicates it            | `false` | `true`, `false`
 VIRTUALENV\_NAME             | Specify Python virtual environment name                        | ""      | "antenv2.7"
 DISABLE\_COLLECTSTATIC       | Disable running `collectstatic` when building Django apps.      | `false` | `true`, `false`
+RUN_BUILD_COMMAND            | Custom command ran at the very end. After virtual envionment, `pip install`, creating python package egg, enabling collecstatic, compresing virtual environment (where applicable)'                                           | ""  | "echo hello world"
```
cormacpayne, line 51: What's the difference between this `RUN_BUILD_COMMAND` and the `POST_BUILD_COMMAND`? Just from reading the descriptions they seem similar, but it's hard to tell the difference.

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -244,4 +243,10 @@ fi
 			fi
 		fi
 	{{ end }}
-{{ end }}
\ No newline at end of file
+{{ end }}
+
+if [ ! -z "$RUN_BUILD_COMMAND" ]
+then
+	echo 'Running $RUN_BUILD_COMMAND'
```
cormacpayne, line 250: nit: just to make the log a little more clear, I would suggest changing the wording here to:

`echo 'Running the following provided RUN_BUILD_COMMAND: $RUN_BUILD_COMMAND`

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -162,6 +162,13 @@ public PlatformDetectorResult Detect(RepositoryContext context)
 
             var manifestFileProperties = new Dictionary<string, string>();
 
+            string runBuildCommand = Environment.GetEnvironmentVariable("RUN_BUILD_COMMAND");
```
cormacpayne, line 165: nit: I'd suggest pulling the `"RUN_BUILD_COMMAND"` string out into its own private `const` defined at the top of this file.

cormacpayne (changes requested):

william-msft:
Closing until we re-visit

PR #1117, opened 2021-09-22
add some missing python modules and update python platform versions with latest patches

arroyc:
- [x] The purpose of this PR is explained in this message or in an issue. >> https://stackoverflow.com/questions/12806122/missing-python-bz2-module/12806325#12806325


gpcastro (approved):

src/BuildScriptGenerator/PythonVersions.cs
```diff
@@ -6,9 +6,9 @@ public static class PythonVersions
     {
         public const string PipVersion = "20.2.3";
         public const string Python27Version = "2.7.18";
-        public const string Python36Version = "3.6.12";
-        public const string Python37Version = "3.7.9";
-        public const string Python38Version = "3.8.6";
-        public const string Python39Version = "3.9.0";
```
william-msft, line 12: @arroyc maybe we should update the unit test to prevent future pipeline failures?
https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/Python/PythonDynamicInstallationTest.cs#L81
https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/CommandTests/PlatformsCommandTest.cs#L79

arroyc: yes we will surely .. i have triggered a nightly build .. thats usually a good way to track which tests are failing and then just fix those tests .. but I guess you already did find part of those tests :) so thanks

qianz2 (approved):

PR #1149, opened 2021-11-02
Williamhe/1394597 add oryx run script

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1394597.


images/build/Dockerfiles/Dockerfile
```diff
@@ -18,9 +18,12 @@ ENV RELEASE_TAG_NAME=${RELEASE_TAG_NAME}
 ENV DEBIAN_FLAVOR="stretch"
 
 RUN ./build.sh dotnetcore /opt/startupcmdgen/dotnet
+RUN ./build.sh golang     /opt/startupcmdgen/golang
+RUN ./build.sh hugo       /opt/startupcmdgen/hugo
 RUN ./build.sh node       /opt/startupcmdgen/nodejs
 RUN ./build.sh php        /opt/startupcmdgen/php
 RUN ./build.sh python     /opt/startupcmdgen/python
+RUN ./build.sh ruby       /opt/startupcmdgen/ruby
```
arroyc, line 26: not needed ?

images/build/Dockerfiles/Dockerfile
```diff
@@ -18,9 +18,12 @@ ENV RELEASE_TAG_NAME=${RELEASE_TAG_NAME}
 ENV DEBIAN_FLAVOR="stretch"
 
 RUN ./build.sh dotnetcore /opt/startupcmdgen/dotnet
+RUN ./build.sh golang     /opt/startupcmdgen/golang
+RUN ./build.sh hugo       /opt/startupcmdgen/hugo
```
arroyc, line 22: not needed?

src/BuildScriptGenerator/Constants.cs
```diff
@@ -20,7 +20,7 @@ public static class Constants
         public const string True = "true";
         public const string False = "false";
 
-        public const string TemporaryInstallationDirectoryRoot = "/tmp/oryx/platforms";
+        public const string TemporaryInstallationDirectoryRoot = "/opt/";
```
arroyc, line 23: that does not look right..  why do you want to change this

src/startupscriptgenerator/src/golang/scriptgenerator.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+type GolangStartupScriptGenerator struct {
+	AppPath					 string
+	UserStartupCommand		 string
+	DefaultAppPath			 string
+	BindPort                 string
+	Manifest                 common.BuildManifest
+	Configuration            Configuration
+}
```
arroyc, line 20: nit: indent

src/startupscriptgenerator/src/python/scriptgenerator.go
```diff
@@ -67,7 +67,7 @@ func (gen *PythonStartupScriptGenerator) GenerateEntrypointScript() string {
 	scriptBuilder.WriteString("echo 'cd $APP_PATH' >> ~/.bashrc\n")
 
 	if gen.Configuration.EnableDynamicInstall && !common.PathExists(pythonInstallationRoot) {
-		scriptBuilder.WriteString(fmt.Sprintf("oryx setupEnv -appPath %s\n", gen.getAppPath()))
+		scriptBuilder.WriteString(fmt.Sprintf("oryx prep --src %s\n", appPath))
```
arroyc, line 70: why? i would not change this  for now

src/startupscriptgenerator/src/ruby/scriptgenerator.go
```diff
@@ -38,7 +38,7 @@ func (gen *RubyStartupScriptGenerator) GenerateEntrypointScript() string {
 
 	enableDynamicInstall := common.GetBooleanEnvironmentVariable(consts.EnableDynamicInstallKey)
 	if enableDynamicInstall && !common.PathExists(rubyInstallationRoot) {
-		scriptBuilder.WriteString(fmt.Sprintf("oryx setupEnv -appPath %s\n", gen.SourcePath))
+		scriptBuilder.WriteString(fmt.Sprintf("oryx prep --src %s\n", appPath))
```
arroyc, line 41: why? i would not change this  for now

tests/BuildScriptGeneratorCli.Tests/BuildCommandTest.cs
```diff
@@ -224,7 +224,8 @@ public void DoesNotShowHelp_EvenIfIntermediateDir_DoesNotExistYet()
         // sure wouldn't change. Since we cannot update product code with test generator we cannot run this test in
         // a docker container. So we run this test on a Linux OS only as build sets execute permission flag and
         // as well as executes a bash script.
-        [EnableOnPlatform("LINUX")]
+        [Fact (Skip = "Skipping test temporarily")]
```
arroyc, line 227: why?

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -75,7 +75,7 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 	nodeBinary := fmt.Sprintf("%s/bin/node", consts.NodeInstallationDir)
 	enableDynamicInstall := common.GetBooleanEnvironmentVariable(consts.EnableDynamicInstallKey)
 	if enableDynamicInstall && !common.PathExists(nodeBinary) {
-		scriptBuilder.WriteString(fmt.Sprintf("oryx setupEnv -appPath %s\n", gen.SourcePath))
+		scriptBuilder.WriteString(fmt.Sprintf("oryx prep --src %s\n", appPath))
```
arroyc, line 78: why? i would not change this  for now

arroyc (commented):
need to take a second look .. but  by then you can take care of these comments, also  try to fix the build error

images/build/Dockerfiles/Dockerfile
william-msft: @arroyc may you elaborate on what you mean? Without this code, the binaries from `go build` that are needed in https://github.com/microsoft/Oryx/blob/21ef7ce9cd1b58b68a8e5da304c9ed0ff6bf1d7d/src/BuildScriptGenerator/DefaultRunScriptGenerator.cs#L92 will cause a failure

images/build/Dockerfiles/Dockerfile
william-msft: Same as https://github.com/microsoft/Oryx/pull/1149/files/ce4f94e5a8fe1c05e3865f38118613f56f67d30d#r746056556

src/BuildScriptGenerator/Constants.cs
william-msft: I thought about this more and could not find a real benefit to us or users installing platforms in `/tmp/oryx/platforms`. `/opt` makes more sense since that is where add-on application software is usually stored. 

Do you foresee this being a breaking change?

src/startupscriptgenerator/src/node/scriptgenerator.go
william-msft: `oryx setupEnv` fails when invoked from the generated `/tmp/run.sh` in the latest build image. `oryx prep` works. Will this be a breaking change for the run image? If not may you elaborate on why we should hold off for now?

src/startupscriptgenerator/src/python/scriptgenerator.go
william-msft: Same as https://github.com/microsoft/Oryx/pull/1149/files/ce4f94e5a8fe1c05e3865f38118613f56f67d30d#r745944937

src/startupscriptgenerator/src/ruby/scriptgenerator.go
william-msft: Same as https://github.com/microsoft/Oryx/pull/1149/files/ce4f94e5a8fe1c05e3865f38118613f56f67d30d#r745944937

tests/BuildScriptGeneratorCli.Tests/BuildCommandTest.cs
william-msft: Accidental addition from merging branches. Revert of this has been made now

src/startupscriptgenerator/src/golang/scriptgenerator.go
william-msft: Updated. Thanks!

src/BuildScriptGenerator/Constants.cs
william-msft: I noticed this caused various pipeline issues like `permissions denied` so I reverted back to original location

images/build/Dockerfiles/Dockerfile
arroyc: im asking why do we need the hugo? do we have startupscriptgenerator for hugo?

william-msft: @arroyc currently there's no startupscriptgenerator for hugo, but I have added it in this PR

src/startupscriptgenerator/src/python/scriptgenerator.go
arroyc: I would   rather investigate why is it failing .. because prep is for different purpose, im guessing we somehow need to enable the setupEnv command in build image

images/build/Dockerfiles/Dockerfile
arroyc: unless we have startupscriptgenerator we should not add this

arroyc: we should only add what we have

images/build/Dockerfiles/Dockerfile
arroyc: same comment

arroyc (changes requested):

images/retry.sh
```diff
@@ -8,13 +8,15 @@
...
-while [ "$r" -le "$retries" ]
-do
-	echo "retry $r"
-	$1 && break
-	echo "error executing: $1"
-	r=$((r+1)) 
-	sleep 15
-done
\ No newline at end of file
+if [[ "$OSTYPE" == "linux-gnu" ]] || [[ "$OSTYPE" == "darwin"* ]]; then
```
arroyc, line 11: why are we checking linux type

arroyc: is it to bypass local windows dev situation

images/retry.sh
```diff
@@ -8,13 +8,15 @@
...
+	r=0
```
arroyc, line 12: instead of r better something like eachretry or something meaningful

src/startupscriptgenerator/src/node/scriptgenerator.go
arroyc: do we know why setupenv fails when its called from run.sh?

oryx prep is prepare build environment , oryx setupEnv is for preparing runtime environment. We should not mix the usage of these two commands, ideally we should use only one command from now on as we are consolidating both build and runtime environment in one image, but I am not expecting that consolidation of prep and setup command in this PR. That can be done later. But I would be curious to know why setupEnv command is failing before I use prep command

images/build/Dockerfiles/Dockerfile
william-msft: @arroyc work item 1394597 requests support for all minus Java. Are asking I remove all Hugo related code? https://github.com/microsoft/Oryx/pull/1149/commits/2cd6d2aa3f037ee05424eaaa711a5d4376e9cd22

images/build/Dockerfiles/Dockerfile
william-msft: @arroyc Ruby currently has startupscriptgenerator code in main: https://github.com/microsoft/Oryx/tree/main/src/startupscriptgenerator/src/ruby
Are we suggesting we remove ruby related startupscriptgenerator code?

images/retry.sh
william-msft: Reverted. Please ignore

images/retry.sh
william-msft: Updated. Thanks!

src/startupscriptgenerator/src/node/scriptgenerator.go
william-msft: @arroyc I see. `oryx setupEnv` since it is not implemented in the `latest` build image, which make sense now. I've reverted back to `oryx setupEnv` so we could consolidate later

images/build/Dockerfiles/Dockerfile
arroyc: all im saying is if you add this line here that means we are assuming there is a startupscriptgenerator for hugo ... which is not there .. if you add it here I expect a startupscriptgenerator for hugo also present in this PR

images/build/Dockerfiles/Dockerfile
arroyc: nah .. i will take back this .. I see ruby startupscript generator is there already .. so you can keep it

images/build/Dockerfiles/Dockerfile
william-msft: @arroyc hugo startupscriptgenerator is in this PR: https://github.com/microsoft/Oryx/blob/417e9d8bcf355dc203c9a7868ed00629d5f245cd/src/startupscriptgenerator/src/hugo/scriptgenerator.go

Are you asking the `orxy setupEnv` get added?

arroyc (approved):

william-msft:
Closing this PR for: https://github.com/microsoft/Oryx/pull/1331/files

PR #1152, opened 2021-11-12
Williamhe/1432952 add jdk17 java support

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1432952.


platforms/java/getJavaSdk.sh
```diff
@@ -46,6 +47,20 @@ downloadJavaSdk()
     IFS='.' read -ra VERSION_PARTS <<< "$JDK_VERSION"
     majorVersion="${VERSION_PARTS[0]}"
 
+    # Version 17 has a different url format than rest of the versions, so special casing it.
+    if [ "$majorVersion" == "17" ]; then
+        local buildNumber="12"
+        local url="https://download.java.net/java/GA/jdk${JDK_VERSION}/2a2082e5a09d4267845be086888add4f/${buildNumber}/GPL/openjdk-${JDK_VERSION}_linux-x64_bin.tar.gz"
```
cormacpayne, line 53: Do you know if this URL (specifically the `2a2082e5a09d4267845be086888add4f` part) will follow the same pattern for later versions of JDK 17? It may be something that we need to keep an eye out for when version 17.0.2 (or greater) of the JDK is released.

platforms/java/getJavaSdk.sh
```diff
@@ -46,6 +47,20 @@ downloadJavaSdk()
...
+        mkdir -p extracted
+        tar -xf $tarFileName --directory extracted
+        tar -zcf "$hostJavaArtifactsDir/$tarFileName" .
+		echo "Version=$JDK_VERSION" >> "$hostJavaArtifactsDir/java-$JDK_VERSION-metadata.txt"
+        return
+    fi
+
     if shouldBuildSdk java $tarFileName || shouldOverwriteSdk || shouldOverwriteJavaSdk; then
         local baseUrl="https://github.com/AdoptOpenJDK/openjdk${majorVersion}-binaries/releases/download"
         if [ "$majorVersion" == "10" ] && [ "$JDK_BUILD_NUMBER" == "13" ]; then
```
cormacpayne, line 66: It seems JDK 17 is available in its own repository within this `AdoptOpenJDK` organization, so even though the URL is different for JDK 17, is there any reason we can't add an additional if-statement like we do here for 10.x.x.13 to special case the URL?

platforms/java/getJavaSdk.sh
william-msft: That's a good point. I removed the URL parameterization and made update minor updates

platforms/java/getJavaSdk.sh
william-msft: `17.0.1` currently isn't supported: https://github.com/AdoptOpenJDK/openjdk17-binaries/releases/
Last release was in May 2021 & 17.0.1 was last released in September 2021

arroyc: look at dotnet core example for supporting multiple download link for sdks ... I'm approving this for one time to unblock you for releasing for vso

arroyc: here is an example https://github.com/microsoft/Oryx/blob/9c093d1d9780093a4c907c4222d2888c8bbe2304/platforms/dotnet/versionsToBuild.txt#L65

arroyc (approved):

PR #1158, opened 2021-11-18
Adding code profiler packages needed for Python

vijaysaayi:
The goal of this Pull request is to add the necessary packages for the code profiler viztracer along with Oryx runtime image.

**Packages needed:**
1. viztracer : this is the code profiler package
2. vizplugins : these help in adding cpu and memory stats to the trace
3. orjson: viztracer package dumps the trace as a json. In their official repo, it is explained that if orjson package is available viztracer would use it as it is better performant when compared to json package.

### DIFF
--- a/images/runtime/python/cgmanifest.json
+++ b/images/runtime/python/cgmanifest.json
@@ -1,0 +1,34 @@
+{
+    "Registrations": [
+        {
+            "Component": {
+                "pip": {
+                    "Name": "viztracer",
+                    "Version": "0.14.2"
+                },
+                "Type": "pip"
+            },
+            "DevelopmentDependency": false
+        },
+        {
+            "Component": {
+                "pip": {
+                    "Name": "vizplugins",
+                    "Version": "0.1.2"
+                },
+                "Type": "pip"
+            },
+            "DevelopmentDependency": false
+        },
+        {
+            "Component": {
+                "pip": {
+                    "Name": "orjson",
+                    "Version": "3.6.4"
+                },
+                "Type": "pip"
+            },
+            "DevelopmentDependency": false
+        }
+    ]
+}
--- a/images/runtime/python/template.Dockerfile
+++ b/images/runtime/python/template.Dockerfile
@@ -39,9 +39,9 @@ RUN ${IMAGES_DIR}/runtime/python/install-dependencies.sh
 RUN pip install --upgrade pip \
     && pip install gunicorn \
     && pip install debugpy \
-    && pip install viztracer \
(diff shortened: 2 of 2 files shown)


images/runtime/python/cgmanifest.json
```diff
@@ -0,0 +1,34 @@
+{
+    "Registrations": [
```
william-msft, line 2: @vijaysaayi may you elaborate on the following:

- Why is this governance file needed?
- Where will `cgmanifest.json` be ingested?
-  If `cgmanifest.json` is ingested outside of Oryx, may we move this file out of Oryx?
- Should `DevelopmentDependency` be `true` since "The vizplugin should be used with viztracer" per: https://github.com/gaogaotiantian/vizplugins

images/runtime/python/template.Dockerfile
```diff
@@ -39,6 +39,9 @@ RUN ${IMAGES_DIR}/runtime/python/install-dependencies.sh
 RUN pip install --upgrade pip \
     && pip install gunicorn \
     && pip install debugpy \
+    && if [ "%PYTHON_MAJOR_VERSION%" = "3" ] && [ "%PYTHON_VERSION%" != "3.6" ]; then pip install viztracer \
+    && pip install vizplugins \
+    && pip install orjson ;fi \
```
william-msft, line 44: nit: May we update semicolon on line 44 to the following: `&& pip install orjson; fi \`

images/runtime/python/cgmanifest.json
vijaysaayi: @william-msft : 
1. We must [register Open source components to use them in Microsoft Products](https://docs.opensource.microsoft.com/using/ ).
For stacks such as Python, the Component Governance injected into devops pipeline would not automatically pick the open source components. Hence We need to add cgmanifest.json file to explicitly register them.
2. [cgmanifest](https://docs.opensource.microsoft.com/tools/cg/features/cgmanifest/) file can be placed anywhere in the repository and Component Governance of DevOps will pick this file to register the components. 
3. Component Governance is injected DevOps pipeline of Oryx . Hence the cgmanifest file should be placed in Oryx only.
4. [DevelopmentDependency](https://docs.opensource.microsoft.com/tools/cg/features/cgmanifest/) indicates that these packages are used for development only. In our case vizplugin is used in production. Hence I set it to false.

images/runtime/python/template.Dockerfile
vijaysaayi: Updated

images/runtime/python/cgmanifest.json
william-msft: Got it, thank you @vijaysaayi!

william-msft (approved):

PR #1195, opened 2022-01-04
Restructure PHP pipelines

william-msft:
- [ ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1449555


qianz2:
Hi William, I see that the reason for doing it is to reduce the space within the pipeline but I think the restructure based on different php versions will make us not easy to maintain all the integration tests and pipelines?  What happened if the old php versions are out of support or new php versions we need to add? And what about other platforms like python, dotnet, are they need to restructure as well?

william-msft:
> Hi William, I see that the reason for doing it is to reduce the space within the pipeline agent but I think the restructure based on different php versions will make us not easy to maintain all the integration tests and pipelines? What happened if the old php versions are out of support or new php versions we need to add? And what about other platforms like python, dotnet, are they need to restructure as well?

Hey  @qianz2 the plan would be to drop integration tests for php versions that php also drops support for (note the removal of old php versions) Unfortunately the category trait `[Fact, Trait("category", "php-8.0")]` needs to be hard coded and resolved at compile time, which prevents us from creating pipelines dynamically. The time and space performance improvements are much better after cleanup/restructure.
If we keep the current setup we'd have to remove test parametrization in the source code or risk of hitting our 10GB space limit. 

The risk of hitting our 10GB is less now, since old test version were removed. This is because every php image, specified in the test, is downloaded. 
image


If we decide to go with this approach I can make the change for the other platforms.

william-msft:
> Hi William, I see that the reason for doing it is to reduce the space within the pipeline agent but I think the restructure based on different php versions will make us not easy to maintain all the integration tests and pipelines? What happened if the old php versions are out of support or new php versions we need to add? And what about other platforms like python, dotnet, are they need to restructure as well?

> Hi William, I see that the reason for doing it is to reduce the space within the pipeline agent but I think the restructure based on different php versions will make us not easy to maintain all the integration tests and pipelines? What happened if the old php versions are out of support or new php versions we need to add? And what about other platforms like python, dotnet, are they need to restructure as well?

Hey  @qianz2 the plan would be to drop integration tests for php versions that php also drops support for (note the removal of old php versions) Unfortunately the category trait `[Fact, Trait("category", "php-8.0")]` needs to be hard coded and resolved at compile time, which prevents us from creating pipelines dynamically. The time and space performance improvements are much better after cleanup/restructure.
If we keep the current setup we'd have to remove test parametrization in the source code or risk of hitting our 10GB space limit. 

The risk of hitting our 10GB is less now, since old test version were removed. This is because every php image, specified in the test, is downloaded. 
image


The other platforms don't need restructuring at the moment. But I could make the change to keep it consistent.

qianz2 (approved):

PR #1200, opened 2022-01-12
Add python 3.10 base image

qianz2:


images/runtime/python/3.10/buster.Dockerfile
```diff
@@ -0,0 +1,39 @@
+ARG DEBIAN_FLAVOR
```
william-msft, line 1: Hi @qianz2, will we be adding a base-image-python pipeline? We might need to add a YAML file similar to: https://github.com/microsoft/Oryx/blob/main/vsts/pipelines/baseImages/php.yml

After it goes through the base-image-python pipeline, the image in the similar format would be generated: `mcr.microsoft.com/oryx/python:3.9-20211214.1`

qianz2: Thanks for reminding, I'll add a base image python pipeline in another PR.

william-msft (approved):

qianz2:
`/opt/python/3.10.0/bin/python3.10: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /opt/python/3.10.0/lib/libpython3.10.so.1.0)`

It looks that debian 10 buster only supports libc6_2.28, but python 3.10 requires libc6_2.29, the way to resolve it is to use debian 11 Bullseye which supports latest version libc6_2.31.

PR #1211, opened 2022-01-27
Support php 8.1 runtime base image

qianz2:


images/runtime/php/8.1/8.1.Dockerfile
```diff
@@ -0,0 +1,261 @@
+ARG DEBIAN_FLAVOR
```
william-msft, line 1: Let's try re-running this script to generate new Dockerfiles:
https://github.com/microsoft/Oryx/blob/main/images/runtime/php/generateDockerfiles.sh#L47

It should generate the MCR base image tag we are looking for

qianz2: I found here is the list of mcr image tags: https://mcr.microsoft.com/v2/oryx/base/tags/list, and php 8.1 was created today through the base image pipeline.

william-msft (approved):

PR #1174, opened 2021-12-02
Populate prebuild and postbuild from procfile

tulikac:
This PR is to read the app.yaml (procfile) and populate the prebuild and postbuild commands.

The app.yaml has a structure like this

version: 1

pre-build: apt-get install xyz

build: pip install requirements.txt
  
post-build: |
  python manage.py makemigrations
  python manage.py migrate
  
run: gunicorn myapp.app --workers 5 --foo bar

If the app.yaml exists, I am overwriting any prebuild and postbuild appsettings that the customer might have.

PR #1235, opened 2022-02-14
User/tulikac/procfiles

tulikac:
This is related to https://github.com/microsoft/Oryx/pull/1174.

I had changed the FileExists method to not use context.SourceRepo.FileExists. However, as a result, the YAML was not found when using the zip deploy method. So updating the method to use context.SourceRepo.RootPath


src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -316,8 +316,13 @@ private void LogScriptIfGiven(string type, string scriptPath)
...
 
-            // Construction and checking the file existence here to allow tests to pass
-            string filePathForAppYaml = Path.Combine(sourceDirInBuildContainer, "app.yaml");
+
+            // Workaround for bug in TestSourceRepo class in validation tests
+            // Should be using context.SourceRepo.FileExists
+            string filePathForAppYaml = Path.Combine(context.SourceRepo.RootPath, "app.yaml");
+
+            _logger.LogDebug("Path to app.yaml " + filePathForAppYaml);
+            _writer.WriteLine("Path to app.yaml " + filePathForAppYaml);
```
william-msft, line 325: @tulikac may we keep only `_writer.WriteLine` or `_writer.LogDebug`?

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -353,6 +358,11 @@ private void LogScriptIfGiven(string type, string scriptPath)
                     _writer.WriteLine("Invalid app.yaml format");
                 }
             }
+            else
+            {
+                _logger.LogDebug("No app.yaml found");
```
william-msft, line 363: @tulikac may we keep only `_writer.WriteLine` or `_writer.LogDebug`?  Same with line 323-352 please

tulikac: Done

daniv-msft (approved):

PR #1251, opened 2022-02-19
Williamhe/1486537 add log warning and fix log message timestamp

william-msft:
- [ ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1486537

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -178,7 +178,6 @@ fi
             LogWarning "${StdWarning} | Exit code: {pythonBuildExitCode} | Please review message | ${moreInformation}"
             exit $pythonBuildExitCode
         fi
-        set -e
     else
         echo $REQS_NOT_FOUND_MSG
     fi
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -159,6 +159,7 @@ fi
         fi
     elif [ -e "pyproject.toml" ]
     then
+        set +e
         echo "Running pip install poetry..."
         InstallPipCommand="pip install poetry"
         printf %s " , $InstallPipCommand" >> "$COMMAND_MANIFEST_FILE"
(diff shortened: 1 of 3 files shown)


src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -171,11 +172,13 @@ fi
         pythonBuildExitCode=${PIPESTATUS[0]}
         ELAPSED_TIME=$(($SECONDS - $START_TIME))
         echo "Done in $ELAPSED_TIME sec(s)."
+        set -e
         if [[ $pythonBuildExitCode != 0 ]]
         then
             LogWarning "${StdWarning} | Exit code: {pythonBuildExitCode} | Please review message | ${moreInformation}"
             exit $pythonBuildExitCode
         fi
+        set -e
```
daniv-msft, line 181: Nit: do you need this one given that you already added one line 175?

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
```diff
@@ -1226,13 +1226,19 @@ public void Django_CollectStaticFailure_DoesNotFailBuild()
                 CommandToExecuteOnRun = "/bin/bash",
                 CommandArguments = new[] { "-c", script }
             });
+            // Regex will match:
+            // "yyyy-mm-dd hh:mm:ss"|ERROR|Failed pip installation with exit code: 1
+            // Example:
+            // "2021-10-27 07:00:00"|ERROR|Failed to pip installation with exit code: 1
+            Regex regex = new Regex(@"""[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1]) (0[0-9]|1[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])""\|WARNING\|.*");
```
daniv-msft, line 1233: I hope I'm not missing something, but are we only handling logs with WARNING here?

daniv-msft: Specifically, given the Assert.Contains we're fixing, shouldn't we be looking for something like "'collectstatic' exited with exit code 1."?

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
william-msft: No, thanks for catching this :)

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
william-msft: The ask changed so the stacktrace of the failure would show; to give better reason of the failure for the user. `collectstatic exited with exit code 1` was too ambiguous. 

However I did update the regex to pattern match for `Exit code 1`

daniv-msft (approved):
Looks good to me!

vijaysaayi (approved):

PR #1266, opened 2022-02-25
Fix to reduce python 3.9 or older version image size

qianz2:

### DIFF
--- a/build/buildPythonSdkByVersion.sh
+++ b/build/buildPythonSdkByVersion.sh
@@ -15,7 +15,7 @@ version="$1"
 
 buildPythonfromSource()
 {
-    pythonVersion=$PYTHON_VERSION 
+    pythonVersion=$PYTHON_VERSION
 
     if [ ! -z "$1" ]; then
        echo "$1"
@@ -26,7 +26,7 @@ buildPythonfromSource()
        echo "$2"
        gpgKey=$2
     fi
-    
+
     wget https://www.python.org/ftp/python/${pythonVersion%%[a-z]*}/Python-$pythonVersion.tar.xz -O /python.tar.xz
     wget https://www.python.org/ftp/python/${pythonVersion%%[a-z]*}/Python-$pythonVersion.tar.xz.asc -O /python.tar.xz.asc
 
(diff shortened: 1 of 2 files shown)


daniv-msft (dismissed):
Looks good! Thanks for the quick turnaround

images/runtime/python/template.Dockerfile
```diff
@@ -47,12 +47,12 @@ RUN chmod +x /tmp/receiveGpgKeys.sh
...
     apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
-        build-essential \ 
+        build-essential \
         tk-dev \
         uuid-dev \
         libgeos-dev
 
-RUN ${BUILD_DIR}/buildPythonSdkByVersion.sh $PYTHON_VERSION $DEBIAN_FLAVOR 
+RUN if [ "%PYTHON_VERSION%" == "3.10" ]; then ${BUILD_DIR}/buildPythonSdkByVersion.sh $PYTHON_VERSION $DEBIAN_FLAVOR; else ${IMAGES_DIR}/installPlatform.sh python $PYTHON_VERSION --dir /opt/python/$PYTHON_VERSION --links false; fi
```
daniv-msft, line 55: Should we change the logic to be `"%PYTHON_VERSION%" >= "3.10"` ? I imagine that the future version of Python will be in the same case as 3.10, right?

qianz2: Thanks for catching that, will update it

build/buildPythonSdkByVersion.sh
```diff
@@ -36,14 +36,16 @@ buildPythonfromSource()
         apt-get update && \
         apt-get upgrade -y && \
         DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+            build-essential \
```
daniv-msft, line 39: Could you please clarify why we're adding build-essential and libgeos-dev? I'm just curious to understand. :)

qianz2: Hi Dan, I removed these dependencies from the template.Dockerfile because it'll apply to all python versions if we put in the template.Dockerfile(It's also the reason why it causes other image size increase). But they're still required dependencies for python 3.10 so I put them here only applies to 3.10.

PR #1278, opened 2022-03-03
Add initial code analysis and rulesets for repository

cormacpayne:
Introduce the initial set of code analysis tooling and rulesets necessary to enforce specific coding guidelines for the repository.

Feel free to let me know if any of the changes don't make sense or don't seem necessary -- we can easily suppress the rule via the `.ruleset` file.

In future updates, we can
- Require the `this.` qualifier for all instance variables
- Remove the `_` prefix for instance variables
- Require a singular newline for all files
- Investigate the amount of work required to
  - add XML documentation to all projects, and
  - enable code analysis and rulesets for test projects


daniv-msft:
@cormacpayne Could you please clarify how the code analysis results would be visible for developers? Do they need to install a specific tool to see it when developing, or is it integrated by default in Visual Studio/VS Code? I imagine that we'll block the build if there are any issues?

cormacpayne:
> @cormacpayne Could you please clarify how the code analysis results would be visible for developers? Do they need to install a specific tool to see it when developing, or is it integrated by default in Visual Studio/VS Code? I imagine that we'll block the build if there are any issues?

Sure thing -- the code analysis is specifically for C# files, and any code that goes against the ruleset in place will result in a compilation error, which is seen in Visual Studio and when building the project externally, so developers will be able to quickly see what needs to be resolved and our validation pipeline will also fail with the corresponding error(s).

daniv-msft (dismissed):
Looks good to me!

daniv-msft (dismissed):

cormacpayne:
@daniv-msft -- looks like the build isn't able to find some of the binaries for the code analysis packages added. I'm reading that this might be due to the version of MSBuild used, so I'm testing out some bumps to the .NET SDK version we're using to bring it up to at least 17.x.x (VS 2022).

william-msft:
> @daniv-msft -- looks like the build isn't able to find some of the binaries for the code analysis packages added. I'm reading that this might be due to the version of MSBuild used, so I'm testing out some bumps to the .NET SDK version we're using to bring it up to at least 17.x.x (VS 2022).

@cormacpayne let's check Oryx-nightly runs successfully before merging please:
https://devdiv.visualstudio.com/DevDiv/_build?definitionId=10393

cormacpayne:
Corresponding Oryx-Nightly pipeline: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5846363&view=results

william-msft (approved):

PR #1279, opened 2022-03-03
Williamhe/1324583 update framework detection with additional requirements

william-msft:
- [ ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1324583

### DIFF
--- a/src/Detector/Node/NodeDetector.cs
+++ b/src/Detector/Node/NodeDetector.cs
@@ -202,6 +202,8 @@ namespace Microsoft.Oryx.Detector.Node
             var detectedFrameworkResult = new List<FrameworkInfo>();
             var packageJson = GetPackageJsonObject(context.SourceRepo, _logger);
             var monitoredDevDependencies = NodeConstants.DevDependencyFrameworkKeyWordToName;
+            // frameworksSet is for preventing duplicates
+            var frameworksSet = new HashSet<string>();
             
             // dev-dependencies
             var devDependencies = packageJson?.devDependencies != null ? packageJson.devDependencies : new string[0];
@@ -211,15 +213,19 @@ namespace Microsoft.Oryx.Detector.Node
 
                 // wild-card dependency
                 (bool isWildCardDependency, string wildCarddependencyName) = GetWildCardDependency(dependencyName);
-
-                if (monitoredDevDependencies.ContainsKey(dependencyName) || isWildCardDependency) 
+                
+                bool isValidDependency = monitoredDevDependencies.ContainsKey(dependencyName) || isWildCardDependency;
+                string frameworkName = isWildCardDependency ? wildCarddependencyName : monitoredDevDependencies[dependencyName];
+                bool isNotDuplicateFramework= !(frameworksSet.Contains(frameworkName));
+                if (isValidDependency && isNotDuplicateFramework) 
                 {
                     var frameworkInfo = new FrameworkInfo
                     {
-                        Framework = isWildCardDependency ? wildCarddependencyName : monitoredDevDependencies[dependencyName],
+                        Framework = frameworkName,
                         FrameworkVersion = dependency.Value.Value
                     };
                     detectedFrameworkResult.Add(frameworkInfo);
+                    frameworksSet.Add(frameworkName);
                 }
             }
 
@@ -233,14 +239,18 @@ namespace Microsoft.Oryx.Detector.Node
                 // wild-card dependency
                 (bool isWildCardDependency, string wildCarddependencyName) = GetWildCardDependency(dependencyName);
 
-                if (monitoredDependencies.ContainsKey(dependencyName) || isWildCardDependency)
+                bool isValidDependency = monitoredDependencies.ContainsKey(dependencyName) || isWildCardDependency;
+                string frameworkName = isWildCardDependency ? wildCarddependencyName : monitoredDependencies[dependencyName];
+                bool isNotDuplicateFramework = !(frameworksSet.Contains(frameworkName));
+                if (isValidDependency && isNotDuplicateFramework)
                 {
                     var frameworkInfo = new FrameworkInfo
                     {
-                        Framework = isWildCardDependency ? wildCarddependencyName : monitoredDependencies[dependencyName],
+                        Framework = frameworkName,
                         FrameworkVersion = dependency.Value.Value
                     };
                     detectedFrameworkResult.Add(frameworkInfo);
+                    frameworksSet.Add(frameworkName);
                 }
             }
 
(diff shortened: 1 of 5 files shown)


src/Detector/Node/NodeDetector.cs
```diff
@@ -211,15 +213,20 @@ private IEnumerable<FrameworkInfo> DetectFrameworkInfos(DetectorContext context)
 
                 // wild-card dependency
                 (bool isWildCardDependency, string wildCarddependencyName) = GetWildCardDependency(dependencyName);
-
-                if (monitoredDevDependencies.ContainsKey(dependencyName) || isWildCardDependency) 
+                
+                bool isValidDependency = monitoredDevDependencies.ContainsKey(dependencyName);
+                string frameworkName = isValidDependency ? monitoredDevDependencies[dependencyName] : "";
+                frameworkName = isWildCardDependency ? wildCarddependencyName : frameworkName;
+                bool isNotDuplicateFramework= !(frameworksSet.Contains(frameworkName));
```
cormacpayne, line 220: nit: to avoid confusion when using it in a conditional, I typically steer away from using `Not` (or other negation words) in `bool` variable names.

In this case, I would either keep the right-hand side logic the same and rename the variable to `isNewFramework`, or drop the leading `!`, rename the variable to `isDuplicateFramework` and change the conditional below to `!isDuplicateFramework`.

src/Detector/Node/NodeDetector.cs
```diff
@@ -211,15 +213,20 @@ private IEnumerable<FrameworkInfo> DetectFrameworkInfos(DetectorContext context)
 
                 // wild-card dependency
                 (bool isWildCardDependency, string wildCarddependencyName) = GetWildCardDependency(dependencyName);
-
-                if (monitoredDevDependencies.ContainsKey(dependencyName) || isWildCardDependency) 
+                
+                bool isValidDependency = monitoredDevDependencies.ContainsKey(dependencyName);
```
cormacpayne, line 217: nit: we should be able to remove the need for this `isValidDependency` variable by checking for the right-hand side in a conditional and continuing through the loop if it doesn't hold true (since we don't take any actions if `isValidDependency` is `false`):

```cs
// Ignore invalid dev dependencies
if (!monitoredDevDependencies.ContainsKey(dependencyName))
{
    continue;
}
```

src/Detector/Node/NodeDetector.cs
```diff
@@ -233,14 +240,19 @@ private IEnumerable<FrameworkInfo> DetectFrameworkInfos(DetectorContext context)
...
+                if (isValidDependency && isNotDuplicateFramework)
                 {
                     var frameworkInfo = new FrameworkInfo
                     {
-                        Framework = isWildCardDependency ? wildCarddependencyName : monitoredDependencies[dependencyName],
+                        Framework = frameworkName,
                         FrameworkVersion = dependency.Value.Value
                     };
                     detectedFrameworkResult.Add(frameworkInfo);
+                    frameworksSet.Add(frameworkName);
```
cormacpayne, line 255: nit: it may be worth looking to see if we can extract this logic in the `foreach` loop out into a separate method and reuse it between the dependencies and dev dependencies checks. It appears the only thing differing between the two cases are (1) iterating over `dependencies` vs. `devDependencies`, and (2) checking for the `dependencyName` in `monitoredDependencies` vs. `monitoredDevDependencies`, both cases we should be able to handle in a single method.

src/Detector/Node/NodeDetector.cs
```diff
@@ -253,6 +265,17 @@ private IEnumerable<FrameworkInfo> DetectFrameworkInfos(DetectorContext context)
                 detectedFrameworkResult.Add(frameworkInfo);
             }
 
+            // remove base frameworks if derived framework exists
+            if (frameworksSet.Contains("Gatsby") || frameworksSet.Contains("Next.js"))
+            {
+                detectedFrameworkResult.RemoveAll(x => x.Framework == "Angular");
+                detectedFrameworkResult.RemoveAll(x => x.Framework == "React");
+            }
+            if (frameworksSet.Contains("VuePress") || frameworksSet.Contains("Nuxt.js"))
```
cormacpayne, line 274: nit: would you mind adding a newline before this `if` block?

src/Detector/Node/NodeDetector.cs
```diff
@@ -202,6 +202,8 @@ private IEnumerable<FrameworkInfo> DetectFrameworkInfos(DetectorContext context)
             var detectedFrameworkResult = new List<FrameworkInfo>();
             var packageJson = GetPackageJsonObject(context.SourceRepo, _logger);
             var monitoredDevDependencies = NodeConstants.DevDependencyFrameworkKeyWordToName;
+            // frameworksSet is for preventing duplicates
```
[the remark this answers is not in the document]
william-msft: Updated :)

src/Detector/Node/NodeDetector.cs
william-msft: Updated

src/Detector/Node/NodeDetector.cs
```diff
@@ -211,15 +213,20 @@ private IEnumerable<FrameworkInfo> DetectFrameworkInfos(DetectorContext context)
 
                 // wild-card dependency
                 (bool isWildCardDependency, string wildCarddependencyName) = GetWildCardDependency(dependencyName);
-
-                if (monitoredDevDependencies.ContainsKey(dependencyName) || isWildCardDependency) 
+                
+                bool isValidDependency = monitoredDevDependencies.ContainsKey(dependencyName);
+                string frameworkName = isValidDependency ? monitoredDevDependencies[dependencyName] : "";
```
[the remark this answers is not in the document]
william-msft: This piece got removed

src/Detector/Node/NodeDetector.cs
william-msft: This piece got removed

src/Detector/Node/NodeDetector.cs
william-msft: I've added a work-item 1493329 to refactor later. I've added points to consider in the story description

src/Detector/Node/NodeDetector.cs
william-msft: I've added a work-item 1493329 to refactor later. I've added points to consider in the story description

src/Detector/Node/NodeDetector.cs
```diff
@@ -212,14 +217,20 @@ private IEnumerable<FrameworkInfo> DetectFrameworkInfos(DetectorContext context)
                 // wild-card dependency
                 (bool isWildCardDependency, string wildCarddependencyName) = GetWildCardDependency(dependencyName);
```
daniv-msft, line 218: Nit: not blocking, but an uppercase is missing in `wildCarddependencyName`

daniv-msft (dismissed):

src/Detector/Node/NodeDetector.cs
william-msft: Ah yes, updated. I will need a new approval :)

daniv-msft (approved):

PR #1292, opened 2022-03-09
Support node 17 and update platform versions docs

qianz2:


daniv-msft (approved):
Looks good, thanks!

platforms/nodejs/versionsToBuild.txt
```diff
@@ -28,4 +28,13 @@
...
\ No newline at end of file
+16.13.2
+16.14.0
+
+17.0.1
+17.1.0
+17.2.0
+17.3.1
+17.4.0
+17.5.0
```
william-msft, line 39: Do we need to update `constants.yaml`  and add new runtime Dockerfile for the new major 17 version?

william-msft: Similar to https://github.com/microsoft/Oryx/pull/572/files

qianz2: We don't need to update the runtime for 17, we will wait for node 18 coming up and update runtime dockerfiles for 18. We update runtime for even numbered versions since they're stable versions.

PR #1294, opened 2022-03-09
Williamhe/1495508 optimize nodejs pipeline integration tests

william-msft:
- [ ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1495508


tests/Oryx.Integration.Tests/Node/NodeAngularAppTest.cs
```diff
@@ -70,11 +118,11 @@ public async Task CanBuildAndRun_Angular6App_WithoutCompressedNodeModules(string
                 });
         }
 
-        [Fact]
-        public async Task CanBuildAndRun_Angular6App_With_NodeModule_Dir_Exists_InRoot_WithoutCompression()
+        [Theory]
+        [InlineData("8")]
```
daniv-msft, line 122: Looking online, the Theory+InlineData tags allow us to have the same test running with different parameters. However here, we're only using one InlineData tag. What's the difference compared to before?

tests/Oryx.Integration.Tests/Node/NodeAngularAppTest.cs
```diff
@@ -23,6 +22,55 @@ public NodeAngularAppTest(ITestOutputHelper output, TestTempDirTestFixture fixtu
...
+        ///     [Fact, Trait("category", "<platform>-<version>")]
+        ///
+        /// The pipeline will invoke these integration tests 
+        /// on the matching category attribute.
+        /// </summary>
+        [Fact, Trait("category", "node-8")]
+        public void PipelineTestInvocationsNode8()
+        {
+            string nodeVersion = "8";
+            CanBuildAndRun_Angular6App_WithoutCompressedNodeModules(nodeVersion);
```
daniv-msft, line 38: At this point, do we even need each function (`CanBuildAndRun_Angular6App_WithoutCompressedNodeModules`, etc.) to be a test? Should they just be a simple function without Theory or InlineData tags?

daniv-msft:
Could you please clarify what is the expected gain in terms of performance?

tests/Oryx.BuildImage.Tests/Python/PythonPackageTest.cs
```diff
@@ -26,7 +26,7 @@ public PythonPackageTest(ITestOutputHelper output) : base(output)
 
         public static IEnumerable<object[]> PythonPackageExamples => new object[][]
         {
-            new object[] { "boto3", "1.14.44",
+            new object[] { "boto3", "1.21.20",
```
cormacpayne, line 29: nit: would you mind pulling the latest changes from `main`? This change should then go away.

tests/Oryx.Integration.Tests/Node/NodeAngularAppTest.cs
cormacpayne: +1 to this -- it seems we're running these functions twice, so unless there's a scenario we're looking to cover where we run the functions one after another, then we should be able to remove the `Theory`/`InlineData` attribute from the functions themselves so that only the pipeline tests functions are called.

vsts/pipelines/templates/_integrationJobTemplate.yml
```diff
@@ -53,60 +69,34 @@ jobs:
...
+    variables:
+      skipComponentGovernanceDetection: true
+    timeoutInMinutes: 300
+    steps:
+        - script: |
+            echo "##vso[task.setvariable variable=BuildBuildImages;]false"
+            echo "##vso[task.setvariable variable=BuildRuntimeImages;]false"
+            echo "##vso[task.setvariable variable=TestBuildImages;]false"
+            echo "##vso[task.setvariable variable=TestRuntimeImages;]false"
+            echo "##vso[task.setvariable variable=TestIntegrationCaseFilter;]category=node-${{ nodeVersion }}"
```
cormacpayne, line 92: Nice 😀

tests/Oryx.Integration.Tests/Node/NodeAngularAppTest.cs
william-msft: That's true. I'll look either:

- removing theory / inlineData for every test method
- or adding a category for every test method  and removing the new methods (like `PipelineTestInvocationsNode8`) that invoke the tests

william-msft: That's true. I'll look either:

- removing theory / inlineData for every test method
- or adding a category for every test method  and removing the new methods (like `PipelineTestInvocationsNode8`) that invoke the tests

tests/Oryx.Integration.Tests/Node/NodeAngularAppTest.cs
william-msft: Ideally we want to test with more versions. Now we have attached it with a trait: https://github.com/microsoft/Oryx/pull/1294/commits/fe4b1aea7727001143caea98f28e3940991e7081#diff-096017d7f5e60abae085d1077d4a75861e2065edf62d853f9cfbf1385d66b706R142

tests/Oryx.Integration.Tests/Node/NodeAngularAppTest.cs
william-msft: Thanks Team, I've addressed with new changes.

william-msft:
> Could you please clarify what is the expected gain in terms of performance?

@daniv-msft improvement: 2.5hrs to 55mins for Node

tests/Oryx.Integration.Tests/Node/NodeAngularAppTest.cs
```diff
@@ -25,8 +24,8 @@ public NodeAngularAppTest(ITestOutputHelper output, TestTempDirTestFixture fixtu
 
         // Official Node.js version that is supported by Angular CLI 6.0+ is 8.9 or greater
         [Theory]
-        [InlineData("8")]
-        [InlineData("9.4")]
+        [InlineData("8"), Trait("category", "node-8")]
+        [InlineData("9.4"), Trait("category", "node-9")]
```
daniv-msft, line 28: Should "node-9" be "node-9.4" here? Or the corresponding InlineData be updated?

daniv-msft (dismissed):

tests/Oryx.Integration.Tests/Node/NodeAngularAppTest.cs
william-msft: Thanks for catching this @daniv-msft! I've updated it and would need re-approval please :)

cormacpayne (approved):

PR #1309, opened 2022-03-17
Add DotNet 6.0.3 SDKs

william-msft:
- [ ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1497390


platforms/dotnet/versionsToBuild.txt
```diff
@@ -70,6 +70,8 @@
 
 6.0.100, cb0d174a79d6294c302261b645dba6a479da8f7cf6c1fe15ae6998bc09c5e0baec810822f9e0104e84b0efd51fdc0333306cb2a0a6fcdbaf515a8ad8cf1af25b,
```
daniv-msft, line 71: As we are removing 6.0.0, is it possible for us to remove it from the versions to build as well?

william-msft: Historically we have not removed old versions from this file. @qianz2  would you happen to know if it'll be ok to remove phased out versions from this file?

qianz2: We have another verionsToBuild_{date}.txt file to store unsupported old version, such as 'php/versionsToBuild_11172020.txt', because we won't need to build those old versions and to avoid issue unable to find old verisons. We can gradually move unsupported old versions to there. But for dotnet 6.0.0, it's comparably latest so we continue to keep it in here.

daniv-msft: Thanks @william-msft and @qianz2. So at this point, once this PR is checked in, could you please confirm that we would still be able to build using version 6.0.0? If that's still the case, then I agree I'm fine with keeping this version in the versionsToBuild.txt file.

qianz2: Yes, we still have dotnet 6.0.0 SDKs in storage account. Here we only updated the minor versions of dotnet 6 runtime image We're able to build dotnet 6.0.0 apps.

daniv-msft: OK, thanks @qianz2!

daniv-msft (approved):

src/BuildScriptGenerator/DotNetCore/DotNetCoreSdkVersions.cs
```diff
@@ -11,7 +11,7 @@ public static class DotNetCoreSdkVersions
         public const string DotNetCore31SdkVersion = "3.1.415";
         public const string DotNetCore50SdkVersion = "5.0.302";
         public const string DotNet50SdkVersion = "5.0.403";
-        public const string DotNet60SdkVersion = "6.0.100";
+        public const string DotNet60SdkVersion = "6.0.201";
         public const string DotNet60RtmSdkVersion = "6-0-100-rtm-21527-11";
```
qianz2, line 15: Do we still need this `DotNet60RtmSdkVersion`

qianz2 (approved):

src/BuildScriptGenerator/DotNetCore/DotNetCoreSdkVersions.cs
william-msft: I can double check. And remove in https://github.com/microsoft/Oryx/pull/1314 if not

PR #1321, opened 2022-03-24
Update PHP 7.4.28 and 8.0.17 runtimes and dynamic installation

qianz2:


images/runtime/php/7.4/base.buster.Dockerfile
```diff
@@ -102,7 +102,7 @@ RUN set -eux; \
 #  - https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server
 # pecl/sqlsrv, pecl/pdo_sqlsrv requires PHP (version >= 7.3.0)
 RUN set -eux; \
-    if [[ $PHP_VERSION == 7.3.* || $PHP_VERSION == 7.4.* || $PHP_VERSION == 8.0.* ]]; then \
+    if [[ $PHP_VERSION == 7.4.* || $PHP_VERSION == 8.0.* ]]; then \
```
cormacpayne, line 105: nit: how does this `PHP_VERSION` environment variable get set? I see at the top of the file that it's declared as `ENV PHP_VERSION 7.4.28`, but is there a way to make it anything other than `7.4.*`? I'm curious how it could be `8.0.*` within this if-statement.

cormacpayne (dismissed):

build/constants.yaml
```diff
@@ -137,12 +137,12 @@
     php81-version: 8.1.0
     php81-keys: 528995BFEDFBA7191D46839EF9BA0ADA31CBD89E 39B641343D8C104B2B146DC3F9C39DC0B9698544
     php81-tar-sha256: a1317eff0723a2b3d3122bbfe107a1158570ea2822dc35a5fb360086db0f6bbc
-    php80-version: 8.0.13
```
william-msft, line 140: Hi @qianz2 do we need to updated the php tag https://github.com/microsoft/Oryx/blob/8ff689840b634cc2bfdd90942d5a1ab808e64b41/build/constants.yaml#L126 for the new runtimes?

qianz2: Updated, thanks

images/runtime/php/7.4/base.buster.Dockerfile
qianz2: We use `images/runtime/php/template.Dockerfile` when supporting a new verison, so basically we copied it from `template.Dockerfile` and set up the `PHP_VERSION `for it, that's why it's more like a general within this if statement.

daniv-msft (approved):

PR #1322, opened 2022-03-29
Update framework logging to StdOut

william-msft:
- [ ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1508351

### DIFF
--- a/src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
+++ b/src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
@@ -139,8 +139,8 @@ namespace Microsoft.Oryx.BuildScriptGenerator.DotNetCore
             {
                 installBlazorWebAssemblyAOTWorkloadCommand = DotNetCoreConstants.InstallBlazorWebAssemblyAOTWorkloadCommand;
                 manifestFileProperties[ManifestFilePropertyKeys.Frameworks] = "blazor";
-                this.logger.LogInformation("Detected the following framework(s): blazor");
-                Console.WriteLine("Detected the following framework(s): blazor");
+                this.logger.LogInformation("Detected the following frameworks: blazor");
+                Console.WriteLine("Detected the following frameworks: blazor");
             }
 
             var templateProperties = new DotNetCoreBashBuildSnippetProperties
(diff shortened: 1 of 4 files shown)


src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -139,7 +139,8 @@ public PlatformDetectorResult Detect(RepositoryContext context)
             {
                 installBlazorWebAssemblyAOTWorkloadCommand = DotNetCoreConstants.InstallBlazorWebAssemblyAOTWorkloadCommand;
                 manifestFileProperties[ManifestFilePropertyKeys.Frameworks] = "blazor";
-                this.logger.LogInformation("Detected the the following framework(s): blazor");
+                this.logger.LogInformation("Detected the following framework(s): blazor");
```
cormacpayne, line 142: nit: it may be worth creating a `GlobalLogger` class in the future that can log a single message to both Application Insights and STDOUT (although I'm not sure how often it would be used outside of the messages in this PR).

cormacpayne, line 142: nit: maybe create a `detectionMessage` variable that stores the string message you have for each change and is reused by the subsequent `LogInformation` and `WriteLine` calls?

cormacpayne (dismissed):

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -323,10 +323,10 @@ public PlatformDetectorResult Detect(RepositoryContext context)
             var frameworksObj = nodePlatformDetectorResult.Frameworks;
             if (frameworksObj != null && frameworksObj.Any())
             {
-                string[] framworks = frameworksObj.Select(p => p.Framework).ToArray();
-                string frameworks = string.Join(",", framworks);
+                string frameworks = string.Join(",", frameworksObj.Select(p => p.Framework).ToArray());
                 manifestFileProperties[ManifestFilePropertyKeys.Frameworks] = frameworks;
                 this.logger.LogInformation($"Detected the following framework(s): {frameworks}");
+                Console.WriteLine($"Detected the following framework(s): {frameworks}");
```
qianz2, line 329: nit: How about other platforms like python, php and golang, should we update those as well?

william-msft: These were the only specified frameworks that were specified by the static websites team

src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
william-msft: Currently these are the only places that are expected to have framework logging. But if we get them for all platforms then I agree we should have a `GlobalLogger`

daniv-msft (approved):

PR #1332, opened 2022-04-07
Update 'oryx dockerfile' command to work as intended

cormacpayne:
Resolves work item 1520774 in our story board.

This PR attempts to resolve the following two issues:
- The current `oryx dockerfile` command hasn't been updated in a while to reflect dynamic installation and can only be used with a small subset of hardcoded platforms and platform versions. In addition, the `oryx create-script` command was introduced to replace the previous `oryx` command which generate the `run.sh` script for a web application, so the generated Dockerfile needed to be updated to reflect this as well.
- The `oryx/cli` image that we produce isn't able to run a simple `oryx dockerfile` command with the base `debian` image it currently uses. The base image should be updated to something that is slim but also allows for dynamic installation to occur so that a simple `oryx dockerfile` scenario can take place with minimal amount of pull wait time.


images/build/Dockerfiles/cli.Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG DEBIAN_FLAVOR
-FROM debian:${DEBIAN_FLAVOR}-slim
+FROM buildpack-deps:stretch-curl
```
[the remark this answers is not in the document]
cormacpayne: Yes we are 😀

william-msft (dismissed):

src/BuildScriptGenerator/Dockerfile.oryx.tpl
```diff
@@ -3,9 +3,10 @@ ARG RUNTIME={{ RuntimeImageName }}:{{ RuntimeImageTag }}
 FROM mcr.microsoft.com/oryx/build:{{ BuildImageTag }} as build
 WORKDIR /app
 COPY . .
-RUN oryx build /app
+RUN oryx build --output /output /app
```
arroyc, line 6: nit: move this --output /output after oryx build /app?

images/build/Dockerfiles/cli.Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG DEBIAN_FLAVOR
-FROM debian:${DEBIAN_FLAVOR}-slim
+FROM buildpack-deps:${DEBIAN_FLAVOR}-curl
```
arroyc, line 2: what was the impact in terms of size?

cormacpayne: From roughly 200MB to 275MB -- this change was needed since the current `debian` base we're using for the CLI image wasn't supporting dynamic installation, and we were looking to create a lightweight CLI image that could support this for the `oryx dockerfile` scenario.

william-msft: @arroyc welcome back 😄

william-msft (approved):

PR #1346, opened 2022-04-19
Run PoliCheck and Rosyln Analyzers as a part of security check pipeline

cormacpayne:
Addressing part of work item 1522230; adding PoliCheck and Roslyn Analyzers to the security check that runs during every pipeline in Azure DevOps.


vsts/pipelines/templates/_securityChecks.yml
```diff
@@ -18,6 +24,11 @@ steps:
     debugMode: false
   condition: always()
 
+- task: securedevelopmentteam.vss-secure-development-tools.build-task-roslynanalyzers.RoslynAnalyzers@2
+  displayName: 'Run Roslyn Analyzers'
+  continueOnError: true
```
daniv-msft, line 29: I'm curious about this `continueOnError: true`. If anything is detected, wouldn't it will be better to let the developer know immediately by stopping the build? Aren't we going to miss some things if we continue by default?

cormacpayne: The perils of copy-paste 😀

daniv-msft: Haha, thanks for clarifying

daniv-msft (dismissed):

daniv-msft (dismissed):

vsts/pipelines/templates/_securityChecks.yml
```diff
@@ -2,26 +2,40 @@ steps:
...
+  displayName: 'Run "nuget restore" on Oryx solution'
+  inputs:
+    command: 'restore'
+    restoreSolution: '$(Build.SourcesDirectory)\Oryx.sln'
+
+- task: VSBuild@1
+  displayName: 'Build the Oryx solution'
+  inputs:
+    solution: '$(Build.SourcesDirectory)\Oryx.sln'
+    configuration: 'debug'
```
daniv-msft, line 27: Could you please clarify why we need the debug configuration? Are there any side effects of doing that?

cormacpayne: When we were running `dotnet publish` previously, we didn't provide a configuration, which then defaults to `debug` for the build. The documentation for the `VSBuild` task wasn't explicit about what the default configuration was if it wasn't configured, so I figured it would be better to provide it then not (though I don't think there would be an issue with doing `debug` vs. `release` in this case, the binaries should be the same).

daniv-msft: OK, thanks for clarifying! Sounds good, I don't see why we couldn't keep as is.

daniv-msft (approved):

PR #1348, opened 2022-04-20
Add java 11.0.14

william-msft:
- [ ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1522369


build/constants.yaml
```diff
@@ -320,7 +320,7 @@
       file-name-prefix: __
 - name: java-versions
   constants:
-    java-version: 17.0.2
+    java-version: 11.0.14
```
daniv-msft, line 323: I understand that we're adding support for this older version, but could you please clarify why we're setting it as the java-version here?

samruddhikhandale: Just to clarify, we will be needing both the Java versions (17.0.2 and 11.0.14)

william-msft: 17.0.2 is already in our storage account. Changes to this file will not remove any SDKs. Since `11.0.14` is an older version, I was considering in using this PR to trigger tests for validation and deployment, then closing the PR after the binaries have been released

daniv-msft: Thanks for clarifying. Even if we revert some of these changes, I imagine we'll still keep the versionsToBuild.txt and JavaDynamicInstallationTest.cs changes, right?

daniv-msft (dismissed):

build/constants.yaml
samruddhikhandale: In `build/__javaVersions.sh` can we then have two constants? One for 17 and another for 11? 
It will be easier to use in `vso.focal.dockerfile` which I am looking forward to modify soon.

samruddhikhandale: or I can add that variable myself when I create the PR. Anything which is easier for you would work.

platforms/java/getJavaSdk.sh
```diff
@@ -30,6 +32,24 @@ downloadJavaSdk()
...
 	fi
     
+    if [ ! -z "$JDK_URL" ]; then
+        echo "JDK_URL: ${JDK_URL}"
+        curl -L "${JDK_URL}" -o $tarFileName
+        echo "$JDK_SHA256 $tarFileName" | sha256sum --check --strict; \
+        
+        rm -rf extracted
+        mkdir -p extracted
+
```
william-msft, line 42: TODO: clean this portion up

build/__javaVersions.sh
```diff
@@ -1,4 +1,4 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
-JAVA_VERSION='17.0.2'
+JAVA_VERSION='11.0.14'
```
qianz2, line 3: Consider having variable names JAVA17_VERSION and JAVA11_VERSION, which is the latest minor/patch version for that major version.

build/constants.yaml
william-msft: I've added a `JAVA11_VERSION`. Thanks for reviewing :)

build/__javaVersions.sh
william-msft: I've added JAVA11_VERSION but will keep JAVA_VERSION for now, we can use that as our LTS version

build/__javaVersions.sh
```diff
@@ -1,4 +1,5 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
 JAVA_VERSION='17.0.2'
+JAVA_VERSION11='11.0.14'
```
daniv-msft, line 4: Could you please clarify why we need to add this field? Why do we need it for these language and not for other languages where we support multiple major versions?

william-msft: Yes, this is because this got removed from the dev storage account. Which blocked our pipelines. As a result, we had to re-build this and update the source for rebuilding.

daniv-msft: OK thanks. So I imagine we'll remove it in the future once we rebuilt it?

daniv-msft (approved):
Looks good to me!

build/__javaVersions.sh
samruddhikhandale: I believe we need to keep it as I'm using it here https://github.com/microsoft/Oryx/blob/8c8aa3034a9e46523af1154a901717983a7400e6/images/build/Dockerfiles/vso.focal.Dockerfile#L242

platforms/java/getJavaSdk.sh
```diff
@@ -17,19 +17,45 @@ downloadJavaSdk()
 {
     local JDK_VERSION="$1"
     local JDK_BUILD_NUMBER="$2"
+    local JDK_SHA256="$3"
```
cormacpayne, line 20: 🎉

platforms/java/versionsToBuild.txt
```diff
@@ -1,10 +1,11 @@
-# version, build number, sha
+# version, buildNumber, sha, jdkUrl, rootDirName
 1.8.0, b01, 1285da6278f2d38a790a21148d7e683f20de0799c44b937043830ef6b57f58c4
```
cormacpayne, line 2: Just to confirm: we're introducing SHA validation for Java when building the binaries, but do we use the SHA anywhere else? I'm curious why we'd be providing them in this file if they were (potentially) not use previously.

cormacpayne (approved):

build/__javaVersions.sh
william-msft: Yes, we'll keeping this since its an active version

platforms/java/versionsToBuild.txt
william-msft: No, there were no traces found in the Oryx repo for Java specifically.

PR #1351, opened 2022-04-21
Add support for Node 12.22.11 binaries

cormacpayne:
Adds the last unsupported version of Node outlined in work item 1514478.


platforms/nodejs/versionsToBuild.txt
```diff
@@ -8,6 +8,7 @@
 12.22.4
 12.22.6
 12.22.9
+12.22.11
```
daniv-msft, line 11: I hope that I'm not commenting on a PR in progress :) Please let me know if it is actually the case and you're already working on this.
Should we actually update the binaries and add the SHA there? Or is this a loop where you need to have a first PR with that to build and then only can update the actual binaries?

william-msft (dismissed):

qianz2 (dismissed):

william-msft (approved):

PR #1359, opened 2022-04-26
Update debian flavor for php SDKs in platform installer

qianz2:

### DIFF
--- a/src/BuildScriptGenerator/PlatformInstallerBase.cs
+++ b/src/BuildScriptGenerator/PlatformInstallerBase.cs
@@ -53,9 +53,15 @@ namespace Microsoft.Oryx.BuildScriptGenerator
                 .AppendLine($"mkdir -p {versionDirInTemp}")
                 .AppendLine($"cd {versionDirInTemp}")
                 .AppendLine("PLATFORM_BINARY_DOWNLOAD_START=$SECONDS")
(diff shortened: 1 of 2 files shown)


src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -53,9 +53,15 @@ public abstract class PlatformInstallerBase
                 .AppendLine($"mkdir -p {versionDirInTemp}")
                 .AppendLine($"cd {versionDirInTemp}")
                 .AppendLine("PLATFORM_BINARY_DOWNLOAD_START=$SECONDS")
+                .AppendLine($"if [ \"$platformName\" = \"php\" ] && [ \"$DEBIAN_FLAVOR\" != \"stretch\" ]; then")
```
daniv-msft, line 56: I thought I added this comment but looks like I missed it: could you please clarify how we're making sure we're only impacting PHP 8 customers here?
Also, who is setting this $DEBIAN_FLAVOR variable?

qianz2: $DEBIAN_FLAVOR was set up in each build image Dockerfile when it get built. It basically reads the environment variable from the image and determine which SDK it will pull. For php 5 and 7, they're stretch based and the change will apply to non stretch based php ones.

cormacpayne (approved):

PR #1367, opened 2022-04-27
Reduce golang build run image

william-msft:
- [ ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1531667


build/buildBuildImages.sh
```diff
@@ -397,6 +397,45 @@ function buildCliImage() {
 	echo "$builtImageName" >> $ACR_BUILD_IMAGES_ARTIFACTS_FILE
 }
 
+function buildAndRunImage() {
```
daniv-msft, line 400: Should this be renamed to `buildFullImage`?

build/buildBuildImages.sh
```diff
@@ -397,6 +397,45 @@ function buildCliImage() {
...
+	buildBuildScriptGeneratorImage
+	
+	local debianFlavor=$1
+	local devImageTag=BuildAndRun
+	local builtImageName="$ACR_BUILD_AND_RUN_IMAGE_REPO"
+
+	if [ -z "$debianFlavor" ] || [ "$debianFlavor" == "stretch" ]; then
+		debianFlavor="stretch"
+	elif  [ "$debianFlavor" == "buster" ]; then
```
daniv-msft, line 409: From the PR, I assume that Bullseye wasn't an easy change?

william-msft: Correct, we'll have to upgrade to golang 1.17+

images/build/Dockerfiles/buildAndRun.Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+RUN ./build.sh golang     /opt/startupcmdgen/golang
+
+### oryx build image
+FROM buildpack-deps:${DEBIAN_FLAVOR}-curl
+ARG DEBIAN_FLAVOR
+ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR
+
+# docker multi-stage builds
+COPY --from=buildscriptgenerator /opt/ /opt/
+COPY --from=startupScriptGens /opt/startupcmdgen/ /opt/startupcmdgen/
```
daniv-msft, line 26: Just to be clear, for other devs/partners reviewing this PR: this is where the size gain happens, right? Basically, we're starting over from the Oryx build image and just copy what we need from previous steps.

build/buildBuildImages.sh
william-msft: Correct, we'll have to upgrade to golang 1.17+ which will also require us to migrate away from `go dep`  to using `go.mod`

build/buildBuildImages.sh
william-msft: This script already has a `full` tag parameter. So I'll have to rename that parameter or figure out a new name for this. I shouldn't add this with the full parameter since it builds all docker images (jamstack, lts-versions, etc...)

images/build/Dockerfiles/buildAndRun.Dockerfile
william-msft: This is where the size reduction happens. And correct, we copy what we need from the previous steps :)

build/buildBuildImages.sh
daniv-msft: Thanks, so what is the plan there moving forward? Do you want to keep the name "buildAndRun"? Do you want to move to "full" later after we rename?
Also, even if we keep this, should we at least rename to `buildBuildAndRunImage`?

william-msft: I've updated the image name for this PR.

daniv-msft (dismissed):
Looks good to me! Thanks

build/buildBuildImages.sh
```diff
@@ -300,11 +300,11 @@ function buildLtsVersionsImage() {
 		.
 }
 
-function buildFullImage() {
+function buildAllImages() {
 	buildLtsVersionsImage
 
 	echo
-	echo "-------------Creating full build image-------------------"
```
cormacpayne, line 307: So this method doesn't appear to be building all of the build images -- it looks like it's just building `$BUILD_IMAGES_DOCKERFILE`, which is the `latest` build image. It would probably make sense to change the method to `buildLatestImage` and update the comment on this line

build/buildBuildImages.sh
```diff
@@ -432,6 +472,8 @@ elif [ "$imageTypeToBuild" == "ltsversions" ]; then
 	buildLtsVersionsImage
 elif [ "$imageTypeToBuild" == "ltsversions-buster" ]; then
 	buildLtsVersionsImage "buster"
+elif [ "$imageTypeToBuild" == "all" ]; then
```
cormacpayne, line 475: Same comment above -- when updating everything to `latest`, include the changes down here

build/buildBuildImages.sh
william-msft: Updated. Thanks!

build/buildBuildImages.sh
```diff
@@ -415,11 +454,12 @@ if [ -z "$imageTypeToBuild" ]; then
 	buildJamStackImage
 	buildLtsVersionsImage "buster"
 	buildLtsVersionsImage	
-	buildFullImage
+	buildlatestImages
```
daniv-msft, line 457: Nit: looks like an uppercase is missing. Not sure if that would create an error?

daniv-msft (dismissed):

build/buildBuildImages.sh
william-msft: This would. Updated. Thanks :)

daniv-msft (approved):
Looks good to me!

PR #1374, opened 2022-05-03
Add nginx fpm config setting

EricSten-MSFT:
php nginx fpm image should not include nginx port in Location: header when redirecting requests.
I've validated this config change by hand in a dev stamp.

### DIFF
--- a/images/runtime/php-fpm/nginx_conf/default.conf
+++ b/images/runtime/php-fpm/nginx_conf/default.conf
@@ -6,6 +6,7 @@ server {
     root /home/site/wwwroot;
     index  index.php index.html index.htm;
     server_name  example.com www.example.com; 
+    port_in_redirect off;
 
     location / {            
         index  index.php index.html index.htm hostingstart.html;


images/runtime/php-fpm/nginx_conf/default.conf
```diff
@@ -6,6 +6,7 @@ server {
     root /home/site/wwwroot;
     index  index.php index.html index.htm;
     server_name  example.com www.example.com; 
+    port_in_redirect off;
```
qianz2, line 9: Thanks for looking into it. Could you explain why we made this change, which issue it's trying to resolve?

EricSten-MSFT: If this is not set, the nginx port is included in the `Location:` header sent back in 301/302 redirect responses.  This winds up going all the way back to the client browser, which then attempts to go to a non-existent port.

e.g.
```
$ curl -D - http://ericstensite1.ericsten2.antares-test.windows-int.net/banana
HTTP/1.1 301 Moved Permanently
Content-Length: 185
Content-Type: text/html
Date: Tue, 03 May 2022 19:06:47 GMT
Server: nginx/1.14.2
Location: http://ericstensite1.ericsten2.antares-test.windows-int.net:8080/banana/
```
The `:8080` shouldn't be returned.

EricSten-MSFT: FYI: Service Desk case #2204080030000257

qianz2: Thanks @EricSten-MSFT for the case id. Triggering a nightly build for this PR right now to make sure nothing break for this change.

daniv-msft (approved):

PR #1380, opened 2022-05-09
Update CLI image and Dockerfile command for more robust builds/runs

cormacpayne:
This PR accomplishes the following:

- Changes the base image of the `oryx/cli` image from `buildpack-deps:{DEBIAN_FLAVOR}-curl` to `buildpack-deps:{DEBIAN-FLAVOR}`
  - Removing the `-curl` at the end of the image tag will increase the size of the image from ~300MB to ~800MB, but will include a set of tooling that allows us to be more confident in the images ability to perform basic build commands with the tooling pre-baked
- Update the `SharedCodeGenerator` to support lists provided in the `constants.yaml` file
  - Rather than allowing only `string:string` property mappings in the provided YAML file, we can now do `string:list` mappings
  - This allows us to specify the list of supported runtime platform versions and have them generated in the corresponding platform C# file to use in the code
- Update the `oryx dockerfile` command to leverage the newly added runtime platform versions and choose the maximum satisfying version over the `dynamic` image tag (will default to `dynamic` if a version cannot be chosen)
  - This allows us to automatically snap to a runtime platform version tag in scenarios where we don't want the application startup overhead that comes from the `dynamic` runtime image tag
  - Users can still provide these values manually and they will take precedence over any other value


daniv-msft:
Thanks Cormac for starting this. A couple of high-level comments:
* I know you're explaining what you did and partly why, but could you please explain the problem we're solving here? That would help understand the solution.
* Could you please clarify why we need to remove the -curl? I saw your comment about being more confident to build, but I imagine that this came from a practical project we couldn't build, right?
At first sight, it makes sense to be strict with what we add to our images, to avoid growing the size (and 500Mb is a non negligible one).
* I don't remember how we ended up the discussion about "What if customers don't want to snap to a specific version, and want to make sure they always get the latest one available?". Forcing a specific runtime version would go against that, and I can't remember what was the argument that made this a non issue.

cormacpayne:
@daniv-msft all good questions, hopefully I can clear them up:

> I know you're explaining what you did and partly why, but could you please explain the problem we're solving here? That would help understand the solution.

There are two main problems I'm attempting to solve with this PR, both relating to some feedback from the Container Apps folks using the ACR Task we advised: (1) slimming down the image we use for _strictly_ dynamic installation builds, and (2) removing the 20-30 startup time that they were seeing when running the application.

Problem (1) is resolved by moving away from building applications with the `oryx/build:azfunc-jamstack` image (~1.9 GB) and using this "skeleton" `oryx/cli:stable` image (~900 MB). The "skeleton" image will be used for both calling the `oryx dockerfile` command and building the user's application, so we only need the ACR runner to pull this smaller image once.

Problem (2) is resolved by attempting to snap to a specific version of a runtime image to avoid pulling the necessary tools during startup (which is what currently occurs with the `dynamic` runtime images).

> Could you please clarify why we need to remove the -curl? I saw your comment about being more confident to build, but I imagine that this came from a practical project we couldn't build, right?
At first sight, it makes sense to be strict with what we add to our images, to avoid growing the size (and 500Mb is a non negligible one).

This was a tricky issue -- from the `buildpack-deps` documentation on the `curl` images:

_This variant includes just the `curl`, `wget`, and `ca-certificates` packages._

This variant was useful _strictly_ for pulling down packages from our storage account during dynamic installation, but it lacked a lot of built-in tooling that some platform packages require under-the-hood (_e.g._, an npm package _could_ require that Python be installed as a prerequisite, but the `curl` variant wouldn't include it and Oryx logic has no way to detect it needs to install Python). As a result, I found it much easier to guarantee we have a larger suite of tooling in the image that we may need for niche scenarios like that; however, as you mentioned, 500 MB is a large hit to the image size to include these tools, so I think it's worth taking a second look at what other options we may have here.

> I don't remember how we ended up the discussion about "What if customers don't want to snap to a specific version, and want to make sure they always get the latest one available?". Forcing a specific runtime version would go against that, and I can't remember what was the argument that made this a non issue.

The user is always able to provide the platform runtime name and version that they want to use for `oryx dockerfile`, but these changes are more focused on guiding them to a valid runtime combination if they provide just the name or neither argument. If we're unable to find a valid runtime version for the user, we'll fall back to using the `dynamic` image tag as we were before, which will dynamically install the tooling needed from our storage account (taking that 20-30 second hit, but ensuring they get something). I'm not sure if that answers your question so just let me know 🙂

images/build/Dockerfiles/cli.Dockerfile
```diff
@@ -1,9 +1,10 @@
 ARG DEBIAN_FLAVOR
-FROM buildpack-deps:${DEBIAN_FLAVOR}-curl
+FROM buildpack-deps:${DEBIAN_FLAVOR}
 ARG DEBIAN_FLAVOR
 ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR
 
 COPY --from=buildscriptgenerator /opt/buildscriptgen/ /opt/buildscriptgen/
+COPY --from=support-files-image-for-build /tmp/oryx/ /opt/tmp
```
william-msft, line 7: I'm curious why is the `COPY` destination in `/opt/` instead of in `/tmp/`

william-msft:
I kicked off a Oryx-Nightly pipeline invocation. But if you happen to push a new commit, please re-run 😄

build/constants.yaml
```diff
@@ -217,6 +221,10 @@
     node-app-insights-sdk-version: 1.8.7
     pm2-version: 4.5.6
     npm-version: 6.14.15
+    runtime-versions: [4.4, 4.5, 4.8, 6, 6.2, 6.6, 6.9, 6.10,
+                      6.11, 8, 8.0, 8.1, 8.2, 8.8, 8.9, 8.11,
+                      8.12, 9.4, 10, 10.1, 10.10, 10.12, 10.14,
+                      12, 14, 16]
```
qianz2, line 227: It might be good to keep all runtime versions consistent by having `major.minor` or only `major` in here.

images/build/Dockerfiles/cli.Dockerfile
```diff
@@ -36,6 +37,13 @@ RUN apt-get update \
...
+RUN tmpDir="/opt/tmp" \
+    && cp -f $tmpDir/images/build/benv.sh /opt/oryx/benv \
+    && chmod +x /opt/oryx/benv
+
 ENV ORYX_SDK_STORAGE_BASE_URL="https://oryx-cdn.microsoft.io"
 ENV ENABLE_DYNAMIC_INSTALL="true"
-ENV PATH="$PATH:/opt/oryx"
\ No newline at end of file
+ENV PATH="$PATH:/opt/oryx"
+ENV DYNAMIC_INSTALL_ROOT_DIR="/opt"
```
qianz2, line 47: setting this `DYNAMIC_INSTALL_ROOT_DIR` to be `/opt` means we don't use the default path `/tmp/oryx/platforms` as other images. I'm wondering if it'll cause similar issues like we see last week?

build/constants.yaml
qianz2: When runnning `oryx dockerfile` command, does the specified platform version need to match with one of listed above?

images/build/Dockerfiles/cli.Dockerfile
cormacpayne: This is the pattern that we follow with a variety of other Dockerfiles when copying the scripts in `build` and `images` folders from the `support-files-image-for-build` intermediate image -- I guess for the CLI image we really only use it to copy over the `benv` script to `/opt/oryx`, so we can change the original copy logic, but I figured we just follow the same pattern as the other Dockerfiles here.

images/build/Dockerfiles/cli.Dockerfile
william-msft: This would be something worth test. I know we have a `TemporaryInstallationDirectoryRoot` pointing to `/tmp/oryx/platforms`

src/BuildScriptGenerator/SemanticVersionResolver.cs
```diff
@@ -6,17 +6,33 @@
...
     {
-        public static string GetMaxSatisfyingVersion(string providedVersion, IEnumerable<string> supportedVersions)
+        /// <summary>
+        /// Gets the latest version from an enumeration that is less than or equal to the provided version.
+        /// </summary>
+        /// <param name="rangeSpec">The provided range specification to match against. For example, "3.9.0", ">=3.9", "~3.0.0".</param>
+        /// <param name="supportedVersions">The enumeration of versions to search through.</param>
+        /// <param name="loose">Determines whether or not a looser RegEx is used when searching the versions.</param>
+        /// <returns>The latest version less than or equal to the provided version.</returns>
+        public static string GetMaxSatisfyingVersion(string rangeSpec, IEnumerable<string> supportedVersions, bool loose = false)
```
william-msft, line 22: May we add a test for verifying the output when the `loose` parameter is passed?

build/constants.yaml
cormacpayne: > It might be good to keep all runtime versions consistent by having major.minor or only major in here.

It does look a little odd, but what's listed for each `runtime-versions` property are the runtime image tags that we have supported for each language, and in some cases (specifically node), there is a single value for the runtime image tag (_i.e._, `node:10` is the runtime image)

> When runnning `oryx dockerfile` command, does the specified platform version need to match with one of listed above?

No -- if the user provides a runtime version, we will use that in the Dockerfile no matter what. This is to allow us to publish new runtime versions and the user not having to update the version of their CLI image to be able to use the new versions; the Dockerfile will be populated with whatever version they provide and we'll respect it (regardless of if it's in the CLI image's supported runtime version list or not).

images/build/Dockerfiles/cli.Dockerfile
cormacpayne: I was following the pattern set by `azureFunctions.JamStack.Dockerfile` -- do you happen to know why this image would want to set the dynamic install root path to `/opt` instead of the path you listed above?

william-msft (approved):

qianz2 (approved):

PR #1392, opened 2022-05-19
Use specified npm version spec from package.json during build

cormacpayne:
Resolves work item 1501504 -- uses the `engines` field within a node project's `package.json` file to update the version of npm used during a build. This allows the user to use a version different from the default version set by Oryx during the build.
 
### DIFF
--- a/src/BuildScriptGenerator/Node/NodePlatform.cs
+++ b/src/BuildScriptGenerator/Node/NodePlatform.cs
@@ -260,6 +260,8 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
                 hasDevDependencies = true;
             }
 
+            var npmVersionSpec = packageJson?.engines?.npm?.Value as string;
+
             var productionOnlyPackageInstallCommand = string.Format(
                 NodeConstants.ProductionOnlyPackageInstallCommandTemplate, packageInstallCommand);
 
@@ -394,6 +396,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
                 LageRunBuildCommand = runBuildLageCommand,
                 NodeBuildProperties = nodeCommandManifestFileProperties,
                 NodeBuildCommandsFile = nodeBuildCommandsFile,
+                NpmVersionSpec = npmVersionSpec,
             };
             string script = TemplateHelper.Render(
                 TemplateHelper.TemplateResource.NodeBuildSnippet,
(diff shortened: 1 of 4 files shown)


src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -260,6 +260,8 @@ public PlatformDetectorResult Detect(RepositoryContext context)
                 hasDevDependencies = true;
             }
 
+            var npmVersionSpec = packageJson?.engines?.npm?.Value as string;
```
william-msft, line 263: Should we add a log statement that Oryx has detected and will use this new version? Or are the logs explicit enough that the npm version in the package.json is being used?

tests/BuildScriptGenerator.Tests/Node/NodeBuildScriptGenerationTest.cs
```diff
@@ -123,6 +123,7 @@ public void GeneratedScript_HasNpmVersion_SpecifiedInPackageJson()
                     {"PlatformWithVersion", "Node.js 10.10.10" },
                 },
                 NodeBuildCommandsFile = FilePaths.BuildCommandsFileName,
+                NpmVersionSpec = "5.4.2",
```
william-msft, line 126: Would we be able to test with a valid and invalid `NpmVersionSpec` and check that `5.4.2` was selected for the build

william-msft: This might be a bit tricky when the user specifies something like this `~5.4.2` or other loose version specification

src/BuildScriptGenerator/Node/NodePlatform.cs
cormacpayne: Per our discussion offline: we have an `echo` statement in the build script that mentions we're using the npm version spec in the `package.json` file right before we run the `npm install -g npm@<spec>` command.

tests/BuildScriptGenerator.Tests/Node/NodeBuildScriptGenerationTest.cs
cormacpayne: Per our discussion offline: we grab the entire version spec that the user provided in the `package.json` file and use that in our `npm install -g npm@<spec>` command.

william-msft (approved):

PR #1393, opened 2022-05-19
Update node 12 and 16 runtimes

cormacpayne:
Updating the node 12 and 16 runtimes to use the latest binaries for each (12.22.11 and 16.14.2, respectively).


build/__nodeVersions.sh
```diff
@@ -7,7 +7,7 @@ YARN_MAJOR_VERSION='1'
 NODE6_VERSION='6.17.1'
 NODE8_VERSION='8.17.0'
 NODE10_VERSION='10.23.0'
-NODE12_VERSION='12.22.6'
+NODE12_VERSION='12.22.11'
 NODE14_VERSION='14.19.1'
 NODE16_VERSION='16.14.2'
```
cormacpayne, line 12: For some reason, this value (and the one below in `NodeVersions.cs`) was set to 16.14.2 already, but was not accurately reflected in the `constants.yaml` file.

william-msft: If this new version isn't released already we'll also have to run BaseImage pipeline for Node against this branch, get the successful pipeline invocation ID and update `node-runtime-base-tag` in constants.yaml and run the `build/generateConstants.sh` so that this change can be prorogated to other files.

cormacpayne: Ran the node base image pipeline and updated the PR 👍

william-msft (approved):

PR #1398, opened 2022-05-26
Update Dockerfile base images for CSSC compliance

cormacpayne:
Resolves work item 1534892 -- update base images (and images tagged locally during builds) to resolve warnings flagged by the CSSC for third-party base images used.


build/buildRunTimeImageBases.sh
```diff
@@ -52,13 +52,6 @@ docker build \
     -t "oryxdevmcr.azurecr.io/private/oryx/$RUNTIME_BASE_IMAGE_NAME-stretch" \
     $REPO_DIR
 
-docker build \
```
daniv-msft: Could you please clarify: Is Buster actually creating the secure supply chain issue? I thought it was the Stretch and PHP.5-6 images.
Won't we have the need to rebuild some images on that base in the future (Buster EOL is still far away)?

daniv-msft:
@cormacpayne > Could you please coordinate with @qianz2 Stretch removal PR, that might have multiple conflicts with yours?

images/build/Dockerfiles/gitHubRunners.BuildPackDepsBullseye.Dockerfile
```diff
@@ -1,2 +1,2 @@
 # https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md
-FROM buildpack-deps:bullseye-scm@sha256:c776f9f4b32b6d3fa36b400434510ccd17a218bb27944b1781e9f823ba39881b
```
daniv-msft, line 2: Nit: should we just update the SHA to use what they're targeting on MCR?

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
```diff
@@ -102,10 +103,16 @@ public string GenerateDockerfile(DockerfileContext ctx)
                 BuildImageTag = dockerfileBuildImageTag,
             };
 
-            return TemplateHelper.Render(
+            var generatedDockerfile = TemplateHelper.Render(
                 TemplateHelper.TemplateResource.Dockerfile,
                 properties,
                 this.logger);
+
+            // Remove the Container Registry Analysis snippet, if it exists in the template.
```
daniv-msft, line 111: Wouldn't this be ignored automatically as a comment? Do we want to hide it from the output dockerfile?

daniv-msft (dismissed):
Looks good to me, and I just added a couple of small comments.

william-msft (dismissed):

images/build/php/7.4/Dockerfile
```diff
@@ -1,3 +1,3 @@
 # Build PHP 7.4
-FROM php-build-prereqs AS php74-build
+FROM oryxdevmcr.azurecr.io/private/oryx/php-build-prereqs AS php74-build
```
qianz2, line 2: Curious how do we push the images to the dev MCR or prod MCR with a `private`, `mirror` or `oss` tag, did we achieve it manually?

qianz2: If we'd like to update these basic layer images, such as a different version of go base image, we'll need to push a new one to the MCR, then update it there?

qianz2: Got it, thanks for explaining.

qianz2 (dismissed):

images/build/Dockerfiles/gitHubRunners.BuildPackDepsBullseye.Dockerfile
cormacpayne: Hmm I'm actually going to revert this change and put the opt-out message here until we determine what the correct course of action is when using a SHA from the mirror registry.

images/runtime/php/5.6/base.stretch.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM php-5.6
+FROM oryxdevmcr.azurecr.io/private/oryx/php-5.6
```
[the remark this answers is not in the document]
cormacpayne: I mentioned something similar in a response to Stella's comment below, but even though I added a reference to `oryxdevmcr` here (and a lot of other places), we were never (and still aren't) pushing these additional images to our ACR. These are "intermediate" images that we build locally and tag in a specific way so that we can re-use them as base layers for multiple images we end up producing. The Secure Supply Chain Analysis treats anything without a registry as "being from Docker Hub", so we have to prefix all of the images with our ACR instance to avoid getting flagged for images we're not even pulling.

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
cormacpayne: The Secure Supply Chain Analysis tooling will flag any file with "Dockerfile" in its name that thinks we're pulling a base layer from a public registry, which includes our template file used for the `oryx dockerfile` command. In this case, I added the opt-out message to the template file, but the `oryx dockerfile` command will use the entire string content, including this message, so I have a quick snippet here to pull it out of the file when used during runtime so the user doesn't have it included in the Dockerfile we generate for them.

william-msft (approved):

PR #1403, opened 2022-06-01
Upgrade Pack to latest version 0.26.0

meip:
The very old pack version is not compatible with the latest builder buildpacks and causes errors like

```
ERROR: failed to build: validating stack mixins: buildpack paketo-buildpacks/bundle-install@0.5.0 does not support stack io.buildpacks.stacks.bionic
```

### DIFF
--- a/images/pack-builder/installPack.sh
+++ b/images/pack-builder/installPack.sh
@@ -4,7 +4,7 @@
 # Licensed under the MIT license.
 # --------------------------------------------------------------------------------------------
 
-declare -r PACK_VERSION='0.18.1'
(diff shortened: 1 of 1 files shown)


qianz2:
> FYI: We built and pushed a docker image for this PR and can confirm that it resolves the issue mentioned in the bug description.

@[meip](https://github.com/meip), thanks for sharing this solution. I've triggered the nightly build for this PR, once it completed we're able to approve it.

MatthiasWinzeler:
@qianz2 thanks for getting back to us so quickly!

we saw that the nightly build failed, but we don't have access to the failed check. Can we somehow support you there?

PR #1406, opened 2022-06-01
Migrate python 3.7, 3.8 and node 14, 16 runtime image to bullseye flavor

qianz2:


tomasreyes (dismissed):

build/buildRunTimeImageBases.sh
```diff
@@ -52,13 +52,6 @@ docker build \
     -t "$RUNTIME_BASE_IMAGE_NAME-stretch" \
     $REPO_DIR
 
-docker build \
```
daniv-msft, line 55: Do you know if we'll be able to switch all new minor versions updates to Bullseye moving forward? For example, as Antares integrate with a specific version of Buster/Bullseye, do you know if it might be OK to move from one to the other for a `patch` update, let's say?

build/buildRunTimeImageBases.sh
```diff
@@ -52,13 +52,6 @@ docker build \
...
-    --pull \
-    --build-arg DEBIAN_FLAVOR=buster \
```
cormacpayne, line 57: I'm curious why we're removing the `docker build` here for the buster runtime base image but keeping the stretch base image above -- are there still stretch images that we're building but none for buster?

build/buildRunTimeImageBases.sh
daniv-msft: Also, I see below that in some cases we kept the Buster logic. Could you please clarify in which case we kept it?

images/runtime/node/generateDockerfiles.sh
```diff
@@ -13,12 +13,28 @@ source $REPO_DIR/build/__nodeVersions.sh
...
+
+		TARGET_DOCKERFILE="$DIR/$VERSION_DIRECTORY/$1.Dockerfile"
+		cp "$DOCKERFILE_TEMPLATE" "$TARGET_DOCKERFILE"
+
+		echo "Generating Dockerfile for bullseye based images..."
+		# Replace placeholders
+		RUNTIME_BASE_IMAGE_NAME="mcr.microsoft.com/oryx/base:node-$VERSION_DIRECTORY-$NODE_RUNTIME_BASE_TAG"
+		sed -i "s|$RUNTIME_BASE_IMAGE_NAME_PLACEHOLDER|$RUNTIME_BASE_IMAGE_NAME|g" "$TARGET_DOCKERFILE"
+	done
+elif [ "$1" == "buster" ];then
```
cormacpayne, line 37: Do we need this buster section if we're no longer building any buster images for Node?

vsts/scripts/tagBaseImagesForRelease.sh
```diff
@@ -65,6 +65,7 @@ mkdir -p $artifactsDir/$imageName
 if [ "$imageName" == "node" ]
 then
   echo ""
+  retagImageWithStagingRepository node-runtimeimage-bases-bullseye.txt $imageName bullseye
   retagImageWithStagingRepository node-runtimeimage-bases-buster.txt $imageName buster
```
cormacpayne, line 69: Similar comment here: do we still need this step for buster if we're no longer creating buster images for Node?

cormacpayne (changes requested):

build/buildRunTimeImageBases.sh
qianz2: Yes, I think it would not be hard for us to achieve that. If Antares ask us to update a specific minor or patch versions for these bullseye images, we can update it easily. From now on, we'll have all minor and patch versions of node 14, 16 and python 3.7 images to be bullseye based.

build/buildRunTimeImageBases.sh
qianz2: That's good point, I keep the stretch docker build here since from node 4 to node 12, those images are still stretch based, I'm not sure if we'd like to migrate those versions to bullseye based one? [daniv-msft](https://github.com/daniv-msft) I'm not sure if Antares team is still using these old versions?

build/buildRunTimeImageBases.sh
daniv-msft: OK thanks! Just to be 100% sure, as I believe this file is shared by all languages: if Antares ask us for an update on a version that we still provide as Buster, would we be able to do that without transitioning this image to Bullseye?
If not, then we might want to check with them if we need to keep the capacity to do that, as I believe that going from Buster to Bullseye for a minor/patch version might be tricky for them.

images/runtime/python/generateDockerfiles.sh
```diff
@@ -21,9 +21,9 @@ declare -r ORYX_PYTHON_IMAGE_BASE_PLACEHOLDER="%BASE_TAG%"
 
 source "$PYTHON_VERSIONS_PATH"
 
-declare -r PYTHON_BULLSEYE_VERSION_ARRAY=($PYTHON310_VERSION)
+declare -r PYTHON_BULLSEYE_VERSION_ARRAY=($PYTHON37_VERSION $PYTHON310_VERSION)
 declare -r PYTHON_BUSTER_VERSION_ARRAY=($PYTHON39_VERSION)
-declare -r PYTHON_STRETCH_VERSION_ARRAY=($PYTHON27_VERSION $PYTHON36_VERSION $PYTHON37_VERSION $PYTHON38_VERSION)
+declare -r PYTHON_STRETCH_VERSION_ARRAY=($PYTHON27_VERSION $PYTHON36_VERSION $PYTHON38_VERSION)
```
daniv-msft, line 26: As discussed separately on Teams, could you please confirm that Python 3.8 is indeed using Stretch? If that's the case, could you please let me know how easy it would be to push it to Bullseye as well?

images/runtime/node/generateDockerfiles.sh
qianz2: Thanks, will update it.

vsts/scripts/tagBaseImagesForRelease.sh
qianz2: Thanks, will update it.

images/runtime/python/generateDockerfiles.sh
qianz2: Currently python 2.7 3.6 and 3.8 are still in stretch, it won't be much efforts for us to update it to bullseye based, do we want to proceed and migrate those as well?

daniv-msft: Thanks for confirming! For 2.7 and 3.6, we don't need to take care of them because they are already end of life. However, 3.8 will still be supported even after June 30 (Stretch EOL), so we need to move it to Bullseye. Could you please migrate this one as well?

build/buildRunTimeImageBases.sh
qianz2: Thanks for pointing out, I think from now on we'll keep these minor and patch versions consistent to be bullseye based, it makes more sense for both Antares team and us to have all images to be bullseye based, not partially buster and partially bullseye. I imagine that having different debian flavor of images for node 14 and 16 could be confusing.

images/runtime/python/generateDockerfiles.sh
qianz2: Ok, will update it.

build/buildRunTimeImageBases.sh
daniv-msft: I agree with you, and my concern is for languages that we didn't move to Bullseye yet. For example, Python 3.9 is built on Buster. Antares is using the Buster based image in their codebase. If they need us to support a new minor or patch version on Buster for Python 3.9, will we be able to do it? Or will we be have to use Bullseye for it, and then create the Buster/Bullseye discrepancy we were talking about above?

build/buildRunTimeImageBases.sh
qianz2: Currently confirm that node 4 to node 12 are stretch based and node 14, 16 are bullseye based, so we don't have any buster based images for node, and we're removing it from the pipelines as well.

build/buildRunTimeImageBases.sh
qianz2: Yes we're able to achieve that, for python 3.9 we're still able to update minor or patch versions based on the buster image. They won't be impacted by this changes.

daniv-msft: Great, thanks for confirming!

images/runtime/node/generateDockerfiles.sh
```diff
@@ -34,22 +34,6 @@ if [ "$1" == "bullseye" ];then
...
-
-		TARGET_DOCKERFILE="$DIR/$VERSION_DIRECTORY/$1.Dockerfile"
-		cp "$DOCKERFILE_TEMPLATE" "$TARGET_DOCKERFILE"
-
-		echo "Generating Dockerfile for buster based images..."
-		# Replace placeholders
-		RUNTIME_BASE_IMAGE_NAME="mcr.microsoft.com/oryx/base:node-$VERSION_DIRECTORY-$NODE_RUNTIME_BASE_TAG"
-		sed -i "s|$RUNTIME_BASE_IMAGE_NAME_PLACEHOLDER|$RUNTIME_BASE_IMAGE_NAME|g" "$TARGET_DOCKERFILE"
-	done
 elif [ "$1" == "stretch" ];then
```
daniv-msft, line 37: Minor comment: do we even need this "stretch" case? Could you please clarify in which case would we use it?
Similar comment for tagBaseImagesForRelease.sh

daniv-msft (approved):
Looks good to me!

cormacpayne (approved):

PR #1411, opened 2022-06-02
Williamhe/1546176 project spotlight golang

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1546176

### DIFF
--- a/images/build/logger.sh
+++ b/images/build/logger.sh
@@ -25,7 +25,7 @@ function LogErrorWithTryCatch()
 
     # try
     set +e
-    output=$( $cmd )
+    output=$( $cmd 2>&1)
     exitCode=${PIPESTATUS[0]}
     set -e
 
--- a/images/build/logger.sh
+++ b/images/build/logger.sh
@@ -25,7 +25,7 @@ function LogErrorWithTryCatch()
 
     # try
     set +e
-    output=$( $cmd 2>&1)
+    output=$( $cmd 2>&1)  # captures STDERR/STDOUT
     exitCode=${PIPESTATUS[0]}
     set -e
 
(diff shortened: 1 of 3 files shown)


src/BuildScriptGenerator/Golang/GolangBashBuildSnippet.sh.tpl
```diff
@@ -23,7 +23,11 @@ echo "   "
 
 # TODO: add support for nested dirs
 echo "building go app..."
-go build -o oryxBuildBinary
+doc="https://aka.ms/troubleshoot-go"
+suggestion="Please check your go.mod is valid. Try building locally first: go build"
```
cormacpayne, line 27: nit: I would slightly rephrase the last sentence here:

_Try building locally first with the following command: go build_

images/build/logger.sh
```diff
@@ -25,7 +25,7 @@ function LogErrorWithTryCatch()
 
     # try
     set +e
-    output=$( $cmd )
+    output=$( $cmd 2>&1)
```
cormacpayne, line 28: Out of curiosity, what is this change needed for?

cormacpayne (dismissed):

src/BuildScriptGenerator/Golang/GolangBashBuildSnippet.sh.tpl
william-msft: I updated, thanks!

images/build/logger.sh
william-msft: The `2>&1` will redirect the STDERR to STDOUT of the `$cmd`. Which is very important for capturing the error message of the command. Without it, only STDOUT gets captured in the `output`. Great question!

william-msft: I updated that line with a comment 😄

daniv-msft (approved):
Looks good to me

PR #1424, opened 2022-06-10
add liblzma-dev to python build

joshspicer:
Relevant issue: https://github.com/microsoft/vscode-dev-containers/issues/1477

The build of oryx that is consumed by Codespaces has a version of python where the `lzma` library was seemingly not present at compile-time, causing issues seen above.

I believe this is the right place to put a build-time dependency for the build we get in the `vso` image. Please correct me if i'm wrong @william-msft !

### DIFF
--- a/images/build/Dockerfiles/vso.Dockerfile
+++ b/images/build/Dockerfiles/vso.Dockerfile
@@ -30,7 +30,7 @@ RUN buildDir="/opt/tmp/build" \
     # Install Conda and related tools
     && apt-get update \
     && apt-get install -y --no-install-recommends \
-        apt-transport-https \
+    apt-transport-https \
     && rm -rf /var/lib/apt/lists/* \
     && curl https://repo.anaconda.com/pkgs/misc/gpgkeys/anaconda.asc | gpg --dearmor > conda.gpg \
     && install -o root -g root -m 644 conda.gpg /usr/share/keyrings/conda-archive-keyring.gpg \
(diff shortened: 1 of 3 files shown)


william-msft:
> Relevant issue: [microsoft/vscode-dev-containers#1477](https://github.com/microsoft/vscode-dev-containers/issues/1477)
> 
> The build of oryx that is consumed by Codespaces has a version of python where the `lzma` library was seemingly not present at compile-time, causing issues seen above.
> 
> I believe this is the right place to put a build-time dependency for the build we get in the `vso` image. Please correct me if i'm wrong @william-msft !

Hi @joshspicer ! Would this be for dynamic installation or having this pre-cached in Codespaces vso image? If the latter, then we'd have to update in the VSO image: https://github.com/microsoft/Oryx/blob/main/images/build/Dockerfiles/vso.focal.Dockerfile

This can also be dynamically installed by running:

- `./build/buildBuildImages.sh -t vso-focal`
- ` docker run -it -v C:\Repos\Oryx\tests\SampleApps\:/tmp/SampleApps oryx/build:vso-focal /bin/bash`

joshspicer:
Hey @william-msft! I think both would probably want to be patched at the same time?  But yes, my specific goal was to fix it in the vso image

william-msft:
> Hey @william-msft! I think both would probably want to be patched at the same time? But yes, my specific goal was to fix it in the vso image

This looks good to me. I would be great if you could double check the library with using the steps displayed above. Would you mind pushing a new commit. The ADO pipeline seems to be stuck since your last commit:
image

william-msft (dismissed):

william-msft (approved):

joshspicer:
hmm, following the testing steps above, this didn't seem to help :/

image

ericLemanissier:
I think this can be closed, now that https://github.com/microsoft/vscode-dev-containers/issues/1477 is fixed

PR #1425, opened 2022-06-10
Removed duplicate package reference from build server csproj file

pauld-msft:
- [x] **The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as \#\<issue\_number\>.**

In the preview version of VS2022, I am seeing the following error:
```
Duplicate 'PackageReference' items found. Remove the duplicate items or use the Update functionality to ensure a consistent restore behavior. The duplicate 'PackageReference' items are: JsonFlatFileDataStore 2.2.3, JsonFlatFileDataStore 2.2.3.
```
This pull request removes the duplicate package reference item from the csproj file.

- [ ] ~Tests are included and/or updated for code changes.~

Minor change, additional testing should not be needed.

- [ ] ~Proper license headers are included in each file.~

N/A, no new files.

PR #1429, opened 2022-06-14
Resolve build warnings for VS Preview and fix detector tests

pauld-msft:
- [x] *The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as \#\<issue\_number\>.*

This pull request resolves 3 separate issues that were throwing up warnings during the build of the Oryx solution on the Visual Studio Preview. 
1. Async functions being called without a `.Result` or `await`
2. Failing tests in the detector due to [default for dotnet projects](https://github.com/microsoft/Oryx/compare/main...pauld/resolve-build-warnings#diff-11fd637ee1899af09fa227e573e8ff0df83f4683ae8078c40ea72a7162b64dabR114) being `Library` and the tests assuming empty string.
3. [Removed unused parameter](https://github.com/microsoft/Oryx/compare/main...pauld/resolve-build-warnings#diff-c5d03228c1b28cf1dc5357b65481c01ad6cd3c0a1041c7b31ec26ff57710b6d7L1414) from a test method, which changed a while back to no longer need the python version as an argument.
- [x] *Tests are included and/or updated for code changes.*
- [ ] *~Proper license headers are included in each file.~*

### DIFF
--- a/tests/Oryx.Integration.Tests/Php/PhpExifTest.cs
+++ b/tests/Oryx.Integration.Tests/Php/PhpExifTest.cs
@@ -26,19 +26,19 @@ namespace Microsoft.Oryx.Integration.Tests
         // platform-version in it's own pipeline agent. This is
         // because our agents currently a space limit of 10GB.
         [Fact, Trait("category", "php-8.0")]
-        public void PipelineTestInvocationsPhp80()
+        public async void PipelineTestInvocationsPhp80()
         {   
             string phpVersion80 = "8.0";
-            ExifExample(phpVersion80);
-            PhpFpmExifExample(phpVersion80);
+            await ExifExample(phpVersion80);
+            await PhpFpmExifExample(phpVersion80);
         }
 
         [Fact, Trait("category", "php-7.4")]
-        public void PipelineTestInvocationsPhp74()
+        public async void PipelineTestInvocationsPhp74()
         {
             string phpVersion74 = "7.4";
-            ExifExample(phpVersion74);
-            PhpFpmExifExample(phpVersion74);
+            await ExifExample(phpVersion74);
+            await PhpFpmExifExample(phpVersion74);
         }
 
         [Theory]
--- a/tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
+++ b/tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
@@ -27,19 +27,19 @@ namespace Microsoft.Oryx.Integration.Tests
         // platform-version in it's own pipeline agent. This is
         // because our agents currently a space limit of 10GB.
         [Fact, Trait("category", "php-8.0")]
-        public void PipelineTestInvocationsPhp80()
+        public async void PipelineTestInvocationsPhp80()
         {
             string phpVersion80 = "8.0";
-            PhpWithWordPress51(phpVersion80);
-            CanBuildAndRun_Wordpress_SampleApp(phpVersion80);
+            await PhpWithWordPress51(phpVersion80);
+            await CanBuildAndRun_Wordpress_SampleApp(phpVersion80);
         }
 
         [Fact, Trait("category", "php-7.4")]
-        public void  PipelineTestInvocationsPhp74()
+        public async void  PipelineTestInvocationsPhp74()
         {
             string phpVersion74 = "7.4";
-            PhpWithWordPress51(phpVersion74);
-            CanBuildAndRun_Wordpress_SampleApp(phpVersion74);
+            await PhpWithWordPress51(phpVersion74);
+            await CanBuildAndRun_Wordpress_SampleApp(phpVersion74);
         }
 
         [Theory]
--- a/src/Detector/DotNetCore/DotnetCoreConstants.cs
+++ b/src/Detector/DotNetCore/DotnetCoreConstants.cs
@@ -29,5 +29,7 @@ namespace Microsoft.Oryx.Detector.DotNetCore
         public const string AzureFunctionsPackageReference = "Microsoft.NET.Sdk.Functions";
 
         public const string AzureBlazorWasmPackageReference = "Microsoft.AspNetCore.Components.WebAssembly";
+
+        public const string DefaultOutputType = "Library";
     }
 }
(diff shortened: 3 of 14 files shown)


tests/Oryx.Integration.Tests/Php/PhpExifTest.cs
```diff
@@ -26,19 +26,19 @@ public PhpExifTest(ITestOutputHelper output, TestTempDirTestFixture fixture)
...
             string phpVersion80 = "8.0";
-            ExifExample(phpVersion80);
-            PhpFpmExifExample(phpVersion80);
+            await ExifExample(phpVersion80);
+            await PhpFpmExifExample(phpVersion80);
         }
 
         [Fact, Trait("category", "php-7.4")]
-        public void PipelineTestInvocationsPhp74()
+        public async void PipelineTestInvocationsPhp74()
```
daniv-msft, line 37: I didn't look into all the new `await` changes, but is it possible that they were run without `await` on purpose, to run them in parallel? I'm not saying that this was good though, because without `await` we would swallow exceptions rather than failing the test.
Should we consider using a `Task.WaitAll()` or similar to keep the parallelism while respecting async calls?

daniv-msft:
As a convention, we're suffixing asynchronous methods with `Async` (e.g. "RunXYZHttpCallAsync"). However, we're not consistently doing this in our tests, and this is likely the reason why we didn't caught the problem before you found here.
Unless it's a huge task, should we consider renaming these methods to have the suffix so that we reduce the risk of this happening again moving forward (and enforcing better the convention)?

pauld-msft:
> As a convention, we're suffixing asynchronous methods with `Async` (e.g. "RunXYZHttpCallAsync"). However, we're not consistently doing this in our tests, and this is likely the reason why we didn't caught the problem you found here. Unless it's a huge task, should we consider renaming these methods to have the suffix so that we reduce the risk of this happening again moving forward (and enforcing better the convention)?

@daniv-msft I personally prefer suffixing the Async methods with the Async suffix, but as you noted its not consistently done throughout the tests. From a brief look there are over 100 tests thar do not follow the convention. More than happy to take this on to help enforce good coding practices, if we think that it would be valuable to do so!

daniv-msft (approved):
Looks good to me!
FYI your pipelines are failing due to the ongoing Scream Test, so we should be good tomorrow to push this if there are no other issues.

william-msft:
I triggered Oryx-Nightly for this branch since Nightly runs the integration tests: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6297043&view=results
If it passes, it looks good to me!

pauld-msft:
> I triggered Oryx-Nightly for this branch since Nightly runs the integration tests: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6297043&view=results If it passes, it looks good to me!

Looks like they failed with some timeouts, so I re-kicked off failed jobs. Hopefully they are transient, and this next run will pass

qianz2 (dismissed):
Looks good to me, please make sure it passed the nightly build before we merging it. Thanks!

pauld-msft:
I think that there may be some integration tests that have had their exceptions swallowed. Looking at the most recent nightly test failures, I've tried running those locally using `await` and am getting the same errors. Will take a look into it further and try to resolve the test failures before merging these changes.

tests/Oryx.Integration.Tests/Php/PhpWordPressTest.cs
```diff
@@ -30,9 +30,7 @@ public PhpWordPressTest(ITestOutputHelper output, TestTempDirTestFixture fixture
         public async Task PipelineTestInvocationsPhp80Async()
         {
             string phpVersion80 = "8.0";
-            await Task.WhenAll(
```
pauld-msft: Wordpress 5.1 and php 8 do not seem compatible https://make.wordpress.org/core/handbook/references/php-compatibility-and-wordpress-versions/

tests/SampleApps/php/wordpress-example/create_wordpress_db.sh
```diff
@@ -3,7 +3,15 @@
 set -ex
 
 apt-get update && apt-get install mariadb-server --yes
-service mysql start
+
+# As of mariadb 10.5, which bullseye is using, the service name is now mariadb
+# instead of mysql, but the mysql command is still symlinked.
+# See: https://mariadb.com/kb/en/changes-improvements-in-mariadb-105/#binaries-named-mariadb-mysql-symlinked
```
pauld-msft, line 9: Just want to call this out explicitly, in case we rely on the mysql service anywhere else but in these tests

cormacpayne (approved):

cormacpayne:
@pauld-msft -- do you know if we're able to add these checks to our [existing ruleset](https://github.com/microsoft/Oryx/blob/main/src/CommonFiles/Oryx.ruleset) so you aren't broken locally if someone were to push a change with the bad format?

pauld-msft:
> @pauld-msft Paul Dorsch FTE -- do you know if we're able to add these checks to our [existing ruleset](https://github.com/microsoft/Oryx/blob/main/src/CommonFiles/Oryx.ruleset) so you aren't broken locally if someone were to push a change with the bad format?

@cormacpayne So the main blocking error was NU1505 which was introduced in Nuget v6.3 (which i have since I'm on the preview of VS). That rule was the one that was actually preventing me from building with nuget v6.3, but I don't think that we want to ignore that rule since it can cause inconsistent restore behavior. Plus, it was a descriptive warning with a [quick fix](https://github.com/microsoft/Oryx/pull/1425/files).

This pull request's changes were not blocking me, rather, I just wanted to resolve the build warnings that stemmed from not awaiting async methods. This is another rule that I believe is valuable to keep around as a warning, since it provided information that I was able to use to fix some tests, and I could have ignored if I wanted to.

daniv-msft (approved):
Looks good to me!

pauld-msft:
> Looks good to me!

@daniv-msft thanks for taking a look! Have to push this last commit to resolve a nightly test failure, but hopefully this is the last one and this can finally be merged after the tests finish

PR #1430, opened 2022-06-14
run script changes

waliMSFT:
This change will consider the customized user run command.


src/BuildScriptGenerator/DefaultRunScriptGenerator.cs
```diff
@@ -123,6 +123,12 @@ public string GenerateBashScript(RunScriptGeneratorContext ctx)
 #pragma warning restore CA2201
             }
 
+            BuildConfigurationFIle buildConfigFile = BuildConfigurationFIle.Create(ctx.SourceRepo.ReadFile("appsvc.yaml"));
```
daniv-msft, line 126: Would there be a way to inject this through `create-script`? Here it feels strange to just append at the end of the file a new command without integrating this with the rest of the script generation. Would there be a way to reconcile these?

daniv-msft:
Just to help me understand: is there an other way for the user to specify a customized run command?

src/BuildScriptGenerator/DefaultRunScriptGenerator.cs
waliMSFT: When talked about this with Tulika we concluded to append the command at the end of run.sh. But your suggestion makes sense. I will check if we can inject the user command through `create-script`.

william-msft: Injecting via `create-script`: If the goal is to take `appsvc.yaml` code and place it into `run.sh` then it would beneficial for Oryx to add the code into `run.sh`. Injecting the code through `create-script` would make sense if we were not reading from `appsvc.yaml`

Appending `appsvc.yaml` to the end of `run.sh`: Adding to `DefaultRunScriptGenerator.cs` adds this to all platforms, not just DotNet. Is this what we want to do? If so, keep in mind that python can run in virtual environment an non-virtual environment which means appending this to the end may not work. Other platforms might have similar "gotchas" so we need to look inside the run.sh and place the `appsvc.yaml` content in the correct order. If you provide a sample `run.sh` we can look at it together. I would suggest revisiting appending this for all platforms. 

Side note: Here is the source code for generating the run.sh for dotnet specifically https://github.com/microsoft/Oryx/tree/main/src/startupscriptgenerator/src/dotnetcore

src/BuildScriptGenerator/DefaultRunScriptGenerator.cs
```diff
@@ -123,6 +123,12 @@ public string GenerateBashScript(RunScriptGeneratorContext ctx)
...
+            if (!string.IsNullOrEmpty(buildConfigFile.Run))
+            {
+                File.AppendAllText(this.tempScriptPath, Environment.NewLine + buildConfigFile.Run);
+            }
+
```
william-msft, line 131: @waliMSFT friendly reminder to add integration tests for this in: https://github.com/microsoft/Oryx/tree/main/tests/Oryx.Integration.Tests/DotNetCore

src/BuildScriptGenerator/DefaultRunScriptGenerator.cs
william-msft: Thanks Wali, but I was asking for the dotnet `run.sh` of the file you plan on appending this to. We'll have to understand what we are appending to. This will help us determine whether we can append to the end or not. 

The other point that should be addressed is whether we want to append to everything or not.  Do we know this? IMO we should only add to DotNet if we don't need this for the other platforms.

waliMSFT: Answering this question-" we want to append to everything or not?": Based on the discussion I had with Tulika we should append everything whatever user specifies as the run command on the `appsvc.yaml` file.

waliMSFT: I will generate the dotnet run.sh file and show it to you.

src/startupscriptgenerator/src/common/fsvalidation.go
```diff
@@ -60,6 +60,31 @@ func WriteScript(filePath string, command string) {
 	ioutil.WriteFile(filePath, []byte(command), 0755)
 }
 
+// Appends command to a file
+func AppendScript(filePath string, command string) {
+	if command == "" {
+		return
+	}
+	fmt.Println("Appending output script to '" + filePath + "'")
```
cormacpayne, line 68: Let's move this print statement to be after we check that both `filePath` exists on disc and that we're able to open it.

I would also be in favor of re-wording this statement to be something along the lines of:

```go
fmt.Println("Appending provided command to '" + filePath +"'")
```

src/startupscriptgenerator/src/common/parseUserCommands.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+)
+
+func ParseUserRunCommand(sourcePath string) string {
+
+	appsvcFile, err := os.Open(sourcePath)
```
cormacpayne, line 17: Do we need to check for the existence of `sourcePath` first, or does `os.Open` do that for us?

src/startupscriptgenerator/src/common/fsvalidation.go
```diff
@@ -60,6 +60,31 @@ func WriteScript(filePath string, command string) {
...
+
+	// Ensures directory
+	dir := filepath.Dir(filePath)
```
cormacpayne, line 71: Is this the same as checking for the existence of the file?

src/startupscriptgenerator/src/common/parseUserCommands.go
```diff
@@ -0,0 +1,56 @@
...
+		//fmt.Println(scanner.Text())
+		indexOfRunHeading := strings.Index(scanner.Text(), runHeading)
+
+		if indexOfRunHeading < 0 && strings.Contains(scanner.Text(), ":") && isRunCommandFound {
+			// runCommand already found
+			// not considering any other customized commands
+			break
+		}
+
+		if isRunCommandFound || (indexOfRunHeading >= 0 && len(scanner.Text()) > len(runHeading)) {
```
cormacpayne, line 38: Would you mind clarifying what's going on in this block? Are we checking for a multiline run command that the user provided in the `appsvc.yaml` file?

tests/Oryx.Integration.Tests/Oryx.Integration.Tests.csproj
```diff
@@ -12,6 +12,11 @@
 
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
     <PlatformTarget>x64</PlatformTarget>
+    <LangVersion>8.0</LangVersion>
```
cormacpayne, line 15: Do we need the changes in this file? I think we should be good to revert them.

tests/Oryx.Tests.Common/DockerVolume.cs
```diff
@@ -53,24 +53,21 @@ private DockerVolume(string originalHostDir, string mountedHostDir, string conta
...
         /// <returns>DockerVolume instance that can be used to mount the new copy of `originalDir`.</returns>
-        public static DockerVolume CreateMirror(string hostDir)
+        /// <param name="writeToHostDir">a boolean which indicates if we want the actual directory or the copy of the actual directory</param>
+        public static DockerVolume CreateMirror(string hostDir, bool writeToHostDir = false)
         {
             if (string.IsNullOrEmpty(hostDir))
             {
                 throw new ArgumentException($"'{nameof(hostDir)}' cannot be null or empty.");
             }
-
```
cormacpayne, line 62: A lot of empty lines were removed in this file -- would you mind adding them back to make the file easier to read?

cormacpayne (changes requested):

src/startupscriptgenerator/src/common/parseUserCommands.go
```diff
@@ -0,0 +1,56 @@
...
+	const runHeading string = "run:" // format of run command- "run: gunicorn myapp.app --workers 5"
+	runCommand := ""
+	var isRunCommandFound bool
+	scanner := bufio.NewScanner(appsvcFile)
+
+	for scanner.Scan() {
+		//fmt.Println(scanner.Text())
+		indexOfRunHeading := strings.Index(scanner.Text(), runHeading)
+
+		if indexOfRunHeading < 0 && strings.Contains(scanner.Text(), ":") && isRunCommandFound {
```
william-msft, line 32: Could we move this logic into a boolean variable?

src/startupscriptgenerator/src/common/parseUserCommands.go
william-msft: Could we increase readability by moving logic into a variable? If you also move the https://github.com/microsoft/Oryx/pull/1430/files#r932680872 into a var you can do something like:
```
foo = ...
bar = ...
if foo & bar {
    // do stuff
}
```

src/startupscriptgenerator/src/common/parseUserCommands.go
```diff
@@ -0,0 +1,56 @@
...
+				isRunCommandFound = true
+				runCommand += strings.TrimSpace(scanner.Text()[indexOfRunHeading+len(runHeading):]) //gets the run command and trim to get rid of the forward and trailing spaces
+			}
+		}
+	}
+
+	fmt.Println("User provided run command: " + runCommand)
+
+	if err := scanner.Err(); err != nil {
+		return ""
```
william-msft, line 52: If an error occurs will the user see any error messages? Here's some recommendations for error handling: https://hussachai.medium.com/error-handling-in-go-a-quick-opinionated-guide-9199dd7c7f76

tests/SampleApps/DotNetCore/NetCore6PreviewMvcApp/appsvc.yaml
```diff
@@ -0,0 +1,3 @@
+version: 1
+
+run: gunicorn myapp.app --workers 5
```
william-msft, line 3: Is there a `gunicorn` command for dotnet?

william-msft:
Gentle reminder there are merge conflicts that need to be resolved.

src/startupscriptgenerator/src/common/fsvalidation.go
waliMSFT: done

src/startupscriptgenerator/src/common/parseUserCommands.go
waliMSFT: if the path does not exist `os.Open` returns a `pathError`

src/startupscriptgenerator/src/common/parseUserCommands.go
waliMSFT: @cormacpayne: it checks multiple scenarios:

1) this validifies only one run command-block with a heading `run:`
2) it checks if the `appsvc.yaml` contains the heading-string `run:`. then, it considers the substring as the run-command followed by the `run:` heading.3) according to the example of a standard appsvc.yaml file(https://github.com/JasonFreeberg/appsvc.yaml-examples) I considered any command-block should have a heading with a colon `:`. E.g. `run:`, `build:`, etc.
here, I did not consider any further command-block when the run command-block is already found.
4)  a multi lined run-command block is considered as a valid run command-block.

tests/SampleApps/DotNetCore/NetCore6PreviewMvcApp/appsvc.yaml
waliMSFT: we actually do not need this file anymore. I will revert it.

src/startupscriptgenerator/src/common/parseUserCommands.go
waliMSFT: no, the user could not see anything if an error occurs. Now, I added `print` statement to show the error.

src/startupscriptgenerator/src/common/fsvalidation.go
```diff
@@ -60,6 +60,29 @@ func WriteScript(filePath string, command string) {
...
+		return
+	}
+	defer file.Close()
+
+	fmt.Println("Appending provided command to '" + filePath + "'")
+	// Appends the command at the end of the file
+	if _, err := file.WriteString("\n" + command); err != nil {
+		return
+	} else {
+		fmt.Println(err)
```
pauld-msft, line 82: won't `err` be nil here? Shouldn't we do something more like we do above?
```
if _, err := file.WriteString("\n" + command); err != nil {
    fmt.Println(err)
    return
}
```

src/startupscriptgenerator/src/common/parseUserCommands.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	defer appsvcFile.Close()
+
+	const runHeading string = "run:" // format of run command- "run: gunicorn myapp.app --workers 5"
+	runCommand := ""
+	var isRunCommandFound bool
+	scanner := bufio.NewScanner(appsvcFile)
+
+	for scanner.Scan() {
+		//fmt.Println(scanner.Text())
```
pauld-msft, line 30: nit: commented out code, do we want this print statement?

src/startupscriptgenerator/src/common/parseUserCommands.go
```diff
@@ -0,0 +1,61 @@
...
+		indexOfRunHeading := strings.Index(scanner.Text(), runHeading)
+		isRunHeadingExists := indexOfRunHeading > 0
```
pauld-msft, line 32: won't this indicate that the run heading doesn't exist if it is in the very first position e.g. `run: a temp command`? It may be better to check if `indexOfRunHeading` is equal to `-1`, indicating that it was not found anywhere in the string?

src/startupscriptgenerator/src/common/parseUserCommands.go
```diff
@@ -0,0 +1,61 @@
...
+		isCurrentLineContainsAnyHeading := strings.Contains(scanner.Text(), ":")
+
+		if !isRunHeadingExists && isCurrentLineContainsAnyHeading && isRunCommandFound {
+			// runCommand already found
+			// not considering any other customized commands
+			break
+		}
+
+		isValidRunCommand := indexOfRunHeading >= 0 && len(scanner.Text()) > len(runHeading)
```
pauld-msft, line 41: Nit: Use `isRunHeadingExists` here

src/startupscriptgenerator/src/common/fsvalidation.go
```diff
@@ -60,6 +60,29 @@ func WriteScript(filePath string, command string) {
...
+// Appends command to a file
+func AppendScript(filePath string, command string) {
+
+	if command == "" {
+		return
+	}
+
+	file, err := os.OpenFile(filePath, os.O_APPEND|os.O_WRONLY, 0755)
+	if err != nil {
+		fmt.Println(err)
```
cormacpayne, line 72: We should provide a little more detail here for the user, something like the following:

```go
fmt.Println("Unable to read provided file to append command to. Error: " + error)
```

src/startupscriptgenerator/src/common/parseUserCommands.go
```diff
@@ -0,0 +1,61 @@
...
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+)
+
+func ParseUserRunCommand(sourcePath string) string {
+
+	appsvcFile, err := os.Open(sourcePath)
+	if err != nil {
```
cormacpayne, line 18: As mentioned in a previous discussion, if the user doesn't have an `appsvc.yaml` file, then we shouldn't display any errors as they shouldn't care about a file they didn't provide not being found.

src/startupscriptgenerator/src/dotnetcore/main.go
```diff
@@ -130,6 +130,9 @@ func main() {
 		}
 
 		common.WriteScript(fullOutputPath, command)
+
+		userRunCommand := common.ParseUserRunCommand(fullAppPath + "/appsvc.yaml")
```
cormacpayne, line 134: For all of the places where you're doing this (creating a full file path), I would recommend using [`filepath.Join()`](https://www.geeksforgeeks.org/filepath-join-function-in-golang-with-examples/) as these types of libraries typically shield you from the OS being used and whether you should be using `/` or `\` in the path.

cormacpayne, line 134: We should probably have the `appsvc.yaml` file string as a constant in one of the commons file since it's used across multiple `.go` files and allows us to easily update it across the board if the name of the file ever changes.

tests/Oryx.Integration.Tests/Php/PhpGreetingsAppTest.cs
```diff
@@ -3,6 +3,7 @@
 // Licensed under the MIT license.
 // --------------------------------------------------------------------------------------------
 
+using System;
```
cormacpayne, line 6: Let's remove this since we made no additional changes to this file.

tests/Oryx.Integration.Tests/Python/PythonEndToEndTests.cs
```diff
@@ -10,6 +10,8 @@
 using Microsoft.Oryx.Tests.Common;
 using Xunit;
 using Xunit.Abstractions;
+using System.IO;
```
cormacpayne, line 13: Same comment here -- let's remove these namespaces.

cormacpayne (changes requested):
Some additional comments to take a look at

cormacpayne:
@waliMSFT -- looks like there are some issues with `common.AppSvcFileName` not being found during builds/tests. Would you mind taking a look?

cormacpayne (approved):

PR #1431, opened 2022-06-14
Draft: Update to move ltsversions & jamstack build image into its own job

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1504900

### DIFF
--- a/vsts/pipelines/validation.yml
+++ b/vsts/pipelines/validation.yml
@@ -55,6 +55,26 @@ jobs:
     parameters:
         imageType: jamstack
 
+- job: Job_BuildImage_latest
+  displayName: Build and Test Build Jamstack Image
+  pool:
+    name: AzurePipelines-EO
+    demands:
+      - ImageOverride -equals AzurePipelinesUbuntu20.04compliant
+  timeoutInMinutes: 480
+  steps:
+  - script: |
+      echo "##vso[task.setvariable variable=BuildBuildImages;]true"
+      echo "##vso[task.setvariable variable=TestBuildImages;]true"
+      echo "##vso[task.setvariable variable=PushBuildImages;]false"
+      echo "##vso[task.setvariable variable=PushRuntimeImages;]false"
+      echo "##vso[task.setvariable variable=EmbedBuildContextInImages;]true"
+      echo "##vso[task.setvariable variable=RELEASE_TAG_NAME;]$(Build.BuildNumber)"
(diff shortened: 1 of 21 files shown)


vsts/pipelines/validation.yml
```diff
@@ -32,6 +32,48 @@ jobs:
...
+      echo "##vso[task.setvariable variable=PushRuntimeImages;]false"
+      echo "##vso[task.setvariable variable=EmbedBuildContextInImages;]true"
+      echo "##vso[task.setvariable variable=RELEASE_TAG_NAME;]$(Build.BuildNumber)"
+    displayName: 'Set variables'
+  - template: templates/_buildTemplate.yml
+    parameters:
+        imageType: jamstack
+
+- job: Job_BuildImage_latest
+  displayName: Build and Test Build Latest Image
```
qianz2, line 59: Trying to understand better how we plan to parallel and divide the `build and test`  part, it looks that we have three parallel jobs for ltsversions, jamstack and latest images respectively.

william-msft: Yes, that's correct. And we'll build and test the same image in the same corresponding job. Still in draft. But do you foresee an issue? 

One pending improvement is to reduce the redundant code we have for all the jobs, by using a loop. Similar to: https://github.com/microsoft/Oryx/blob/main/vsts/pipelines/templates/_integrationJobTemplate.yml#L96

For now I'm making sure this will work.

qianz2: Thanks William, I'm thinking what about some other build images such as github-actions, vso images for the jobs, should we add them as parallel jobs or they will get built in the existing jobs? 

For the testing parts, one thing we can improve is to remove unsupported EOL image versions from the pipeline, I'll also start working on cleaning up EOL stretch references from the codebase, hopefully it will help too.

william-msft: The plan is to have github-action, vso, etc in their own job. Right now tests are failing and we may need to keep them with the latest (largest) image but I'm trying to isolate as much as I can to get the most improvements. 

Agreed on the EOL will need to drop support. We have a story for that to look at that next: https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1509318

But if you see anything concerning that comes up do let me know 😸

PR #1433, opened 2022-06-15
Pauld/1554526 add support for ostype

pauld-msft:
https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1554526

### DIFF
--- a/tests/Oryx.Integration.Tests/Database/PulledBuildOsTypeTest.cs
+++ b/tests/Oryx.Integration.Tests/Database/PulledBuildOsTypeTest.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Oryx.Integration.Tests
         [InlineData(Settings.JamStackBuildImageName, "stretch")]
         [InlineData(Settings.JamStackBullseyeBuildImageName, "bullseye")]
         [InlineData(Settings.JamStackBusterBuildImageName, "buster")]
-        [InlineData(Settings.VsoUbuntuBuildImageName, "focal")]
+        [InlineData(Settings.VsoUbuntuBuildImageName, "focal-scm")]
         public void PulledBuildImages_Contains_BUILDOS_TYPE_Info(string buildImageName, string expectedBuildOsType)
         {
             // Arrange and Act
@@ -66,7 +66,7 @@ namespace Microsoft.Oryx.Integration.Tests
                 () =>
                 {
                     Assert.True(result.IsSuccess);
-                    Assert.Contains(expectedBuildOsType, actualOutput);
+                    Assert.Equal(expectedBuildOsType, actualOutput);
                 },
                 result.GetDebugInfo());
         }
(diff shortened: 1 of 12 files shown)


tests/Oryx.Integration.Tests/Database/PulledBuildOsTypeTest.cs
```diff
@@ -0,0 +1,74 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Oryx.Integration.Tests
+{
+    [Trait("category", "db")]
+    public class PulledBuildOsTypeTest
```
pauld-msft, line 16: Based this test heavily on https://github.com/microsoft/Oryx/blob/main/tests/Oryx.Integration.Tests/Database/PulledBuildImageTypeTest.cs

tests/Oryx.RuntimeImage.Tests/Ruby/RubyImagesTest.cs
```diff
@@ -18,7 +18,7 @@ public RubyImagesTest(ITestOutputHelper output) : base(output)
         {
         }
 
-        [Fact]
```
william-msft, line 21: I'm curious to why we've changed to `SkippableFact`?

william-msft:
I've triggered nightly pipeline: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6297085&view=results
We'll need to re-trigger if a new commit is pushed. For now it's manual, but we've discussed on having it be auto-triggered

tests/Oryx.RuntimeImage.Tests/Ruby/RubyImagesTest.cs
pauld-msft: ah, yes, so I am somewhat unfamiliar with xunit and so this may not have been the correct way to approach the problem but changed this because `Skip.If` was throwing an exception locally that flat out failed the test. This change allowed the test to instead skip instead of failing when executed locally (I believe that it should still execute unchanged on the ci build agent)
image

pauld-msft:
> I've triggered nightly pipeline: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6297085&view=results We'll need to re-trigger if a new commit is pushed. For now it's manual, but we've discussed on having it be auto-triggered

Thanks for that! In what cases do we need to trigger the nightly tests before merging?

william-msft:
> > I've triggered nightly pipeline: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6297085&view=results We'll need to re-trigger if a new commit is pushed. For now it's manual, but we've discussed on having it be auto-triggered
> 
> @william-msft Thanks for that! In what cases do we need to trigger the nightly tests before merging?

If we are touching any files in the integration tests we'll need to run nightly. Which usually also includes adding new versions, since the `./build/generateConstants` automatically bumps up version variables that are referenced in integration tests.

tests/Oryx.RuntimeImage.Tests/Ruby/RubyImagesTest.cs
william-msft: Got it!

william-msft (dismissed):

qianz2 (dismissed):

daniv-msft (approved):

PR #1449, opened 2022-06-27
Project spotlight node

waliMSFT:
added error and warning logs


tests/Oryx.BuildImage.Tests/Node/NodeJSSampleAppsTestBase.cs
```diff
@@ -24,5 +28,43 @@ public class NodeJSSampleAppsTestBase : SampleAppsTestBase
...
+            var appOutputDir = "/tmp/" + SampleAppName + "-output";
+            var script = new ShellScriptBuilder()
+                .AddDefaultTestEnvironmentVariables()
+                .AddCommand($"echo RandomText >> {appDir}/Program.cs") // triggers a failure
+                .AddBuildCommand(
+                $"{appDir} -o {appOutputDir}")
+                .ToString();
+            // Regex will match:
+            // "yyyy-mm-dd hh:mm:ss"|ERROR|Micro
+            Regex regex = new Regex(@"""[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1]) (0[0-9]|1[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])""\|ERROR\|.*");
```
william-msft, line 47: Could we fine tune this regex to pattern match a more granular message. Similar to: https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/golang/GolangDynamicInstallationTest.cs#L91

tests/Oryx.BuildImage.Tests/Node/NodeJSSampleAppsTestBase.cs
```diff
@@ -24,5 +28,43 @@ public class NodeJSSampleAppsTestBase : SampleAppsTestBase
...
+            var appDir = volume.ContainerDir;
+            var appOutputDir = "/tmp/" + SampleAppName + "-output";
+            var script = new ShellScriptBuilder()
+                .AddDefaultTestEnvironmentVariables()
+                .AddCommand($"echo RandomText >> {appDir}/Program.cs") // triggers a failure
+                .AddBuildCommand(
+                $"{appDir} -o {appOutputDir}")
+                .ToString();
+            // Regex will match:
+            // "yyyy-mm-dd hh:mm:ss"|ERROR|Micro
```
william-msft, line 46: Will the regex actually match for `"yyyy-mm-dd hh:mm:ss"|ERROR|Micro` per comment above?

william-msft:
Side note: you'll need to merge with https://github.com/microsoft/Oryx/pull/1398 after that PR gets merged to resolve OSS compliant failures

tests/Oryx.BuildImage.Tests/Node/NodeJSSampleAppsTestBase.cs
waliMSFT: `Micro` is missing. So, it should not match.

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
```diff
@@ -184,6 +186,7 @@ cd "$SOURCE_DIR"
 	echo "Switching to package directory provided: '{{ PackageDirectory }}'..."
 	if [ ! -d "$SOURCE_DIR/$packageDirName" ]; then
 		echo "Package directory '$SOURCE_DIR/$packageDirName' does not exist." 1>&2
```
cormacpayne, line 188: Do we need this `echo` call anymore? The `LogWarning` call is providing the same text to the user with additional information, so I think we should be safe to remove this to avoid redundancy.

cormacpayne (dismissed):

tests/Oryx.BuildImage.Tests/Node/NodeJSSampleAppsTestBase.cs
```diff
@@ -24,5 +28,43 @@ public class NodeJSSampleAppsTestBase : SampleAppsTestBase
             }))
         {
         }
+
+        [Fact]
```
william-msft, line 32: @Wali I noticed this test doesn't run in the pipeline since there's no assigned category. Could you add a category to this function and move it outside the TestBase with other tests. Example: https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/Node/NodeJsSampleAppsOtherTests.cs#L124

william-msft: Please also merge with `main`  before making the change since BuildImage tests have noticeable changes

src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
waliMSFT: yah I was thinking the same. But kept it as it was already there.

tests/Oryx.BuildImage.Tests/Node/NodeJSSampleAppsTestBase.cs
waliMSFT: Also, it should be a warning.

tests/Oryx.BuildImage.Tests/Node/NodeJSSampleAppsTestBase.cs
waliMSFT: I made the warning message as specific as possible.

william-msft (approved):

PR #1451, opened 2022-06-29
Pauld/add ostype to build output dir

pauld-msft:
Instead of the customers using the /opt/oryx/ directory for the .ostype images, they asked us to put that .ostype file in the same directory as the manifest file, as LWAS only has access to that and the output directories.


src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
```diff
@@ -216,6 +216,15 @@ echo "Manifest file created."
 {{ end }}
 {{ end }}
 
+OS_TYPE_SOURCE_DIR="/opt/oryx/.ostype"
+if [ -f "$OS_TYPE_SOURCE_DIR" ]
```
daniv-msft, line 220: Just for my understanding: is there any legitimate case where we wouldn't have the .ostype file at this point? This change will only apply to new images we generate, that should have the .ostype file as well, right?

pauld-msft: Yes I believe that we should have the OS type file. I suppose that we could alternatively have this script fail when it tries to copy over the ostype file if it doesn't exist, its up to you all. I'm not sure how we generally want to handle those sorts of cases

daniv-msft: Overall, the approach taken is almost always "Fail fast": as soon as something unexpected happens, let the program fail/throw an exception/etc. rather than handling the case proactively and hiding the problem when it does happen.
That said, here we want to unblock our partners quickly so I'm all good with keeping this as is to not restart the pipelines.

daniv-msft: Should you remove the Draft so that we can merge it?

pauld-msft: > Overall, the approach taken is almost always "Fail fast": as soon as something unexpected happens, let the program fail/throw an exception/etc. rather than handling the case proactively and hiding the problem when it does happen. That said, here we want to unblock our partners quickly so I'm all good with keeping this as is to not restart the pipelines.

Sounds good, I can add a followup PR to address this and have it "fail fast"

daniv-msft (approved):

PR #1453, opened 2022-06-30
Upgrade NGINX to latest stable version in PHP Fpm images

qianz2:


images/runtime/php-fpm/template.base.Dockerfile
```diff
@@ -10,7 +10,15 @@ ENV NGINX_RUN_USER www-data
...
-    && apt-get install nano nginx -y
+    && apt-get install nano curl gnupg2 ca-certificates lsb-release debian-archive-keyring -y \
+    && curl https://nginx.org/keys/nginx_signing.key | gpg --dearmor \
+        | tee /usr/share/keyrings/nginx-archive-keyring.gpg >/dev/null \
+    && gpg --dry-run --quiet --import --import-options import-show /usr/share/keyrings/nginx-archive-keyring.gpg \
+    && echo "deb [signed-by=/usr/share/keyrings/nginx-archive-keyring.gpg] \
+        http://nginx.org/packages/debian `lsb_release -cs` nginx" \
+        | tee /etc/apt/sources.list.d/nginx.list \
+    && apt update \
+    && apt install nginx
```
daniv-msft, line 21: Could you please clarify which NGINX version we'll get by default from this?

images/runtime/php-fpm/template.base.Dockerfile
```diff
@@ -10,7 +10,15 @@ ENV NGINX_RUN_USER www-data
 ENV NGINX_DOCUMENT_ROOT /home/site/wwwroot
 # Install NGINX 
 RUN apt-get update \
-    && apt-get install nano nginx -y
+    && apt-get install nano curl gnupg2 ca-certificates lsb-release debian-archive-keyring -y \
+    && curl https://nginx.org/keys/nginx_signing.key | gpg --dearmor \
+        | tee /usr/share/keyrings/nginx-archive-keyring.gpg >/dev/null \
+    && gpg --dry-run --quiet --import --import-options import-show /usr/share/keyrings/nginx-archive-keyring.gpg \
```
daniv-msft, line 16: I might be missing something, but if this is a dry run, then could you please clarify what is this command doing?

images/runtime/php-fpm/template.base.Dockerfile
```diff
@@ -10,7 +10,15 @@ ENV NGINX_RUN_USER www-data
 ENV NGINX_DOCUMENT_ROOT /home/site/wwwroot
```
daniv-msft, line 10: If NGINX only installed in PHP-FPM images? For my knowledge, could you please clarify the difference between PHP-FPM and the standard PHP?

images/runtime/php-fpm/template.base.Dockerfile
qianz2: The default version it installs will be 1.22.

images/runtime/php-fpm/template.base.Dockerfile
qianz2: It basically verifies the downloaded files contains the proper key. The step will simply output a key for us to verify and do nothing else.

daniv-msft: OK, thanks for clarifying

images/runtime/php-fpm/template.base.Dockerfile
qianz2: Yes, NGINX only exists in PHP FPM runtime images. The main difference is that standard PHP images use Apache server while FPM images use NGINX as a load balancing server.

daniv-msft: Thanks for clarifying

daniv-msft (approved):

daniv-msft (dismissed):

cormacpayne (dismissed):
:shipit:

build/buildRunTimeImageBases.sh
```diff
@@ -45,6 +45,13 @@ echo
 # Build the common base image first, so other images that depend on it get the latest version.
 # We don't retrieve this image from a repository but rather build locally to make sure we get
 # the latest version of its own base image.
+docker build \
```
daniv-msft, line 48: This change seems to impact more than just this PR. Were we not building the right runtime images anymore?
Also, should we also update in this file the `if [ "$runtimeSubDir" == "node" ]; then` block with the buster docker build, in this case?

daniv-msft (approved):

build/buildRunTimeImageBases.sh
qianz2: Yes we found the baseImages pipelines failed since Jun, here we put back docker build buster based runtime images so that the ACR images will get built in the pipeline. Node contains only bullseye based images now, so we won't need to put one under there.

daniv-msft: Thanks for clarifying!

cormacpayne (approved):
:shipit: x2

Issue #1456, opened 2022-07-01
Disable ASCII color codes for ASP.NET apps


bradygaster:
> At what date and time did you most recently experience the problem?

Any time a customer deploys a .NET app to App Service for Linux using the ORYX images, the default log behavior emits ASCII color codes that are injected into the logs. This makes the logs quite difficult to read. This screenshot shows that the log output looks like when customers deploy an out-of-the-box ASP.NET Core app to App Service using an ORYX image:

![image](https://user-images.githubusercontent.com/41929050/176956476-1fd9b251-ca0a-4fc6-a475-ac3eb7d2b9d6.png)

Once a customer sets the `ASPNETCORE_LOGGING__CONSOLE__DISABLECOLORS` environment variable to `true` in their App Service Configuration blade:

![image](https://user-images.githubusercontent.com/41929050/176956602-deaaac9f-90ec-4a71-9694-172f35550776.png)

The log output is much friendlier:

![image](https://user-images.githubusercontent.com/41929050/176956741-b7514f48-eeeb-482b-97a7-f08777966f1d.png)

Ideally, ORYX team could set the `ASPNETCORE_LOGGING__CONSOLE__DISABLECOLORS` to `true` in the base image for .NET apps.

simonjj:
This issue seems resolved. Will be closing this issue. If this is still relevant and impacting you please open a new issue and link back to this issue for reference. Thanks @bradygaster for the fix.

PR #1457, opened 2022-07-01
disaabling colors via the envvar

bradygaster:
This is a fix (or at least _partial fix_) for #1456. cc @daniv-msft, with whom I've been chatting about this. Should there be any additional tweaks I'd need to make, let me know.


daniv-msft:
@waliMSFT FYI, this is the PR referred to in the separate work item.

cormacpayne:
@bradygaster @daniv-msft I'm not sure if this is user-facing from App Service, but do we know if this also affects users when building their app (_e.g._, during `dotnet publish`)? We could also set this environment variable in our build images if that's the case, but if it's only reported during runtime, then this should be good 👍

bradygaster:
> @bradygaster @daniv-msft I'm not sure if this is user-facing from App Service, but do we know if this also affects users when building their app (_e.g._, during `dotnet publish`)? We could also set this environment variable in our build images if that's the case, but if it's only reported during runtime, then this should be good 👍

good question @cormacpayne - it shouldn't impact customers when using `dotnet publish`. This is definitely more of a "runtime" sort of issue. Folks just can't really grok their logs with all the ASCII codes.

waliMSFT:
@bradygaster, I have tested this locally in a dotnet Image. it works fine. logs are coming without any color.

bradygaster:
YAS! When will this ship? :)

cormacpayne (approved):
:shipit:

PR #1458, opened 2022-07-05
Build script hard fail when ostype file does not exist

pauld-msft:
https://github.com/microsoft/Oryx/pull/1451#discussion_r910344791
We want the build script to fail when ostype is not present, as that represents an invalid state for the container

### DIFF
--- a/src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
+++ b/src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
@@ -222,7 +222,8 @@ then
 	echo "Copying .ostype to manifest output directory."
 	cp "$OS_TYPE_SOURCE_DIR" "$MANIFEST_DIR/.ostype"
 else
-	echo "File $OS_TYPE_SOURCE_DIR does not exist. Cannot copy to manifest directory."
+	echo "File $OS_TYPE_SOURCE_DIR does not exist. Cannot copy to manifest directory." 1>&2
+	exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
```diff
@@ -222,7 +222,8 @@ then
 	echo "Copying .ostype to manifest output directory."
 	cp "$OS_TYPE_SOURCE_DIR" "$MANIFEST_DIR/.ostype"
 else
-	echo "File $OS_TYPE_SOURCE_DIR does not exist. Cannot copy to manifest directory."
+	echo "File $OS_TYPE_SOURCE_DIR does not exist. Cannot copy to manifest directory." 1>&2
+	exit 1
```
pauld-msft, line 226: Was looking to add a test that created a state where this was hit, but I was having trouble removing the ostype file before executing the build script. I can spend a bit more time trying to get this failure case to work if the team thinks it would be good to have.

daniv-msft (dismissed):

cormacpayne (approved):
:shipit:

src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
pauld-msft: @daniv-msft luckily there was a test failure that alerted me to a particular test suite I hadn't seen before since it only runs on linux machines (so didn't run locally). I was able to take these tests and base a new one off of those in order to test this functionality explicitly.

PR #1461, opened 2022-07-06
Update BuildServer to prevent command injection

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1566795

### DIFF
--- a/src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
+++ b/src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
@@ -34,7 +34,7 @@ namespace Microsoft.Oryx.BuildServer.Services.ArtifactBuilders
                 StartInfo = new ProcessStartInfo
                 {
                     FileName = "/bin/bash",
-                    Arguments = $"-c $\'{cmd}\'",
+                    Arguments = $"-c \'{cmd}\'",
                     RedirectStandardOutput = true,
                     UseShellExecute = false,
                     CreateNoWindow = true,
--- a/src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
+++ b/src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
@@ -29,7 +29,6 @@ namespace Microsoft.Oryx.BuildServer.Services.ArtifactBuilders
                 $"--output {outputPath} --platform {build.Platform} " +
                 $"--platform-version {build.Version}";
             cmd = cmd.Replace("'", "\\'");
-            var escapedArgs = cmd.Replace("\"", "\\\"");
             var process = new Process()
             {
                 StartInfo = new ProcessStartInfo
--- a/src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
+++ b/src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
@@ -5,6 +5,7 @@
 
 using System.Diagnostics;
 using System.IO;
+using System.Text.RegularExpressions;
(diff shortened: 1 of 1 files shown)


src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
```diff
@@ -28,13 +28,14 @@ public bool Build(Build build)
             var cmd = $"oryx build {sourcePath} --log-file {logFilePath} " +
                 $"--output {outputPath} --platform {build.Platform} " +
                 $"--platform-version {build.Version}";
+            cmd = cmd.Replace("'", "\\'");
             var escapedArgs = cmd.Replace("\"", "\\\"");
```
surenderssm, line 32: Where are we using `escapedArgs `?

daniv-msft (dismissed):

qianz2 (dismissed):

src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
william-msft: Removed the un-used variable. Thanks!

src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
```diff
@@ -59,5 +64,19 @@ public bool Build(Build build)
...
+            Regex regex = new Regex(@"^[a-zA-Z0-9./\-_]*$");
+            Match match = regex.Match(parameter);
+            if (match.Success)
+            {
+                return parameter;
+            }
+
+            this.logger.LogError($"Invalid parameter. " +
+                $"Only alpha-numeric, '/', '.', '-', '_' characters are allowed");
+            return null;
```
cormacpayne, line 79: nit: should we throw an exception here instead of returning `null`? Is there a scenario where we'd want to continue with a command different from what the user intended instead of terminating early?

src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
```diff
@@ -59,5 +64,19 @@ public bool Build(Build build)
...
+        private string ValidateParameter(string parameter)
+        {
+            Regex regex = new Regex(@"^[a-zA-Z0-9./\-_]*$");
+            Match match = regex.Match(parameter);
+            if (match.Success)
+            {
+                return parameter;
+            }
+
+            this.logger.LogError($"Invalid parameter. " +
```
cormacpayne, line 77: nit: what do you think about adding the provided parameter to this `LogError` call so we're more clear about what parameter we're referring to?

```cs
this.logger.LogError($"Invalid parameter provided: '{parameter}'. Only alpha-numeric, '/', '.', '-', '_' characters are allowed");
```

cormacpayne (dismissed):
Two minor comments, otherwise LGTM :shipit:

qianz2 (dismissed):

src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
william-msft: Ah I meant to add this, thanks!

src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
william-msft: We might change the way we are handling errors in the future, for now bash's error message will suffice

cormacpayne (approved):

PR #1471, opened 2022-07-08
Pauld/1551977 allow php fpm configuration

pauld-msft:
https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1551977
This work will allow the customers to provide environment variables through App Settings that will configure the fpm service on app startup.


tests/Oryx.Tests.Common/EndToEndTestHelper.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.Oryx.Tests.Common
     public static class EndToEndTestHelper
     {
         private const int MaxRetryCount = 100;
-        private static readonly TimeSpan DelayBetweenRetries = TimeSpan.FromSeconds(100);
+        private static readonly TimeSpan DelayBetweenRetries = TimeSpan.FromSeconds(10);
```
pauld-msft, line 23: @william-msft was there a reason that we needed to delay a full 100 seconds AND retry 100 times? I thought that checking for liveliness every 10 seconds, for a total of ~16 minutes would be enough to prevent timeouts, and it saves us time on the tests that I've tried it with (since we check for the first time after 10 seconds instead of a full 100).

pauld-msft: Ok so the nightly tests have answered this for me with a suite of failures. I still have some questions regarding this wait, namely, why are we waiting so long before checking for container existence? This wait seems to me like it could be adding significant time to our testing suite.

I think a better approach would be to:
1. Initiate container run
2. Check for container existence every few seconds for a certain period of time, if it doesn't exist after ~2-3 minutes then the test fails
3. Then we begin checking for the app to be up every few seconds for a certain period of time

It seems like this would also separate the 2 processes of checking for the container to exist and the app to startup. It also would allow for containers that are from cached images and don't need to be pulled down (so they will start up much faster) to not need to be waiting for a full 100 seconds before even checking for the application to be up.

Happy to give this a shot, but let me know if its already been explored at all or any thoughts you have for it!

cormacpayne: The above approach sounds good to me -- as long as there is no performance hit or downside to checking the container existence and application on a faster interval than we should go ahead with it.

tests/Oryx.Integration.Tests/Php/PhpFpmConfigTest.cs
```diff
@@ -0,0 +1,226 @@
...
+            await PhpFpmConfigTestFailuresAsync(
+                "7.4",
+                fpmMaxChildren,
+                fpmStartServers,
+                fpmMaxSpareServers,
+                fpmMinSpareServers,
+                failureOutputText);
+        }
+
+        private async Task PhpFpmConfigTestAsync(
```
cormacpayne, line 104: nit: it looks like this method and `PhpFpmConfigTestFailuresAsync` share a lot of the same code -- I think it could be worth looking into consolidating them into a single method that have the same arguments with an additional optional argument for the `failureOutputText` whose existence changes the assertion portion of the code.

cormacpayne (approved):
A couple of minor comments, otherwise LGTM :shipit:

tests/Oryx.Tests.Common/DockerCli.cs
```diff
@@ -147,6 +150,75 @@ public DockerRunCommandProcessResult RunAndDoNotWaitForProcessExit(DockerRunArgu
                 $"{DockerCmd} {string.Join(" ", arguments)}");
         }
 
+        public async Task<DockerRunCommandProcessResult> RunAndWaitForContainerStartAsync(DockerRunArguments dockerRunArguments)
```
pauld-msft, line 153: - [ ] TODO: utilize existing RunAndDoNotWaitForProcessExit method to prevent copied code

tests/Oryx.Integration.Tests/Php/PhpGreetingsAppTest.cs
```diff
@@ -12,7 +12,6 @@
 
 namespace Microsoft.Oryx.Integration.Tests
 {
-    [Collection("Php integration")]
```
pauld-msft, line 15: Removed these from the tests, now that I've found that they can run in parallel as long as we give them some more time to start up, which I did with the changes that also check more often for container existence

tests/Oryx.Integration.Tests/Php/PhpFpmConfigTest.cs
```diff
@@ -0,0 +1,170 @@
...
+            string phpVersion, 
+            string fpmMaxChildren, string expectedFpmMaxChildren,
+            string fpmStartServers, string expectedFpmStartServers,
+            string fpmMaxSpareServers, string expectedFpmMaxSpareServers,
+            string fpmMinSpareServers, string expectedFpmMinSpareServers,
+            string failureOutputText)
```
cormacpayne, line 80: nit: I would have `failureOutputText` be optional so a test that doesn't expect a failure doesn't need to worry about providing something for this argument. I would also apply this to your tests above so that the `InlineData` sets that don't expect a failure don't have to provide a `null` value to the argument, they just don't provide anything.

```cs
[Theory]
[InlineData("foo", "FAILED")]
[InlineData("bar")]
public async Task TestSomethingAsync(
    string name,
    string failureOutputText = default)
{
    // Test
}
```

tests/Oryx.Integration.Tests/Php/PhpFpmConfigTest.cs
```diff
@@ -0,0 +1,170 @@
...
+            var runScript = new ShellScriptBuilder()
+                .AddCommand($"oryx create-script -appPath {appOutputDir} -output {RunScriptPath} -bindPort {ContainerPort}")
+                .AddCommand("mkdir -p /home/site/wwwroot")
+                .AddCommand($"cp -rf {appOutputDir}/* /home/site/wwwroot")
+                .AddCommand(RunScriptPath)
+                .ToString();
+
+            var phpimageVersion = string.Concat(phpVersion, "-", "fpm");
+
+            if (failureOutputText == null)
```
cormacpayne, line 104: nit: I generally like to do `string.IsNullOrEmpty` for string checks, even if we explicitly set the value to `null` somewhere (helps keep consistent behavior if you were to change that `null` to `default` or `string.Empty`)

tests/Oryx.Tests.Common/DockerCli.cs
```diff
@@ -147,6 +150,24 @@ public DockerRunCommandProcessResult RunAndDoNotWaitForProcessExit(DockerRunArgu
...
+        {
+            var dockerCommand = RunAndDoNotWaitForProcessExit(dockerRunArguments);
+
+            var retry = 0;
+            while(!dockerCommand.Process.HasExited && retry < TotalContainerStartupRetries)
```
cormacpayne, line 158: nit: space between `while` and `(` 🙂

tests/Oryx.Tests.Common/DockerCli.cs
```diff
@@ -147,6 +150,24 @@ public DockerRunCommandProcessResult RunAndDoNotWaitForProcessExit(DockerRunArgu
...
+            {
+                if (!string.IsNullOrEmpty(GetContainerStatus(dockerCommand.ContainerName)))
+                {
+                    break;
+                }
```
cormacpayne, line 163: nit: newline after a closing brace 🙂

cormacpayne (approved):
A few minor comments, otherwise LGTM :shipit:

cormacpayne (dismissed):

tests/Oryx.Tests.Common/EndToEndTestHelper.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.Oryx.Tests.Common
     public static class EndToEndTestHelper
     {
         private const int MaxRetryCount = 100;
-        private static readonly TimeSpan DelayBetweenRetries = TimeSpan.FromSeconds(100);
+        private static readonly TimeSpan DelayBetweenRetries = TimeSpan.FromSeconds(2);
```
qianz2, line 23: What does this timespan set up for? Does it reduce the retry waiting time when timeout occurs?

william-msft: Yeah I changed it a while back because this was very flaky: https://github.com/microsoft/Oryx/commit/0cd0b89843e4fcc32a3141fe77b503282e23747f#

I do wonder if 2 seconds is too little. Network ports usually seem to take a few seconds to retry (open and close)

tests/Oryx.Tests.Common/DockerCli.cs
```diff
@@ -147,6 +150,25 @@ public DockerRunCommandProcessResult RunAndDoNotWaitForProcessExit(DockerRunArgu
...
         }
 
+        public async Task<DockerRunCommandProcessResult> RunAndWaitForContainerStartAsync(DockerRunArguments dockerRunArguments)
+        {
+            var dockerCommand = RunAndDoNotWaitForProcessExit(dockerRunArguments);
+
+            var retry = 0;
+            while (!dockerCommand.Process.HasExited && retry < TotalContainerStartupRetries)
+            {
+                if (!string.IsNullOrEmpty(GetContainerStatus(dockerCommand.ContainerName)))
```
qianz2, line 160: Trying to have better understanding for this part, It'd be great if you could help clarifying a bit what does this part do.

tests/Oryx.Integration.Tests/Php/PhpGreetingsAppTest.cs
qianz2: If we remove "php integration" collection from these tests, would they show up in the php integration jobs for nightly build or CI build? Just want to double check they won't removed from the pipeline jobs if we remove this line.

tests/Oryx.Tests.Common/EndToEndTestHelper.cs
william-msft: Could we bump this up from 2 seconds to 10 seconds (what it was originally)? My concern is the when perfroming a retry, the port that it's trying to close and re-open hasn't been closed completely and cause errors when trying to re-open. Unless we're sure 2 seconds will suffice

pauld-msft: Sure I can bump back to the original delay. 

@qianz2 this timespan is the delay in between each retry of trying to get the host port, and since we wait this long [before we try to get the host port for the first time](https://github.com/microsoft/Oryx/pull/1471/files/7d2d507ce5e559dc8e4c3b17f0457a42bf7aaec6#diff-98cee7e79c6b4790f7b445c7de5c47e1fc95b79b19941e94cf70a13013e10b0cR412), we were introducing a minimum of a 100 second delay before we even attempted to see if the container was up and had a port that we could access. We needed this delay in the past since it gave time (100 seconds) for the container to start up. One of the changes in this PR is to introduce a new method that waits for the container to start up (checking every few seconds) before we try to connect to the host port, so in theory we can then attempt to reconnect to the host port here every few seconds and since the container exists we will not error out on the first attempt.

tests/Oryx.Tests.Common/DockerCli.cs
pauld-msft: I added a comment to explain why we are getting the container status, but essentially the entire method is checking every few seconds for the existence of the container, and as soon as it exists (or the process exits, indicating a failure) then we return the container result. This method allows us to go into the GetHostPortAsyc method, knowing that either the container exists and is running or something has errored. I tried to explain this flow in this comment https://github.com/microsoft/Oryx/pull/1471#discussion_r917934755

tests/Oryx.Integration.Tests/Php/PhpGreetingsAppTest.cs
pauld-msft: Yes they will show up in the integration tests, since the tests themselves have the correct categories. The only thing that this collection was doing was making sure that the php tests were not running in parallel, which was a temporary fix for some transient failures we were seeing.

qianz2 (approved):

PR #1477, opened 2022-07-13
Revert Nginx version from PHP Fpm images

qianz2:
Temporarily revert Nginx version in PHP Fpm images to unblock the pipelines. 
Nginx web server was not successfully installed on the latest php fpm image during the integration tests.


build/__phpVersions.sh
```diff
@@ -1,7 +1,7 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
```
william-msft, line 2: I noticed there are more changes in the merged PR: https://github.com/microsoft/Oryx/pull/1453/files#diff-96c90487c35a34cd754813eae9cf3009f38549fe4f0a03ddcb55489ceffee8caR50

I just wanted to double check we are not leaving anything out. Otherwise, look good to me

PR #1482, opened 2022-07-18
Bugfix: storage account tests stopping at first empty line

pauld-msft:
This pull request addresses a bug where we stop checking the `versionsToBuild.txt` file after the first empty line, which as you can imagine is an issue for [a version file like our dotnet one](https://github.com/microsoft/Oryx/blob/main/platforms/dotnet/versionsToBuild.txt).

------
This PR is marked as draft because I am additionally looking for some feedback from the team as to how we should address rollout cases, which will make these tests fail as they assert that every version in the versionsToBuild.txt exists in the dev and prod storage accounts.
1. Leave test as is, and reformat the versionsToBuild files to be like:
```# version, sha, download url
3.1.811, sha,
5.0.203, sha,

# newest version
7.0.100-preview.5.22307.18, sha
```
This format, with the whitespace between the other versions and the version we rollout will cause the test to stop checking before it gets to the newest version.

2. Hardcode versions that are rolling out in the test and add a step to the version release instructions to make sure that the list is up to date.

4. Add some sort of a threshold to allow for failing versions that are rolling out. Cons are that we could hide issues up to that threshold, and we would not be able to roll out more than that number of versions at the same time.

More than open to any other suggestions as to how we could get around this issue!

### DIFF
--- a/tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
+++ b/tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
@@ -331,10 +331,10 @@ namespace Oryx.Integration.Tests
             using (var streamReader = new StreamReader(versionFile))
             {
                 string line = null;
-                while ((line = streamReader.ReadLine()) != null && !string.IsNullOrEmpty(line))
+                while ((line = streamReader.ReadLine()) != null)
                 {
-                    // ignore comments
-                    if (line.StartsWith("#"))
+                    // ignore comments or empty lines
+                    if (line.StartsWith("#") || string.IsNullOrEmpty(line))
                     {
                         continue;
                     }
@@ -363,10 +363,10 @@ namespace Oryx.Integration.Tests
             using (var streamReader = new StreamReader(file))
             {
                 string line = null;
-                while ((line = streamReader.ReadLine()) != null && !string.IsNullOrEmpty(line))
+                while ((line = streamReader.ReadLine()) != null)
                 {
-                    // ignore comments
-                    if (line.StartsWith("#"))
+                    // ignore comments or empty lines
+                    if (line.StartsWith("#") || string.IsNullOrEmpty(line))
                     {
                         continue;
                     }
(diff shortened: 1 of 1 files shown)


tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
```diff
@@ -331,10 +331,10 @@ private List<string> GetListOfVersionsToBuild(params string[] platformPath)
             using (var streamReader = new StreamReader(versionFile))
             {
                 string line = null;
-                while ((line = streamReader.ReadLine()) != null && !string.IsNullOrEmpty(line))
+                while ((line = streamReader.ReadLine()) != null)
                 {
-                    // ignore comments
-                    if (line.StartsWith("#"))
+                    // ignore comments or empty lines
+                    if (line.StartsWith("#") || string.IsNullOrEmpty(line))
```
cormacpayne, line 337: This is a good catch, thanks for making this change -- one thing I'm noticing we could do to expand this a little is for trim the `line` variable of any extra whitespace it is (specifically the case where the line _only_ has whitespace) to prevent it from skipping over the conditional we have here.

```cs
// Remove extraneous whitespace
line = line.Trim();

// Ignore comments or empty lines
if (line.StartsWith("#") || string.IsNullOrEmpty(line))
```

cormacpayne (dismissed):

tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
william-msft: Yes, here is the related work item: https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1577218

william-msft (dismissed):

tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
william-msft: All are good suggestions. Parameterizing the pipeline does seem like it will solve the rollout case you are referring, Paul.

pauld-msft: Ok yeah that seems like it would fix the issue as long as we also make sure to have it override whatever is passed into the SDK storage tests, such as the prod url here https://github.com/microsoft/Oryx/blob/main/tests/Oryx.Integration.Tests/ProdStorageAccountSanityTest.cs#L20

PR #1483, opened 2022-07-18
Update SDKS debian type for PHP and PHP composer

qianz2:
As we don't support bullseye SDKs for now, when pulling SDKs from a bullseye based build image, it'll look for bullseye SDKs which doesn't exist. So we replace it with buster SDKs to resolve an issue from partner team.

### DIFF
--- a/src/BuildScriptGenerator/PlatformInstallerBase.cs
+++ b/src/BuildScriptGenerator/PlatformInstallerBase.cs
@@ -55,6 +55,8 @@ namespace Microsoft.Oryx.BuildScriptGenerator
                 .AppendLine($"cd {versionDirInTemp}")
                 .AppendLine("PLATFORM_BINARY_DOWNLOAD_START=$SECONDS")
                 .AppendLine($"platformName=\"{platformName}\"")
+                // Temporarily use buster SDKs in bullseye image until bullseye SDKs are supported in storage account.
+                // Bug item: 1578477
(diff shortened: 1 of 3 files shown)


src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -54,10 +55,13 @@ public abstract class PlatformInstallerBase
                 .AppendLine($"cd {versionDirInTemp}")
                 .AppendLine("PLATFORM_BINARY_DOWNLOAD_START=$SECONDS")
                 .AppendLine($"platformName=\"{platformName}\"")
+                .AppendLine($"if [ \"$DEBIAN_FLAVOR\" = \"bullseye\" ]; then")
```
daniv-msft, line 58: Could you please clarify how this only applies to PHP 7.4?
Also, this would definitely deserve a comment in the codebase to clarify why we're doing that, and the link to the work item to revert it, etc. Could you please add this?

qianz2: Hi Dan, I believe the issue not only happens for PHP 7.4, but also other PHP and PHP composer versions. As we know of, we added an if else statement for PHP and PHP composer for an old ICM issue, so it will select SDKs according to its image debian type. But now we have bullseye image, so it'll apply to bullseye SDKs for PHP and PHP composer, which will definitely fail. So we need to select buster SDKs until bullseye SDKs are supported.

Yes sure, I can add comment and link above it.

cormacpayne (approved):

PR #1484, opened 2022-07-20
Pauld/1559705 improve cli to detect viable sdks based on debian flavor

pauld-msft:
Hi all, sorry for the monster of a PR. Luckily, a lot of it is small changes. This is the first of 3 pull requests, which are documented at the links below. 

Link to ticket: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1559705
Link to high level plan: https://microsoft-my.sharepoint.com/:o:/p/pauldorsch/EjlwXiCawbVJpdhAIrlGwEIBKF43t43jgXebfVuTh0fXJQ?e=vo1VZG

TLDR; 
PR 1 (This one): Updates what we upload to the sdk storage, only lets cli pull sdks for its debian flavor, and pushes some bullseye sdks.
PR 2: Updates logic to use bullseye sdks, which won't exist in the storage account until we merge this PR.
PR 3: Release some new bullseye-only sdks (like Node 18, Python 3.10, and Python 3.11)

-----------

#### SDK Storage Upload
- [x] Updated file structure to support specifying versions for each os type
  - [Image of new file structure](https://user-images.githubusercontent.com/107068277/179870391-e5541dd4-b306-4433-b3d4-61a78fa91db9.png)

- [x] Storage accounts will now contain default files like `defaultVersion.<debian flavor>.txt`
- [x] Add metadata to each item we upload
  - SDK_Version
  - Os_type
  - Version (only stretch images will be uploaded with this, it is what is currently used to generate the list of compatible versions)
- [x] TODO: determine what SDKs we want uploaded for each os type
#### Shared code generator
- [x] Support for new file structure
- [x] Fixed bugs with supported platforms documentation file generation, so we can now keep all changes that come from executing the `generateConstants.sh` script
- [x] Support for a legacy versions file, which allows us to keep the supportedVersion.md file up to date automatically
#### CLI
- [x] Support for platform installation only pulling sdks that are associated with its current debian flavor
- [x] Added regex to support determining compatible sdks (for non-stretch) for now, to be deprecated once the sdk metadata is uploaded to the sdk storage
#### Tests
- [x] Update to storage account tests to make sure that we have the expected versions and files
- [x] Update cli tests to ensure that debian flavor is being respected in what is being pulled
#### Small enhancements
- [x] Allow for buildPlatformBinaries.sh to move tar files into artifacts directory for node and java when executed on windows machine
- [x] Added php fpm config to constants file and made relevant changes so they aren't overwritten when generateConstants is run


build/__extVarNames.sh
```diff
@@ -5,6 +5,16 @@ APPLICATION_INSIGHTS_INSTRUMENTATION_KEY_ENV_VAR_NAME='ORYX_AI_INSTRUMENTATION_K
 APP_SERVICE_APP_NAME_ENV_VAR_NAME='APPSETTING_WEBSITE_SITE_NAME'
 SCM_COMMIT_ID_ENV_VAR_NAME='SCM_COMMIT_ID'
 PHP_ORIGIN='PHP_ORIGIN'
+PHP_FPM_CONFIGURATION_FILE='/usr/local/etc/php-fpm.d/www.conf'
```
pauld-msft, line 8: I added these to the `constants.yaml` file, since these variables were getting overwritten in other files when the `generateConstants.sh` script was run.

build/tools/SharedCodeGenerator/VersionComparer.cs
```diff
@@ -0,0 +1,52 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+
+namespace SharedCodeGenerator
+{
+    internal class VersionComparer : IComparer<string>
```
pauld-msft, line 11: Ah man, I just stumbled across the https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator.Common/VersionInfo.cs class, which I can probably steal to do this in a cleaner way

build/tools/SharedCodeGenerator/Program.cs
```diff
@@ -106,30 +107,100 @@ private static void GenerateSupportedPlatformsReadmeFile(string repoDir)
...
+                    string line;
+                    while ((line = reader.ReadLine()) != null)
+                    {
+                        if (string.IsNullOrWhiteSpace(line) || line.Trim().StartsWith("#"))
+                        {
+                            continue;
+                        }
+
+                        var parts = line.Split(",", StringSplitOptions.RemoveEmptyEntries);
+                        var versionPart = parts[0];
```
cormacpayne, line 173: Do we need to trim potential whitespace from `parts[0]`?

build/tools/SharedCodeGenerator/Program.cs
```diff
@@ -106,30 +107,100 @@ private static void GenerateSupportedPlatformsReadmeFile(string repoDir)
...
+
+                            if (string.IsNullOrWhiteSpace(line) || line.StartsWith("#"))
                             {
                                 continue;
                             }
 
-                            var parts = line.Split(",", StringSplitOptions.RemoveEmptyEntries);
-                            var versionPart = parts[0];
-                            sw.WriteLine($"- {versionPart}");
+                            supportedVersions.Add(line);
```
cormacpayne, line 192: Do we need to modify the `line` in any way before adding it to `supportedVersions`? Are the lines written to `legacyVersions.txt` guaranteed to always be _just_ a version from what you've seen?

build/tools/SharedCodeGenerator/VersionComparer.cs
cormacpayne: I wrote a `SemanticVersionComparer` in Terrapin that we could easily port over if you want to reuse that code.

images/runtime/python/3.10/base.bullseye.Dockerfile
```diff
@@ -13,6 +13,7 @@ ENV BUILD_NUMBER=${BUILD_NUMBER}
 RUN ./build.sh python /opt/startupcmdgen/startupcmdgen
 
 FROM oryxdevmcr.azurecr.io/private/oryx/oryx-run-base-${DEBIAN_FLAVOR} as main
+ARG DEBIAN_FLAVOR
```
cormacpayne, line 16: If we're adding `DEBIAN_FLAVOR` to this Dockerfile, we should probably drop the `bullseye` from the file name (unless this is a change happening to every Dockerfile independent of what Debian flavors it will actually be built against).

platforms/dotnet/versions/buster/defaultVersion.txt
```diff
@@ -0,0 +1 @@
+3.1
```
cormacpayne, line 1: nit: just curious, are the default versions across Debian flavors for a given platform going to start off the same? I'm wondering if there's a case that we need to handle where the default for bullseye is different from buster, for example.

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -54,14 +55,19 @@ public abstract class PlatformInstallerBase
                 .AppendLine($"cd {versionDirInTemp}")
                 .AppendLine("PLATFORM_BINARY_DOWNLOAD_START=$SECONDS")
                 .AppendLine($"platformName=\"{platformName}\"")
-                .AppendLine($"if [[ \"$platformName\" = \"php\" || \"$platformName\" = \"php-composer\" ]] && [[ \"$DEBIAN_FLAVOR\" != \"stretch\" ]]; then")
+
+                // TODO: PR2 remove this if statement, as bullseye images will be supported
+                .AppendLine($"if [ \"$DEBIAN_FLAVOR\" = \"bullseye\" ]; then")
+                .AppendLine($"SDK_DEBIAN_TYPE=\"buster\"")
```
cormacpayne, line 61: Should this be `bullseye` or should the conditional above be checking for `buster`?

cormacpayne: Also do we want to use the constants generated in the `OsTypes` class for `buster` and `bullseye` strings in this file like we do below for `OsTypes.DebianStretch`?

src/BuildScriptGenerator/SdkStorageVersionProviderBase.cs
```diff
@@ -46,17 +47,46 @@ public class SdkStorageVersionProviderBase
...
+                    {
+                        supportedVersions.Add(versionElement.Value);
+                    }
+
+                    continue;
+                }
+
+                // TODO: PR2 Remove this logic, as we will be able to depend on bullseye images existing in the sdk storage
+                var patternText = this.commonOptions.DebianFlavor == OsTypes.DebianBullseye
+                    ? $"{platformName}-buster-(?<version>.*?).tar.gz"
```
cormacpayne, line 71: Same comment here: is `bullseye` being treated as `buster` for the time being?

(also same comment about using the `OsTypes` constants)

platforms/dotnet/versions/buster/defaultVersion.txt
pauld-msft: @cormacpayne  So this file was added for each OS type exactly for that case. I was imagining a scenario in the future where we have stopped releasing versions to stretch as it is EOL, and we want to update the bullseye image's default version for a language BUT that new version doesn't exist for stretch.

This file will get uploaded to the storage account as `defaultVersion.buster.txt`, and our CLI can pull it down if its running on buster.

src/BuildScriptGenerator/PlatformInstallerBase.cs
pauld-msft: @cormacpayne  This was added when I saw Stella's changes in this PR: https://github.com/microsoft/Oryx/pull/1483/files 

These changes are here because we don't have any bullseye sdks, so bullseye images should pull buster SDKs until this PR adds bullseye sdks to our storage account

src/BuildScriptGenerator/SdkStorageVersionProviderBase.cs
pauld-msft: Yes thats right, for the time being we want bullseye images to pull buster sdks, until this PR releases bullseye sdks to our storage account.

build/tools/SharedCodeGenerator/VersionComparer.cs
pauld-msft: @cormacpayne As long as it can support some non-semantic versions, such as golang which we have as 1.17 (no patch version). I was able to modify the VersionInfo to fallback to the built in Version class for non-semantic-versioned platforms (right now just golang), I just haven't pushed any changes quite yet.

images/runtime/python/3.10/base.bullseye.Dockerfile
pauld-msft: I _think_ that the buildRuntimeImageBases is depending on the bullseye to be there in order to find it, right? https://github.com/microsoft/Oryx/blob/main/build/buildRunTimeImageBases.sh#L67-L68

build/constants.yaml
```diff
@@ -245,9 +268,19 @@
...
     default-version-file-name: defaultVersion.txt
+    default-version-file-prefix: defaultVersion
+    default-version-file-type: txt
     versions-to-build-file-name: versionsToBuild.txt
     container-metadata-url-format: '{0}/{1}?restype=container&comp=list&include=metadata'
     sdk-download-sentinel-file-name: .oryx-sdkdownload-sentinel
+    # The following metadata names control the field that is uploaded to sdk storage alongside each sdk.
+    # Legacy is used to provide backwards compatibility for images with an older CLI, as they depend on identifying
+    # supported versions by searching for all SDKs that contain a Version metadata field.
+    sdk-version-metadata-name: SDK_Version
```
qianz2, line 279: nit: `Sdk_version` to be consistent with other metadata-name. Or we could have all name in lowered case.

build/tools/SharedCodeGenerator/Program.cs
```diff
@@ -106,30 +106,100 @@ private static void GenerateSupportedPlatformsReadmeFile(string repoDir)
...
+                                break;
+                            case "maven":
+                                sw.WriteLine($"## {platformName} maven");
+                                sw.WriteLine();
+                                AddVersions(sw, Path.Join(platformSubDirPath, "versions"), "maven");
+                                break;
+                            case "composer":
+                                sw.WriteLine($"## {platformName} composer");
+                                sw.WriteLine();
+                                AddVersions(sw, Path.Join(platformSubDirPath, "versions"), "composer");
```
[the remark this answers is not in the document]
pauld-msft: We only have those 2 since they are the only languages with their own version files embedded in a language directory. So the structure for php and java looks like

- php/
  - versions/
  - composer/
    - versions/
- java/
  - versions/
  - maven/
    - versions/

pauld-msft:
Regarding https://github.com/microsoft/Oryx/pull/1484#issuecomment-1190304289, @daniv-msft and I had an offline conversation about it and we are going to try to keep the sdks that we upload for buster to a minimum. For now, we are trying to just upload the ones that we know we will need for runtime, and I made a change to the jamstack build image so that it uses the supported node 14 instead of node 12, which all OS's have an associated sdk for.

images/runtime/python/3.10/base.bullseye.Dockerfile
cormacpayne: Hmm looks that way, so now I'm curious why `DEBIAN_FLAVOR` is needed in this file if we always know what flavor it'll be 🤔 Let's just leave it as-is since it's not bothering anything, just something interesting in this file.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -12,7 +12,7 @@ ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
     LANG="C.UTF-8"
 
 ARG IMAGES_DIR="/opt/tmp/images"
-RUN oryx prep --skip-detection --platforms-and-versions nodejs=12 --debug \
+RUN oryx prep --skip-detection --platforms-and-versions nodejs=14 --debug \
```
qianz2, line 15: Is there a reason we upgrade nodejs version to 14 on the jamstack image?

images/runtime/python/3.10/base.bullseye.Dockerfile
qianz2: I have same question as Cormac for this new added argument `DEBIAN_FLAVOR`. I think when these `base.bullseye.Dockerfile` getting built, the `DEBIAN_FLAVOR=bullseye` environment variable was set up in the image already. The `generateDockerfiles.sh` script will look for any `base.<debianFlavor>.Dockerfile` under versions folder.

cormacpayne (approved):
Minor comments otherwise LGTM :shipit:

images/runtime/python/template.Dockerfile
```diff
@@ -13,6 +13,7 @@ ENV BUILD_NUMBER=${BUILD_NUMBER}
 RUN ./build.sh python /opt/startupcmdgen/startupcmdgen
 
 FROM oryxdevmcr.azurecr.io/private/oryx/%BASE_TAG% as main
+ARG DEBIAN_FLAVOR
 ARG IMAGES_DIR=/tmp/oryx/images
 ARG BUILD_DIR=/tmp/oryx/build
 ENV DEBIAN_FLAVOR=${DEBIAN_FLAVOR}
```
qianz2, line 19: Same here, I'm just a bit curious why we have debian argument at line 16 when we previously use line 19 to set up environment variable.

platforms/dotnet/versions/bullseye/versionsToBuild.txt
```diff
@@ -0,0 +1,15 @@
+# version, sha, download url
```
qianz2, line 1: An open question that we could discuss: 
Which way we should add these debian specific txt files, adding debian folders or putting them in a single folder but naming these like this: `defaultVersions-bullseye.txt`, `defaultVersions.bullseye.txt`.

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -56,19 +56,20 @@ public abstract class PlatformInstallerBase
                 .AppendLine("PLATFORM_BINARY_DOWNLOAD_START=$SECONDS")
                 .AppendLine($"platformName=\"{platformName}\"")
 
+                // TODO: PR2 remove this if statement, as bullseye images will be supported
                 // Temporarily use buster SDKs in bullseye image until bullseye SDKs are supported in storage account.
                 // Bug item: 1578477
-                .AppendLine($"if [ \"$DEBIAN_FLAVOR\" = \"bullseye\" ]; then")
-                .AppendLine($"SDK_DEBIAN_TYPE=\"buster\"")
+                .AppendLine($"if [ \"$DEBIAN_FLAVOR\" = \"{OsTypes.DebianBullseye}\" ]; then")
+                .AppendLine($"SDK_DEBIAN_TYPE=\"{OsTypes.DebianBuster}\"")
```
qianz2, line 63: `SDK_DEBIAN_TYPE` can be removed and use only `DEBIAN_FLAVOR` in the file.

src/CommonFiles/Oryx.ruleset
```diff
@@ -21,6 +21,9 @@
     <Rule Id="SA1611" Action="None" /> <!-- ElementParametersMustBeDocumented -->
     <Rule Id="SA1615" Action="None" /> <!-- ElementReturnValueMustBeDocumented -->
     <Rule Id="SA1618" Action="None" /> <!-- GenericTypeParametersMustBeDocumented -->
+    
+    <!-- Documentation text should end in a period -->
+    <Rule Id="SA1629" Action="None" />
```
qianz2, line 26: Here did we make ruleset stricter or looser? Curious does `Action="None"` mean that we start to ignore such cases?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
pauld-msft: Yes! This is because we use this Jamstack dockerfile for all of the various ostypes. Since nodejs 12 is deprecated, we don't want to force bullseye to have to support node 12 SDKS. Instead, we can upgrade to 14 here, which is available for all of the OS types

build/__extVarNames.sh
```diff
@@ -5,6 +5,16 @@ APPLICATION_INSIGHTS_INSTRUMENTATION_KEY_ENV_VAR_NAME='ORYX_AI_INSTRUMENTATION_K
...
+PHP_FPM_MAX_CHILDREN_ENV_VAR_NAME='FPM_MAX_CHILDREN'
+PHP_FPM_START_SERVERS_ENV_VAR_NAME='FPM_START_SERVERS'
+PHP_FPM_MIN_SPARE_SERVERS_ENV_VAR_NAME='FPM_MIN_SPARE_SERVERS'
+PHP_FPM_MAX_SPARE_SERVERS_ENV_VAR_NAME='FPM_MAX_SPARE_SERVERS'
+PHP_FPM_MAX_CHILDREN_SETTING_NAME='pm.max_children'
+PHP_FPM_START_SERVERS_SETTING_NAME='pm.start_servers'
+PHP_FPM_MIN_SPARE_SERVERS_SETTING_NAME='pm.min_spare_servers'
```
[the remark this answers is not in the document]
pauld-msft: Let me know if this answers your question: https://github.com/microsoft/Oryx/pull/1484#discussion_r925063050

src/CommonFiles/Oryx.ruleset
pauld-msft: Yes, this change ignores the errors that appeared when documentation text didn't end in a period. I wanted to document some methods that I wrote, and this error kept appearing. I disabled it since its my personal preference to allow docs without punctuation, but if people have strong feelings I will happily revert this change and add punctuation to the method docs

tests/Oryx.BuildImage.Tests/Python/CondaTests.cs
```diff
@@ -89,7 +89,7 @@ public void CanBuildAppWhichHasCondaEnvironmentYmlFile()
                 result.GetDebugInfo());
         }
 
-        [Fact, Trait("category", "vso-focal")]
+        [Fact, Trait("category", "latest")]
```
qianz2, line 92: Is these a reason we change the test image from vso focal to latest build image?

tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
```diff
@@ -29,7 +29,7 @@ public VersionInformationTest(ITestOutputHelper output)
             _dockerCli = new DockerCli();
         }
 
-        [Fact, Trait("category", "latest")]
+        [SkippableFact, Trait("category", "latest")]
```
qianz2, line 32: Will `SkippableFact` go skipping these test if they failed, any reason we add it and should we revert it?

build/tools/SharedCodeGenerator/VersionInfo.cs
```diff
@@ -0,0 +1,67 @@
...
+            var semanticVersionStr = displayVersion;
+
+            // The display version is in preview version format or it's a non-preview version
+            // Both cases can be handled by SemVer library.
+            // Throws ArgumentException if SemVer library found invalid version format.
+            if (displayVersion.Contains('-') || !displayVersion.Any(c => char.IsLetter(c)))
+            {
+                if (Version.TryParse(semanticVersionStr, out var dotnetVersion))
+                {
+                    return new SemVer.Version(dotnetVersion.ToString(3));
```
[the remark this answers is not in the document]
pauld-msft: @cormacpayne  So this is a little strange, since the library overrides the built in ToString method. Its behavior is that the argument passed in actually determines the depth of release version to print. See attached picture for details. So what is actually happening here is that we are taking any string that can be parsed by the built in library, but might not be able to be parsed by semver library, and converting it to something that the semver library can parse. 

That being said, you raise a valid point about why this block is even there, and it was leftover from before we handled golang as non-semver. Now, we should not need this since that is being handled appropriately. I will remove.

image

platforms/__common.sh
```diff
@@ -129,7 +129,7 @@ getSdkFromImage() {
 	echo
 	docker run \
 		--rm \
-		-v $hostVolumeDir:$volumeContainerDir \
+		-v /$hostVolumeDir:$volumeContainerDir \
```
william-msft, line 132: @pauld-msft is the `/` a typo?

vsts/scripts/publishSdksFromDevToProdStorageAccount.sh
```diff
@@ -51,12 +51,31 @@ function copyBlob() {
 
 function copyPlatformBlobsToProd() {
```
qianz2, line 52: We have another script for publishing SDKs to dev, just a reminder in case we also need to make any changes there.

images/runtime/python/template.Dockerfile
pauld-msft: This is because in the added line below, I could not find a way that the env var could be used in the COPY docker command. [It looked like](https://stackoverflow.com/a/55255444) I needed to place the debian flavor in the build context by making it an argument.

platforms/dotnet/versions/bullseye/versionsToBuild.txt
pauld-msft: @qianz2  I did try doing it this way, however to me it felt really cluttered. There were a whole bunch of similarly named files in the same directory. 

I can make this change, but it would be a bit of work on my end so I would like to have a great reason to do so!

tests/Oryx.BuildImage.Tests/VersionInformationTest.cs
pauld-msft: @qianz2  No, it will still fail if it exits with anything other than a skip exception. All this does is allow us to run it locally (where the test returns a Skip exception) and the test will mark it as skipped instead of failed when it receives the skip exception

tests/Oryx.BuildImage.Tests/Python/CondaTests.cs
pauld-msft: @qianz2 Yes! This is just because VSO-focal doesn't actually have python 2 SDKs associated with it. We only have python 2 for stretch images, like latest

platforms/__common.sh
pauld-msft: @william-msft nope, this is to allow buildPlatformBinaries.sh to move tar files into the artifacts directory for php, node, and java when executed on windows machine. Without this change, it doesn't work locally for those languages.

vsts/scripts/publishSdksFromDevToProdStorageAccount.sh
pauld-msft: @qianz2 Correct me if I'm wrong, but I think that SDKs get to dev by the following:
1. PlatformBinaries job is made with `../templates/_platformBinariesTemplate.yml`
2. `../templates/_platformBinariesTemplate.yml` invokes the `./build/buildPlatformBinaries.sh` script
3. `buildPlatformBinaries.sh` invokes the build{Platform}.sh scripts in the `platforms/` dir that I have modified to support this change
4. Those scripts place the SDKs in a local `artifacts/` directory
5. `/scripts/publishFilesToAzureStorage.sh` is run by the platform binaries job, which finds the contents of the `artifacts/` directory and publishes them

If all that is correct, then I believe that everything has been updated appropriately. I won't know for sure until I merge though, i think

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
qianz2: Ok, thanks for clarifying!

build/tools/SharedCodeGenerator/VersionInfo.cs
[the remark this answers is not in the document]
cormacpayne: Oops, read `.ToString` as `.Substring` 😄 Thanks for clarifying this -- if it was working before, then I think we're OK to keep it here.

src/BuildScriptGenerator/PlatformInstallerBase.cs
pauld-msft: I'll remove this in PR2, when we completely remove the logic that keeps Bullseye from pulling its own SDKs

build/tools/SharedCodeGenerator/VersionInfo.cs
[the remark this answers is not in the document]
pauld-msft: I removed it anyways since it just added code that made things more complicated, and was completely unnecessary haha

cormacpayne (dismissed):
:shipit:

william-msft (dismissed):

qianz2 (approved):

build/tools/SharedCodeGenerator/VersionInfo.cs
```diff
@@ -0,0 +1,62 @@
...
+        public Version Version { get; }
+
+        public int CompareTo(VersionInfo other) =>
+            this.nonSemverPlatforms.Contains(this.Platform)
+            ? this.Version.CompareTo(other.Version)
+            : this.SemanticVersion.CompareTo(other.SemanticVersion);
+
+        private static SemVer.Version ToSemanticVersion(string displayVersion)
+        {
+            var semanticVersionStr = displayVersion;
```
cormacpayne, line 45: Would you mind outlining some example scenarios for this method? I'm trying to understand a few things going on in this method, but just want to make sure I first understand its purpose.

platforms/__common.sh
```diff
@@ -139,6 +139,7 @@ buildPlatform() {
 	local funcToCall="$2"
 	while IFS= read -r VERSION_INFO || [[ -n $VERSION_INFO ]]
 	do
+		VERSION_INFO="$(echo -e "${VERSION_INFO}" | sed -e 's/^[[:space:]]*//')"
```
cormacpayne, line 142: nit: would you mind adding a comment above this line to explain what's happening to `VERSION_INFO`?

cormacpayne (approved):
Minor comments otherwise LGTM :shipit:

build/tools/SharedCodeGenerator/VersionInfo.cs
pauld-msft: So I pulled this from elsewhere in the code base, but I investigated the purpose of this method and it looks like adding the loose parameter will allow certain invalid formats, such as pre-release versions without a leading hyphen e.g. 3.10.0a2. And most of the method seems to be adding the hyphen into a version number in some cases when it wasn't found. So I ended up just removing the whole thing in favor of always using the loose param for semver versions. Tested it and it works identically.

pauld-msft: library docs: https://github.com/adamreeve/semver.net

cormacpayne: Got it -- thanks for the context

cormacpayne (approved):

PR #1497, opened 2022-07-29
Pauld/1559705 PR2 cleanup and utilize bullseye images

pauld-msft:
Link to ticket: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1559705
Link to high level plan: https://microsoft-my.sharepoint.com/:o:/p/pauldorsch/EjlwXiCawbVJpdhAIrlGwEIBKF43t43jgXebfVuTh0fXJQ?e=Hw81u4

TLDR;
PR 1: Updates what we upload to the sdk storage, only lets cli pull sdks for its debian flavor, and pushes some bullseye sdks.
***PR 2 (This one): Updates logic to use bullseye sdks, which won't exist in the storage account until we merge PR 1***
PR 3: Release some new bullseye-only sdks (like Node 18, Python 3.10, and Python 3.11)

*Note that these tests will not pass until we have moved the bullseye / newly formatted SDKs into the prod storage account*


images/build/installDotNetCore.sh
```diff
@@ -17,14 +17,12 @@ debianFlavor="$DEBIAN_FLAVOR"
 
 fileName="dotnet.tar.gz"
 
-# TODO: PR2 configure this to account for the different debian flavors once the Version metadata has
```
cormacpayne, line 20: No actual changes were made in this file right?

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreSdkStorageVersionProvider.cs
```diff
@@ -58,24 +59,43 @@ public void GetVersionInfo()
...
+                            .FirstOrDefault();
+
+                        var osTypeElement = childElements.Where(e => string.Equals(
+                                SdkStorageConstants.OsTypeMetadataName,
                                 e.Name.LocalName,
                                 StringComparison.OrdinalIgnoreCase))
                             .FirstOrDefault();
 
-                        supportedVersions[runtimeVersionElement.Value] = sdkVersionElement.Value;
+                        if (this.commonOptions.DebianFlavor == OsTypes.DebianStretch || this.commonOptions.DebianFlavor == osTypeElement.Value)
```
cormacpayne, line 95: Just to confirm: in addition to the platform binaries corresponding to the current Debian flavor of the image, we're _always_ going to say that stretch binaries are supported too, correct? It may be worth adding a comment above this if-statement calling out what's happening here just to be 100% clear to anyone who runs across this code in the future.

src/BuildScriptGenerator/SdkStorageVersionProviderBase.cs
```diff
@@ -47,46 +44,29 @@ protected PlatformVersionInfo GetAvailableVersionsFromStorage(string platformNam
...
+                var childElements = metadataElement.Elements();
+                var versionElement = childElements
+                    .Where(e => string.Equals(sdkVersionMetadataName, e.Name.LocalName, StringComparison.OrdinalIgnoreCase))
+                    .FirstOrDefault();
+
+                var osTypeElement = childElements
+                    .Where(e => string.Equals(SdkStorageConstants.OsTypeMetadataName, e.Name.LocalName, StringComparison.OrdinalIgnoreCase))
+                    .FirstOrDefault();
+
+                // if the os type is stretch and we find a blob with the correct version metadata, we add as a supported version
```
cormacpayne, line 64: nit: I'd rephrase this comment so that it prioritizes the fact that the version metadata needs to match since the result of `versionElement != null` is `&&`'d against the rest of the conditional statement, so it must always be `true` if we're to add this as a supported version.

tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
```diff
@@ -197,45 +203,33 @@ private XDocument GetMetadata(string platformName)
...
+                var childElements = metadataElement.Elements();
+                var versionElement = childElements
+                    .Where(e => string.Equals(sdkVersionMetadataName, e.Name.LocalName, StringComparison.OrdinalIgnoreCase))
+                    .FirstOrDefault();
+
+                var osTypeElement = childElements
+                    .Where(e => string.Equals(SdkStorageConstants.OsTypeMetadataName, e.Name.LocalName, StringComparison.OrdinalIgnoreCase))
+                    .FirstOrDefault();
+
+                // if the os type is stretch and we find a blob with the correct version metadata, we add as a supported version
```
cormacpayne, line 227: nit: same comment as above

images/build/installDotNetCore.sh
pauld-msft: @cormacpayne  correct, I just wanted to make sure that I went through and double checked that the logic in this method was going to be correct following the change to the storage account

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreSdkStorageVersionProvider.cs
pauld-msft: Yes, so the specific case this is addressing is the stretch platform binaries that no longer are uploaded (and therefore aren't overwritten with new metadata). So we can't rely on os type to pull stretch SDKs. Its also a little tricky, since we have to keep in mind that stretch SDKs have a different version metadata than all others (Version instead of Sdk_version).

We determine stretch supported versions by looking for SDKs with the `Version` metadata (only stretch SDKs will have this). we determine all other os type supported versions by looking for `Sdk_version` metadata for all SDKs with a matching `Os_type` metadata.

pauld-msft: @cormacpayne though you do raise a good point that we aren't checking to make sure that SDK version exists. I added this check, which shouldn't be possible to reach if we upload using our scripts... but adding the check means that this would simply not add a mis-formatted sdk as a supported version, instead of breaking the oryx build.

PR #1498, opened 2022-07-29
Point python 3.6 get-pip to new source

pauld-msft:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6490101&view=logs&j=949f3752-01a7-599c-273e-99de61b000fe&t=46adcf0e-eea5-5352-25f6-dcd70cf91fc3

failing because 3.6's get-pip was removed from the original source

### DIFF
--- a/platforms/python/prereqs/build.sh
+++ b/platforms/python/prereqs/build.sh
@@ -47,6 +47,10 @@ if [ "$debianFlavor" == "stretch" ]; then
     echo "Hack flavor is: "$debianHackFlavor
 
     pythonSdkFileName=python-$PYTHON_VERSION.tar.gz
+
+    if [[ $PYTHON_VERSION == 3.6* ]]; then
+        PYTHON_GET_PIP_URL="https://bootstrap.pypa.io/pip/3.6/get-pip.py"
+    fi
     PYTHON_GET_PIP_URL="https://bootstrap.pypa.io/get-pip.py"
     PIP_VERSION="20.2.3"
 else
--- a/platforms/python/prereqs/build.sh
+++ b/platforms/python/prereqs/build.sh
@@ -47,11 +47,12 @@ if [ "$debianFlavor" == "stretch" ]; then
     echo "Hack flavor is: "$debianHackFlavor
 
     pythonSdkFileName=python-$PYTHON_VERSION.tar.gz
-
+    PYTHON_GET_PIP_URL="https://bootstrap.pypa.io/get-pip.py"
+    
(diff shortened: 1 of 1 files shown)


platforms/python/prereqs/build.sh
```diff
@@ -47,6 +47,10 @@ if [ "$debianFlavor" == "stretch" ]; then
     echo "Hack flavor is: "$debianHackFlavor
 
     pythonSdkFileName=python-$PYTHON_VERSION.tar.gz
+
+    if [[ $PYTHON_VERSION == 3.6* ]]; then
+        PYTHON_GET_PIP_URL="https://bootstrap.pypa.io/pip/3.6/get-pip.py"
+    fi
     PYTHON_GET_PIP_URL="https://bootstrap.pypa.io/get-pip.py"
```
daniv-msft, line 54: Isn't the `PYTHON_GET_PIP_URL` value going to be overwritten by the next line?
Also, should we expect the `PIP_VERSION` to still be 20.2.3 in that case?

platforms/python/prereqs/build.sh
```diff
@@ -47,6 +47,10 @@ if [ "$debianFlavor" == "stretch" ]; then
     echo "Hack flavor is: "$debianHackFlavor
 
     pythonSdkFileName=python-$PYTHON_VERSION.tar.gz
+
+    if [[ $PYTHON_VERSION == 3.6* ]]; then
```
daniv-msft, line 51: Naive question, but is there a way for us to make this more generic than just targeting 3.6? Please ignore if this problem actually impacts only 3.6, and we don't care about other versions.

platforms/python/prereqs/build.sh
pauld-msft: yes, absolutely correct

platforms/python/prereqs/build.sh
pauld-msft: from the job: 
`ERROR: This script does not work on Python 3.6 The minimum supported Python version is 3.7. Please use https://bootstrap.pypa.io/pip/3.6/get-pip.py instead.`

daniv-msft: OK, thanks for fixing it!

platforms/python/prereqs/build.sh
daniv-msft: OK, thanks for fixing it!

daniv-msft (approved):
Looks good to me!

qianz2 (approved):

PR #1499, opened 2022-07-30
Add job to move sdks from prod to another storage account

pauld-msft:
This job will replace sdks in a destination storage account with whatever is in the prod storage account.

This should allow us to create a backup of what is currently in the production storage account, as well as restore the dev storage account from what exists in prod if we run into another issue like what we saw earlier this week

------------

This pull request also includes a job that will push from a set storage account to our production storage account.


vsts/scripts/copySdksFromProdToStorageAccount.sh
```diff
@@ -0,0 +1,85 @@
...
+
+function copyBlobContainerFromProdToDestination() {
+    local platformName="$1"
+
+    if shouldOverwriteSdk || shouldOverwritePlatformSdk $platformName; then
+        echo
+        echo "Overwriting blob container '$platformName' in storage account '$destinationSdk'."
+        # azcopy copy [source] [destination] [flags]
+        "$azCopyDir/azcopy" copy \
+            "$PROD_SDK_STORAGE_BASE_URL/$platformName/$PROD_STORAGE_SAS_TOKEN" \
```
cormacpayne, line 37: Even though you have it documented a couple of lines above, it would personally make me feel more comfortable if we used the explicit parameter name when providing these two arguments, just so there's no ambiguity as to what argument is associated with what parameter, and in the (hopefully) very very very slim chance a breaking change was made that flipped positional arguments, we are shielded from it.

vsts/scripts/copySdksFromProdToStorageAccount.sh
```diff
@@ -0,0 +1,85 @@
...
+            "$destinationSdk/$platformName/$sasToken" --overwrite true --recursive true
+    elif blobContainerExistsInDestination $platformName; then
+        echo
+        echo "Blob container '$platformName' already exists in storage account '$destinationSdk'. Skipping copying it..."
+    else
+        echo
+        echo "Blob container '$platformName' does not exist in storage account '$destinationSdk'. Copying it from $PROD_SDK_STORAGE_BASE_URL..."
+        # azcopy copy [source] [destination] [flags]
+        "$azCopyDir/azcopy" copy \
+            "$PROD_SDK_STORAGE_BASE_URL/$platformName/$PROD_STORAGE_SAS_TOKEN" \
```
cormacpayne, line 47: Same comment here about using named parameters vs. positional.

cormacpayne (commented):
Just a few comments I want to make sure we're on the same page about before approving 🙂

vsts/pipelines/copySdksFromProdToStorageAccount.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - task: ShellScript@2
+        displayName: 'Copy SDKs from the prod storage account to a destination storage account'
+        env:
+          DEV_STORAGE_SAS_TOKEN: $(DEV-STORAGE-SAS-TOKEN)
+          SANDBOX_STORAGE_SAS_TOKEN: $(SANDBOX-STORAGE-SAS-TOKEN)
+          PROD_STORAGE_SAS_TOKEN: $(PROD-STORAGE-SAS-TOKEN)
+        inputs:
+          scriptPath: ./vsts/scripts/copySdksFromProdToStorageAccount.sh
+          args:  ${{ variables.destinationStorageAccount }}
```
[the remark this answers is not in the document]
pauld-msft: Great point, I thought they would appear as variables in the pipeline that could be modified, but I think that may not be correct. I went ahead and updated it to be a parameter which will appear on the main blade when we try to kick off a pipeline manually, e.g. 
image

vsts/scripts/copySdksFromProdToStorageAccount.sh
pauld-msft: hmmm so I'm actually not sure if there is a way to do this? I've been looking around at the docs and such and can't find "source" or "destination" under any flags that I can pass in. did some testing locally and still couldn't determine the correct way to do it:
image

vsts/scripts/copySdksFromProdToStorageAccount.sh
pauld-msft: In the off chance that there is no way to explicitly do this, I've added a dry run parameter which will allow us to run the job without uploading or downloading anything, so that we can verify that we are inputting the correct parameters beforehand

vsts/pipelines/copySdksFromProdToStorageAccount.yml
```diff
@@ -0,0 +1,39 @@
...
+      
+      - task: ShellScript@2
+        displayName: 'Test Prod storage account'
```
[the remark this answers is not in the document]
pauld-msft: Yeah probably dont need the step... I just kept it around to make sure that everything looks ok after the script runs. But I think that we would have to add sandbox tests to oryx if we wanted this to do something meaningful for that account as well, since this is just running oryx tests

vsts/scripts/copyAllBlobsToProdStorageAccount.sh
```diff
@@ -0,0 +1,103 @@
...
+sasToken=""
+
+if [ "$1" = $SANDBOX_SDK_STORAGE_BASE_URL ]; then
+    sourceSdk=$SANDBOX_SDK_STORAGE_BASE_URL
+    sasToken=$SANDBOX_STORAGE_SAS_TOKEN
+elif [ "$1" = $DEV_SDK_STORAGE_BASE_URL ]; then
+    sourceSdk=$DEV_SDK_STORAGE_BASE_URL
+    sasToken=$DEV_STORAGE_SAS_TOKEN
+else
+	echo "Error: $1 is an invalid storage account url."
```
cormacpayne, line 85: nit: for just a little more clarity, "is an invalid _source_ storage account URL"

vsts/scripts/copyAllBlobsToProdStorageAccount.sh
```diff
@@ -0,0 +1,103 @@
...
+    local platformName="$1"
+
+    if shouldOverwriteSdk || shouldOverwritePlatformSdk $platformName; then
+        echo
+        echo "Overwriting blob container '$platformName' in storage account '$destinationSdk'."
+        # azcopy copy [source] [destination] [flags]
+        if [ $dryRun == "false" ] ; then
+            "$azCopyDir/azcopy" copy \
+                "$sourceSdk/$platformName$sasToken" \ 
+                "$PROD_SDK_STORAGE_BASE_URL/$platformName$PROD_STORAGE_SAS_TOKEN" --overwrite true --recursive
```
cormacpayne, line 39: Any reason we need the `--recursive` flag? I don't see that being used for the dev --> prod script we have today, so just curious what it's needed for here

cormacpayne (approved):
:shipit:

cormacpayne (approved):

PR #1502, opened 2022-08-03
Add document for supported runtime versions

cormacpayne:
Introduce a document that lists all runtime versions supported by Oryx to be present alongside the existing `supportedPlatformVersions.md` document that lists all platform versions in our storage account.

In the future, we should work to make this document dynamic like the platform versions document by pulling the versions from  `https://mcr.microsoft.com/v2/oryx/{platform}/tags/list`


pauld-msft (approved):
LGTM

PR #1511, opened 2022-08-10
Bugfix: platform binaries job requires defaultVersion.txt file deletion

pauld-msft:
This pull request should keep us from having to delete default version files when releasing platform binaries, as it will allow them to be overwritten.

Examples of the error this should resolve:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6534004&view=results
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6537900&view=results


cormacpayne (approved):
:shipit:

qianz2 (approved):

PR #1519, opened 2022-08-17
Install Nginx latest stable version from sources

qianz2:


images/runtime/php-fpm/template.base.Dockerfile
```diff
@@ -8,15 +8,20 @@ ENV PHP_ORIGIN php-fpm
...
-# Install NGINX 
-RUN apt-get update \
-    && apt-get install nano nginx -y
+# Install NGINX latest stable version using APT Method with Nginx Repository instead of distribution-provided one:
+# - https://www.linuxcapable.com/how-to-install-latest-nginx-mainline-or-stable-on-debian-11/
+RUN apt-get update
+RUN apt install curl nano -y
+RUN curl -sSL https://packages.sury.org/nginx/README.txt | bash -x
+RUN apt-get update
+RUN yes '' | apt-get install nginx-core nginx-common nginx nginx-full -y
```
daniv-msft, line 17: I saw that apt-get previously had a `--force-yes` option, that is now deprecated but seems to be replaced by even more precise `--allow-unauthenticated`, `--allow-downgrades`, `--allow-remove-essential`, `--allow-change-held-packages` parameters. Would it be possible to use these to be more granular than a blanket `yes` command?

images/runtime/php-fpm/template.base.Dockerfile
```diff
@@ -8,15 +8,20 @@ ENV PHP_ORIGIN php-fpm
...
 RUN ls -l /etc/nginx
 COPY images/runtime/php-fpm/nginx_conf/default.conf /etc/nginx/sites-available/default
 COPY images/runtime/php-fpm/nginx_conf/default.conf /etc/nginx/sites-enabled/default
 RUN sed -ri -e 's!worker_connections 768!worker_connections 10068!g' /etc/nginx/nginx.conf
 RUN sed -ri -e 's!# multi_accept on!multi_accept on!g' /etc/nginx/nginx.conf
 RUN ls -l /etc/nginx
+RUN nginx -t
```
daniv-msft, line 24: I saw from the doc that this will test the configuration file. Is this something we still need to keep in our final dockerfile?

images/runtime/php/8.0/base.buster.Dockerfile
```diff
@@ -19,6 +19,12 @@ RUN echo -e '<FilesMatch "\.(?i:ph([[p]?[0-9]*|tm[l]?))$">\n SetHandler applicat
 # Disable Apache2 server signature
 RUN echo -e 'ServerSignature Off' >> /etc/apache2/apache2.conf
 RUN echo -e 'ServerTokens Prod' >> /etc/apache2/apache2.conf
+RUN { \
```
daniv-msft, line 22: Could you please add a comment in the Dockerfile to explain what this is doing?

build/constants.yaml
```diff
@@ -140,7 +140,7 @@
 - name: php-versions
   constants:
     php-runtime-base-tag: 20220803.3
-    php-fpm-runtime-base-tag: 20220803.2
+    php-fpm-runtime-base-tag: 202208017.1
```
daniv-msft, line 143: Should this be `20220817.1` (one less zero)?

qianz2: Thanks for catching it!

images/runtime/php-fpm/template.base.Dockerfile
qianz2: This was a tricky place that I was working. There
are two prompts that showing options and let users enter either `"N"(default)` or some other letters. This is different from other `Y/N `prompts that we've seen. I tried lots of ways and the `force-yes` doesn't seem to work. And I found that `yes ' '` will perform similar as we tab the enter key and it will work.

images/runtime/php-fpm/template.base.Dockerfile
qianz2: I'd prefer to leave this line here. It would output the status of configuration file and make it easier to check and debug next time.

images/runtime/php/8.0/base.buster.Dockerfile
qianz2: It was auto generated when I run the `./buildRuntimeImages.sh` or `./buildRuntimeImageBases.sh` script locally. It happened previously and I think we're good to add these to the `conf` file as the default configuration.

cormacpayne (approved):

images/runtime/php-fpm/template.base.Dockerfile
daniv-msft: OK, thanks for clarifying!

cormacpayne:
@qianz2 Are there any additional tests that we can run here to verify that existing scenarios using nginx continue to work while also supporting the new targeted scenario(s)?

daniv-msft (approved):

qianz2:
> @qianz2 Are there any additional tests that we can run here to verify that existing scenarios using nginx continue to work while also supporting the new targeted scenario(s)?

@cormacpayne , sorry missed this comment before checking the PR in. Currently the PHP FPM integration tests are based on the success installation of Nginx web server and there're outputs from the BaseImages pipelines I can check that the Nginx new versions was installed there, so I didn't add more tests to our current PHP FPM runtime images tests, but let me know if you'd like us to have one for confirmation I can surly add it in another PR.

PR #1521, opened 2022-08-17
Pull Python 3.10 SDK from storage account for runtime image

cormacpayne:
After fixing the ODBC library issue for Python runtime images, the Antares team reported a bug in just the Python 3.10 runtime image where the `pyodbc` module is not able to be found. Looking at the stack trace, a different version of the Python SDK is being used for some of the calls (from the `/usr/local/lib` folder), but none of the other Python runtime images contain this additional SDK. I found this was due to an additional step in our `buildPythonSdkByVersion.sh` script that handles versions after and including 3.10 separately and attempts to build the SDK from scratch rather than pulling it from the storage account. Updating this logic to handle 3.10 like the older versions appears to resolve this issue locally with an app that repro'd the error on the current 3.10 runtime image, but not on the locally built copy with this fix.


pauld-msft:
Do we also need to update the other places in code that we are adding code specifically for python 3.10?
https://github.com/microsoft/Oryx/blob/main/platforms/python/prereqs/build.sh#L35
https://github.com/microsoft/Oryx/blob/main/platforms/python/buildPython.sh#L28-L29

cormacpayne:
> Do we also need to update the other places in code that we are adding code specifically for python 3.10? https://github.com/microsoft/Oryx/blob/main/platforms/python/prereqs/build.sh#L35 https://github.com/microsoft/Oryx/blob/main/platforms/python/buildPython.sh#L28-L29
> 
> They are related to platform binaries for build scripts, so maybe not since this is a runtime issue.

I don't think either of those need to be updated in this case -- the first link seems to be installing additional packages that may be needed as of Python 3.10.x, probably will investigate that later, and the second link is related to the stretch release of the 3.10 binaries, which wouldn't be used here since the runtime is bullseye-based, but also worth looking into later.

qianz2:
Did it reduce the size of python 3.10 SDK or runtime image once we build it as previous way? As I remember 3.10 image has much larger size, and we were trying to reduce it.

cormacpayne:
> Did it reduce the size of python 3.10 SDK or runtime image once we build it as previous way? As I remember 3.10 image has much larger size, and we were trying to reduce it.

Yes, previously it was just under 1 GB and now it's just under 600 MB, around the same size as the other Python runtimes.

build/buildPythonSdkByVersion.sh
```diff
@@ -150,12 +150,13 @@ getPythonGpgByVersion() {
...
 echo
-echo "Building python 3.10 or newer from source code..."
+# TODO: Determine if we need to continue building newer versions of Python from scratch
+echo "Building python 3.11 or newer from source code..."
 
 getPythonGpgByVersion "/tmp/versionsToBuild.txt" $version
 IFS='.' read -ra SPLIT_VERSION <<< "$PYTHON_VERSION"
 
-if  [ "${SPLIT_VERSION[0]}" == "3" ] && [ "${SPLIT_VERSION[1]}" -ge "10" ]
+if  [ "${SPLIT_VERSION[0]}" == "3" ] && [ "${SPLIT_VERSION[1]}" -ge "11" ]
```
qianz2, line 159: What about python 3.11 version, will it also have ODBC library issue as I'm currently working on it, is there any way we could test if the ODBC library issue will occur or not?

cormacpayne: @qianz2 sorry for the late reply -- I added a new sample app that has the `pyodbc` library as a dependency and can reproduce the error if the platform version in the runtime cannot handle it, so I'm optimistic you should be able to reuse it with your 3.11 changes to verify if the issue will be solved.

qianz2: Thanks, I'll use it for testing on python 3.11

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -182,6 +183,15 @@ public virtual void InstallPlatformSpecificSkeletonDependencies(StringBuilder st
                 .AppendLine($"mkdir -p /home/codespace/.ruby")
                 .AppendLine($"ln -sfn /opt/ruby/{version} /home/codespace/.ruby/current")
                 .AppendLine("fi")
+                .AppendLine($"if [ -f \"$oryxImageDetectorFile\" ] && [ -f \"$oryxOsDetectorFile\" ] && [ \"$platformName\" = \"python\" ] && grep -q \"githubactions\" \"$oryxImageDetectorFile\" && grep -q \"BULLSEYE\" \"$oryxOsDetectorFile\"; then")
```
qianz2, line 186: I'm thinking rather than having these commands in `PlatformInstallerBase.cs` file, would it make sense to put them under the corresponding build or runtime dockerfiles?

cormacpayne: @qianz2 the corresponding runtime images have these changes, so we're good there, but the corresponding bullseye images require them when the user is trying to build a Python 3.7 or 3.8 application, which we won't know until build time, and I didn't want to "bloat" the bullseye image with additional dependencies that may or may not be needed. I don't mind migrating them over if you feel the bullseye image Dockerfile would be a better location for it though.

qianz2 (approved):
Looks good to me!

PR #1523, opened 2022-08-18
Snehapar/1583653 add support for dotnet 6.0.8 and 3.1.28 images and sdks

snehapar9:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1583653


vsts/scripts/publishFilesToAzureStorage.sh
```diff
@@ -41,7 +41,7 @@ uploadFiles() {
             checksum=$(sha256sum $fileToUpload | cut -d " " -f 1)
         fi
         
-        if shouldOverwriteSdk || shouldOverwritePlatformSdk $platform || [[ "$fileToUpload" == defaultVersion* ]]; then
+        if shouldOverwriteSdk || shouldOverwritePlatformSdk $platform || [[ "$fileToUpload" == *defaultVersion*txt ]]; then
```
pauld-msft, line 44: Thanks 🥇

platforms/dotnet/versions/stretch/versionsToBuild.txt
```diff
@@ -36,6 +36,8 @@
 
 3.1.420, b3bdd964182f9edc3c2976541e657fcc43b0eaf9bc97197597c7ecb8b784d79e3efb9e0405c84e1dcb434cf4cd38ddc4af628c5df486c3d7ae8a23e5254796e3,
 
+3.1.422, 690759982b12cce7a06ed22b9311ec3b375b8de8600bd647c0257c866d2f9c99d7c9add4a506f4c6c37ef01db85c0f7862d9ae3de0d11e9bec60958bd1b3b72c,
```
cormacpayne, line 39: Maybe a question for @daniv-msft: do we need to produce new platform binaries built against stretch? I guess Antares may need these binaries as long as they support stretch for building images correct?

cormacpayne (dismissed):
:shipit:

platforms/dotnet/versions/stretch/versionsToBuild.txt
daniv-msft: Unless I'm missing something, I don't believe we need to add any new versions for Stretch as it is now EOL. @qianz2 Would you know if we can remove the /stretch/versionsToBuild.txt file, or if there are any reasons for us to keep it around?

pauld-msft: The version updates in files like [src/BuildScriptGenerator/DotNetCore/DotNetCoreRunTimeVersions.cs](https://github.com/microsoft/Oryx/pull/1523/files#diff-ff5dc1ba7b8c7a8732db8a5c958f5d1db8d40427f2e641282fd5a632c5c1df4c) could raise potential issues. Specifically, if we have any tests that are using stretch-based images and attempt to use those variables to determine the most up to date version we will have a problem if those SDK versions do not exist for stretch as well.

pauld-msft: If we deprecate tests depending on those variables, AND make sure that nowhere in code is using that to determine the version to pull, then I think we will be fine to stop updating stretch with patches

qianz2: @daniv-msft When partner teams use the EOL stretch images with older image tags, it will pull stretch SDKs from our Prod storage account. And since we've already uploaded all stretch SDKs into Prod, it should be fine if we remove stretch/versionsToBuild.txt file from the repo as we no longer need to add or remove stretch based binaries.

qianz2: @pauld-msft Oryx has stretch-based build image tests for the build image since we prebaked lots of stretch SDK into the build image. And all stretch based runtime images and corresponding tests are removed from the repo. 
The variables are usually fetched from `build/constants.yaml `or `_dotnetCoreVersions.sh`, and it pulls directly from storage accounts, so it won't related to the versions under `stretch/versionsToBuild.txt` file.

cormacpayne (dismissed):
:shipit:

qianz2 (dismissed):

cormacpayne (approved):
:shipit: x2

PR #1524, opened 2022-08-18
Pauld/bugfix issues using oryx as base image

pauld-msft:
New DEBIAN_FLAVOR detection strategy for when Oryx is being used as a base image:
1. Try to use DEBIAN_FLAVOR environment variable
2. If DEBIAN_FLAVOR does not exist, parse `/opt/oryx/.ostype` for flavor. This should resolve the bug we are currently seeing.
3. If neither works (which only happens if they use Oryx as base and don't copy `/opt/oryx/` directory to final image), then we hard fail

Follow-up work (less time sensitive / more investigation needs to be done):
1. Create docs around using Oryx as a base image
2. Update error messages to reference those docs


src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -109,8 +109,19 @@ public virtual void InstallPlatformSpecificSkeletonDependencies(StringBuilder st
                 .AppendLine($"cd {versionDirInTemp}")
                 .AppendLine("PLATFORM_BINARY_DOWNLOAD_START=$SECONDS")
                 .AppendLine($"platformName=\"{platformName}\"")
-                .AppendLine("echo \"Detecting image debian flavor: $DEBIAN_FLAVOR.\"")
-                .AppendLine($"if [ \"$DEBIAN_FLAVOR\" == \"{OsTypes.DebianStretch}\" ]; then")
+                .AppendLine($"if [ -z \"$DEBIAN_FLAVOR\" ]; then")
```
cormacpayne, line 112: Just brainstorming here: is there any reason we can't just set `DEBIAN_FLAVOR` equal to the value of `this.CommonOptions.DebianFlavor` here? At this point in the code, `this.CommonOptions.DebianFlavor` will either be set to the `DEBIAN_FLAVOR` environment variable, the value of the `.ostype` file, or will have thrown an exception that terminated the process if neither were set/found. Is it possible for the user to have changed either of these conditions between the time we checked `this.CommonOptions.DebianFlavor` in the `BuildCommand` class and when this build script is ran? Both code snippets are executed _after_ the user runs `oryx build`, so I don't see a way that the behavior could change, but would love to hear what you think because I might be missing something.

tests/Oryx.Tests.Common/TestTempDirTestFixture.cs
```diff
@@ -18,6 +18,8 @@ public TestTempDirTestFixture()
             RootDirPath = Path.Combine(Path.GetTempPath(), "oryxtests", Guid.NewGuid().ToString());
 
             Directory.CreateDirectory(RootDirPath);
+
+            Environment.SetEnvironmentVariable("DEBIAN_FLAVOR", "stretch");
```
cormacpayne, line 22: nit: would you mind leaving a comment that mentions why we're setting this to stretch? (I assume this is just the default value we want to use, but just want the reader to be sure).

tests/Oryx.BuildImage.Tests/BaseImage/BaseImageTest.cs
```diff
@@ -0,0 +1,238 @@
...
+                    NetCoreApp31MvcApp,
+                    imageHelper.GetGitHubActionsAsBaseBuildImage("github-actions-bullseye-base"),
+                    "bullseye");
+
+                return data;
+            }
+        }
+
+        /// <summary>
+        /// This test uses an image that has Oryx githubactions as a base image, but does not provide
```
cormacpayne, line 68: Awesome, I love these as tests 😎

cormacpayne (approved):
:shipit:

tests/Oryx.Tests.Common/TestTempDirTestFixture.cs
pauld-msft: Yeah the flavor here doesn't really matter, we just need the CLI to find _some_ debian flavor so it doesn't fail. Could also have provided an .ostype file for it to parse. I'll add a comment for that

src/BuildScriptGenerator/PlatformInstallerBase.cs
pauld-msft: Yeah this is definitely correct! It should not be possible to get to this stage and not have the CommonOptions.DebianFlavor populated. It's a good point to just simplify this logic and set it in this script from the get-go.

cormacpayne (approved):
:shipit:

PR #1492, opened 2022-07-26
Pauld/1568462 cut down skeleton/cli image

pauld-msft:
This PR is an amalgamation of @cormacpayne 's work to: 
- determine the dependencies for each language
- pull the required dependencies for each language into the CLI Dockerfile

And my work to:
- pull the required dependencies for each language from the CLI into the build script

This work, altogether, allows the CLI image that is ~288MBs to install required dependencies for each language at build time.

-----
### Details
Baking the dependencies into the Dockerfile resulted in:
- cli-buster image size: 979.96 MBs [Image Layers](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1568462#3182437)
- cli image size: 866.87 MB [Image Layers](https://user-images.githubusercontent.com/107068277/181095691-c2e6ddf7-50f8-4755-9f5d-5632d5c01086.png)

Moving the logic into the build script resulted in:
- cli-buster image size: 317.14 MBs [Image Layers](https://user-images.githubusercontent.com/107068277/182624337-11b92881-77c3-4d0c-80d2-f55ea709a3cd.png)
- cli image size: 296.76 MBs [Image Layers](https://user-images.githubusercontent.com/107068277/182624217-6ffe1718-a771-4562-be3e-9efe54890649.png)

#### Testing with Sample Apps
[cli](https://github.com/pauld-msft/acr-test/actions/runs/2825331993)
[cli-buster](https://github.com/pauld-msft/acr-test/actions/runs/2825332460)

PR #1528, opened 2022-08-23
Pauld/reduce cli test time

pauld-msft:
The CLI tests introduced in https://github.com/microsoft/Oryx/pull/1492 have been taking about 20-30 min longer than the next longest category, increasing our validation and nightly tests by that length. This splits up those cli tests into 2 categories for the 2 different base OSs, which allows for parallelization


cormacpayne (approved):
:shipit:

PR #1529, opened 2022-08-23
Pauld/1581775 include ostype in image tag

pauld-msft:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1581775
Updates the images to have tags like:
image


qianz2:
@pauld-msft Instead of putting OS type, image type and release tag all together in the image tag suffix, a suggestion would make it looks cleaner if we move OS type to the prefix like this:
```
mcr.microsoft.com/oryx/build/debian-buster:github-actions-20220818.2
mcr.microsoft.com/oryx/build/ubuntu-focal:lts-versions-20220818.2
mcr.microsoft.com/oryx/dotnetcore/debian-buster:3.1-20220818.2
```

pauld-msft:
> @pauld-msft Instead of putting OS type, image type and release tag all together in the image tag suffix, a suggestion would make it looks cleaner if we move OS type to the prefix like this:
> 
> ```
> mcr.microsoft.com/oryx/build/debian-buster:github-actions-20220818.2
> mcr.microsoft.com/oryx/build/ubuntu-focal:lts-versions-20220818.2
> mcr.microsoft.com/oryx/dotnetcore/debian-buster:3.1-20220818.2
> ```

@qianz2 thanks for the suggestion. I made these changes with this format, as that what was proposed in [the design document](https://microsoft-my.sharepoint.com/:w:/p/pauldorsch/EYwkPYc-SfNAnblq83_lOAMBEboPRZA2Ex1GPumboApWlg?e=uti4ie) that I outlined for this work. We iterated on the naming additionally in the `[Antares] Include OS type in Oryx image tag` email thread to add the os type + os version.

I think that one of the main issues that I see with having it as the prefix is that we would need to completely re-organize our [defined MCR repo structure](https://github.com/microsoft/mcr/tree/main/teams/oryx) to release these new image names to the public. Instead of releasing this change as a new 'version', like the tag indicates, we would release a whole new image.

I will say that this is a separate conversation that @daniv-msft and I had offline, particularly concerning why cli and cli-buster have their own repos, while the rest of the build images didn't. AFAIK, docker image tagging should coorespond to the image _version_ which would almost lead me to think we should have the github-actions images separate from the lts images and so on? From [docker best practices](https://docs.docker.com/develop/dev-best-practices/):
`When building images, always tag them with useful tags which codify version information, intended destination (prod or test, for instance), stability, or other information that is useful when deploying the application in different environments
`

I think that down the line we could definitely look at the structure of our images and re-organize them, but I believe that creating new docker repositories is out of scope for this particular work item. Happy to discuss!!

build/__variables.sh
```diff
@@ -74,7 +74,7 @@ declare -r ACR_AZURE_FUNCTIONS_JAMSTACK_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:azfun
 declare -r ACR_BUILD_FULL_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:full"
 declare -r ACR_BUILD_LTS_VERSIONS_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:lts-versions"
 declare -r ACR_BUILD_GITHUB_ACTIONS_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:github-actions"
-declare -r ACR_BUILD_VSO_FOCAL_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:vso-focal"
+declare -r ACR_BUILD_VSO_FOCAL_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:ubuntu-vso-focal"
```
cormacpayne, line 77: Should the format of the tag here be `vso-ubuntu-focal`? From the changes made in the `buildBuildImages.sh` script, it looked like we were doing `<type>-<OS>-<flavor>`, but correct me if I'm wrong.

build/buildBuildImages.sh
```diff
@@ -310,11 +307,15 @@ function buildLtsVersionsImage() {
...
+	local debianFlavor=$1
+	if [ -z "$debianFlavor" ] ; then
+		debianFlavor="stretch"
+	fi
+	buildLtsVersionsImage $debianFlavor
 
 	echo
 	echo "-------------Creating latest build images-------------------"
-	local builtImageName="$ACR_BUILD_IMAGES_REPO"
+	local builtImageName="$ACR_BUILD_IMAGES_REPO:debian-$debianFlavor"
```
cormacpayne, line 318: I'm wondering if we need to think about how to handle the `latest` image since it's currently always built using `stretch`, and this would tag the image as `debian-stretch` instead of `latest` or `latest-debian-stretch`, correct?

build/buildBuildImages.sh
```diff
@@ -361,28 +362,26 @@ function buildVsoFocalImage() {
 	echo "$builtImageName image history"
 	docker history $builtImageName
 
-	docker tag $builtImageName "$DEVBOX_BUILD_IMAGES_REPO:vso-focal"
-
-	echo
-	echo "$builtImageName" >> $ACR_BUILD_IMAGES_ARTIFACTS_FILE
+	docker tag $builtImageName "$DEVBOX_BUILD_IMAGES_REPO:ubuntu-vso-focal"
```
cormacpayne, line 365: Same comment here about if the format of the tag is correct.

build/buildBuildImages.sh
```diff
@@ -361,28 +362,26 @@ function buildVsoFocalImage() {
...
-	elif  [ "$debianFlavor" == "buster" ]; then
-		debianFlavor="buster"
-		devImageTag=$devImageTag-$debianFlavor
-		echo "dev image tag: "$devImageTag
-		builtImageName=$builtImageName-$debianFlavor
-		echo "built image name: "$builtImageName
+		builtImageName="$builtImageName:debian-$debianFlavor"
+	else
+		builtImageName="$builtImageName-$debianFlavor:debian-$debianFlavor"
+		devImageRepo="$DEVBOX_CLI_BUILD_IMAGE_REPO-buster"
```
cormacpayne, line 381: nit: separate from your work but would be nice to change the hard-coded `buster` here to `$debianFlavor` just to be safe for future flavor adoptions in the CLI image.

images/runtime/php/8.1/base.bullseye.Dockerfile
```diff
@@ -1,11 +1,9 @@
-# DisableDockerDetector "Below image not yet supported in the Docker Hub mirror"
-FROM php:8.1-bullseye
+FROM oryxdevmcr.azurecr.io/private/oryx/php-8.1
```
cormacpayne, line 1: Is this the image we want to use here? It was previously being pulled from Docker Hub, but hadn't been migrated to the [approved MCR mirror](https://eng.ms/docs/more/containers-secure-supply-chain/approved-images) yet

images/runtime/ruby/2.5/buster.Dockerfile
```diff
@@ -20,9 +20,10 @@ ENV RUBY_VERSION 2.5.8
 RUN ${IMAGES_DIR}/installPlatform.sh ruby $RUBY_VERSION --dir /opt/ruby/$RUBY_VERSION --links false
 RUN set -ex \
  && cd /opt/ruby/ \
- && ln -s 2.5.8 2.5
+ && ln -s 2.5.8 2.5 \
+ && ln -s 2.5 2
 
-ENV PATH="/opt/ruby/2.5/bin:${PATH}"
+ENV PATH="/opt/ruby/2/bin:${PATH}"
```
cormacpayne, line 26: nit: just curious about the need for these changes in the Ruby runtimes.

cormacpayne (commented):
Just a few clarifying questions before signing-off

images/runtime/ruby/2.5/buster.Dockerfile
pauld-msft: These changes were added automatically when I [fixed the generateDockerfiles script for ruby](https://github.com/microsoft/Oryx/pull/1529/files#diff-f379ee5e93788de18e787b3f0dea9986fc1c78276ff03ebf7d73188a470b9eb5) in order to test my changes. Turns out that the generate dockerfiles script wasn't actually going over the versions directories due to a small typo.

build/buildBuildImages.sh
pauld-msft: Yes thats right. I was wondering about that particular case as well. Even though we call it the latest image, I think our partners were just using `build:<releasetag>` (https://mcr.microsoft.com/v2/oryx/build/tags/list). `latest` was never a part of the image tag, so I decided not to include it here. I could really go either way on this naming schema though

build/__variables.sh
pauld-msft: Yeah great point. I'm not sure why but I guess I just assumed the flavor was `vso-focal` even though its `focal-scm` in our code. I'll fix

tests/Oryx.BuildImage.Tests/BenvTest.cs
```diff
@@ -66,7 +66,7 @@ public void PipelineTestInvocationVsoFocal()
         {
             var imageTestHelper = new ImageTestHelper();
             BuildImagesHaveOryxPathsEnvironmentVariableAvailable(
-                imageTestHelper.GetVsoBuildImage("vso-focal"));
+                imageTestHelper.GetVsoBuildImage(ImageTestHelperConstants.VsoUbuntu));
```
qianz2, line 69: consider renaming `ImageTestHelperConstants.VsoUbuntu` -> `ImageTestHelperConstants.VsoFocal`, which is consistent with `ImageTestHelperConstants.LtsVersionsStretch`

vsts/scripts/pullAndTag.sh
```diff
@@ -39,19 +39,19 @@ function tagBuildImageForIntegrationTest() {
...
+tagBuildImageForIntegrationTest "$imagefilter/build" "debian-stretch"
+tagBuildImageForIntegrationTest "$imagefilter/build" "lts-versions-debian-stretch" 
+tagBuildImageForIntegrationTest "$imagefilter/build" "lts-versions-debian-buster"
+tagBuildImageForIntegrationTest "$imagefilter/build" "azfunc-jamstack-debian-stretch"
+tagBuildImageForIntegrationTest "$imagefilter/build" "azfunc-jamstack-debian-buster"
+tagBuildImageForIntegrationTest "$imagefilter/build" "azfunc-jamstack-debian-bullseye" 
+tagBuildImageForIntegrationTest "$imagefilter/build" "github-actions-debian-stretch"
+tagBuildImageForIntegrationTest "$imagefilter/build" "github-actions-debian-buster"
+tagBuildImageForIntegrationTest "$imagefilter/build" "github-actions-debian-bullseye"
+tagBuildImageForIntegrationTest "$imagefilter/build" "ubuntu-vso-focal"
```
qianz2, line 51: `"ubuntu-vso-focal"` -> `"vso-ubuntu-focal"`

vsts/scripts/tagBuildImagesForRelease.sh
```diff
@@ -45,17 +45,17 @@ function tagBuildImage() {
...
+tagBuildImage "oryxdevmcr.azurecr.io/public/oryx/build:azfunc-jamstack-debian-bullseye-$BUILD_DEFINITIONNAME.$RELEASE_TAG_NAME" "azfunc-jamstack-debian-bullseye" "azfunc-jamstack-debian-bullseye-$RELEASE_TAG_NAME"
+tagBuildImage "oryxdevmcr.azurecr.io/public/oryx/build:github-actions-debian-stretch-$BUILD_DEFINITIONNAME.$RELEASE_TAG_NAME" "github-actions-debian-stretch" "github-actions-debian-stretch-$RELEASE_TAG_NAME"
+tagBuildImage "oryxdevmcr.azurecr.io/public/oryx/build:github-actions-debian-buster-$BUILD_DEFINITIONNAME.$RELEASE_TAG_NAME" "github-actions-debian-buster" "github-actions-debian-buster-$RELEASE_TAG_NAME"
+tagBuildImage "oryxdevmcr.azurecr.io/public/oryx/build:github-actions-debian-bullseye-$BUILD_DEFINITIONNAME.$RELEASE_TAG_NAME" "github-actions-debian-bullseye" "github-actions-debian-bullseye-$RELEASE_TAG_NAME"
+tagBuildImage "oryxdevmcr.azurecr.io/public/oryx/build:ubuntu-vso-focal-$BUILD_DEFINITIONNAME.$RELEASE_TAG_NAME" "ubuntu-vso-focal" "ubuntu-vso-focal-$RELEASE_TAG_NAME"
```
qianz2, line 57: same here:)

tests/Oryx.Integration.Tests/Python/PythonSqlServerIntegrationTests.cs
```diff
@@ -27,9 +27,9 @@ public PythonSqlServerIntegrationTests(ITestOutputHelper output) : base(output,
         }
 
         [Theory(Skip = "Bug #1274414")]
-        [InlineData("github-actions")]
-        [InlineData("github-actions-buster")]
-        [InlineData("latest")]
+        [InlineData(ImageTestHelperConstants.GitHubActionsStretch)]
+        [InlineData(ImageTestHelperConstants.GitHubActionsBuster)]
+        [InlineData(ImageTestHelperConstants.LatestTag)]
```
qianz2, line 32: nit: looks that we tagged `latest `image to `debian-stretch`, we can change `ImageTestHelperConstants.LatestTag` to `ImageTestHelperConstants.DebianStretch` or `ImageTestHelperConstants.LatestDebianStretch`?

images/runtime/php/8.1/base.bullseye.Dockerfile
```diff
@@ -1,11 +1,9 @@
...
 SHELL ["/bin/bash", "-c"]
 ENV PHP_VERSION 8.1.6
 
 RUN a2enmod rewrite expires include deflate remoteip headers
-#Bake in client certificate path into image to avoid downloading it
-ENV PATH_CA_CERTIFICATE="/etc/ssl/certs/ca-certificate.crt"
```
qianz2, line 8: Should we keep this line?

pauld-msft: yeah I think I'm going to revert this file to its previous state

build/buildBuildImages.sh
```diff
@@ -175,16 +178,13 @@ function buildGitHubActionsImage() {
...
 		debianFlavor="stretch"
-		echo "Debian Flavor is: "$debianFlavor
-	elif [ "$debianFlavor" == "buster" ] || [ "$debianFlavor" == "bullseye" ]; then
-		debianFlavor=$debianFlavor
-		devImageTag=$devImageTag-$debianFlavor
-		echo "dev image tag: "$devImageTag
-		builtImageName=$builtImageName-$debianFlavor
-		echo "built image name: "$builtImageName
 	fi
+	devImageTag=$devImageTag-debian-$debianFlavor
```
qianz2, line 184: I'm wondering if it's good place to hardcode `debian` here, as we also have `ubuntu` OS type, so when $debianFlavor is `focal-scm`, it would tag wrongly?
https://github.com/microsoft/Oryx/blob/0405a71cca5640a8d2bb41fa7a234271989b09dc/src/BuildScriptGenerator.Common/OsTypes.cs#L10
Another place I think we'll probably address is here in vso focal dockerfile we have `ENV DEBIAN_FLAVOR="focal-scm"` ->` ENV UBUNTU_FLAVOR="focal-scm"` or simply rename all `DEBIAN_FLAVOR` to `OS_FLAVOR`? 
https://github.com/microsoft/Oryx/blob/0405a71cca5640a8d2bb41fa7a234271989b09dc/images/build/Dockerfiles/vso.focal.Dockerfile#L72

build/buildBuildImages.sh
cormacpayne: I could go either way as well as long as it's communicated to partners that this specific tag is changing "drastically" relative to other tags that are just getting the OS added in the tag.

images/runtime/ruby/2.5/buster.Dockerfile
cormacpayne: Got it -- if there were changes added as a result of running a script in our repository, then I think we _should_ be safe to keep them in this PR.

build/buildBuildImages.sh
pauld-msft: Are we planning on releasing an ubuntu github actions image? Since this is taking debian flavor as an input, I think that it is ok to keep here hardcoded as debian. I only say this since it seems to fall under a larger effort that you mentioned, and we hardcode "Debian" throughout our repository.

But yeah its a super good point that we are essentially using DEBIAN_FLAVOR as a synonym to OS_TYPE in certain places in our code which creates confusion. It would be valuable to the longevity of our code to rename debian flavor to os type, but I think that we should have a ticket devoted solely to that work. It would likely be good to do before the introduction of the Mariner OS type?

build/buildBuildImages.sh
pauld-msft: Ok, take a peek at this commit: https://github.com/microsoft/Oryx/pull/1529/commits/721b6d43c92256978468dd96797512d4cdd6218b

I've added some manual tagging of the latest image with no tag, which adds `build`, `build:latest`, and `build:<releaseTag>` methods of pulling the latest image. I think that this was a good point since this allows us to intentionally set exactly what the "latest" tag means and it stays up to date. We also don't have to update the various places in our docs where we have tutorials for people having them pull the "latest" image.

Can definitely revert the commit if we decide that going with the original strategy is better

cormacpayne: This approach makes sense to me 👍

cormacpayne (approved):
:shipit:

PR #1533, opened 2022-08-26
Snehapar/1588872 upgrade golang version to 1.19 and change to bullseye

snehapar9:
[User Story 1588872](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1588872): [Due 2022-08-31] Upgrade Golang version to 1.19 latest in runtime images


tests/Oryx.Integration.Tests/Golang/GolangBuildRunTests.cs
```diff
@@ -22,28 +22,36 @@ public GolangBuildRunTests(ITestOutputHelper output, TestTempDirTestFixture test
...
                 .AddCommand(
-                $"oryx run-script {appOutputDir} --output {DefaultStartupFilePath} --debug")
+                $"oryx run-script --platform {GolangConstants.PlatformName} --platform-version {golangVersion} {appOutputDir}  --output {DefaultStartupFilePath} --debug")
                 .AddCommand(DefaultStartupFilePath)
                 .ToString();
 
             // Assert
             await EndToEndTestHelper.RunAndAssertAppAsync(
-                imageName: "oryxdevmcr.azurecr.io/public/oryx/build:full",
+                imageName: $"oryxdevmcr.azurecr.io/public/oryx/build:full-{debianFlavor}",
```
cormacpayne, line 54: Would you mind replacing this hard-coded image name with a call to a new method in the `ImageTestHelper` class that will construct the `full` image name?

cormacpayne (dismissed):
One minor comment, otherwise LGTM :shipit:

tests/Oryx.Integration.Tests/Golang/GolangBuildRunTests.cs
```diff
@@ -22,28 +22,36 @@ public GolangBuildRunTests(ITestOutputHelper output, TestTempDirTestFixture test
...
+
+
+        [Theory]
+        [InlineData("1.17", "buster")]
+        [InlineData("1.17", "bullseye")]
+        [InlineData("1.18", "buster")]
+        [InlineData("1.18", "bullseye")]
+        [InlineData("1.19", "buster")]
+        [InlineData("1.19", "bullseye")]
+        public async Task CanRunApp_WithoutBuildManifestFileAsync(string golangVersion, string debianFlavor)
```
william-msft, line 34: Very nice! 👍

william-msft (dismissed):

tests/Oryx.Tests.Common/ImageTestHelper.cs
```diff
@@ -346,5 +352,10 @@ private string GetTestTag()
 
             return _tagSuffix;
         }
+
+        private string GetFullBuildImage()
```
cormacpayne, line 356: nit: since we have to account for `full`, `full-buster` and `full-bullseye`, I think it would be best if we followed the same model that the above `GetGitHubActionsBuildImage` method uses to handle these different Debian flavors (_i.e._, provide a `debianFlavor` argument that we change the returned image name based on).

Also, we will need to account for the `_tagSuffix` at the end of the image name for tests in our pipeline that append the date-time to the end.

cormacpayne (commented):
One last comment 🙂

qianz2 (approved):

PR #1534, opened 2022-08-26
Pauld/1597308 parameterize pipelines

pauld-msft:
Proposal 1 from https://microsoft-my.sharepoint.com/:w:/p/pauldorsch/EYFJkV2JFOtHrfmklKs2r88BZls8mNkHwIhBa5TwNulOfQ?e=7o3JN4

- parameterizes the platform binaries, copyFromProd, and validation/nightly pipelines to allow users to pass in a storage account
- adds an ORYX_TEST_SDK_STORAGE_URL environment variable that controls the storage account used for testing (DEV by default)


doc/supportedPlatformVersions.md
```diff
@@ -16,6 +16,25 @@
 - 3.1.419
 - 3.1.420
 - 3.1.422
+- 6.0.100-preview.1.21103.13
```
pauld-msft, line 19: unrelated to this change, were added when `generateConstants.sh` was run

build/constants.yaml
```diff
@@ -216,6 +216,7 @@
     bash: /bin/bash
     benv: /opt/oryx/benv
     pre-run-command-env-var-name: PRE_RUN_COMMAND
+    app-svc-file-name: appsvc.yaml
```
pauld-msft, line 219: `generateConstants.sh` was overwriting a go file that removed this constant that is needed for the startupscriptgenerator

vsts/pipelines/ci.yml
```diff
@@ -1,4 +1,10 @@
 parameters:
+  - name: storageAccountUrl
+    displayName: Storage Account Url
+    type: string
+    default: https://oryx-cdn.microsoft.io
+    values:
+    - https://oryx-cdn.microsoft.io
```
pauld-msft, line 7: Currently means that the Oryx-CI job will _only_ be able to use the prod CDN...

This comes back to the question of whether or not we want to be pre-baking with the dev or prod storage accounts

build/testBuildImages.sh
```diff
@@ -12,6 +12,13 @@ declare -r testProjectName="Oryx.BuildImage.Tests"
 
 # Load all variables
 source $REPO_DIR/build/__variables.sh
+source $REPO_DIR/build/__sdkStorageConstants.sh
+
+if [ -z "$ORYX_TEST_SDK_STORAGE_URL" ]; then
+    echo
+    echo "Setting environment variable 'ORYX_TEST_SDK_STORAGE_URL' to default value '$DEV_SDK_STORAGE_BASE_URL'."
```
cormacpayne, line 19: nit: since this message is repeated in a few spots, it would be nice for clarity to add in _which_ test suite this environment variable is being set for.

"Setting environment variable 'ORYX_TEST_SDK_STORAGE_URL' to default value '$DEV_SDK_STORAGE_BASE_URL' _for build image tests_."

vsts/pipelines/ci.yml
pauld-msft: If we do not want this behavior, I will need to add in options to specify a pre-bake storage account and a dynamic installation storage account, as the 2 will need to be able to be different.

src/BuildScriptGenerator.Common/ShellScriptBuilder.cs
```diff
@@ -158,11 +158,19 @@ public ShellScriptBuilder AddStringDoesNotExistInFileCheck(string searchString,
                 "exit 1; fi");
         }
 
+        /// <summary>
+        /// Append a command to the shell script that sets the ORYX_SDK_STORAGE_BASE_URL to the value
+        /// of ORYX_TEST_SDK_STORAGE_URL if ORYX_TEST_SDK_STORAGE_URL exists in the environment that is executing this code.
```
cormacpayne, line 163: nit: out of curiosity, any reason we introduce a new `ORYX_TEST_SDK_STORAGE_URL` environment variable rather than just setting the existing `ORYX_SDK_STORAGE_BASE_URL` one in the calling scripts? Is this just to explicitly differentiate what's being set for test execution and what is actually being used within the container?

vsts/pipelines/PlatformBinaries/dotnetcore.yml
```diff
@@ -1,6 +1,12 @@
 variables:
   - group: Oryx
 
+parameters:
+- name: destinationStorageAccount
```
cormacpayne, line 5: nit: it's probably a lot of copy-paste, but I would defer to having this parameter called `destinationStorageAccountName`, just so we have the differentiation between the storage account resource name and the URL of the storage account.

vsts/pipelines/ci.yml
```diff
@@ -1,4 +1,10 @@
 parameters:
+  - name: storageAccountUrl
+    displayName: Storage Account Url
```
cormacpayne, line 3: nit: if we can use this `displayName` property as a "description", it would be nice to include some additional information about how this URL is being used

_e.g._, "Storage Account URL for Production Images and Testing"

vsts/scripts/publishFilesToAzureStorage.sh
```diff
@@ -70,6 +72,18 @@ uploadFiles() {
     done
 }
 
+storageAccountUrl="https://$storageAccount.blob.core.windows.net"
```
cormacpayne, line 75: nit: similar comment as before but specific to this file, would you mind renaming `storageAccount` to `storageAccountName`?

cormacpayne (dismissed):
Just some minor comments, otherwise LGTM :shipit:

src/BuildScriptGenerator.Common/ShellScriptBuilder.cs
pauld-msft: I couldn't think of a better way to essentially get the functionality of passing in a parameter to the tests

The reason that I went with introducing the `ORYX_TEST_SDK_STORAGE_URL` environment variable is so that we do not need to re-generate every image in order to change the storage account url used for testing.

It allows us to use DEV by default, which is what happens now. But if we want to use something other than DEV I had 2 solutions:
- remove the "AddDefaultTestEnvironmentVariables" from our tests and have it use the default that exists in the container (requires re-generation of images to pass in the new `ORYX_SDK_STORAGE_BASE_URL`)
- add a new env var that tests will use to determine the url to use (or default to dev)

cormacpayne: Yep makes sense to me, good call 👍

cormacpayne (approved):
:shipit:

pauld-msft:
Pulled main into here so I can verify that we can run nightly and validation against the production storage account without issues now that pre-baking will use the prod storage account....

PR #1536, opened 2022-08-26
update dotnet-monitor version

waliMSFT:
Update .NET 6 and .NET Core 3.1 so that they include dotnet-monitor 6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -264,6 +264,9 @@ images/runtime/dotnetcore/*/Dockerfile
 images/runtime/dotnetcore/*/buster.Dockerfile
 images/runtime/dotnetcore/*/stretch.Dockerfile
 images/runtime/python/*/Dockerfile
+images/runtime/python/*/buster.Dockerfile
+images/runtime/python/*/stretch.Dockerfile
+images/runtime/python/*/bullseye.Dockerfile
 images/runtime/php/*/Dockerfile
 images/runtime/php/*/buster.Dockerfile
 images/runtime/php/*/stretch.Dockerfile
@@ -282,7 +285,3 @@ src/startupscriptgenerator/src/**/__debug_bin
 src/startupscriptgenerator/src/**/run.sh
 !**/dynamic/Dockerfile
 !**/dynamic/buster.Dockerfile
-/images/runtime/python/3.10/bullseye.Dockerfile
-images/runtime/python/*/bullseye.Dockerfile
-/images/runtime/python/3.7/bullseye.Dockerfile
-/images/runtime/python/3.8/bullseye.Dockerfile
--- a/.gitignore
+++ b/.gitignore
@@ -284,3 +284,6 @@ src/startupscriptgenerator/src/**/__debug_bin
 src/startupscriptgenerator/src/**/run.sh
 !**/dynamic/Dockerfile
 !**/dynamic/buster.Dockerfile
+/images/runtime/python/3.10/bullseye.Dockerfile
+/images/runtime/python/3.7/bullseye.Dockerfile
+/images/runtime/python/3.8/bullseye.Dockerfile
--- a/.gitignore
+++ b/.gitignore
@@ -264,8 +264,6 @@ images/runtime/dotnetcore/*/Dockerfile
 images/runtime/dotnetcore/*/buster.Dockerfile
 images/runtime/dotnetcore/*/stretch.Dockerfile
 images/runtime/python/*/Dockerfile
-images/runtime/python/*/buster.Dockerfile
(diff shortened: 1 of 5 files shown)


images/runtime/dotnetcore/3.1/base.bullseye.Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-trace --version 5.0
 RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-dump --version 5.0.236902
 RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-counters --version 5.0.236902
 RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-gcdump --version 5.0.236902
-RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-monitor --version 6.1.*
+RUN dotnet tool install --tool-path /dotnetcore-tools dotnet-monitor --version 6.*
```
pauld-msft, line 9: I don't have access to the email chain, but just want to confirm that we do not also want to make this change for [.NET 5](https://github.com/microsoft/Oryx/blob/main/images/runtime/dotnetcore/5.0/base.buster.Dockerfile#L9) as well?

pauld-msft (dismissed):

images/runtime/dotnetcore/3.1/base.bullseye.Dockerfile
pauld-msft: also, :shipit:

.gitignore
```diff
@@ -284,3 +284,6 @@ src/startupscriptgenerator/src/**/__debug_bin
 src/startupscriptgenerator/src/**/run.sh
 !**/dynamic/Dockerfile
 !**/dynamic/buster.Dockerfile
+/images/runtime/python/3.10/bullseye.Dockerfile
```
cormacpayne, line 287: There are two lines above in this file that have the following:

```
images/runtime/python/*/buster.Dockerfile
images/runtime/python/*/stretch.Dockerfile
```

Let's delete the lines you've added here and add the following line underneath these two:

```
images/runtime/python/*/bullseye.Dockerfile
```

images/runtime/dotnetcore/3.1/base.bullseye.Dockerfile
cormacpayne: It looks like from the [work item](https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1582251) that it's just .NET 6 and .NET Core 3.1, but good question if we should update .NET 5 as well 🤔

cormacpayne (changes requested):
Would you mind adding some testing to the existing `DotnetMonitorTool_IsPresentInTheImage` test in the `DotNetCoreLibraryVerificationTest` class?

images/runtime/dotnetcore/3.1/base.bullseye.Dockerfile
william-msft: Looks like .NET 5 is EOL: 
image

waliMSFT: I ignored .net 5 because of two reasons:
1) .net 5 was not mentioned anywhere in the story and the email thread.
2) .net 5 is not supported anymore.

.gitignore
```diff
@@ -264,8 +264,6 @@ images/runtime/dotnetcore/*/Dockerfile
 images/runtime/dotnetcore/*/buster.Dockerfile
 images/runtime/dotnetcore/*/stretch.Dockerfile
 images/runtime/python/*/Dockerfile
-images/runtime/python/*/buster.Dockerfile
```
cormacpayne, line 267: These lines shouldn't be removed -- my original comment was intended for you to remove the changes you originally made, and add the following line underneath these lines:

```
images/runtime/python/*/bullseye.Dockerfile
```

The only change in this `.gitignore` file should be the addition of this `bullseye.Dockerfile` line I mentioned above to be right underneath line 268 (which is the `stretch.Dockerfile` line for `python`)

cormacpayne (commented):
`.gitignore` needs to be cleaned up, so I can approve once that's taken care of

.gitignore
waliMSFT: yah, this file was committed unintentionally. Now, I have fixed this as you suggested.

cormacpayne (approved):

william-msft (approved):

PR #1539, opened 2022-08-29
Final stretch version constants for dotnet 3.1 and 6.0

pauld-msft:
In order to stop uploading patch SDK versions for stretch, we need to lock the stretch images into using the last available version when pre-baking.

Our CLI is smart enough to figure out the versions available in the storage account, so that should not need these changes.


build/__finalStretchVersions.sh
```diff
@@ -0,0 +1,8 @@
+# This file was auto-generated from 'constants.yaml'. Changes may be overridden.
+
+FINAL_STRETCH_DOT_NET_CORE_31_SDK_VERSION='3.1.420'
```
cormacpayne, line 3: Any reason we only need to consider the final stretch version for .NET and no other platforms? Is it because we aren't pre-baking any version of other platforms that are currently supported anymore? (_e.g._, Python 3.7 and 3.8 in the `lts-versions` build image)

cormacpayne (approved):
:shipit:

build/__finalStretchVersions.sh
pauld-msft: Oh yeah some much needed context is that there is [a ticket](https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1599477) coming down the line which @waliMSFT will be working on to do this for any other versions / platforms that are needed.

This PR is just a stopgap until then since if we currently use the prod storage account to pre-bake images it will break, and my parameterization pull request switches Oryx CI to use prod to pre-bake.

Once this PR is merged, we should be free to remove the dotnet 3.1.422 and 6.0.400 stretch versions from our dev storage account, reconciling it with what exists in our versionsToBuild.txt file

cormacpayne: Cool, sounds good :shipit:

src/BuildScriptGenerator/FinalStretchVersions.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Microsoft.Oryx.BuildScriptGenerator.Common
+{
+    public static class FinalStretchVersions
+    {
+        public const string FinalStretchDotNetCore31SdkVersion = "3.1.420";
+        public const string FinalStretchNetCoreApp31RunTimeVersion = "3.1.26";
+        public const string FinalStretchAspnetCoreApp31RunTimeVersion = "3.1.26";
+        public const string FinalStretchDotNet60SdkVersion = "6.0.301";
+        public const string FinalStretchNetCoreApp60RunTimeVersion = "6.0.6";
```
daniv-msft, line 11: Nit: should this include a "Dot"? :)

daniv-msft: Similar comment above of course

daniv-msft (dismissed):

src/BuildScriptGenerator/FinalStretchVersions.cs
pauld-msft: I tried to maintain the naming to be consistent with what existed in the constants.yaml, e.g. 
image


but yeah I like this more and I need to make some changes anyways to this PR, so I'll update

daniv-msft: OK thanks, I didn't realize we used this somewhere else already. Likely something we forgot at the time?
Thanks for updating the name!

daniv-msft (approved):

PR #1540, opened 2022-08-29
Publish sdks to prod improvements - approval required

pauld-msft:
You all likely received an email with "Run stage approval pending" in the subject. This is the pull request that will introduce this functionality to the publish to prod job. This allows us to have a manual approval before making SDKs available to production (since hopefully everything before this step will be automated eventually)

I manually denied the non-dry run job from running in this case.
image

### DIFF
--- a/vsts/pipelines/publishSdkToProd.yml
+++ b/vsts/pipelines/publishSdkToProd.yml
@@ -71,6 +71,7 @@ stages:
               inputs:
                 scriptPath: ./vsts/scripts/publishSdksFromDevToProdStorageAccount.sh
                 args: False # no dry run
+
             - task: ShellScript@2
               displayName: 'Test prod storage account'
               inputs:
--- a/vsts/pipelines/publishSdkToProd.yml
+++ b/vsts/pipelines/publishSdkToProd.yml
@@ -2,9 +2,9 @@ variables:
 - group: Oryx
 
 stages:
-  - stage: Component Detection
+  - stage: ComponentDetection
     jobs:
-    - job: Component Detection
+    - job: ComponentDetection
       pool:
         name: AzurePipelines-EO
         demands:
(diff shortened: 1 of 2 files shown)


vsts/pipelines/publishSdkToProd.yml
```diff
@@ -16,21 +18,62 @@ jobs:
...
-        inputs:
-          scriptPath: ./build/testIntegration.sh
-          args: StorageAccountTests=Prod
\ No newline at end of file
+          args: True # dry run
+
+  - stage: PublishSDKsToProduction
+    jobs:
+    - deployment: PublishSDKsToProduction
+      environment: oryx-prod
```
cormacpayne, line 49: nit: it would be nice to include a comment on this line to call out that the `environment` property is needed to require approvals for this stage to run

cormacpayne (dismissed):
:shipit:

daniv-msft:
Looks good to me but, out of curiosity, could you please clarify where are defined the fact that we'll wait for an approver before pushing? Where is the list of approvers defined? I couldn't find it in the code change, but maybe I missed it. Or is it set up in the pipeline instead?

daniv-msft (approved):

vsts/scripts/publishSdksFromDevToProdStorageAccount.sh
```diff
@@ -4,13 +4,19 @@
 # Licensed under the MIT license.
 # --------------------------------------------------------------------------------------------
 
-set -ex
+set -e
```
william-msft, line 7: Curious why we're removing the `x`? The trace sometimes helps with debugging

pauld-msft: @william-msft I removed the x since in my experience with these jobs, the extra printing creates some noise in the job. Its a lot easier to review what is going to change during the dry-run (which we will be looking at before manually approving), and in this script we are quite verbose with our printing, to the point where we can quickly determine what step/blob it may have failed on.

william-msft: Sounds good

william-msft (approved):

PR #1541, opened 2022-08-30
Pauld/add personal storage account - improved

pauld-msft:
These changes allow a personal storage account to be added without ANY code changes. Instead, the secret itself just needs to have an enforced naming convention of `<storageAccountName>-PERSONAL-STORAGE-SAS-TOKEN`. Thanks to @cormacpayne for the suggestion!

I've also updated https://microsoft-my.sharepoint.com/:w:/p/pauldorsch/ESoImW2tpFdFk5GKJ3q1_98B3g-fFko6Ly0wi4THkvL54g?e=am4PcZ to reflect this change.


vsts/pipelines/copySdksFromProdToStorageAccount.yml
```diff
@@ -37,9 +37,10 @@ jobs:
           DEV_STORAGE_SAS_TOKEN: $(DEV-STORAGE-SAS-TOKEN)
           SANDBOX_STORAGE_SAS_TOKEN: $(SANDBOX-STORAGE-SAS-TOKEN)
           PROD_STORAGE_SAS_TOKEN: $(PROD-STORAGE-SAS-TOKEN)
+          PERSONAL_STORAGE_SAS_TOKEN: $(${{ parameters.destinationStorageAccountName }}-PERSONAL-STORAGE-SAS-TOKEN)
```
cormacpayne, line 40: nit: I'd be in favor of making sure the entire variable is capitalized independent of what we provide to the pipeline by using the [`upper` function](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#upper):

```
$(${{ upper(parameters.destinationStorageAccountName) }}-PERSONAL-STORAGE-SAS-TOKEN )
```

I'd make the same comment in all instances of `PERSONAL_STORAGE_SAS_TOKEN` in this PR.

vsts/scripts/copySdksFromProdToStorageAccount.sh
```diff
@@ -72,15 +72,18 @@ if [ ! -f "$azCopyDir/azcopy" ]; then
...
+destinationSdkUrl="https://$1.blob.core.windows.net"
 sasToken=""
 
-if [ "$1" == $SANDBOX_SDK_STORAGE_BASE_URL ]; then
+if [ "$destinationSdkUrl" == $SANDBOX_SDK_STORAGE_BASE_URL ]; then
     sasToken=$SANDBOX_STORAGE_SAS_TOKEN
-elif [ "$1" == $DEV_SDK_STORAGE_BASE_URL ]; then
+elif [ "$destinationSdkUrl" == $DEV_SDK_STORAGE_BASE_URL ]; then
     sasToken=$DEV_STORAGE_SAS_TOKEN
+# check if the peronal sas token has been found in the oryx key vault
```
cormacpayne, line 82: nit, typo: "peronal" --> "personal"

vsts/scripts/copySdksFromProdToStorageAccount.sh
```diff
@@ -72,15 +72,18 @@ if [ ! -f "$azCopyDir/azcopy" ]; then
...
+elif [ "$PERSONAL_STORAGE_SAS_TOKEN" != "\$($1-PERSONAL-STORAGE-SAS-TOKEN)" ]; then
```
cormacpayne, line 83: Would you mind clarifying what we're comparing here with these two values?

cormacpayne (approved):
A few clarifying comments, otherwise :shipit:

vsts/scripts/copySdksFromProdToStorageAccount.sh
pauld-msft: we are checking to see if PERSONAL_STORAGE_SAS_TOKEN was contains a SAS token or not

in the template, we have
`PERSONAL_STORAGE_SAS_TOKEN: $(${{ parameters.destinationSdkStorageAccountName }}-PERSONAL-STORAGE-SAS-TOKEN)`

lets take `oryxsdkspauldorsch` for example. There exists an oryxsdkspauldorsch-PERSONAL-STORAGE-SAS-TOKEN in the keyvault, so PERSONAL_STORAGE_SAS_TOKEN env var will contain the SAS token.

if we try with `oryxsdksfake`, where there is no oryxsdksfake-PERSONAL-STORAGE-SAS-TOKEN in the keyvault, PERSONAL_STORAGE_SAS_TOKEN env var will just equal the literal string value `"$(oryxsdksfake-PERSONAL-STORAGE-SAS-TOKEN)"`

so this elif statement is checking to see if the SAS token was found in the key vault, or if the PERSONAL_STORAGE_SAS_TOKEN env var was set to the literal string value

cormacpayne (approved):
:shipit:

PR #1545, opened 2022-09-01
Concatenated blobs from multiple pages

snehapar9:
[Bug 1601443](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1601443): Make Oryx SDK retrieval resilient to storage account partitioning


build/constants.yaml
```diff
@@ -275,7 +275,7 @@
     default-version-file-prefix: defaultVersion
     default-version-file-type: txt
     versions-to-build-file-name: versionsToBuild.txt
-    container-metadata-url-format: '{0}/{1}?restype=container&comp=list&include=metadata'
+    container-metadata-url-format: '{0}/{1}?restype=container&comp=list&include=metadata&marker={2}'
```
pauld-msft, line 278: When changes are made to the constants.yaml file, generally it's good to run the `./build/generateConstants.sh` script. You can see [here](https://github.com/microsoft/Oryx/blob/main/build/constants.yaml#L314-L315) that this change should also be propagating to our go script, and running the script will ensure that it does!

There are some fixes to the constants.yaml file in the current main branch, so make sure to pull down main and merge its changes into your branch before running the script!

tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
```diff
@@ -198,9 +198,19 @@ private void AssertExpectedListOfBlobs(string platformName, params string[] expe
...
+            var url = string.Format(SdkStorageConstants.ContainerMetadataUrlFormat, _storageUrl, platformName, string.Empty);
             var blobList = _httpClient.GetStringAsync(url).Result;
-            return XDocument.Parse(blobList);
+            XDocument xdoc = XDocument.Parse(blobList);
+            var marker = xdoc.Root.Element("NextMarker").Value;
+            do
+            {
+                url = string.Format(SdkStorageConstants.ContainerMetadataUrlFormat, _storageUrl, platformName, marker);
+                var blobListFromNextMarker = _httpClient.GetStringAsync(url).Result;
+                marker = XDocument.Parse(blobListFromNextMarker).Root.Element("NextMarker").Value;
```
pauld-msft, line 209: nit: we are parsing the string to XDocument twice (here and next line). Let's pull that logic out so we only do it once

build/constants.yaml
pauld-msft: hmmm yes I would think that we should also see [this line](https://github.com/microsoft/Oryx/blob/main/src/startupscriptgenerator/src/common/consts/sdk_storage_constants.go#L16) have this change. Perhaps it is on your local branch and hasn't been commited?

snehapar9: Correct, it was not staged. Thanks for pointing it out!

src/BuildScriptGenerator/SdkStorageVersionProviderBase.cs
```diff
@@ -50,19 +50,26 @@ protected PlatformVersionInfo GetAvailableVersionsFromStorage(string platformNam
...
             var sdkVersionMetadataName = isStretch
                 ? SdkStorageConstants.LegacySdkVersionMetadataName
                 : SdkStorageConstants.SdkVersionMetadataName;
-
+            do
+            {
+                url = string.Format(SdkStorageConstants.ContainerMetadataUrlFormat, sdkStorageBaseUrl, platformName, marker);
+                var blobListFromNextMarker = httpClient.GetStringAsync(url).Result;
+                marker = XDocument.Parse(blobListFromNextMarker).Root.Element("NextMarker").Value;
+                xdoc.Descendants("Blobs").LastOrDefault().AddAfterSelf(XDocument.Parse(blobListFromNextMarker).Descendants("Blobs"));
```
cormacpayne, line 70: nit: would you mind adding a comment above this line describing what's occurring? It may not be obvious that we're adding the collection of blobs from the next page to the `xdoc` object that we iterate over later.

cormacpayne (dismissed):
Small comment along with a +1 to Paul's comment, otherwise :shipit:

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreSdkStorageVersionProvider.cs
```diff
@@ -57,11 +57,13 @@ public void GetVersionInfo()
             {
                 var httpClient = this.HttpClientFactory.CreateClient("general");
                 var sdkStorageBaseUrl = this.GetPlatformBinariesStorageBaseUrl();
+                var url = string.Format(SdkStorageConstants.ContainerMetadataUrlFormat, sdkStorageBaseUrl, "dotnet", string.Empty);
```
pauld-msft, line 60: nit: use `DotNetCoreConstants.PlatformName` instead of hardcoded "dotnet"

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreSdkStorageVersionProvider.cs
```diff
@@ -75,6 +77,18 @@ public void GetVersionInfo()
                     runtimeVersionMetadataName = SdkStorageConstants.LegacyDotnetRuntimeVersionMetadataName;
                 }
 
+                // if <NextMarker> element's value is not empty, we iterate through every page by appending marker value to the url
+                // and consolidate blobs from all the pages.
+                do
+                {
+                    url = string.Format(SdkStorageConstants.ContainerMetadataUrlFormat, sdkStorageBaseUrl, "dotnet", marker);
```
pauld-msft, line 84: same nit as above

pauld-msft (dismissed):
super minor, otherwise lgtm :shipit:

src/BuildScriptGenerator/SdkStorageVersionProviderBase.cs
```diff
@@ -63,6 +63,17 @@ protected PlatformVersionInfo GetAvailableVersionsFromStorage(string platformNam
                 ? SdkStorageConstants.LegacySdkVersionMetadataName
                 : SdkStorageConstants.SdkVersionMetadataName;
 
+            // if <NextMarker> element's value is not empty, we iterate through every page by appending marker value to the url
+            // and consolidate blobs from all the pages.
+            do
+            {
+                url = string.Format(SdkStorageConstants.ContainerMetadataUrlFormat, sdkStorageBaseUrl, platformName, marker);
```
daniv-msft, line 70: Nit: should we consider putting this logic in an helper, so that we don't repeat it?

daniv-msft:
More of a high level comment, but as mentioned in the standup: this change is good and is more future proof, but we won't be able to make the change to Production as we cannot upgrade all customers, right (or maybe we can, but it will be long)?
We should discuss on what options we have that don't include updating the code.

snehapar9:
> More of a high level comment, but as mentioned in the standup: this change is good and is more future proof, but we won't be able to make the change to Production as we cannot upgrade all customers, right (or maybe we can, but it will be long)? We should discuss on what options we have that don't include updating the code.

My take on this is that - we would need to upgrade all customers if this change goes to production. Need to do more research on this - but this may be the way to go if we change our replication strategy. 
On the other hand, we could manage two LRS accounts but would also have to manage load balancing and make modifications to our pipelines. There seems to be a tradeoff between the two options.

pauld-msft:
> More of a high level comment, but as mentioned in the standup: this change is good and is more future proof, but we won't be able to make the change to Production as we cannot upgrade all customers, right (or maybe we can, but it will be long)? We should discuss on what options we have that don't include updating the code.

Yeah all I can think of is seeing if there is a way that we could control or configure the partitioning strategy to ensure that contents of containers are guaranteed to exist on the same partition. If that's not possible, then I'm not sure how we could guarantee support for older images' dynamic installation.

qianz2 (dismissed):

cormacpayne (approved):
:shipit:

daniv-msft (approved):

PR #1546, opened 2022-09-01
telemetry for dockerfile command

waliMSFT:


src/BuildScriptGeneratorCli/Commands/DockerfileCommand.cs
```diff
@@ -78,33 +79,49 @@ internal class DockerfileCommand : CommandBase
...
         {
-            ILoggerFactory loggerFactory = serviceProvider.GetRequiredService<ILoggerFactory>();
-            var sourceRepo = new LocalSourceRepo(this.SourceDir, loggerFactory);
-            var ctx = new DockerfileContext
+            var buildEventProps = new Dictionary<string, string>()
             {
-                SourceRepo = sourceRepo,
+                { "platform", this.PlatformName },
+                { "platformVersion", this.PlatformVersion },
+                { "runtimePlatformName", this.RuntimePlatformName },
```
pauld-msft, line 86: Some other inputs it could be useful to log are the `RuntimePlatformVersion` and `BuildImage`

cormacpayne: Agreed -- even though not all of these arguments will be provided to the `oryx dockerfile` call, it's still worth seeing what percentage of calls are being made with or without certain arguments, and when arguments are provided, what their value is. The other arguments don't need to be tracked in this case.

src/BuildScriptGeneratorCli/Commands/DockerfileCommand.cs
```diff
@@ -4,6 +4,7 @@
 // --------------------------------------------------------------------------------------------
```
cormacpayne, line 4: We should also include a `LogTimedEvent` call wrapping the execution of the `GenerateDockerfile` command in `DefaultDockerfileGenerator` so we can log information about the build and runtime images and tags that we detect and what their result was.

src/BuildScriptGeneratorCli/Commands/DockerfileCommand.cs
waliMSFT: alright. I can add these.

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
```diff
@@ -47,90 +47,104 @@ internal class DefaultDockerfileGenerator : IDockerfileGenerator
...
+                generatedDockerfile = Regex.Replace(generatedDockerfile, pattern, string.Empty);
+
+                var buildEventProps = new Dictionary<string, string>()
+                {
+                    { "runtimeImageName", properties.RuntimeImageName },
+                    { "runtimeImageTag", properties.RuntimeImageTag },
+                    { "buildImageName", properties.BuildImageName },
+                    { "buildImageTag", properties.BuildImageTag },
+                    { "createScriptArguments", properties.CreateScriptArguments },
+                    { "generatedDockerfile", generatedDockerfile },
```
cormacpayne, line 142: Let's remove the `createScriptArguments` and `generatedDockerfile` properties from this `buildEventProps` dictionary -- they won't be useful for us and will bloat the logging.

cormacpayne (commented):
One minor comment otherwise :shipit:

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
waliMSFT: alright.

cormacpayne (approved):

qianz2 (approved):

PR #1547, opened 2022-09-02
Install and add imagick package test coverage for PHP 8.0 and 8.1

qianz2:
For resolving IcM: https://portal.microsofticm.com/imp/v3/incidents/details/330398778/home

Looking into the PHP 8.* image dockerfiles and found this part of code:

It used pecl to install imagick and its extensions, but it may not work as according to the above issue link, imagick for PHP 8.* should be installed through source. I see Oryx has imagick test coverage for 7.4 but not for 8.0 and 8.1.

Adding test coverage for 8.0 and 8.1 and install it with the methods mentioned in the GH issue links.

```
# https://github.com/Imagick/imagick/issues/331
RUN set -eux; \
    if [[ $PHP_VERSION != 8.* ]]; then \
        pecl install imagick && docker-php-ext-enable imagick; \
    fi
```


images/runtime/php-fpm/template.base.Dockerfile
```diff
@@ -93,7 +93,7 @@ RUN set -eux; \
 #  - https://docs.microsoft.com/en-us/sql/connect/php/installation-tutorial-linux-mac
 #  - https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server
 RUN set -eux; \
-    if [[ $PHP_VERSION == 7.4.* || $PHP_VERSION == 8.0.* ]]; then \
+    if [[ $PHP_VERSION == 7.4.* || $PHP_VERSION == 8.* ]]; then \
```
daniv-msft, line 96: Could you please clarify how updating this will install imagick? Is this related to the OBDC driver mentioned here?

images/runtime/php-fpm/template.base.Dockerfile
```diff
@@ -93,7 +93,7 @@ RUN set -eux; \
 #  - https://docs.microsoft.com/en-us/sql/connect/php/installation-tutorial-linux-mac
```
daniv-msft, line 93: When looking above in the source file, line 80, I can still see the condition:
`if [[ $PHP_VERSION != 8.* ]]; then \`
Is this only going to work for PHP 7.4 in that case?

qianz2: I verified that the php imagick sample apps passed for 7.4, 8.0 and 8.1 with these installations. So here I kept the original way to install imagick library for 8.0 and 8.1.

images/runtime/php-fpm/template.base.Dockerfile
qianz2: It's the pdo_sqlsrv library that 8.1 didn't include, so I simply added it and corresponding tests.

images/runtime/php-fpm/template.base.Dockerfile
daniv-msft: Sorry but I'm a bit confused: could you please clarify where in the code we are installing imagick for PHP 8.0 and 8.1? The only place where I see a reference to imagick in this file is line 81, that explicitely excludes PHP 8.*.
Could you please clarify what I'm missing?

images/runtime/php-fpm/template.base.Dockerfile
daniv-msft: OK, and so imagick was already installed separately? I guess I'm just confused as I didn't see where (cf. my other comment).

images/runtime/php-fpm/template.base.Dockerfile
qianz2: Good catch, this is very strange. Here's the place where we install the imagick library but I wonder why the tests all passing even with this condition. So I'll remove the condition here and check if the tests are correctly covered.

images/runtime/php-fpm/template.base.Dockerfile
qianz2: Yes it's a seperate one.

images/runtime/php-fpm/template.base.Dockerfile
daniv-msft: Thanks!

cormacpayne (approved):

PR #1550, opened 2022-09-06
Bugfix: Fix if statements for publishing to prod

pauld-msft:
Fixes `line 58: [: missing ``]'` error seen in [this job](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6652549&view=logs&j=9b0d530e-cf64-58a6-9150-38de442bc22a&t=70dad668-6562-5a56-a9b4-58284a3cfd77)


qianz2 (approved):
Thanks for fixing it!

daniv-msft:
I actually cannot see the fix: all I see are spaces added. Are these the fix? I didn't expect that this would create the failure.

daniv-msft (approved):

pauld-msft:
> I actually cannot see the fix: all I see are spaces added. Are these the fix? I didn't expect that this would create the failure.

Yes that is the bug! I did not expect it either 
image

william-msft (approved):

PR #1552, opened 2022-09-07
added USER_CALLER_ID env variable and telemetry in the build

waliMSFT:


src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -129,6 +129,9 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
             var oryxVersion = Program.GetVersion();
             var oryxCommitId = Program.GetMetadataValue(Program.GitCommit);
             var oryxReleaseTagName = Program.GetMetadataValue(Program.ReleaseTagName);
+
+            var options = serviceProvider.GetRequiredService<IOptions<BuildScriptGeneratorOptions>>().Value;
```
cormacpayne, line 133: This change is only going to be available for the `oryx build` command -- we want to make sure that the `USER_CALLER_ID` environment variable is read for all of our commands in the case that we want to check the frequency by which any partner teams are calling our commands. We should find a way to generalize this change you've added such that it's readily available for all of our commands (my vote is to find a way to put it in the `CommandsBase` class).

cormacpayne (changes requested):

src/BuildScriptGenerator/Options/BuildScriptGeneratorOptions.cs
```diff
@@ -82,5 +82,7 @@ public class BuildScriptGeneratorOptions
         public string CustomRequirementsTxtPath { get; set; }
 
         public string DebianFlavor { get; set; }
+
+        public string UserCallerId { get; set; }
```
daniv-msft, line 86: From our discussion, I thought that we would go with "CallerId", as anything "User" related looks like it actually corresponds to the end-users, our customers, and not the partner team relying on Oryx.
Did you think about another scenario where having User in the name would be valuable?

waliMSFT: Dan, during our discussion I thought by `CallerId` you tried to explain the purpose of this environment variable. Did not realize I should change the name of the variable.
Getting rid of the `user` prefix from the variable name makes more sense to me as well. I will make this change.

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
waliMSFT: thanks for the suggestion, Cormac. When I discussed with Dan regarding the potential reports we can generate using this telemetry, the main focus was to get telemetry in the `build` command. Hence, I added it in the `build` command.
your suggestion is also a nice one. Let me find a way to do it.

waliMSFT: @cormacpayne In the `CommandBase` class we have an event telemetry `logger?.LogTimedEvent(this.GetType().Name))`. I have used this event and added the `callerId` here because this event is created whenever a `*Command` class object is executed (example: BuildCommand, RunScriptCommand, PrepareEnvironmentCommand, etc).  
However, for other command the usage is quite low compared to the `BuildCommand`. Here is a quick stat I got from the AppInsight (in last 24 hours):
count of usage of `BuildCommand`:  30,476
count of usage of other Commands: 8

Based on that data I am just wondering do we really need to consider other `commands`?

cormacpayne: We do want to consider other Oryx CLI commands for two reasons: (1) we just collaborated with the Container App folks to integrate a call to `oryx dockerfile` into their `az containerapp up` command that's part of the Azure CLI, so we want to make sure that we're recording how often that command is being called by the Azure CLI, and (2) the telemetry is already in place in the `CommandBase` class to allow us to track this information across all of our commands, so it's better to generalize our telemetry collection where we can rather than have it collected for only a single command.

src/BuildScriptGeneratorCli/SettingsKeys.cs
```diff
@@ -68,5 +68,6 @@ public static class SettingsKeys
         public const string DisableRecursiveLookUp = "DISABLE_RECURSIVE_LOOKUP";
         public const string CustomRequirementsTxtPath = "CUSTOM_REQUIREMENTSTXT_PATH";
         public const string DebianFlavor = "DEBIAN_FLAVOR";
+        public const string CallerId = "USER_CALLER_ID";
```
cormacpayne, line 71: We need to also change the environment variable itself to be `CALLER_ID`

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -139,12 +142,11 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
                     string.Join(' ', environment.GetCommandLineArgs())
                 },
                 { "sourceRepoCommitId", sourceRepoCommitId },
+                { "platformName", this.PlatformName },
```
cormacpayne, line 145: Are we able to revert the changes in this file?

src/BuildScriptGenerator/Options/BuildScriptGeneratorOptions.cs
daniv-msft: Thanks!

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
waliMSFT: makes sense. I already added it inside `CommandBase `.

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
waliMSFT: In our discussion with Dan, we concluded to add the `platform` telemetry in the build. That is why I added it. But I will keep rest as it was.

src/BuildScriptGeneratorCli/SettingsKeys.cs
waliMSFT: In the case, our partners need to aware of that as well. isn't it? They have to send the env variable `CALLER_ID`. how this works?

cormacpayne: Correct they have to set the `CALLER_ID` environment variable to some value, but in this file, the environment variable we'll be reading is `USER_CALLER_ID` since we didn't change it yet after the comment from Dan to drop the `USER` part of it.

cormacpayne (approved):

PR #1554, opened 2022-09-08
Williamhe/1562986 add vso bullseye image

william-msft:

**Goal:**
Add lzma module for python 3.10

**Problem:**
lzma is not supported for the existing vso (githubrunners-buildpackdeps-focal based) image. We know this because lzma comes prebuilt with python versions but not with 3.10. Additionally when we manually add lzma we get kernel related errors.

**Solution:**
Introduce a new vso image that is bullseye based. Bullseye is a newer image with longer support and Oryx is upgrading other images to bullseye too.
This comes with contingencies like removing python 2 (EOL) and the other 2 ruby gems (ruby-debug-ide and debase). Bullseye is a newer image with longer support. This means introducing a new image in Oryx for Codespaces.
https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1562986


build/__variables.sh
```diff
@@ -75,6 +76,7 @@ declare -r ACR_BUILD_FULL_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:full"
 declare -r ACR_BUILD_LTS_VERSIONS_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:lts-versions"
 declare -r ACR_BUILD_GITHUB_ACTIONS_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:github-actions"
 declare -r ACR_BUILD_VSO_FOCAL_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:vso-ubuntu-focal"
+declare -r ACR_BUILD_VSO_BULLSEYE_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:vso-bullseye"
```
pauld-msft, line 79: We now include both OS type and OS version in the image tags. Could you update the tag to be `vso-debian-bullseye`?

build/buildBuildImages.sh
```diff
@@ -514,6 +556,8 @@ elif [ "$imageTypeToBuild" == "full" ]; then
 	buildFullImage "buster"
 elif [ "$imageTypeToBuild" == "vso-focal" ]; then
 	buildVsoFocalImage
```
pauld-msft, line 558: Looks like the `buildVsoImage` method you made can support a debian bullseye OS version or a ubuntu focal OS version. Should we remove the `buildVsoFocalImage` method and switch this to the `buildVsoImage` you made?

images/installPlatform.sh
```diff
@@ -43,17 +43,14 @@ VERSION="$2"
 
 debianFlavor=$DEBIAN_FLAVOR
 fileName="$PLATFORM_NAME-$VERSION.tar.gz"
-
-sdkStorageAccountUrl="$ORYX_SDK_STORAGE_BASE_URL"
```
pauld-msft, line 47: We need this logic so that we can control where the pre-baked binaries come from (dev during testing, prod during actual image release)

tests/Oryx.Tests.Common/ImageTestHelper.cs
```diff
@@ -319,9 +333,10 @@ public string GetGitHubActionsAsBaseWithEnvBuildImage(string debianFlavor = null
             }
         }
 
-        public string GetVsoBuildImage(string debianFlavor=null)
+        public string GetVsoBuildImage(string debianFlavor = null)
         {
-            return $"{_repoPrefix}/{_buildRepository}:{_vsoUbuntu}{_tagSuffix}";
+            string _vsoImage = !string.IsNullOrEmpty(debianFlavor) && debianFlavor == "bullseye" ? _vsoBullseye : _vsoUbuntu;
```
pauld-msft, line 338: Can we make this consistent with the other test methods? They have the image test helper constants (e.g. ImageTestHelperConstants.VsoBullseye) passed in as the `debianFlavor` parameter, instead of the actual debian flavor. This naming can probably be updated as its confusing. 

They then check to see if what is passed in matches the constant e.g.
`string.Equals(debianFlavor.ToLower(), _azureFunctionsJamStackBullseye)` which in this case would be `_vsoBullseye` or `_vsoUbuntu`

All of this allows us to only have to update the `ImageTestHelperConstants` value when an image tag changes, instead of find every place that `GetVsoBuildImage` or `GetBuildImage` is used and seeing if it also needs to be changed

images/installPlatform.sh
pauld-msft: Looks like you already reverted this change. Thanks!

images/build/Dockerfiles/vso.bullseye.Dockerfile
```diff
@@ -0,0 +1,325 @@
+FROM mcr.microsoft.com/mirror/docker/library/buildpack-deps:bullseye AS main
```
cormacpayne, line 1: Any reason we don't base this off of our intermediate `oryx/githubrunners-buildpackdeps-bullseye` image?

images/build/Dockerfiles/vso.bullseye.Dockerfile
```diff
@@ -0,0 +1,325 @@
...
+RUN LANG="C.UTF-8" \
+    && apt-get update \
+    && apt-get upgrade -y \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        git \
+        make \
+        unzip \
+        # The tools in this package are used when installing packages for Python
+        build-essential \
+        # swig3.0 \
```
cormacpayne, line 15: There are a couple of lines that are now commented-out in this file (lines 15, 49, 53-55, 159, 280-285, 296) -- are we able to just delete them?

images/build/Dockerfiles/vso.bullseye.Dockerfile
```diff
@@ -0,0 +1,325 @@
...
+# Even though this adds a new docker layer we are doing this 
+# because we want to avoid duplication (which is always error-prone)
+
+RUN set -ex \
+    && tmpDir="/opt/tmp" \
+    && imagesDir="$tmpDir/images" \
+    && buildDir="$tmpDir/build" \
+    # https://github.com/docker-library/python/issues/147
+    && PYTHONIOENCODING="UTF-8" \    
+    # new next line
```
cormacpayne, line 90: I see this comment a few times in this file now -- are we able to remove it or is it needed for documentation purposes? I'm in the boat that if we want to keep a comment on this line, it should explain what's going on with the `add-apt-repository` command called below it.

images/build/Dockerfiles/vso.bullseye.Dockerfile
```diff
@@ -0,0 +1,325 @@
...
+    && add-apt-repository --remove ppa:xapienz/curl34 \
+    && apt-get update \
+    && apt-get upgrade -y \
+    # It's not clear whether these are needed at runtime...
+    && apt-get install -y --no-install-recommends \
+        tk-dev \
+        uuid-dev \
+        curl \
```
cormacpayne, line 98: nit: is `curl` not included in the base `buildpack-deps` image?

images/build/Dockerfiles/vso.bullseye.Dockerfile
```diff
@@ -0,0 +1,325 @@
...
+    && ln -s $YARN_MINOR_VERSION $YARN_MAJOR_VERSION \
+    # Install Python SDKs
+    # Upgrade system python
+    && mkdir -p /home/codespace/.python \
+    && pip install --upgrade cython \
+    && pip3 install --upgrade cython \
+    && . $buildDir/__pythonVersions.sh \
+    && $imagesDir/installPlatform.sh python $PYTHON39_VERSION \
+    # && $imagesDir/installPlatform.sh python $PYTHON310_VERSION \
+    # build python directly from source
```
cormacpayne, line 160: nit: just curious why we would want to build Python 3.10 from source here rather than use the `installPlatform.sh` script to do this

images/build/Dockerfiles/vso.bullseye.Dockerfile
```diff
@@ -0,0 +1,325 @@
...
+# Now adding remaining of VSO platform features
+RUN buildDir="/opt/tmp/build" \
+    && imagesDir="/opt/tmp/images" \
+    # new next line
+    && add-apt-repository --remove ppa:xapienz/curl34 \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
+        apt-transport-https \
+        nano \
+        sed \
```
cormacpayne, line 232: nit: are the `sed` commands you added needed?

images/build/Dockerfiles/vso.bullseye.Dockerfile
```diff
@@ -0,0 +1,325 @@
...
+    && ls . \
+    && php go-pear.phar \
+    #&& mkdir -p pear/tmp/pear/install \
+    #&& wget https://pear.php.net/go-pear.phar \
+    #&& export PHP_PEAR_PHP_BIN=/usr/local/php74/bin/php \
+    #&& export PATH=${HOME}/pear/bin:/usr/local/php74/bin:${PATH} \
+    #&& ls /usr/local/php74/bin/ \
+    #&& /opt/php/8.1.6/bin/php go-pear.phar \
+    && pecl version \
+    && which pecl \
```
cormacpayne, line 287: nit: leftover debugging commands here?

images/build/php/prereqs/installPrereqs.sh
```diff
@@ -22,9 +22,10 @@ PHPIZE_DEPS="autoconf dpkg-dev file g++ gcc libc-dev make pkg-config re2c"
 # libcurl3 and libcurl4 both needs to be supported in ubuntu focal for php
 # https://github.com/xapienz/curl-debian-scripts
 add-apt-repository ppa:xapienz/curl34 -y \
-&& apt-get update \
+apt-get update \
```
cormacpayne, line 25: nit: this works without the `&&` here?

images/build/php/prereqs/installPrereqs.sh
```diff
@@ -22,9 +22,10 @@ PHPIZE_DEPS="autoconf dpkg-dev file g++ gcc libc-dev make pkg-config re2c"
...
 && apt-get upgrade -y \
 && apt-get install -y \
+        sed \
```
cormacpayne, line 28: nit: similar comment as before -- is `sed` (and `grep`) needed?

images/build/Dockerfiles/vso.bullseye.Dockerfile
william-msft: Yeah since this isn't in our approved images: https://eng.ms/docs/more/containers-secure-supply-chain/approved-images

images/build/Dockerfiles/vso.bullseye.Dockerfile
william-msft: Thanks Cormac, I've moved to the PR to draft while I cleanup a little more 😃

images/build/Dockerfiles/vso.bullseye.Dockerfile
william-msft: I was testing with `mcr.microsoft.com/mirror/docker/library/debian:bullseye` which didn't have `curl`. This will be part of my cleanup

images/build/Dockerfiles/vso.bullseye.Dockerfile
william-msft: Thanks for pointing this out, this line should no longer be present.

images/build/Dockerfiles/vso.bullseye.Dockerfile
william-msft: Correct, this has been removed.

build/__variables.sh
william-msft: Yes, update has been made.

build/buildBuildImages.sh
william-msft: Thank Paul, update has been made.

tests/Oryx.Tests.Common/ImageTestHelper.cs
william-msft: Agreed, I've updated this method.

images/build/Dockerfiles/vso.bullseye.Dockerfile
cormacpayne: That shouldn't affect us since the intermediate images that we build locally are tagged with the prefix `oryxdevmcr.azurecr.io/private/oryx`, so this Dockerfile scan doesn't flag these images because it thinks they're later pushed to our ACR instance, but we just use that naming convention strictly to bypass the check.

images/build/Dockerfiles/vso.bullseye.Dockerfile
william-msft: Confirmed no longer needed

images/build/php/prereqs/installPrereqs.sh
william-msft: Confirmed no longer needed

images/build/php/prereqs/installPrereqs.sh
william-msft: This actually only works without the `&&` in the front which is strange.

images/build/Dockerfiles/vso.bullseye.Dockerfile
william-msft: Originally using `installPlatform.sh` for python 3.10 was failing, but I figured out why and have updated this. 

The reason this was failing was because the Debain flavor was still set as focal, which caused focal SDK to be downloaded, which is incompatible with this bullseye image.

images/build/Dockerfiles/vso.bullseye.Dockerfile
william-msft: Confirmed this is no longer needed.

images/build/Dockerfiles/vso.bullseye.Dockerfile
william-msft: I ran into ` php: error while loading shared libraries: libxml2.so.2: cannot open shared object file: No such file or directory` which is similar to what I saw with `mcr.microsoft.com/mirror/docker/library/debian:bullseye` . Checking to see if I can make it work with this image

images/build/php/prereqs/installPrereqs.sh
cormacpayne, line 28: nit: looks like `sed` and `grep` are still being installed with this script -- you mentioned we at least don't need `grep`, so can we revert these changes?

cormacpayne (dismissed):
One minor comment, otherwise :shipit:

images/build/php/prereqs/installPrereqs.sh
william-msft: Thanks for catching this! Its been removed now.

platforms/php/versions/bullseye/versionsToBuild.txt
```diff
@@ -9,6 +9,7 @@
 7.4.28, 9cc3b6f6217b60582f78566b3814532c4b71d517876c25013ae51811e65d8fce, 42670A7FE4D0441C8E4632349E4FDC074A4EF02D 5A52880781F755608BF815FC910DEB46F53EA312,
 7.4.30, ea72a34f32c67e79ac2da7dfe96177f3c451c3eefae5810ba13312ed398ba70d, 42670A7FE4D0441C8E4632349E4FDC074A4EF02D 5A52880781F755608BF815FC910DEB46F53EA312,
 
+8.0.22, 130937c0fa3050cd33d6c415402f6ccbf0682ae83eb8d39c91164224ddfe57f1, 1729F83938DA44E27BA0F4D3DBDB397470D12172 BFDDD28642824F8118EF77909B67A5C12229118F,
```
william-msft, line 12: We need to add this so the `vso-debian-bullseye` image can download this SDK

src/startupscriptgenerator/src/common/Gopkg.toml
```diff
@@ -41,6 +41,10 @@
   name = "github.com/spf13/afero"
   version = "=1.8.2"
 
+[[override]]
+  name = "golang.org/x/sys"
```
daniv-msft, line 45: Isn't this coming from Paul's latest PR? Do you need to rebase, or are you adding this to other files?

william-msft: I merge Paul's branch `bugfix/lock-sys-unix-package` with my branch since I also need his changes and since they're not in `main` yet

daniv-msft (approved):

PR #1556, opened 2022-09-09
Draft:Snehapar/1612862 create env variable for connection string for app insights

snehapar9:
User Story 1612862](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1612862): Create a new env variable for Connection String for AppInsights


images/build/Dockerfiles/full.Dockerfile
```diff
@@ -21,8 +21,10 @@ RUN ./build.sh golang     /opt/startupcmdgen/golang
 FROM buildpack-deps:${DEBIAN_FLAVOR}-curl
 ARG DEBIAN_FLAVOR
 ARG SDK_STORAGE_BASE_URL_VALUE="https://oryx-cdn.microsoft.io"
+ARG AI_CONNECTION_STRING
```
cormacpayne, line 24: nit: I'd sync with William to see if there's a reason why the instrumentation key was never included in this image just in case we don't want to have telemetry here for some reason

src/BuildScriptGenerator.Common/LoggingConstants.cs
```diff
@@ -13,6 +13,7 @@ public static class LoggingConstants
         public const string ApplicationInsightsInstrumentationKeyEnvironmentVariableName
             = "ORYX_AI_INSTRUMENTATION_KEY";
 
+        public const string ApplicationInsightsConnectionStringKeyEnvironmentVariableName = "ORYX_AI_CONNECTION_STRING";
```
cormacpayne, line 16: nit: let's follow the same pattern with these longer variables/strings and have the value be on a new line:

```cs

public const string ApplicationInsightsConnectionStringKeyEnvironmentVariableName
    = "ORYX_AI_CONNECTION_STRING";

public const string OryxDisableTelemetryEnvironmentVariableName
    = "ORYX_DISABLE_TELEMETRY";
```

src/BuildScriptGeneratorCli/ServiceProviderBuilder.cs
```diff
@@ -72,6 +72,18 @@ private static LoggingConfiguration BuildNLogConfiguration([CanBeNull] string lo
 
             var aiKey = disableTelemetry ? string.Empty : Environment.GetEnvironmentVariable(
                 LoggingConstants.ApplicationInsightsInstrumentationKeyEnvironmentVariableName);
+            var aiConnectionString = disableTelemetry ? string.Empty : Environment.GetEnvironmentVariable(LoggingConstants.ApplicationInsightsConnectionStringKeyEnvironmentVariableName);
+            if (!string.IsNullOrEmpty(aiConnectionString))
+            {
+                var aiTarget = new ApplicationInsights.NLogTarget.ApplicationInsightsTarget()
+                {
+                    Name = "ai",
+                    InstrumentationKey = aiConnectionString,
```
cormacpayne, line 81: We are able to provide the connection string to this `InstrumentationKey` property?

src/BuildScriptGeneratorCli/ServiceProviderBuilder.cs
```diff
@@ -72,6 +72,18 @@ private static LoggingConfiguration BuildNLogConfiguration([CanBeNull] string lo
...
+                };
+                config.AddTarget(aiTarget);
+                config.AddRuleForAllLevels(aiTarget);
+            }
+
             if (!string.IsNullOrWhiteSpace(aiKey))
```
cormacpayne, line 87: We should make this an `else if` so that we only use the instrumentation key when the connection string isn't provided

src/startupscriptgenerator/src/common/consts/ext_var_names.go
```diff
@@ -4,6 +4,7 @@ package consts
 
 const EnvironmentType string = "ORYX_ENV_TYPE"
 const ApplicationInsightsInstrumentationKeyEnvVarName string = "ORYX_AI_INSTRUMENTATION_KEY"
+const ApplicationInsightsConnectionStringEnvVarName string = "ORYX_AI_CONNECTION_STRING"
```
cormacpayne, line 7: There are some changes we'll also need to make in the [`logging.go`](https://github.com/microsoft/Oryx/blob/main/src/startupscriptgenerator/src/common/logging.go#L53-L62) file to actually use this connection string for the runtime images.

cormacpayne (changes requested):

src/BuildScriptGeneratorCli/ServiceProviderBuilder.cs
snehapar9: I only see Instrumentation Key in ApplicationInsightsTarget class. Not seeing Connection String.
Not sure if I need to install an updated version from Nuget Package Manager?

snehapar9:
Closing this PR as we are not able to integrate with Application Insights with these changes. Support for Connection String has not been added in ApplicationInsights.NLogTarget.ApplicationInsightsTarget 
Please see User Story 1612862](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1612862) for more details.

PR #1557, opened 2022-09-09
Added env variable for client certificate path in dotnet 7 and node 18

snehapar9:
User Story 1612862](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1612862): Create a new env variable for Connection String for AppInsights

### DIFF
--- a/images/runtime/dotnetcore/7.0/base.buster.Dockerfile
+++ b/images/runtime/dotnetcore/7.0/base.buster.Dockerfile
@@ -34,6 +34,8 @@ ENV ASPNETCORE_URLS=http://+:80 \
     # Enable detection of running in a container
     DOTNET_RUNNING_IN_CONTAINER=true \
     PATH="/opt/dotnetcore-tools:${PATH}"
+#Bake in client certificate path into image to avoid downloading it
+ENV PATH_CA_CERTIFICATE="/etc/ssl/certs/ca-certificate.crt"
 
 COPY --from=tools-install /dotnetcore-tools /opt/dotnetcore-tools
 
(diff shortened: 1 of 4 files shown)


images/runtime/dotnetcore/7.0/base.buster.Dockerfile
```diff
@@ -34,6 +34,8 @@ ENV ASPNETCORE_URLS=http://+:80 \
     # Enable detection of running in a container
     DOTNET_RUNNING_IN_CONTAINER=true \
     PATH="/opt/dotnetcore-tools:${PATH}"
+#Bake in client certificate path into image to avoid downloading it
+ENV PATH_CA_CERTIFICATE="/etc/ssl/certs/ca-certificate.crt"
```
pauld-msft, line 38: Maybe I missed this in a previous PR or email thread or something, but how are these environment variables actually used? I don't see anywhere that we are using them, and I'm also not sure why we don't supply it in all of our runtime images (e.g. php-fpm and ruby don't have this, other dotnet base files don't have it, etc.)

snehapar9: This request is from the App Services team to create an environment variable for client certs so they are baked into runtime images for Python 3.10, PHP 8.1, Node 18 and .Net 7 Preview 7

Please see email thread - Bundling Azure MySQL/Postgres client cert in App Service Oryx Images for more context.

pauld-msft: Hmm looking at that email thread, I'm not finding `PATH_CA_CERTIFICATE`. Did we send that to our partners to use? So they would be using it, and we don't need to do anything but give them the path of the certificate?

pauld-msft (approved):
Just curious how its being used / whether it needs to be updated for other runtime images. Other than that, LGTM

qianz2 (approved):

images/runtime/dotnetcore/7.0/base.buster.Dockerfile
snehapar9: Looks like they want env for PATH_CA_CERTIFICATE baked into images Antares would release with ANT:99

Antares provided the path to the certificate file. Essentially, the ask on this work item was to just add an env variable for easy reference.

PR #1560, opened 2022-09-12
Williamhe/1578887 automate dotnet sdk release

william-msft:
Draft: More updates to come.

- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1578887


.editorconfig
```diff
@@ -0,0 +1,16 @@
+﻿[*.cs]
```
william-msft, line 1: Draft: More changes to come

platforms/dotnet/versions/bullseye/versionsToBuild.txt
```diff
@@ -1,5 +1,10 @@
-# version, sha, download url
```
william-msft, line 1: Only a format change to be in sorted in alpha-numeric order

platforms/dotnet/versions/buster/versionsToBuild.txt
```diff
@@ -1,101 +1,103 @@
-# version, sha, download url
```
william-msft, line 1: Only a format change to be in sorted in alpha-numeric order

platforms/dotnet/versions/focal-scm/versionsToBuild.txt
```diff
@@ -1,101 +1,103 @@
-# version, sha, download url
```
william-msft, line 1: Only a format change to be in sorted in alpha-numeric order

platforms/dotnet/versions/stretch/versionsToBuild.txt
```diff
@@ -1,97 +1,99 @@
-# version, sha, download url
```
william-msft, line 1: Only a format change to be in sorted in alpha-numeric order

build/constants.yaml
```diff
@@ -1,19 +1,19 @@
 - name: ext-var-names
```
william-msft, line 1: Only a format change. The C# `YamlDotNet` library removes comments, removes double quotes, has specific spacing requirements, and converts list with `[]` to `-`. Example: https://github.com/microsoft/Oryx/pull/1560/files#diff-47c28d7a6c8135707f46b624b5913e35beea6dfbe7a8be2db7efefde606eba59R47

.github/workflows/main.yaml
```diff
@@ -0,0 +1,71 @@
+name: CI
+
+on:
+  # Run every day at 10 pm
+  #schedule:
+  #  # Temporary cron for testing.
+  #  - cron: '*/15 * * * *'
+  
+  push:
```
william-msft, line 9: Temporary change: We will be running every once every day

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,236 @@
...
+        ///     - If new release, then store release information into PlatformConstants
+        ///        Otherwise don't store anything
+        /// </Summary>
+        /// <returns>PlatformConstants used later to update constants.yaml</returns>
+        public override async Task<List<PlatformConstant>> GetPlatformConstantsAsync(string dateTarget)
+        {
+
+            // check dotnet releases' meta data
+            var response = await Request.RequestAsync(Constants.DotNetReleasesMetaDataUrl);
+            var json = JsonConvert.DeserializeObject<ReleaseNotes>(response);
```
cormacpayne, line 43: I'd rename this variable to `releaseNotes` since it's not a JSON string but the deserialization of the JSON string into the `ReleaseNotes` object

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,236 @@
...
+        private static string GetSha(List<FileObj> files)
+        {
+            HashSet<string> tarFileNames = new HashSet<string>() {
+                "dotnet-sdk-linux-x64.tar.gz",
+                "dotnet-runtime-linux-x64.tar.gz",
+                "aspnetcore-runtime-linux-x64.tar.gz",
+            };
+            foreach (var file in files)
+            {
+                if (tarFileNames.Contains(file.Name))
```
cormacpayne, line 224: nit: I think it would be easier if we check that the file matched the pattern `*-linux-x64.tar.gz` using RegEx (expression would need to be updated, but you get the idea)

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,236 @@
...
+                {
+                    return file.Hash;
+                }
+            }
+
+            // TODO: special exception if sha not found
+            Console.WriteLine("No sha found");
+
+            return string.Empty;
```
cormacpayne, line 233: We should throw an exception now and handle it by the caller so that they can choose to skip processing the current version if it doesn't contain a 64-bit Linux release that we could use since right now we're still going ahead and creating a `PlatformConstant` object with an empty string `Sha` property.

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,236 @@
...
+            Dictionary<string, Constant> dotnetYamlConstants = GetYamlDotNetConstants(yamlConstants);
+
+            // update dotnetcore sdks and runtimes
+            foreach (var platformConstant in platformConstants)
+            {
+                string version = platformConstant.Version;
+                string sha = platformConstant.Sha;
+                string versionType = platformConstant.VersionType;
+                string dotNetConstantKey = GenerateDotNetConstantKey(platformConstant);
+                Console.WriteLine($"version: {version} versionType: {versionType} sha: {sha} dotNetConstantKey: {dotNetConstantKey}");
```
cormacpayne, line 128: nit: same comment here for `Console.Write*`

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,236 @@
...
+
+                // sort
+                Console.WriteLine($"Updating {versionsToBuildTxtFile}...");
+                var contents = File.ReadAllLines(versionsToBuildTxtFile);
+                Array.Sort(contents);
+                File.WriteAllLines(versionsToBuildTxtFile, contents);
+            }
+        }
+
+        private static Dictionary<string, Constant> GetYamlDotNetConstants(List<Constant> yamlContents)
```
cormacpayne, line 176: We should replace this method with a LINQ query:

```cs
var dotnetConstants = yamlContents.Where(c => c.Name == Constants.DotNetSdkKey || c.Name == Constants.DotNetRuntimeKey)
                                  .ToDictionary(c => c.Name, c => c);
```

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,236 @@
...
+            }
+            return dotNetConstants;
+        }
+
+        private static string GenerateDotNetConstantKey(PlatformConstant platformConstant)
+        {
+            string[] splitVersion = platformConstant.Version.Split('.');
+            string majorVersion = splitVersion[0];
+            string minorVersion = splitVersion[1];
+            string majorMinor = majorVersion + minorVersion;
```
cormacpayne, line 195: I wouldn't use `majorVersion + minorVersion` as the key because we could easily be bitten in the scenario where we have `11.0` and `1.10` and both keys turn out to be `110`. We should include as much information as we can into the key to make it a unique string compared to any other version release.

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,236 @@
...
+                .WithNamingConvention(UnderscoredNamingConvention.Instance)
+                .Build();
+
+            var stringResult = serializer.Serialize(yamlConstants);
+            File.WriteAllText(Constants.ConstantsYaml, stringResult);
+        }
+
+        private static void UpdateVersionsToBuildTxt(PlatformConstant platformConstant)
+        {
+            List<string> versionsToBuildTxtFiles = new List<string>() {
```
cormacpayne, line 157: Similar comment to one I have elsewhere about having file paths and using that with `File.ReadAll*` methods

cormacpayne (changes requested):

.github/workflows/main.yaml
```diff
@@ -0,0 +1,71 @@
...
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v3
+
+      - name: Run SDK automation
+        run: time dotnet run --project build/tools/Automation/Automation.csproj
+
+        # [Important]
+        # Currently fails silently due to malformed path for Hugo.
+        # Unhandled exception. System.IO.DirectoryNotFoundException: Could not find a part of the path '/home/runner/work/Oryx/Oryx/build\detector/__detectorNugetPackagesVersions.props'.
```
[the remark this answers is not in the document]
pauld-msft: Yep the paths are from [this block](https://github.com/microsoft/Oryx/blob/main/build/constants.yaml#L353-L379). Lets just try to change the 3 `directory: build\detector` instances to `directory: build/detector` (forward slash) and see if that works?

.github/workflows/main.yaml
```diff
@@ -0,0 +1,71 @@
...
+        run: ./build/generateConstants.sh || true
+      
+      # TODO: replace William with Oryx credentials
+      - name: Create commits
+        run: |
+          git status
+          git config user.name 'William Hernandez'
+          git config user.email 'williamhe@microsoft.com'
+          git add build/*
+          git add src/BuildScriptGenerator/*
```
pauld-msft, line 51: Why don't we just add all changes? I've updated generateConstants / constants.yaml so that we should be able to commit all changes from running `generateConstants.sh` without incident. I just don't know if there are usually some other changes that happen which we aren't trying to have committed? To be clear, I'm fine with only adding these 2 paths, more just curious if you noticed something

Oryx.sln
```diff
@@ -100,6 +100,13 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BuildServer", "src\BuildSer
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BuildServer.Tests", "tests\BuildServer.Tests\BuildServer.Tests.csproj", "{5BD3E03A-3086-4CF6-950B-215C1E76B7F5}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Automation", "build\tools\Automation\Automation.csproj", "{18F7BB74-0A1E-4289-9161-515E65E667CE}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{EC38B4A4-CC12-4D60-AE29-C9C7CD4A8ED5}"
```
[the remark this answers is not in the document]
william-msft: Updated

Oryx.sln
```diff
@@ -100,6 +100,13 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "BuildServer", "src\BuildSer
...
+	ProjectSection(SolutionItems) = preProject
```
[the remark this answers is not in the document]
william-msft: Updated

build/tools/Automation/DotNet/DeserializerConstants.cs
```diff
@@ -0,0 +1,135 @@
...
+        public string Product { get; set; } = string.Empty;
+
+        [JsonProperty(PropertyName = "support-phase")]
+        public string SupportPhase { get; set; } = string.Empty;
+
+        [JsonProperty(PropertyName = "eol-date")]
+        public string EolDate { get; set; } = string.Empty;
+
+        [JsonProperty(PropertyName = "releases.json")]
+        public string ReleasesJson { get; set; } = string.Empty;
```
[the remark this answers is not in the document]
william-msft: Updated

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,236 @@
...
+        /// <Summary>
+        /// Gets DotNet's new release version and sha.
+        ///
+        /// This is accomplished by:
+        ///     - Checking release meta data url if there's a new release
+        ///     - If new release, then store release information into PlatformConstants
+        ///        Otherwise don't store anything
+        /// </Summary>
+        /// <returns>PlatformConstants used later to update constants.yaml</returns>
+        public override async Task<List<PlatformConstant>> GetPlatformConstantsAsync(string dateTarget)
```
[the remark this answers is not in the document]
william-msft: Updated

build/tools/Automation/DotNet/DotNet.cs
william-msft: Updated

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,236 @@
...
+        {
+
+            // check dotnet releases' meta data
+            var response = await Request.RequestAsync(Constants.DotNetReleasesMetaDataUrl);
+            var json = JsonConvert.DeserializeObject<ReleaseNotes>(response);
+            var releasesMetaData = json == null ? new List<ReleaseNote>() : json.ReleasesIndex;
```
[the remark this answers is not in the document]
william-msft: Updated

.github/workflows/main.yaml
william-msft: I noticed changes in additional files when running `./build/generateConstants.sh` while having no changes to constants.yaml
image

build/tools/Automation/DotNet/DotNet.cs
william-msft: Updated. Thanks for the code snippet!

build/tools/Automation/Program.cs
```diff
@@ -0,0 +1,145 @@
...
+    {
+        public static int Main(string[] args)
+        {
+            // TODO: use dotnet parameters instead and handle invalid date
+            string dateTarget = args.Length > 0 ? args[0] : string.Empty;
+            if (string.IsNullOrEmpty(dateTarget))
+            {
+                Console.WriteLine("No dateTarget provided.");
+                dateTarget = DateTime.Today.ToString();
+            }
```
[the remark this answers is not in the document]
william-msft: Updated

build/tools/Automation/Program.cs
```diff
@@ -0,0 +1,145 @@
...
+            Console.WriteLine($"dateTarget: {dateTarget}");
```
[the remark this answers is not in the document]
william-msft: Agreed. Ideally I'd like log this in `Logger.Debug` and we have a ToDo in the project. Would it be ok if we kept these for phase 0?

build/tools/Automation/Program.cs
```diff
@@ -0,0 +1,145 @@
...
+        }
+
+        /// <Summary>
+        /// Adds new platform constants to Oryx repo
+        /// </Summary>
+        public static async Task AddNewPlatformConstantsAsync(string dateTarget)
+        {
+            DotNet dotNet = new DotNet();
+            List<PlatformConstant> platformConstants = await dotNet.GetPlatformConstantsAsync(dateTarget).ConfigureAwait(true);
+            List<Constant> yamlConstants = await DeserializeConstantsYamlAsync().ConfigureAwait(true);
```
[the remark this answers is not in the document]
william-msft: We do not, I've updated this

build/tools/Automation/Request.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Microsoft.Oryx.Automation
+{
+    public static class Request
```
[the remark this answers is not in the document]
william-msft: Updated

build/tools/Automation/Request.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        private static readonly HttpClient Client = new HttpClient();
+
+        /// <Summary>
+        /// Performs a request for a given URL.
+        /// </Summary>
+        public static async Task<string> RequestAsync(string url)
```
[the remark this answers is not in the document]
william-msft: I renamed the method. I think there is value in having a `Logger.Debug` statement in case a future dev needs to troubleshoot.

build/constants.yaml
william-msft: Correct. There is no simple approach on saving the comments. The good thing is that this should only be modified by Automation

.github/workflows/main.yaml
```diff
@@ -0,0 +1,71 @@
+name: CI
+
+on:
+  # Run every day at 10 pm
```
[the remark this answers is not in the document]
william-msft: Per discussion, we'll run twice a day for now

build/constants.yaml
pauld-msft: That isn't the case though, right? Whenever we want to create a new constant that spans our scripts and go files and csharp projects, this is the place we want to create that constant so its consistent, which would require manually editing this file? Also, we keep a lot of settings and environment variable information in here as well, which could require manual editing in the future unless I"m mistaken

.github/workflows/main.yaml
pauld-msft: I've noticed that many times sourcetree will mark a regenerated file as modified when unstaged, and then when I go to stage it then it realizes there are no updates and the file disappears. How many of those changes are meaningful?

.github/workflows/main.yaml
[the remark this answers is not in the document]
pauld-msft: Changing from `\` to `/` in the constants yaml file has no impact on the location that the file is placed when generateConstants.sh is run, and all projects still build. I was thinking that it would also fix this issue, but are you saying that it doesn't?

build/constants.yaml
william-msft: Yeah perhaps there can still be minor editing after all automation is still in place. If missing comments are critical, I curretnly see a couple options which would be to have a separate doc with comments for this, or we can have 2 constants.yaml one that can be manually updated with comments and one that is automatically updated without comments.

.github/workflows/main.yaml
william-msft: I've switched to add all files since there might be times when we want to add files other than the two specified folder such as  `build/`. For example when .Net8 rolls out, more files such as `build/__dotNetCoreSdkVersions.sh` and `build/dotNetCoreRunTimeVersions.sh` will automatically get updated

.github/workflows/main.yaml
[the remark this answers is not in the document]
william-msft: Sorry Paul I didn't see both of your messages, perhaps the page didn't refresh. But I updated the slashes and it works fine now. Thanks!

.editorconfig
```diff
@@ -0,0 +1,22 @@
...
```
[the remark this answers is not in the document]
cormacpayne: I still don't think we should have this file as it could be affecting the rules we have in place for other projects set by StyleCop. I checked locally and if we remove this file, you'd need to do three things:

- Add the [`<DisableDocRequirement>`](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/BuildScriptGenerator.csproj#L17) tag to your `Automation.csproj` file to dismiss the errors about the XML comments
- Add inline attribute suppressions to the file complaining about the class types not matching the file name (the file where you have all of the nested deserialization classes)
- Clean up the additional errors that remain (curly braces on a new line, empty lines needing to be removed, etc.)

build/tools/Automation/Request.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            Console.WriteLine($"url: {url}");
```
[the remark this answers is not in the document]
cormacpayne: If we're going to keep the `Console.WriteLine` calls, we should make them more descriptive so we're aware of where we are in the code and what the context is. For example:

```cs
Console.WriteLine($"[GetRequestStringAsync] Making GET request against provided URL '{url}'.");
```

build/tools/Automation/Request.cs
```diff
@@ -0,0 +1,35 @@
...
+                response.EnsureSuccessStatusCode();
+                string responseBody = await response.Content.ReadAsStringAsync();
+
+                return responseBody ?? string.Empty;
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine("\nException Caught!");
+                Console.WriteLine("Message :{0} ", e.Message);
+                return string.Empty;
```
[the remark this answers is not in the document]
cormacpayne: I still feel this comment applies, but let me know what you think.

build/tools/Automation/Program.cs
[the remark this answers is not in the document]
cormacpayne: Sure -- ideally if we kept them, the messages would be a bit more descriptive so we had more context and knew where in the code we were calling from, but if there isn't time to clean that up, then it's OK.

build/tools/Automation/Program.cs
```diff
@@ -0,0 +1,145 @@
...
+
+        /// <Summary>
+        /// Updates:
+        ///     - Constants.ConstantsYaml
+        ///
+        /// Deserializes Constants.ConstantsYaml for platforms, that have new releases, can update.
+        /// </Summary>
+        public static async Task<List<Constant>> DeserializeConstantsYamlAsync()
+        {
+            string fileContents = await File.ReadAllTextAsync(Constants.ConstantsYaml).ConfigureAwait(true);
```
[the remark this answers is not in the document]
cormacpayne: I still feel we should take a look at this since file I/O can be tricky across different OS's and when we're running this from an executable.

build/tools/Automation/Program.cs
```diff
@@ -0,0 +1,145 @@
...
+            var deserializer = new DeserializerBuilder()
+                .WithNamingConvention(UnderscoredNamingConvention.Instance)
+                .Build();
+            var yamlContents = deserializer.Deserialize<List<Constant>>(fileContents);
+            return yamlContents;
+        }
+
+        public static bool DatesMatch(string dateTarget, string dateReleased)
```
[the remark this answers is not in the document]
cormacpayne: This comment still applies -- curious why we need this method.

build/tools/Automation/Program.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// <param name="platformConstants">List of PlatformConstant containing platform release information</param>
+        /// <param name="yamlConstants">Deserialized Constants.ConstantsYaml which is ready for editing</param>
+        /// </Summary>
+        public abstract void UpdateConstants(List<PlatformConstant> platformConstants, List<Constant> yamlConstants);
+
+        /// <Summary>
+        /// Stores platform release information so it can be referenced when updating:
+        /// constants.yaml and versionsToBuild.txt files
+        /// </Summary>
+        public class PlatformConstant
```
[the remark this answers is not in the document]
cormacpayne: This comment still applies, but it's something that we can take a look at refactoring in the next phase.

build/tools/Automation/DotNet/DotNet.cs
cormacpayne: I feel this is probably something we should prioritize for now and get it out of the way to not leave us vulnerable for missing releases.

.github/workflows/main.yaml
```diff
@@ -0,0 +1,56 @@
+name: CI
+
+on:
+  # runs on 12:55 (5:55AM PST) & 23:55 UTC
+  schedule:
+    - cron:  '55 12,23 * * *'
```
cormacpayne, line 6: Just to be 100% safe and ensure we're running each 12 hours a part, we should either run the first flow at 11:55 UTC or the second flow at 00:55 UTC

cormacpayne (commented):
Just one additional comment to go with the responses I made to a few other comments.

build/tools/Automation/DotNet/DotNet.cs
william-msft: Updated

build/tools/Automation/Program.cs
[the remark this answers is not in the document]
william-msft: I updated with the function name that is calling them

build/tools/Automation/DotNet/DotNet.cs
william-msft: Updated: https://github.com/microsoft/Oryx/pull/1560/commits/fdead7e4bf24cd498f6876ae9916cd12cbeef86e#diff-947811409334f4664b049add37fc8940c9c822b2cd039cd24b3d9a7cd410716dR200

One way this falls short is if we have .NET version 100+ 
I considered appending the releaseDate if the major version is above 9 but it seems unnecessary.

build/tools/Automation/DotNet/DotNet.cs
william-msft: Updated with regex.

.editorconfig
```diff
@@ -0,0 +1,20 @@
+﻿# Default severity for analyzer diagnostics with category 'StyleCop.CSharp.OrderingRules'
```
cormacpayne, line 1: Are we not able to remove this file?

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,241 @@
...
+                }
+            }
+
+            var serializer = new SerializerBuilder()
+                .WithNamingConvention(UnderscoredNamingConvention.Instance)
+                .Build();
+
+            var constantsYamlAbsolutePath = Path.Combine(this.repoAbsolutePath, "build", Constants.ConstantsYaml);
+            var stringResult = serializer.Serialize(yamlConstants);
+            File.WriteAllText(Constants.ConstantsYaml, stringResult);
```
cormacpayne, line 167: We should be using `constantsYamlAbsolutePath` here

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -0,0 +1,241 @@
...
+            foreach (string versionsToBuildTxtFile in versionsToBuildTxtFiles)
+            {
+                string line = $"\n{platformConstant.Version}, {platformConstant.Sha},";
+                File.AppendAllText(versionsToBuildTxtFile, line);
+
+                // sort
+                Console.WriteLine($"[UpdateVersionsToBuildTxt] Updating {versionsToBuildTxtFile}...");
+                var contents = File.ReadAllLines(versionsToBuildTxtFile);
+                Array.Sort(contents);
+                File.WriteAllLines(versionsToBuildTxtFile, contents.Distinct());
```
cormacpayne, line 189: We need to make sure `versionsToBuildTxtFile` is prepended with `repoAbsolutePath`

cormacpayne (dismissed):
A few minor comments, otherwise LGTM :shipit:

build/tools/Automation/DotNet/DotNet.cs
william-msft: Updated

cormacpayne (approved):
:shipit:

PR #1567, opened 2022-09-14
Add Environment variable to setup yarn timeout config

qianz2:

### DIFF
--- a/doc/configuration.md
+++ b/doc/configuration.md
@@ -43,7 +43,7 @@ ENABLE\_NODE\_MONOREPO\_BUILD| Apply node monorepo build if repo indicates it   
 COMPRESS\_DESTINATION\_DIR   | Indicates if the entire output directory needs to be compressed.   | ""      | `false` | `true`, `false`
 PRUNE\_DEV\_DEPENDENCIES     | Only the prod dependencies are copied to the output for Node apps. | ""      | `false` | `true`, `false`
 NPM\_REGISTRY\_URL           | Specify the npm registry url.                                | ""      | "http://foobar.com/"
-CONFIG\_YARN\_TIMEOUT           | Specify the yarn timeout config with a delay in milliseconds.                                | ""      | "600000"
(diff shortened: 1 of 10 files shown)


src/BuildScriptGenerator/Node/NodeConstants.cs
```diff
@@ -35,6 +35,7 @@ public static class NodeConstants
         public const string HugoVersionCommand = "echo Using Hugo version: && hugo version";
         public const string YarnPackageInstallCommand = "yarn install --prefer-offline";
         public const string Yarn2PackageInstallCommand = "yarn workspaces focus --all || yarn install";
+        public const string YarnTimeoutConfigCommandTemplate = "yarn config set network-timeout {0} -g";
```
cormacpayne, line 38: nit: is this used anywhere?

doc/configuration.md
```diff
@@ -43,6 +43,8 @@ ENABLE\_NODE\_MONOREPO\_BUILD| Apply node monorepo build if repo indicates it
 COMPRESS\_DESTINATION\_DIR   | Indicates if the entire output directory needs to be compressed.   | ""      | `false` | `true`, `false`
 PRUNE\_DEV\_DEPENDENCIES     | Only the prod dependencies are copied to the output for Node apps. | ""      | `false` | `true`, `false`
 NPM\_REGISTRY\_URL           | Specify the npm registry url.                                | ""      | "http://foobar.com/"
+CONFIG\_YARN\_TIMEOUT           | Specify the yarn timeout config with a delay in milliseconds.                                | ""      | "600000"
```
cormacpayne, line 46: It looks like we're reading `YARN_TIMEOUT_CONFIG` in the code, but the docs here say otherwise -- just want to make sure we're both are pointing to the same environment variable.

cormacpayne (changes requested):
Two minor comments, one of which we should fix, otherwise LGTM :shipit:

doc/configuration.md
qianz2: Updated

cormacpayne (approved):
:shipit:

PR #1568, opened 2022-09-15
Snehapar/1599029 add support for dotnet7 rc1

snehapar9:
[User Story 1599029](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1599029): Add support for .NET 7 RC1, .NET 6.0.9, .NET Core 3.1.29


doc/supportedPlatformVersions.md
```diff
@@ -139,6 +141,7 @@
 - 7.0.100-preview.5.22307.18
 - 7.0.100-preview.6.22352.1
 - 7.0.100-preview.7.22377.5
+- 7.0.100-rc.1.22431.12
```
pauld-msft, line 144: If we are not going to release for stretch, we should create a constant that defines .NET 7's final stretch version. This constant will need to be used to update tests like [this](https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs#L571-L575) that use a stretch based image. 

See [this pull request](https://github.com/microsoft/Oryx/pull/1539) for an example of this being done for .NET 3.1 and .NET 6

cormacpayne (dismissed):
Agree with Paul's comment, otherwise LGTM :shipit:

build/constants.yaml
```diff
@@ -109,6 +109,10 @@
     final-stretch-dot-net-60-sdk-version: 6.0.301
     final-stretch-dot-net-core-app-60-run-time-version: 6.0.6
     final-stretch-aspnet-core-app-60-run-time-version: 6.0.6
+
+    final-stretch-dot-net-60-sdk-version: 7.0.100-preview.6.22352.1
```
pauld-msft, line 113: can you update these to be `final-stretch-dot-net-70-sdk-version` etc. Having the same name seems to be overwriting the actual .NET 6 constants.

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
```diff
@@ -55,7 +55,7 @@ public DotNetCoreDynamicInstallTest(ITestOutputHelper output) : base(output)
         [InlineData(NetCoreApp21WebApp, "2.1")]
         [InlineData(NetCoreApp31MvcApp, "3.1")]
         [InlineData(NetCoreApp50MvcApp, "5.0")]
-        [InlineData(NetCore7PreviewMvcApp, "7.0.0-preview.7.22375.6")]
+        [InlineData(NetCore7PreviewMvcApp, "7.0.100-rc.1.22431.12")]
```
pauld-msft, line 58: stretch image, needs to use the constant we are creating. Same with the github action test right above

build/constants.yaml
snehapar9: Thanks for pointing it out!

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
snehapar9: Makes sense! For my own clarity -- these are the images that would have dotnet 7 prebaked and would not use dynamic installation. Correct?

pauld-msft: @snehapar9 The CLI images are the opposite, they only use dynamic installation.

I'm seeing `dotnet-7.0.100-preview.7.22377.5.tar.gz` in the [dev storage account](https://oryxsdksdev.blob.core.windows.net/dotnet?restype=container&comp=list&include=metadata)

build/constants.yaml
snehapar9: Makes sense! I'm probably missing something -- I don't remember off the top of my head, but I think for dotnet 7 preview 7, we decided to delete versionsToBuild.txt for stretch after publishing dotnet 7 preview 7 sdks to the dev storage account. I think this is the reason why we are seeing these sdks in the dev storage account.

pauld-msft: we did that for .NET 3.1 and .NET 6, which is why I released the PR with those final stretch versions. Looks like we did not remove the preview 7 version from [dotnet stretch versionsToBuild.txt](https://github.com/microsoft/Oryx/blob/main/platforms/dotnet/versions/stretch/versionsToBuild.txt#L95), which is why we see it in the dev storage account.

pauld-msft (approved):
:shipit:

qianz2 (approved):

PR #1575, opened 2022-09-20
reverted base image to Buster for Node

waliMSFT:


build/buildRunTimeImageBases.sh
```diff
@@ -13,11 +13,11 @@ source $REPO_DIR/build/__variables.sh
 source $REPO_DIR/build/__functions.sh
 source $REPO_DIR/build/__nodeVersions.sh
 
-declare -r NODE_BULLSEYE_VERSION_ARRAY=($NODE18_VERSION $NODE16_VERSION $NODE14_VERSION)
+declare -r NODE_BUSTER_VERSION_ARRAY=($NODE18_VERSION $NODE16_VERSION $NODE14_VERSION)
```
daniv-msft, line 16: Here this would also transition Node 18 to Buster, whereas we need this one to stick to Bullseye, I believe you'll need to keep the two things in parallel in the end.

qianz2 (approved):

PR #1580, opened 2022-09-22
Add performance flags to Python build script

tonybaloney:
See https://docs.python.org/3/using/configure.html#performance-options

This enables PGO and LTO, which improves runtime performance by about 20%.

These flags are already enabled for "official" binaries from python.org.

It does also extend the build time by a couple of minutes, because it has to run the profiler to optimize the binary.

### DIFF
--- a/platforms/python/prereqs/build.sh
+++ b/platforms/python/prereqs/build.sh
@@ -78,6 +78,8 @@ else
         --build=$(dpkg-architecture --query DEB_BUILD_GNU_TYPE) \
         --enable-loadable-sqlite-extensions \
         --enable-shared \
+	--enable-optimizations \
+	--with-lto \
         --with-system-expat \
         --with-system-ffi \
         --without-ensurepip


daniv-msft:
Thanks @tonybaloney for opening this PR! I just triggered our Validation pipeline to run our tests on this: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6748086&view=results

daniv-msft:
@tonybaloney Thanks again for opening this PR. @btardif and I were discussing about the impact of this change this morning, and especially the impact in terms of build time. We will schedule a separate discussion about this, but it would be great if you could clarify the couple of minutes impact: how much time are we talking about?

btardif:
tagging @JasonFreeberg so that he has more context here

daniv-msft (approved):
Just realized this one wasn't merged despite our discussion a couple of weeks (month?) back. Merging it

PR #1582, opened 2022-09-23
Fix issue with custom requirements.txt path not being used

cormacpayne:
Fix for IcM 336712282 -- the `CUSTOM_REQUIREMENTSTXT_PATH` environment variable is used to notify Oryx that a different `requirements.txt` file should be used to build a Python application rather than the root `requirements.txt` file, but is currently not being set as a part of the build script we generate.

This PR ensures that the path provided to the environment variable exists within the repository and is passed to the build script, as well as adds test coverage to ensure that this path is included in the build script.


src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -243,7 +243,16 @@ public PlatformDetectorResult Detect(RepositoryContext context)
                     = compressedVirtualEnvFileName;
             }
 
-            this.TryLogDependencies(pythonVersion, context.SourceRepo);
+            var customRequirementsTxtPath = this.pythonScriptGeneratorOptions.CustomRequirementsTxtPath;
+            var requirementsTxtPath = customRequirementsTxtPath == null ? PythonConstants.RequirementsFileName : customRequirementsTxtPath;
+            if (!context.SourceRepo.FileExists(requirementsTxtPath))
```
daniv-msft, line 248: Could you please confirm that having a requirements.txt file is actually required for all Python applications?

daniv-msft: The code we had previously wasn't throwing an exception, and was just exiting TryLogDependencies instead

cormacpayne: Ah good catch -- let me clean that up.

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -243,7 +243,12 @@ public PlatformDetectorResult Detect(RepositoryContext context)
                     = compressedVirtualEnvFileName;
             }
 
-            this.TryLogDependencies(pythonVersion, context.SourceRepo);
+            var customRequirementsTxtPath = this.pythonScriptGeneratorOptions.CustomRequirementsTxtPath;
+            var requirementsTxtPath = customRequirementsTxtPath == null ? PythonConstants.RequirementsFileName : customRequirementsTxtPath;
+            if (context.SourceRepo.FileExists(requirementsTxtPath))
+            {
+                this.TryLogDependencies(requirementsTxtPath, pythonVersion, context.SourceRepo);
+            }
```
daniv-msft, line 251: Here I'm asking for a behavioral change, so feel free to ignore if you believe that would be too risky. Would it still make sense to throw an exception if the env variable was set but that the file doesn't exist? We could argue that if a customer set the environment variable, then he/she WANTS this file to be taken into account, and that it would be better to fail early than just silently ignoring this.

daniv-msft (dismissed):

src/BuildScriptGenerator/Python/PythonPlatform.cs
cormacpayne: Discussed offline -- moved to throw an exception when `CUSTOM_REQUIREMENTSTXT_PATH` is provided but isn't a valid path to fail-fast for the user and given them an opportunity to update the environment variable to the correct path, or scrap it altogether.

daniv-msft (dismissed):

daniv-msft (dismissed):

daniv-msft (approved):

PR #1587, opened 2022-09-26
Split Python integration tests, remove Ruby from integration test job

cormacpayne:
This PR looks to resolve work item 1632680.

To help with the disk space issues we're seeing in pipelines that runs integration tests, this PR does the following:

- Splits up the existing Python integration tests based on their versions, ensuring that all of the runtimes won't be pulled down during the "Pull and re-tag" job, which has been using additional disk space for unused images
- Removes Ruby from being a part of the integration test job since it currently has no tests run in this suite and splitting it up into three separate flows based on the version tested would use three additional workers for no additional testing


daniv-msft (approved):
Looks good to me! I didn't have any specific comments.

qianz2 (approved):

PR #1588, opened 2022-09-27
Revert "Remove stretch and buster multi-stage tests"

daniv-msft:
Reverts microsoft/Oryx#1581

This was added temporarily to unblock our pipelines due to size constraints. Now that the better fix was pushed by @cormacpayne, we should be good to add these tests back. + @pauld-msft for FYI.


daniv-msft:
Closing as discussed with @cormacpayne

PR #1592, opened 2022-09-29
Split integration tests by platform version and build image

cormacpayne:
Integration tests run in a container that has pulled down all previously built build images, as well as generates new test images from these build images, which use up a sizeable amount of disk space in the container. This PR aims to split the integration tests to run based on both the platform version and build image targeted in the test, allowing for the containers to only pull down and build the necessary images within the container.

From trial runs of the Nightly build, this process will add about 30-35m to the pipeline execution time as a large amount of agents are required for all of the combinations of platform versions and build images, and there is a limit to how many we're able to use at a given moment, so some processes become serial rather than parallel.


daniv-msft:
That's a lot of manual changes! I understand why it's surprising only two tests didn't work afterwards. :)

daniv-msft (dismissed):

daniv-msft (dismissed):

daniv-msft (approved):

PR #1593, opened 2022-09-29
Add support for bindPort2 parameter for .NET

qianz2:

Oryx side change:
•        Surface a new optional -bindPort2 parameter.
•        When this parameter is set, pass the following env variables to the .NET 7 process:
o   Kestrel__Endpoints__Http2__Url=http://*:8585
(Port value here comes from the -bindPort2 parameter)
o   Kestrel__Endpoints__Http2__Protocols=Http2
o   Kestrel__Endpoints__Http__Url=http://*:8080
(Port value here comes from the -bindPort parameter)
o   Kestrel__Endpoints__Http__Protocols=Http1
•        We won’t change the existing behavior if -bindPort2 is not provided.

### DIFF
--- a/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
+++ b/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
@@ -69,6 +69,10 @@ namespace Microsoft.Oryx.BuildScriptGenerator
                 {
                     createScriptArguments.Add("bindPort", this.commonOptions.BindPort);
                 }
+                if (!string.IsNullOrEmpty(this.commonOptions.BindPort2))
+                {
+                    createScriptArguments.Add("bindPort2", this.commonOptions.BindPort2);
+                }
 
                 var compatiblePlatforms = this.GetCompatiblePlatforms(ctx);
                 if (!compatiblePlatforms.Any())
--- a/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
+++ b/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator
                 {
                     createScriptArguments.Add("bindPort", this.commonOptions.BindPort);
                 }
-                
+
                 if (!string.IsNullOrEmpty(this.commonOptions.BindPort2))
                 {
                     createScriptArguments.Add("bindPort2", this.commonOptions.BindPort2);
--- a/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
+++ b/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
@@ -69,6 +69,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator
                 {
                     createScriptArguments.Add("bindPort", this.commonOptions.BindPort);
                 }
+                
                 if (!string.IsNullOrEmpty(this.commonOptions.BindPort2))
                 {
                     createScriptArguments.Add("bindPort2", this.commonOptions.BindPort2);
--- a/src/startupscriptgenerator/src/dotnetcore/main.go
+++ b/src/startupscriptgenerator/src/dotnetcore/main.go
@@ -41,7 +41,7 @@ func main() {
 		"[Optional] Path to the directory where build manifest file can be found. If no value is provided, then "+
 			"it is assumed to be under the directory specified by 'appPath'.")
 	bindPortPtr := scriptCommand.String("bindPort", "", "[Optional] Port where the application will bind to. Default is 8080")
-	bindPort2Ptr := scriptCommand.String("bindPort2", "", "[Optional] .NET only - HTTP/2 port where the application will bind to. By default, no binding will be made if this is not set")
+	bindPort2Ptr := scriptCommand.String("bindPort2", "", "[Optional] .NET only - HTTP/2 port where the application will bind to. By default, no HTTP/2 binding will be made")
 	userStartupCommandPtr := scriptCommand.String(
 		"userStartupCommand",
 		"",
--- a/src/startupscriptgenerator/src/dotnetcore/main.go
+++ b/src/startupscriptgenerator/src/dotnetcore/main.go
@@ -41,7 +41,7 @@ func main() {
 		"[Optional] Path to the directory where build manifest file can be found. If no value is provided, then "+
 			"it is assumed to be under the directory specified by 'appPath'.")
 	bindPortPtr := scriptCommand.String("bindPort", "", "[Optional] Port where the application will bind to. Default is 8080")
-	bindPort2Ptr := scriptCommand.String("bindPort2", "", "[Optional] Port 2 where the application will bind to.")
+	bindPort2Ptr := scriptCommand.String("bindPort2", "", "[Optional] .NET only - HTTP/2 port where the application will bind to. By default, no binding will be made if this is not set")
 	userStartupCommandPtr := scriptCommand.String(
 		"userStartupCommand",
 		"",
--- a/src/startupscriptgenerator/src/dotnetcore/main.go
+++ b/src/startupscriptgenerator/src/dotnetcore/main.go
@@ -41,6 +41,7 @@ func main() {
 		"[Optional] Path to the directory where build manifest file can be found. If no value is provided, then "+
 			"it is assumed to be under the directory specified by 'appPath'.")
 	bindPortPtr := scriptCommand.String("bindPort", "", "[Optional] Port where the application will bind to. Default is 8080")
+	bindPort2Ptr := scriptCommand.String("bindPort2", "", "[Optional] Port where the application will bind to. Default is 8080")
 	userStartupCommandPtr := scriptCommand.String(
 		"userStartupCommand",
 		"",
@@ -118,6 +119,7 @@ func main() {
 			AppPath:            fullAppPath,
 			RunFromPath:        fullRunFromPath,
 			BindPort:           *bindPortPtr,
+			BindPort2:          *bindPort2Ptr,
 			UserStartupCommand: *userStartupCommandPtr,
 			DefaultAppFilePath: fullDefaultAppFilePath,
 			Manifest:           buildManifest,
--- a/src/startupscriptgenerator/src/dotnetcore/main.go
+++ b/src/startupscriptgenerator/src/dotnetcore/main.go
@@ -41,7 +41,7 @@ func main() {
 		"[Optional] Path to the directory where build manifest file can be found. If no value is provided, then "+
 			"it is assumed to be under the directory specified by 'appPath'.")
 	bindPortPtr := scriptCommand.String("bindPort", "", "[Optional] Port where the application will bind to. Default is 8080")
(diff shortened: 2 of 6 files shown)


src/startupscriptgenerator/src/dotnetcore/main.go
```diff
@@ -41,6 +41,7 @@ func main() {
 		"[Optional] Path to the directory where build manifest file can be found. If no value is provided, then "+
 			"it is assumed to be under the directory specified by 'appPath'.")
 	bindPortPtr := scriptCommand.String("bindPort", "", "[Optional] Port where the application will bind to. Default is 8080")
+	bindPort2Ptr := scriptCommand.String("bindPort2", "", "[Optional] Port where the application will bind to. Default is 8080")
```
daniv-msft, line 44: I know it's only a draft PR, so you would have probably fixed it anyway, but here the description should be updated. I'd recommend something like: "[Optional] .NET only - HTTP/2 port where the application will bind to. By default, no binding will be made if this is not set"

qianz2: Updated

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -69,6 +69,39 @@ func (gen *DotnetCoreStartupScriptGenerator) shouldApplicationInsightsBeConfigur
...
+	
+	if dotNetRuntimeVersion != "" {
+		dotNetRuntimeVersionConstraint, err := semver.NewConstraint(">= " + versionConstraint)
+		if err != nil {
+    		fmt.Printf("\nError in creating semver constraint %s", err)
+		}
+
+		dotNetCurrentVersion, err := semver.NewVersion(dotNetRuntimeVersion)
+		if err != nil {
+    		fmt.Printf("\nError in parsing current version to semver version %s", err)
```
cormacpayne, line 92: nit: same two comments here as above

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -69,6 +69,39 @@ func (gen *DotnetCoreStartupScriptGenerator) shouldApplicationInsightsBeConfigur
...
+		}
+		// Check if the version meets the constraints. The a variable will be true.
+		constraintCheckResult := dotNetRuntimeVersionConstraint.Check(dotNetCurrentVersion)
+				
+	    if constraintCheckResult  {
+			   fmt.Printf("\nBefore returning true")
```
cormacpayne, line 98: nit: can we remove this message? In addition, if we don't need to log any additional information, then I'd replace this `if constraintCheckResult` block with the following:

```go
return constraintCheckResult
```

cormacpayne (dismissed):
Minor comments, otherwise LGTM :shipit:

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
qianz2: Updated

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -69,6 +69,39 @@ func (gen *DotnetCoreStartupScriptGenerator) shouldApplicationInsightsBeConfigur
...
+	} else {
+		dotNetRuntimeVersion = os.Getenv("DOTNET_VERSION")
+	}
+	
+	fmt.Printf("\nDotNet Runtime %s", dotNetRuntimeVersion)
+	
+	if dotNetRuntimeVersion != "" {
+		dotNetRuntimeVersionConstraint, err := semver.NewConstraint(">= " + versionConstraint)
+		if err != nil {
+    		fmt.Printf("\nError in creating semver constraint %s", err)
```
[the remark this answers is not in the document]
qianz2: Added the return false here. For the indentation not sure why github shows differently with what I see from VS code:
Screenshot 2022-10-05 144922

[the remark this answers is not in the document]
cormacpayne: Hmm I've seen GitHub do this before -- probably some weird tab vs. spaces display thing in their UI 🤔

cormacpayne (dismissed):

src/startupscriptgenerator/src/dotnetcore/main.go
```diff
@@ -41,6 +41,7 @@ func main() {
 		"[Optional] Path to the directory where build manifest file can be found. If no value is provided, then "+
 			"it is assumed to be under the directory specified by 'appPath'.")
 	bindPortPtr := scriptCommand.String("bindPort", "", "[Optional] Port where the application will bind to. Default is 8080")
+	bindPort2Ptr := scriptCommand.String("bindPort2", "", "[Optional] .NET only - HTTP/2 port where the application will bind to. By default, no binding will be made if this is not set")
```
daniv-msft, line 44: Minor comment: I know I'm the one who suggested that particular sentence, but please consider updating to: `By default, no HTTP/2 binding will be made`.
If this is the only comment preventing you from merging, not a big deal, feel free to ignore.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -81,6 +113,15 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
 	common.SetEnvironmentVariableInScript(scriptBuilder, "PORT", gen.BindPort, DefaultBindPort)
 	scriptBuilder.WriteString("export ASPNETCORE_URLS=http://*:$PORT\n\n")
 
+	logger.LogInformation("Setting up Kestrel Endpoints with BindPort and BindPort2 Env variables if .NET runtime version >= 7")
```
daniv-msft, line 116: I actually don't believe we need to restrict it to .NET 7. I believe the info in the email was to limit it to .NET 7 if it was easier, but that fundamentally we wanted it for all .NET versions. Could you please clarify: is there any way to limit it to .NET 7?

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -81,6 +113,15 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
+	if gen.isDotnetRuntimeVersionMeetConstraint("7.0.0-0") {
+		scriptBuilder.WriteString("if [ -z \"$" + gen.BindPort2 + "\" ]; then" + "\n")
+		scriptBuilder.WriteString("		export Kestrel__Endpoints__Http2__Url=http://*:" + gen.BindPort2 + "\n")
+		scriptBuilder.WriteString("		export Kestrel__Endpoints__Http2__Protocols=Http2\n")
+		scriptBuilder.WriteString("		export Kestrel__Endpoints__Http1__Url=http://*:$PORT\n\n")
```
daniv-msft, line 121: The $PORT value is actually expected to be replaced by the value provided in the BindPort parameter, if I remember correctly.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -81,6 +113,15 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
 	common.SetEnvironmentVariableInScript(scriptBuilder, "PORT", gen.BindPort, DefaultBindPort)
 	scriptBuilder.WriteString("export ASPNETCORE_URLS=http://*:$PORT\n\n")
 
+	logger.LogInformation("Setting up Kestrel Endpoints with BindPort and BindPort2 Env variables if .NET runtime version >= 7")
+	if gen.isDotnetRuntimeVersionMeetConstraint("7.0.0-0") {
+		scriptBuilder.WriteString("if [ -z \"$" + gen.BindPort2 + "\" ]; then" + "\n")
```
daniv-msft, line 118: We should also validate that BindPort is set as well, as we'll need to read its value to set Kestrel__Endpoints__Http1__Url

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
daniv-msft: Oh my bad, I actually just saw that we were doing this in the file before:
`common.SetEnvironmentVariableInScript(scriptBuilder, "PORT", gen.BindPort, DefaultBindPort)`
So the env variable should already be set properly there.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
daniv-msft: Or the $PORT value if it's easier, as it's already defined above?

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
qianz2: Yes we're actually limited to .NET 7 with the function at line 118. My understanding from the email thread they're asking to scope to .NET 7 first, then .NET 6.

`if gen.isDotnetRuntimeVersionMeetConstraint("7.0.0-0") {`

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
qianz2: The $PORT value was setup here, and it has default value of 8080:
`	common.SetEnvironmentVariableInScript(scriptBuilder, "PORT", gen.BindPort, DefaultBindPort)
`

src/startupscriptgenerator/src/dotnetcore/main.go
qianz2: Updated

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
qianz2: Or we should remove this line? Please let me know if I understand it wrongly.

daniv-msft: Yes, it's ambiguous. Jenny and I mentioned that we should support all .NET versions if it's possible, but Surender mentioned that having only .NET 7 would be better to reduce the blast radius.

Let's keep what you have, and we can remove the line as needed in the future.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
daniv-msft: OK fair enough! Thanks for clarifying

daniv-msft (approved):

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
```diff
@@ -70,6 +70,11 @@ public string GenerateDockerfile(DockerfileContext ctx)
                     createScriptArguments.Add("bindPort", this.commonOptions.BindPort);
                 }
 
+                if (!string.IsNullOrEmpty(this.commonOptions.BindPort2))
```
surenderssm, line 73: NIT: Checking for IsNullOrWhiteSpace might add more value here.

PR #1595, opened 2022-09-30
Update appservice.md to mention the AzureFrontDoor.Frontend service tag

daniv-msft:
Update the AppService documentation to mention the service tag allowing to allow outbound access to the Oryx CDN.


daniv-msft:
Closing for now, as I believe this guidance might actually not be correct.

PR #1604, opened 2022-10-12
Moved cert env so it's included in the final docker layer

snehapar9:
Created this PR to move PATH_CA_CERTIFICATE to template.Dockerfile in dotnetcore, php, python and node so it's included in the final docker layer.


daniv-msft (dismissed):

images/runtime/node/template.Dockerfile
```diff
@@ -12,15 +12,15 @@ ENV RELEASE_TAG_NAME=${RELEASE_TAG_NAME}
 ENV GIT_COMMIT=${GIT_COMMIT}
 ENV BUILD_NUMBER=${BUILD_NUMBER}
 #Bake in client certificate path into image to avoid downloading it
```
daniv-msft, line 14: Could you please remove this comment? #Bake in client certificate path into image to avoid downloading it

cormacpayne (dismissed):
:shipit:

images/runtime/node/template.Dockerfile
snehapar9: Done. Any reason why it may be good to remove it?

daniv-msft: Oh sorry I'm now looking at the code change: I didn't mean to remove the comment everywhere. Just the one that wasn't referring to anything.

snehapar9: Got it. Thanks :)

cormacpayne (approved):
:shipit:

daniv-msft (approved):

PR #1605, opened 2022-10-12
Wali/1638426 add suupport for dotnet7 rc2

waliMSFT:
add support for dotnet 7 cr2


build/__dotNetCoreRunTimeVersions.sh
```diff
@@ -26,7 +26,7 @@ NET_CORE_APP_60='6.0.9'
 NET_CORE_APP_60_SHA='a6df2cfef73047247bd36f51eaf696f616c6aa9b428e42f219bf91dcf05c03dff817a8ec826740002c8aa83df2fce8a7ace562ad2e2956789542f0b8ab8b1173'
```
daniv-msft, line 26: Should we also update `DOT_NET_CORE_RUNTIME_BASE_TAG` (line 3) in this file? I saw that it was updated in previous PRs to update .NET 7.
Similar comment for src/BuildScriptGenerator/DotNetCore/DotNetCoreRunTimeVersions.cs

pauld-msft: @waliMSFT This should be automatically updated if you just run the `generateConstants.sh` script

waliMSFT: Yah, this should update automatically. I ran the `generateConstants.sh` scripts. let's see why it did not get updated.

waliMSFT: yah, `generateConstants.sh` fixes this issue. By following the doc, I ran the `generateConstants.sh` before updating the tag and which is why it is not updated here. It would be better to update the doc to address this.

daniv-msft (approved):

build/__dotNetCoreRunTimeVersions.sh
daniv-msft: @waliMSFT Thanks, looks good to me. Could you please update the doc so that moving forward we don't get hit by this again?

qianz2 (approved):

PR #1620, opened 2022-10-20
Williamhe/1578887 update pipeline cron

william-msft:
.NET team recommended we query for new SDKs at noon.


.github/workflows/main.yaml
```diff
@@ -1,9 +1,9 @@
 name: Oryx-SDK-Automation
 
 on:
-  # runs on 12:55 (5:55AM PST) & 23:55 UTC
+  # runs at 12pm PST per .NET team recommendation
```
pauld-msft, line 4: This means that if an SDK is released after 12pm PST on any given day, then our automation won't catch it, correct? Did the .NET team guarantee that it will always be released before then / did they understand that if it doesn't detect it at noon then it won't detect it at all?

pauld-msft (approved):
Just want to confirm the dotnet team guarantees that we will not miss any releases, as this time is less forgiving. If so, then LGTM

.github/workflows/main.yaml
cormacpayne: Depends on how our logic handles previous day releases -- I remember having a conversation with @william-msft about how this was implemented, and we ultimately settled on running the function twice a day since it may have difficulty with previous day releases. Is this still the case @william-msft ?

william-msft: Keeping current logic for now (adding checking if version exists in SDK soon)

william-msft:
Keeping current logic for now (adding checking if version exists in SDK soon)

PR #1623, opened 2022-10-21
Add sqlsrv to PHP FPM 8.1 runtime image

cormacpayne:
Regenerate the PHP FPM bullseye runtime images with the latest changes from the template file, included updated logic that will include `sqlsrv` with the PHP FPM 8.1 runtime image. Also include tests to ensure that `sqlsrv` is available across all PHP FPM runtime versions.


images/runtime/php-fpm/8.1/base.bullseye.Dockerfile
```diff
@@ -76,10 +76,7 @@ RUN docker-php-ext-configure pdo_odbc --with-pdo-odbc=unixODBC,/usr \
         xsl
 RUN pecl install redis && docker-php-ext-enable redis
 # https://github.com/Imagick/imagick/issues/331
```
daniv-msft, line 78: Nit: should we remove this comment from the template file? Similar comment for sqlsvr

images/runtime/php-fpm/8.1/base.bullseye.Dockerfile
```diff
@@ -93,7 +90,7 @@ RUN set -eux; \
 #  - https://docs.microsoft.com/en-us/sql/connect/php/installation-tutorial-linux-mac
 #  - https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server
 RUN set -eux; \
-    if [[ $PHP_VERSION == 7.4.* || $PHP_VERSION == 8.0.* ]]; then \
+    if [[ $PHP_VERSION == 7.4.* || $PHP_VERSION == 8.* ]]; then \
```
pauld-msft, line 93: The only PHP versions we have right now are 7.4, 8.0, and 8.1 right? So is there a reason we don't just remove this IF statement?

daniv-msft:
> Regenerate the PHP FPM bullseye runtime images with the latest changes from the template file, included updated logic that will include `sqlsrv` with the PHP FPM 8.1 runtime image. Also include tests to ensure that `sqlsrv` is available across all PHP FPM runtime versions.
> 
> * [x]  The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as #<issue_number>.
> * [x]  Tests are included and/or updated for code changes.
>   ~- [ ] Proper license headers are included in each file.~

Yes, I can see that the PR who made that changes updated the runtime images for PHP, but not for PHP-FPM. Thanks for catching.

images/runtime/php-fpm/8.1/base.bullseye.Dockerfile
```diff
@@ -76,10 +76,7 @@ RUN docker-php-ext-configure pdo_odbc --with-pdo-odbc=unixODBC,/usr \
         xsl
 RUN pecl install redis && docker-php-ext-enable redis
 # https://github.com/Imagick/imagick/issues/331
-RUN set -eux; \
-    if [[ $PHP_VERSION != 8.* ]]; then \
-        pecl install imagick && docker-php-ext-enable imagick; \
-    fi
+RUN pecl install imagick && docker-php-ext-enable imagick
```
daniv-msft, line 80: Naive question: do you know why we don't have a similar change for PHP 8.0?

images/runtime/php-fpm/7.4/base.bullseye.Dockerfile
```diff
@@ -93,7 +90,7 @@ RUN set -eux; \
 #  - https://docs.microsoft.com/en-us/sql/connect/php/installation-tutorial-linux-mac
 #  - https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server
 RUN set -eux; \
-    if [[ $PHP_VERSION == 7.4.* || $PHP_VERSION == 8.0.* ]]; then \
+    if [[ $PHP_VERSION == 7.4.* || $PHP_VERSION == 8.* ]]; then \
```
pauld-msft, line 93: Same comment as above. Also, don't we want to make these changes in [the template.base dockerfiles](https://github.com/microsoft/Oryx/blob/main/images/runtime/php/template.base.Dockerfile#L103-L112) as well? Also in the [php-fpm template.base](https://github.com/microsoft/Oryx/blob/main/images/runtime/php-fpm/template.base.Dockerfile#L89-L97) dockerfiles?

tests/Oryx.RuntimeImage.Tests/Php-fpm/PhpFpmImageTest.cs
```diff
@@ -169,5 +169,30 @@ public void Redis_IsInstalled(string imageTag)
                 result.GetDebugInfo());
 
         }
+
```
daniv-msft, line 172: I see that in the previous PR, https://github.com/microsoft/Oryx/pull/1547/files, some tests were added to validate the change change for Imagick (tests/Oryx.Integration.Tests/Php/PhpImagickExampleTest.cs). I would expect these tests to fail given that the runtime images were not regenerated. Just asking in case you already know: would you have an idea of why this works?

images/runtime/php-fpm/8.1/base.bullseye.Dockerfile
pauld-msft: So I think this is just because Cormac regenerated the bullseye images. PHP-FPM 8.0 is buster based

daniv-msft: Thanks!

daniv-msft (dismissed):

tests/Oryx.RuntimeImage.Tests/Php-fpm/PhpFpmImageTest.cs
cormacpayne: The tests that were added were for the PHP images (not the PHP FPM images), which did have their version-specific Dockerfiles updated in that case.

images/runtime/php-fpm/8.1/base.bullseye.Dockerfile
cormacpayne: I'll update this in a follow-up PR (or if we need to make additional changes to this PR) if that is OK.

tests/Oryx.RuntimeImage.Tests/Php-fpm/PhpFpmImageTest.cs
daniv-msft: I agree the tests were added in the "PHP" section, but they did contain calls to methods such as "PhpFpmImagickExampleAsync". I would have expected this to fail

images/runtime/php-fpm/8.1/base.bullseye.Dockerfile
daniv-msft: Of course, thanks!

images/runtime/php-fpm/generateDockerfiles.sh
```diff
@@ -45,6 +45,6 @@ do
 	sed -i "s|$PHP_VERSION_PLACEHOLDER|$PHP_VERSION|g" "$TARGET_DOCKERFILE_BASE"
 
 	# RUNTIME_BASE_IMAGE_TAG="php-$VERSION_DIRECTORY-fpm-$PHP_FPM_RUNTIME_BASE_TAG-$ImageDebianFlavor"
-    RUNTIME_BASE_IMAGE_TAG="php-$VERSION_DIRECTORY-fpm-$PHP_FPM_RUNTIME_BASE_TAG"
+    RUNTIME_BASE_IMAGE_TAG="php-$VERSION_DIRECTORY-fpm-debian-$ImageDebianFlavor-$PHP_FPM_RUNTIME_BASE_TAG"
```
pauld-msft, line 48: Ah, good catch. The fact that this base image is changed was unintentional. None of the base images were intended to be changed to include the OS type. Unfortunately, php-fpm uses [a txt file](https://github.com/microsoft/Oryx/blob/main/images/runtime/php-fpm/8.1/tag.txt) to determine what the tag is going to be, and I did not account for the fact that this file would also change the base image tag. I believe that this is the only type that does this, so the other bases should continue to not have the os type in their tags. We may want to update all of them to be consistent, but for now this should work

cormacpayne: No worries 👍 I just filed a work item [here](https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1656098) to investigate if this is a problem for other platforms' runtime images.

daniv-msft (approved):

PR #1624, opened 2022-10-23
Node version update 1638365

harryli0108:
Updating node 14 and 16 to the newest version.


pauld-msft (changes requested):
There are a few more things that need to happen for a successful version update.

We have documented the steps in the internal [Oryx OneNote](https://microsoft.sharepoint.com/:o:/t/ProjectOryx/EmnxrbuBwQdAhmH8XSc5gkABDS1hAYSjYPeBWGyUMXyXRg?e=tocIYF), in the section entitled "Oryx - How to's" -> "How to add support for a new platform version". 

Furthermore, there are a few extra steps that need to happen in order to "lock in" the most recent stretch versions, as documented by https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1599477. 

Feel free to reach out to me with any questions regarding these various steps, I know the process is a bit confusing at first.

platforms/nodejs/versions/stretch/versionsToBuild.txt
```diff
@@ -18,6 +18,7 @@
 14.17.6
 14.18.3
 14.19.1
+14.20.1
```
pauld-msft, line 21: So since stretch is deprecated, we don't want to release any more versions of it. So we will want to remove the changes from this file, since it causes the stretch binaries to be uploaded to the storage account

harryli0108: I see. Just reverted the change on stretch. Thanks!

pauld-msft (approved):

PR #1627, opened 2022-10-24
Lock in final stretch versions for node 14 and node 16

pauld-msft:
See https://github.com/microsoft/Oryx/pull/1624#pullrequestreview-1153127032 for context. This required some prior knowledge and should help to unblock that PR


daniv-msft (approved):
Looks good to me. Thanks

daniv-msft (approved):

PR #1631, opened 2022-10-25
Fixed node base image script

pauld-msft:
Looks like [node base image pipeline jobs](https://devdiv.visualstudio.com/DevDiv/_build?definitionId=12627) have been broken for a while now, and they need to work for us to release the new versions. I've tested this locally with `./build/buildRunTimeImageBases.sh node buster/bullseye` and the script no longer fails


cormacpayne (approved):
:shipit:

cormacpayne (approved):

PR #1643, opened 2022-10-31
Pauld/add support for staging runtime images

pauld-msft:
Adds support to stage base runtime and runtime images. Currently this only supports dotnet runtimes, but it should be fairly straightforward to add others in the future if needed.

Docs surrounding the staging of runtime images: https://microsoft-my.sharepoint.com/:w:/p/pauldorsch/Ec-UXhtaBQ5BtjyyGmg95x8B-K53cmpvGtNV64WRSg2Ltw?e=kszqgb


build/__dotNetCoreSdkVersions.sh
```diff
@@ -10,3 +10,4 @@ DOT_NET_50_SDK_VERSION='5.0.408'
 DOT_NET_60_SDK_VERSION='6.0.401'
 DOT_NET_60_RTM_SDK_VERSION='6-0-100-rtm-21527-11'
 DOT_NET_70_SDK_VERSION='7.0.100-rc.2.22477.23'
+RUNTIME_VERSIONS=("1.0" "1.1" "2.0" "2.1" "2.2" "3.0" "3.1" "5.0" "6.0" "7.0")
```
pauld-msft, line 13: Adding lists to shell output created these `RUNTIME_VERSIONS` lists. I realize that they will overwrite each other because they are the same name, but they are never used and so I think that it should be ok

images/runtime/dotnetcore/7.0/base.buster.staging.Dockerfile
```diff
@@ -0,0 +1,67 @@
+# syntax=docker/dockerfile:1.3
```
pauld-msft, line 1: Visual representation of the changes between this file and the non-staging dockerfile
image

images/runtime/dotnetcore/7.0/base.buster.staging.Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+# mount the secret sas token to pull the binaries, and make sure we do not print to docker build logs
+RUN --mount=type=secret,id=DOTNET_PRIVATE_STORAGE_ACCOUNT_ACCESS_TOKEN \
+    set -e \
+    # based on resolution on https://github.com/NuGet/Announcements/issues/49#issue-795386700
+    && apt-get remove ca-certificates -y \
+    && apt-get purge ca-certificates -y \
+    && apt-get update \
+    && apt-get install -f ca-certificates=20200601~deb10u2 -y --no-install-recommends \
+    && . ${BUILD_DIR}/__dotNetCoreRunTimeVersions.sh \
+    && curl -SL --output dotnet.tar.gz https://dotnetcli.blob.core.windows.net/dotnet-private/internal/$NET_CORE_APP_70/dotnet-runtime-$NET_CORE_APP_70-linux-x64.tar.gz$(cat /run/secrets/DOTNET_PRIVATE_STORAGE_ACCOUNT_ACCESS_TOKEN) \
```
pauld-msft, line 53: I originally had code that would rewrite base.buster.Dockerfile based on whether the version was to be staged or not, but honestly there were enough changes like comments and things that this ended up being a more straightforward way to implement. In the future I think we could try using templates or something to swap values of a single file, but with the time crunch for this, I didn't think the added complexity was worth it.

build/buildRunTimeImageBases.sh
```diff
@@ -123,24 +124,45 @@ for dockerFile in $dockerFiles; do
     platformName="${PARTS[0]}"
     platformVersion="${PARTS[1]}"
 
-    # Set $localImageTagName to the following format: oryxdevmcr.azurecr.io/public/oryx/base:{platformName}-{platformVersion}
-    localImageTagName="$BASE_IMAGES_REPO:$platformName-$platformVersion"
+    if shouldStageRuntimeVersion $platformName $platformVersion ; then
```
pauld-msft, line 127: I originally had code that would rewrite base.buster.Dockerfile based on whether the version was to be staged or not, but honestly there were enough changes like comments and things that this ended up being a more straightforward way to implement. In the future I think we could try using templates or something to swap values of a single file, but with the time crunch for this, I didn't think the added complexity was worth it.

images/runtime/python/3.10/bullseye.Dockerfile
```diff
@@ -66,6 +66,12 @@ ENV PATH="/opt/python/3/bin:${PATH}"
 # Bake Application Insights key from pipeline variable into final image
 ARG AI_KEY
 ENV ORYX_AI_INSTRUMENTATION_KEY=${AI_KEY}
+#Bake in client certificate path into image to avoid downloading it
```
pauld-msft, line 69: These python changes are not related to this update, I assume they are from running the build runtime images commands that runs the generate dockerfiles scripts

build/__functions.sh
```diff
@@ -78,3 +78,23 @@ function showDockerImageSizes()
...
+	source $REPO_DIR/build/__stagingRunTimeConstants.sh
+
+	case $platformName in
+	'dotnet'|'dotnetcore')
+		if [[ " ${DOTNETCORE_STAGING_RUNTIME_VERSIONS[*]} " =~ " ${platformRuntimeVersion} " ]]; then
+			return 0
+		fi
+		;;
+	*) 
+		echo "Platform '$platformName' does not support staging."
```
daniv-msft, line 96: I'm mostly curious here: why should a platform "support" staging? What is required for this, and that make it non trivial?
The answer might be in the rest of the PR, so I might delete the comment later.

pauld-msft: It will just need a few things to be added to "support" staging, such as a list constant to the constants.yaml, such as what dotnet has with `dotnetcore-staging-runtime-versions`. There will also need to be updates to their platform specific generateDockerfiles scripts https://github.com/microsoft/Oryx/pull/1643/files#diff-db172e07e3c1672513a6c8ce0c9248f8522602a54038c6459aefb8af98946a39 and probably the build runtime bases script https://github.com/microsoft/Oryx/pull/1643/files#diff-bccc0a4c8d28421a96eefefb692ff896fabcb915f0b2ae9e31cceb9b04de0e48 

So not a ton of work to make all platforms support staging, but non-trivial enough that it seemed out of scope for the immediate effort?

build/__dotNetCoreSdkVersions.sh
daniv-msft, line 13: Could you please clarify where this (and the similar one for other languages) is used?

pauld-msft: Adding support for lists in constants.yaml to be outputted to shell scripts added this: see https://github.com/microsoft/Oryx/pull/1643#discussion_r1009494706

Basically, I needed to be able to create a list of dotnet versions that we want to upload to a staging repo, and unfortunately adding that functionality meant that these lists from the constants.yaml are also created. But per that comment, i don't see them as being much of an issue since they are never used

build/__variables.sh
```diff
@@ -79,6 +80,7 @@ declare -r ACR_BUILD_VSO_FOCAL_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:vso-ubuntu-foc
 declare -r ACR_BUILD_VSO_BULLSEYE_IMAGE_NAME="$ACR_BUILD_IMAGES_REPO:vso-debian-bullseye"
 
 declare -r BASE_IMAGES_REPO="$ACR_DEV_NAME/public/oryx/base"
```
daniv-msft, line 82: Nit comment, so please feel free to ignore if you believe it's not worth it/going to create a lot of noise.
Now that we created the STAGING equivalent, should we rename this one to BASE_IMAGES_PRODUCTION_REPO or BASE_IMAGES_PUBLIC_REPO?

build/buildRunTimeImageBases.sh
```diff
@@ -123,24 +124,48 @@ for dockerFile in $dockerFiles; do
...
+        # pass in env var as a secret, which is mounted during a single run command of the build
+        # https://github.com/docker/buildx/blob/master/docs/reference/buildx_build.md#secret
+        DOCKER_BUILDKIT=1 docker build -f $dockerFile \
+            -t $localImageTagName \
+            --build-arg CACHEBUST=$(date +%s) \
+            --build-arg NODE14_VERSION=$NODE14_VERSION \
+            --build-arg NODE16_VERSION=$NODE16_VERSION \
+            --build-arg NODE18_VERSION=$NODE18_VERSION \
+            --build-arg DEBIAN_FLAVOR=$runtimeImageDebianFlavor \
+            --secret id=DOTNET_PRIVATE_STORAGE_ACCOUNT_ACCESS_TOKEN,env=DOTNET_PRIVATE_STORAGE_ACCOUNT_ACCESS_TOKEN \
```
daniv-msft, line 156: Just to be sure: is this the actual access token, or is this variable corresponding to the ID of the secret where the access token is stored? I assume this is the ID of the secret, but then it could be a good thing to add an _ID suffix to the env variable or something?

images/runtime/dotnetcore/7.0/base.buster.staging.Dockerfile
daniv-msft: Appreciate it! Thanks

images/runtime/python/3.10/bullseye.Dockerfile
daniv-msft: Yes, seems likely indeed. Tagging @snehapar9 and @cormacpayne just in case they see any issues with it.

build/__functions.sh
daniv-msft: Makes sense. We can probably keep this as a future effort once we merge staging with the dev environment maybe (until then, there wouldn't be a lot of interests in expanding this outside of .NET anyway).

build/buildRunTimeImageBases.sh
pauld-msft: see: https://github.com/microsoft/Oryx/pull/1643/commits/c290aaaec36180ae5710cc42294b6672703e82db

pauld-msft: the `env=DOTNET_PRIVATE_STORAGE_ACCOUNT_ACCESS_TOKEN` tells the code to look at the DOTNET_PRIVATE_STORAGE_ACCOUNT_ACCESS_TOKEN env var where this was run, and use that as the secret

build/__variables.sh
pauld-msft: https://github.com/microsoft/Oryx/pull/1643/commits/718ed1ba5d4cf201b8f204c6007b7d21cf27143c

build/__dotNetCoreSdkVersions.sh
pauld-msft: Ended up excluding these!

build/buildRunTimeImageBases.sh
daniv-msft: Thanks Paul!

daniv-msft (approved):
Looks good to me!

PR #1644, opened 2022-10-31
Wali/1634821 adopting private dotnet binaries

waliMSFT:
adopting private binaries for dotnet


build/constants.yaml
```diff
@@ -324,6 +324,7 @@
     testing-sdk-storage-url-key-name: ORYX_TEST_SDK_STORAGE_URL
     dev-sdk-storage-base-url: https://oryxsdksdev.blob.core.windows.net
     sandbox-sdk-storage-base-url: https://oryxsdkssandbox.blob.core.windows.net
+    private-sdk-storage-base-url: https://oryxsdksprivate.blob.core.windows.net
```
pauld-msft, line 327: You'll want to run `generateConstants.sh` to propagate this for your pipeline script to use

platforms/dotnet/versions/bullseye/versionsToBuild.txt
```diff
@@ -34,4 +34,7 @@
 7.0.100-preview.6.22352.1, e49a2119021e4069f1193898536cc59628336e656f2f7e49d454a593a330d8e437acf2f4efb70925bc16a9c900c2e49f4a6c2fb5f69e696b09a91ebccd2c9307,
 7.0.100-preview.7.22377.5, c16d452dbe4f097b75d304c8bc19892017847768bf2e8a0a72fafd6d6b46c3dd77e0c251b80c245197f47fdeafc2c18db255af8a1a5c30be982de19129874390
 7.0.100-rc.1.22431.12, 7ecebf284bd1546e1575b9561f4e64bbb8d69680a7bbda6f06ff6fbf687d3a6c653b0e6a6c569241455c6f83620f0ddbe193ca5cd52384ac062c8565dd22b859,
-7.0.100-rc.2.22477.23, 22db5d1d16f6fcedfc46f87896920425b5d9b61d09c47d254d6ac731c6d853657882b21faf21f313ed20b33e6331d01b9f723b2c586f0e0cf5acc5ed570b0260,
\ No newline at end of file
+7.0.100-rc.2.22477.23, 22db5d1d16f6fcedfc46f87896920425b5d9b61d09c47d254d6ac731c6d853657882b21faf21f313ed20b33e6331d01b9f723b2c586f0e0cf5acc5ed570b0260,
+7.0.100-rc.2.22477.23, 1941cd6ea3bea31d970029c34acf0d60b63d1e7fba39086be1a79177a6f87f2b0300bf4008e96a5235d52bc74a41503b21b088143cd306058d42dd3ce8252af0, https://dotnetcli.blob.core.windows.net/dotnet-private/internal/7.0.0-RC.2/aspnetcore-runtime-7.0.0-rc.2.22476.2-linux-x64.tar.gz,
```
pauld-msft, line 38: These are `aspnetcore-runtime-` tarballs, but we need to push dotnet SDKs to the storage account. Here is the default download url for reference
`https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$sdkVersion/dotnet-sdk-$sdkVersion-linux-x64.tar.gz`

src/BuildScriptGenerator.Common/SdkStorageConstants.cs
```diff
@@ -16,7 +16,7 @@ public static class SdkStorageConstants
         public const string DefaultVersionFilePrefix = "defaultVersion";
         public const string DefaultVersionFileType = "txt";
         public const string VersionsToBuildFileName = "versionsToBuild.txt";
-        public const string ContainerMetadataUrlFormat = "{0}/{1}?restype=container&comp=list&include=metadata&marker={2}";
+        public const string ContainerMetadataUrlFormat = "{0}/{1}?restype=container&comp=list&include=metadata&marker={2}&{3}";
```
pauld-msft, line 19: Generally SAS tokens have a `?` at the beginning, so we should maybe check for its existence and remove it if it is there?

pauld-msft (changes requested):

platforms/dotnet/getDotNetSdk.sh
```diff
@@ -47,7 +47,12 @@ getDotNetCoreSdk() {
 		if [ -z "$downloadUrl" ]; then
 			# Use default download url file
 			downloadUrl="https://dotnetcli.blob.core.windows.net/dotnet/Sdk/$sdkVersion/dotnet-sdk-$sdkVersion-linux-x64.tar.gz"
+		elif  [ "$downloadUrl" == *"dotnet-private"* ]; then
```
pauld-msft, line 50: I think you need to wrap this elif in double brackets `[[  * ]]` instead of `[ * ]`, as they have different behavior
image

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -117,7 +118,7 @@ public virtual void InstallPlatformSpecificSkeletonDependencies(StringBuilder st
                 $"--output {tarFile} >/dev/null 2>&1")
                 .AppendLine("else")
                 .AppendLine(
-                $"curl -D headers.txt -SL \"{sdkStorageBaseUrl}/{platformName}/{platformName}-$DEBIAN_FLAVOR-{version}.tar.gz\" " +
+                $"curl -D headers.txt -SL \"{sdkStorageBaseUrl}/{platformName}/{platformName}-$DEBIAN_FLAVOR-{version}.tar.gz{sdkStorageAccountAccessToken}\" " +
```
pauld-msft, line 121: It looks to me like we are getting the string value of the SAS token here, and actually printing it out to the build script here. Probably more secure to make sure that we don't actually print it into the file. 

In cases where partners use our image in a multi-stage build, they usually copy the oryx files over to their final image which would include the sas token. I think that it would be better to figure out a way to make sure this script just uses the `ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN` env var during execution without printing the string here.

pauld-msft: For example, just using `$ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN` here instead at least works to keep it from being printed in my limited testing of $DEBIAN_FLAVOR

doc/supportedPlatformVersions.md
```diff
@@ -39,6 +39,7 @@
 - 7.0.100-preview.7.22377.5
 - 7.0.100-rc.1.22431.12
 - 7.0.100-rc.2.22477.23
+- 7.0.100-rc.2.22477.23
```
daniv-msft, line 42: (Please ignore) Adding a comment just so that we don't ignore these once we update the version to .NET 7 GA.

platforms/dotnet/versions/bullseye/versionsToBuild.txt
waliMSFT: the SAS token is already added with the URL.

vsts/pipelines/templates/_platformBinariesTemplate.yml
```diff
@@ -19,6 +19,8 @@ steps:
 
 - task: ShellScript@2
   displayName: 'Building platform binaries'
+  env:
+    PRIVATESTORAGESASTOKEN: $(PrivateStorageSASToken)
```
pauld-msft, line 23: Just realized that this name is almost exactly the same as PRIVATE_STORAGE_SAS_TOKEN that we set to the value in our key vault (in _platformBinariesReleaseTemplate.yml). I would lean towards updating this to be the name that we discussed in our meeting:
```
    env:
      DOTNET_PRIVATE_STORAGE_ACCOUNT_ACCESS_TOKEN: $(DotnetPrivateStorageAccountAccessToken)
```

this would also align it with the variable that I'm using for the base images job

pauld-msft (approved):
:shipit:

PR #1648, opened 2022-11-02
Williamhe/1641601 add storage account check

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1641601

Part 1 of addressing TODOs before I head out on vacation.
-  Added storage account check
    - helps remove redundant PRs
- Added date to the new branch creation
    - prevents PRs from getting overwritten with an empty PR when there is no release the following day
    - prevents recycling old PR branch
- Stop workflow from running if no `git` changes are detected before creating a PR.


build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -77,13 +81,13 @@ public override async Task<List<PlatformConstant>> GetPlatformConstantsAsync(str
                 {
                     // check releasedToday again since there
                     // are still releases from other dates.
-                    if (!DatesMatch(dateTarget, release.ReleaseDate))
+                    string sdkVersion = release.Sdk.Version;
+                    if (!DatesMatch(dateTarget, release.ReleaseDate) || this.prodSdkVersions.Contains(sdkVersion))
```
william-msft, line 85: We still need to check if the `dateTarget` and `ReleaseDate` match otherwise really old SDKs that are not in our storage account (like .NET2.1) get added.

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -127,6 +131,27 @@ public override async Task<List<PlatformConstant>> GetPlatformConstantsAsync(str
             return platformConstants;
         }
 
+        public override async Task CacheSdkVersionsAsync(string platform)
+        {
+            string url = $"https://oryxsdks.blob.core.windows.net/{platform}" +
```
william-msft, line 136: Decided to check prod SDK account instead of dev. This will help us re-generate a new fresh PR with updated main changes. If integrating new SDKs ever takes longer than usuals without having to delete new SDKs from dev storage account. Or if we ever want to easily rebuild SDK binaries, in dev, that don't exist in prod.

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -127,6 +131,27 @@ public override async Task<List<PlatformConstant>> GetPlatformConstantsAsync(str
...
+            "?restype=container&comp=list&include=metadata";
```
william-msft, line 137: Decided to check prod SDK account instead of dev. This will help us re-generate a new fresh PR with updated main changes. If integrating new SDKs ever takes longer than usuals without having to delete new SDKs from dev storage account.

.github/workflows/main.yaml
```diff
@@ -51,10 +63,11 @@ jobs:
...
+          branch: automation/add-new-dotnet-release-${{ env.releaseDate }}
+          title: New DotNetCore Sdks ${{ env.releaseDate }}
 
       - name: Run Azure Pipeline 
         run: |
           echo  "${{ secrets.AZURE_DEVOPS_TOKEN }}" | az devops login --organization https://devdiv.visualstudio.com/
-          az pipelines run --organization https://devdiv.visualstudio.com/ --project DevDiv --name Oryx-PlatformBinary-DotNetCore --branch automation/add-new-dotnet-release
\ No newline at end of file
+          az pipelines run --organization https://devdiv.visualstudio.com/ --project DevDiv --name Oryx-PlatformBinary-DotNetCore --branch automation/add-new-dotnet-release-${{ env.releaseDate }}
+          az pipelines run --organization https://devdiv.visualstudio.com/ --project DevDiv --name Oryx-Validation --branch automation/add-new-dotnet-release-${{ env.releaseDate }}
```
william-msft, line 73: Appending the date allows us to prevent new SDKs to get overwritten by no other changes. For example, if V1 is released today and V2 is released tomorrow, then V1 will have it's own branch and V2 will have it's own branch. Preventing V2 from overwriting V1 changes.

william-msft, line 73: Appending the date allows us to prevent new SDKs to get overwritten by other/empty changes. For example, if V1 is released today and V2 is released tomorrow, then V1 will have it's own branch and V2 will have it's own branch. Preventing V2 from overwriting V1 changes.

.github/workflows/main.yaml
```diff
@@ -29,8 +29,19 @@ jobs:
             echo "Defaulting to todays date ${releaseDate}"
           fi
 
+          echo "releaseDate=${releaseDate}" >> $GITHUB_ENV
+
           dotnet run --project build/tools/Automation/Automation.csproj ${releaseDate} ${{ github.workspace }}
 
+      - name: Check if any changes exist.
```
cormacpayne, line 36: nit: would you mind updating this text to be a little more specific as to what changes you're checking for? (in this case, it's Git changes, but I could see it being confused for SDK changes at the package manager level)

.github/workflows/main.yaml
```diff
@@ -39,6 +50,7 @@ jobs:
         run: |
           git config user.name 'William Hernandez'
           git config user.email 'williamhe@microsoft.com'
+          git checkout doc/supportedPlatformVersions.md
```
cormacpayne, line 53: nit: what is this command doing? I typically associate `git checkout` with switching branches, but this command references a file, so just want to ensure I understand what's happening here

.github/workflows/main.yaml
```diff
@@ -29,8 +29,19 @@ jobs:
             echo "Defaulting to todays date ${releaseDate}"
           fi
 
+          echo "releaseDate=${releaseDate}" >> $GITHUB_ENV
```
cormacpayne, line 32: nit: as a reminder, does the cron job provide a value for the `releaseDate` input?

.github/workflows/main.yaml
```diff
@@ -51,10 +63,11 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@171dd555b9ab6b18fa02519fdfacbb8bf671e1b4  # locking to a commit version for now
         with:
-          branch: automation/add-new-dotnet-release
-          title: New DotNetCore Sdks
+          branch: automation/add-new-dotnet-release-${{ env.releaseDate }}
```
cormacpayne, line 66: nit: how does appending the `releaseDate` input to the branch affect the number of PRs that will be opened? Will this logic continue to push to the same branch that a PR is opened for, or will one be opened every 12 hours while an SDK change is needed to be made?

build/tools/Automation/Program.cs
```diff
@@ -31,7 +31,9 @@ namespace Microsoft.Oryx.Automation
     public abstract class Program
     {
         private static string repoAbsolutePath = string.Empty;
+        private static HashSet<string> prodSdkVersions = new HashSet<string>();
 
+        // private static HashSet<string> sdkVersions = new HashSet<string>();
```
cormacpayne, line 36: nit: remove this comment

build/tools/Automation/Program.cs
```diff
@@ -95,6 +99,14 @@ public static bool DatesMatch(string dateTarget, string dateReleased)
             return match;
         }
 
+        /// <Summary>
+        /// Caches the storage account sdk versions to avoid pulling all SDKs everytime
+        /// we check if a version exists in the storage account.
+        /// </Summary>
+        /// <param name="platform">dotnet, nodejs, python, etc
+        /// This dateTarget can be passed through github actions through an argument</param>
+        public abstract Task CacheSdkVersionsAsync(string platform);
```
cormacpayne, line 108: nit: just for my understanding, how did we go about "pulling all SDKs" previously? I think it's good for us to read the prod storage account when the `DotNet` object is instantiated so we can freely use the list of published production SDKs in that lifespan of the object. I do wonder if we should replace `Cache` with `Pull` since the verb `Cache` leads me to believe that these SDK versions we pull down will persist across sessions so that we don't have to pull from prod next time we run the cron job.

cormacpayne (commented):
Just a few minor comments before I approve

build/tools/Automation/Program.cs
william-msft: Removed.

.github/workflows/main.yaml
william-msft: Updated. Hm do you have a more specific scenario of when this will behave unexpectedly?

.github/workflows/main.yaml
pauld-msft: @william-msft do you mind sharing an example of that bug? In theory, I feel like we should be updating that file as we release new versions which the code will support

cormacpayne: Agree with @pauld-msft -- we should quickly check out the bug in the generator to see why it's producing more supported versions than expected so that this cron job will help to accurately reflect the platform versions we support.

.github/workflows/main.yaml
cormacpayne: If .NET releases a new 7.0 version, and our noon cron job picks up the change and creates a PR based on it, but that PR isn't merged by the end of the day, would the nightly cron job pick up the same change from .NET (since it's not yet in our repository) and try to create another branch based on the same changes that are already in the PR?

william-msft: In that case it would use the same branch and PR.

The problem I do see is if it get's merged to `main` but the `Oryx-PublishSdkToProd` doesn't get triggered before the nightly cron. Then it would re-open the same PR on the same branch. We might have just have to keep the nightly cron.

.github/workflows/main.yaml
william-msft: Surprisingly I'm not able to reproduce locally. Maybe the github workflow triggers had an older version of the .`md` or something. I created a story just in case: https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1664336

I've removed the `git checkout` now.

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -63,7 +67,7 @@ public override async Task<List<PlatformConstant>> GetPlatformConstantsAsync(str
             {
                 // TODO: check if SDK already exists in storage account
```
pauld-msft, line 68: remove todo now?

pauld-msft (dismissed):
few minor things, other than that LGTM

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -63,7 +67,7 @@ public override async Task<List<PlatformConstant>> GetPlatformConstantsAsync(str
...
                 var dateReleased = releaseIndex.LatestReleaseDate;
-                if (!DatesMatch(dateTarget, dateReleased))
+                if (!DatesMatch(dateTarget, dateReleased) || this.prodSdkVersions.Contains(releaseIndex.LatestSdk))
```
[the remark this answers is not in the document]
william-msft: Currently 6.0.6 would be added. So long as a matching date (date in DotNet's releases.json and target version) match.

build/tools/Automation/DotNet/DotNet.cs
william-msft: Removed

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -127,6 +131,27 @@ public override async Task<List<PlatformConstant>> GetPlatformConstantsAsync(str
             return platformConstants;
         }
 
+        public override async Task PullSdkVersionsAsync(string platform)
+        {
+            string url = $"https://oryxsdks.blob.core.windows.net/{platform}" +
```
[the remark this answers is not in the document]
william-msft: Updated to use CDN

build/tools/Automation/DotNet/DotNet.cs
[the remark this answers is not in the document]
pauld-msft: @william-msft if the most recent 6.0 version exists in our prod storage account wouldn't this logic skip the full 6.0 release JSON though (and therefore would not loop through the .NET 6.0 releases where 6.0.6 is)?

build/tools/Automation/DotNet/DotNet.cs
william-msft: I moved the SDK CDN base URL into the `constants.yaml`. I kept prod since there is value in being explicit.

pauld-msft (approved):

build/tools/Automation/DotNet/DotNet.cs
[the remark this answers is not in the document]
pauld-msft: spoke about this offline, and found that the case I was mentioning will need some more work to be valid. If we try and add a version that is not the latest with this code as is (even after removing this if statement), then the Oryx repo will be changed to use the outdated version as well (6.0.6 would be updated to be the dotnet version in the constants.yaml file, for example)

PR #1660, opened 2022-11-04
Pauld/net 7 ga updates

pauld-msft:
PR that will release .NET 7 publicly once it is available.

Once .NET 7 reaches GA, this is what we will run the pipelines off of to push .NET 7 SDKs to DEV and then release to prod. We will also run the base images pipeline off of this PR, and then update the runtime image base tag. I've also added a new sample web app and some tests, and ensured they work by retagging the build and runtime images that currently can use the .NET 7 binaries for local testing.


build/constants.yaml
```diff
@@ -448,7 +448,7 @@
 - name: staging-runtime-constants
   constants:
     dotnetcore-staging-runtime-versions:
```
cormacpayne, line 450: Are we able to remove this value so the corresponding file(s) isn't generated, or is our code reliant on this file existing?

build/tools/SharedCodeGenerator/Outputs/CSharpOutput.cs
```diff
@@ -58,7 +58,11 @@ public string GetContent()
                 Name = this.className,
                 Scope = scope,
                 StringConstants = this.collection.StringConstants?.ToDictionary(pair => pair.Key.Camelize(), pair => pair.Value),
-                ListConstants = this.collection.ListConstants?.ToDictionary(pair => pair.Key.Camelize(), pair => $"{{ \"{string.Join("\", \"", pair.Value)}\" }}"),
+                ListConstants = this.collection.ListConstants?.ToDictionary(
+                    pair => pair.Key.Camelize(),
+                    pair => pair.Value.Any(s => !string.IsNullOrEmpty(s?.ToString()))
+                        ? $"{{ \"{string.Join("\", \"", pair.Value)}\" }}"
+                        : "{ }"),
```
cormacpayne, line 65: nit: just confirming that this handles the case where no item is provided to a list in `constants.yaml`

tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
```diff
@@ -45,127 +45,109 @@ public abstract class StorageAccountSanityTestBase
             _sdkStorageAccountAccessToken = token;
         }
 
-        [Fact(Skip = "Skipping test temporarily till we resolve the issue #1663008")]
```
cormacpayne, line 48: nit: what are the changes in this file for?

build/constants.yaml
pauld-msft: So I wanted the runtime tests to automatically switch over to the staging image name/tags for the versions that apply, and so I added https://github.com/microsoft/Oryx/blob/main/tests/Oryx.Tests.Common/ImageTestHelper.cs#L161 into the tests. Thats the only place that I think uses the constant. I felt that leaving this empty so it still registers as a list wasn't too egregious, but if you feel strongly about it I can try and brainstorm another way to approach it.

build/tools/SharedCodeGenerator/Outputs/CSharpOutput.cs
pauld-msft: yep, this addition makes it so that the list will be actually empty in the c sharp output if all of the list contents are empty

tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
pauld-msft: As we had a bit of a time crunch to get the binaries out to partners, we ignored these tests in order to disable them because we had added a version (7.0.100) to the versionToBuild.txt file that never made it to the DEV or PROD storage accounts, as those are publicly accessible. That meant that these tests would fail for the nightly and CI jobs. So we skipped them and made a [ticket](https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1663008) to address the case fully. 

I'm removing the skip here, since we should be able to get around to addressing that ticket before the next release of private binaries (we don't even know when that will be).

A quick 1-2 line alternative I could implement here is to configure the test to ignore any lines in the versionToBuild.txt that supply a download url (or even a download url that contains "dotnet-private"). But if we want to think about it some more, then I was just going to leave it for the ticket to handle.

build/constants.yaml
cormacpayne: Got it -- I'm good with it as-is since we may re-use it in the future, I was just curious if we'd be able to remove it now since it wasn't actively being used.

build/tools/SharedCodeGenerator/Outputs/CSharpOutput.cs
cormacpayne: Sweet -- thanks for the confirmation 👍

cormacpayne (approved):
:shipit:

tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
pauld-msft: OH yeah good catch, thanks!

cormacpayne (approved):

daniv-msft (dismissed):
Thanks! Looks good to me

daniv-msft (approved):

PR #1661, opened 2022-11-04
Bump up viztracer version and pull binaries from storage account for …

snehapar9:
…python 3.11

- [<del> ] Tests are included and/or updated for code changes.
- [ <del>] Proper license headers are included in each file.

The purpose of this PR is to install python 3.11 tools to /opt/python/3/bin instead of usr/local/bin folder as requested by Antares.

However, looks like vizplugins is not supported in python 3.11 and this package could not be transitioned. 
https://pypi.org/project/vizplugins/

Edit: pip install vizplugins fails for python 3.11 because of building wheel for psutil failure.
image

Here is a screenshot to validate this change.

image


images/runtime/python/template.Dockerfile
```diff
@@ -77,8 +77,9 @@ RUN ${IMAGES_DIR}/runtime/python/install-dependencies.sh
 RUN pip install --upgrade pip \
     && pip install gunicorn \
     && pip install debugpy \
-    && if [ "%PYTHON_MAJOR_VERSION%" = "3" ] && [ "%PYTHON_VERSION%" != "3.6" ]; then pip install viztracer==0.14.3 \
-    && pip install vizplugins==0.1.2 \
+    && if [ "%PYTHON_MAJOR_VERSION%" = "3" ] && [ "%PYTHON_VERSION%" != "3.6" ] && [ "%PYTHON_VERSION%" != "3.11" ]; then pip install viztracer==0.14.3; fi\
```
cormacpayne, line 80: Investigating this a little more -- from the change logs for the releases between `0.14.3` and `0.15.5` in the `viztracer` repository, it doesn't appear that anything notable has changed other than dropped support for Python 3.6, but that version has reached end-of-life and Antares doesn't consume it anymore. I'd be OK with us adopting `0.15.5` for all versions of Python that we publish now, letting Antares know about this bump and having them test out any `viztracer` functionality they expect to be supported with their Python app tests. Worst case scenario, they can still use the new Python 3.11 runtime (since `0.15.5` is the only version support for Python 3.11), and we can bring back this conditional to install `0.14.3` for Python 3.7 to 3.10.

@daniv-msft -- any thoughts on this? (more context in the description of the PR)

cormacpayne, line 80: nit: I think we're good to remove checks for Python 3.6 in this file now that we no longer support it

images/runtime/python/cgmanifest.json
```diff
@@ -4,7 +4,7 @@
             "Component": {
                 "pip": {
                     "Name": "viztracer",
-                    "Version": "0.14.3"
+                    "Version": "0.15.5"
```
cormacpayne, line 7: Was this a manual update? We may need to make another entry in this file if we support both `viztracer@0.14.3` and `viztracer@0.15.5`

gaogaotiantian:
viztracer dev here - VizPlugin is a pure Python implementation so it should support 3.11. The reason that 3.11 is not shown on the github page is the label is never updated. However, I'll test if it actually works on 3.11 and update the version.

The attach mechanism is updated between 0.14.3 and the current 0.15.5, which I believe was used? VizTracer now supports gdb based attach by default, which means you don't need a pre-installed process to be attached. However, you can still use the backward-compatible command with ``viztracer --attach_installed`` to keep the old behavior where `SIGUSR1` and `SIGUSR2` are used as the trigger signals.

LMK if there's anything I can help with this.

snehapar9:
> viztracer dev here - VizPlugin is a pure Python implementation so it should support 3.11. The reason that 3.11 is not shown on the github page is the label is never updated. However, I'll test if it actually works on 3.11 and update the version.
> 
> The attach mechanism is updated between 0.14.3 and the current 0.15.5, which I believe was used? VizTracer now supports gdb based attach by default, which means you don't need a pre-installed process to be attached. However, you can still use the backward-compatible command with `viztracer --attach_installed` to keep the old behavior where `SIGUSR1` and `SIGUSR2` are used as the trigger signals.
> 
> LMK if there's anything I can help with this.

Building wheels for psutil fails while installing vizplugins 0.1.2 for python 3.11 in Windows.

gaogaotiantian:
> Building wheel for psutil fails while installing vizplugins 0.1.2 for python 3.11 in Windows.

This seems like a `psutil` issue - https://github.com/giampaolo/psutil/issues/2165. The feature for VizPlugins requires `psutil` to record CPU/memory usage. If that's a necessary feature here, then we both need to wait for `psutil` to figure this out.

gaogaotiantian:
Just FYI, there are some fundamental issues with 0.15.5 and the plugins, besides the fact that `psutil` does not support 3.11 now. I'm planning to check in the fix and bump the version to 0.15.6 as soon as `psutil` adds the support.

images/runtime/python/cgmanifest.json
snehapar9: Since we are installing viztracer version 0.15.5 across all python versions, do we still need  this?

cormacpayne: Since we're installing `0.15.5` across the board, we should be good to keep this change as-is 👍

cormacpayne (approved):
:shipit:

PR #1663, opened 2022-11-07
Updating character enconding from POSIX to UTF-8

harryli0108:
Updated language setting env from POSIX to C.UTF-8 or en_US.UTF-8.
Most images don't have locale installed. Therefore can only use C.UTF-8 without installing it. The supported characters of C.UTF-8 and en_US are same. The only difference is the orders of a-z, A-Z, numbers, etc. en_US is more user friendly, while C.UTF-8 is more compatible to machine. 
Therefore, this PR hasn't installed locale to every img but to choose whichever character encoding the image supported.
Only python runtime img is using en_US.UTF-8. All the others are using C.UTF-8

### DIFF
--- a/images/build/Dockerfiles/gitHubActions.Dockerfile
+++ b/images/build/Dockerfiles/gitHubActions.Dockerfile
@@ -3,8 +3,7 @@ FROM oryxdevmcr.azurecr.io/private/oryx/githubrunners-buildpackdeps-${DEBIAN_FLA
 ARG DEBIAN_FLAVOR
 ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR
 # Install basic build tools
-RUN LANG="C.UTF-8" \
-    && apt-get update \
+RUN apt-get update \
     && apt-get upgrade -y \
     && apt-get install -y --no-install-recommends \
         git \
@@ -160,6 +159,8 @@ RUN tmpDir="/opt/tmp" \
 ENV ORYX_PATHS="/opt/oryx:/opt/yarn/stable/bin:/opt/hugo/lts"
 
 ENV LANG="C.UTF-8" \
+    LANGUAGE="C.UTF-8" \
+    LC_ALL="C.UTF-8" \
     ORIGINAL_PATH="$PATH" \
     PATH="$ORYX_PATHS:$PATH" \
     NUGET_XMLDOC_MODE="skip" \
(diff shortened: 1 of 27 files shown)


images/build/Dockerfiles/gitHubActions.Dockerfile
```diff
@@ -3,7 +3,9 @@ FROM oryxdevmcr.azurecr.io/private/oryx/githubrunners-buildpackdeps-${DEBIAN_FLA
 ARG DEBIAN_FLAVOR
 ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR
 # Install basic build tools
-RUN LANG="C.UTF-8" \
+RUN export LANG="C.UTF-8" \
```
cormacpayne, line 6: When setting environment variables that we want to persist to the container that the image is ran against, we should use `ENV` in the Dockerfile rather than `RUN export` as the latter only persists for the single command call.

[This post](https://stackoverflow.com/a/33379487) should provide a little more information.

cormacpayne (changes requested):

images/build/Dockerfiles/gitHubActions.Dockerfile
harryli0108: Thanks for the comment! I've just changed the RUN to ENV variables. It's currently in validation pipeline.

images/build/Dockerfiles/gitHubActions.Dockerfile
```diff
@@ -3,8 +3,7 @@ FROM oryxdevmcr.azurecr.io/private/oryx/githubrunners-buildpackdeps-${DEBIAN_FLA
 ARG DEBIAN_FLAVOR
 ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR
 # Install basic build tools
-RUN LANG="C.UTF-8" \
```
cormacpayne, line 6: Do you see any issue with keeping this `LANG="C.UTF-8"` statement here (and in the other places that you have it removed)? I know it may be redundant to have it defined here _and_ in the `ENV` command that you've added below in these files, but I'm wondering if there was a reason that this environment variable was set here because it was needed for the duration of the `RUN` command when installing packages via `apt-get install`.

cormacpayne (approved):
One question, otherwise LGTM :shipit:

images/build/Dockerfiles/gitHubActions.Dockerfile
harryli0108: That's the part I was kinda worried as well. I'm not sure if this is required for apt-get install. It seems like other runtime imgs doesn't have LANG env defined prior to apt-install tho. Also, it didn't have any error thrown when building the base images without the env defined there.

harryli0108: just did a quick search. seems like setting LANG before apt-get is a recommended procedure for troubleshooting in Ubuntu guide. I think it'll be fine leaving `LANG="C.UTF-8"` out before the `apt-get install`

PR #1670, opened 2022-11-08
remove .NET7 condition on bindport2

waliMSFT:


daniv-msft (approved):
Looks good to me!

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -117,15 +118,14 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
...
-		scriptBuilder.WriteString("\n\n")
-	}
+	logger.LogInformation("Setting up Kestrel Endpoints with BindPort and BindPort2 Env variables")
+
+	scriptBuilder.WriteString("if [ ! -z \"$PORT2\" ]; then" + "\n")
+	scriptBuilder.WriteString("		export Kestrel__Endpoints__Http2__Url=http://*:$PORT2\n")
+	scriptBuilder.WriteString("		export Kestrel__Endpoints__Http2__Protocols=Http2\n")
+	scriptBuilder.WriteString("		export Kestrel__Endpoints__Http1__Url=http://*:$PORT\n\n")
+	scriptBuilder.WriteString("fi")
+	scriptBuilder.WriteString("\n\n")
```
snehapar9, line 129: Just to make sure I understand this -- why are we removing this condition?

daniv-msft: That was because the App Service partner team wanted this to be only available on .NET 7 at first, for testing purposes. We were not sure if they would have a way to limit the change to .NET 7, so we decided at the time to check the version on our side.
Now that everything looks good on .NET 7, we have no reasons to limit this anymore and can expand it to other .NET versions.

snehapar9: Thanks @daniv-msft!

waliMSFT: our partners asked for it. Previously bindPort2 was only available for .NET7.

cormacpayne (approved):
:shipit:

PR #1671, opened 2022-11-09
Pauld/switch go packages to go mod

pauld-msft:
Finishing the conversion from `dep` -> `go.mod`, following the steps laid out in https://stackoverflow.com/questions/55664630/how-do-i-migrate-from-dep-to-go-modules/55664631#55664631

We did this because modules are how go manages dependencies, and we were already creating a fresh [go.mod file](https://go.dev/doc/modules/gomod-ref) that defines a module whenever we built our go modules in the `.build.sh` script. This update deletes the [deprecated dep package manager](https://golang.github.io/dep/docs/introduction.html) in favor of managing our dependencies in the supported way.

Tested with the `./testStartupScriptGenerators.sh` script and by successfully building images where the `./build.sh` script is executed.


cormacpayne (approved):
The changes look good to me pending the test changes that are needed -- one thing I would suggest adding if you have time: at the src/startupscriptgenerator/src` level, it would be nice if we had a `README.md` file outlining how to quickly set up the Go environment and the steps required to generate the `go.mod` and `go.sum` files if we ever run into a dependency issue again and someone needs to bump the versions of dependencies used.

pauld-msft:
> The changes look good to me pending the test changes that are needed -- one thing I would suggest adding if you have time: at the `src/startupscriptgenerator/src` level, it would be nice if we had a `README.md` file outlining how to quickly set up the Go environment and the steps required to generate the `go.mod` and `go.sum` files if we ever run into a dependency issue again and someone needs to bump the versions of dependencies used.

@cormacpayne I'll add that README! But there aren't actually any testing changes that need to go through, I was just explaining why I needed to replace the `go test ./... -v` command with a script

cormacpayne:
> > The changes look good to me pending the test changes that are needed -- one thing I would suggest adding if you have time: at the `src/startupscriptgenerator/src` level, it would be nice if we had a `README.md` file outlining how to quickly set up the Go environment and the steps required to generate the `go.mod` and `go.sum` files if we ever run into a dependency issue again and someone needs to bump the versions of dependencies used.
> 
> @cormacpayne I'll add that README! But there aren't actually any testing changes that need to go through, I was just explaining why I needed to replace the `go test ./... -v` command with a script

Ah whoops, I misread _needed_ as _need_ in the original comment you made about testing 😁

cormacpayne (approved):
:shipit:

PR #1685, opened 2022-11-15
Wali/oryx build using yarn when yarn is specified in the package.json 1676966

waliMSFT:

### DIFF
--- a/src/BuildScriptGenerator/Node/NodePlatform.cs
+++ b/src/BuildScriptGenerator/Node/NodePlatform.cs
@@ -187,11 +187,13 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
                 throw new InvalidUsageException("Multiple monorepo package management tools are found, please choose to use either Lerna or Lage.");
             }
 
+            string yarnVersionSpec = null;
             if (ctx.SourceRepo.FileExists(NodeConstants.YarnLockFileName) || packageJson?.engines?.yarn != null)
             {
                 packageManagerCmd = NodeConstants.YarnCommand;
                 configureYarnCache = false;
                 packageInstallerVersionCommand = NodeConstants.YarnVersionCommand;
+                yarnVersionSpec = packageJson?.engines?.yarn?.Value as string;
 
                 // In Yarn 2+ and .yarnrc.yml file replaces .yarnrc in Yarn 2+.
                 // Applying yarn 2 cache folder name and package install command.
@@ -398,6 +400,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
                 NodeBuildProperties = nodeCommandManifestFileProperties,
                 NodeBuildCommandsFile = nodeBuildCommandsFile,
                 NpmVersionSpec = npmVersionSpec,
+                YarnVersionSpec = yarnVersionSpec,
             };
             string script = TemplateHelper.Render(
                 TemplateHelper.TemplateResource.NodeBuildSnippet,
--- a/src/BuildScriptGenerator/Node/NodePlatform.cs
+++ b/src/BuildScriptGenerator/Node/NodePlatform.cs
@@ -193,7 +193,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
                 packageManagerCmd = NodeConstants.YarnCommand;
                 configureYarnCache = false;
                 packageInstallerVersionCommand = NodeConstants.YarnVersionCommand;
-                yarnVersionSpec = packageJson?.engines?.yarn?.Value as string;
+                yarnVersionSpec = ctx.SourceRepo.FileExists(NodeConstants.YarnLockFileName) ? "latest" : packageJson?.engines?.yarn?.Value as string;
 
                 // In Yarn 2+ and .yarnrc.yml file replaces .yarnrc in Yarn 2+.
                 // Applying yarn 2 cache folder name and package install command.
--- a/src/BuildScriptGenerator/Node/NodePlatform.cs
+++ b/src/BuildScriptGenerator/Node/NodePlatform.cs
@@ -187,13 +187,12 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Node
                 throw new InvalidUsageException("Multiple monorepo package management tools are found, please choose to use either Lerna or Lage.");
             }
 
-            string yarnVersionSpec = null;
-            if (ctx.SourceRepo.FileExists(NodeConstants.YarnLockFileName) || packageJson?.engines?.yarn != null)
+            string yarnVersionSpec = packageJson?.engines?.yarn;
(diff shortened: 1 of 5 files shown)


waliMSFT:
Below is the context:
"Customer is using Node.js v14.19.1 with his static web app “HkdirZubaierUSA”, and has specified “yarn” in “engines” field in package.json:

image


And he has both yarn.lock and package.json in the repo root as shown above.

 

However Oryx still uses npm as shown below:

image

I have suggested customer to remove the line with npm and it worked. However customer hesitates to apply the same in his production environment and asks if there is another way to avoid oryx to use npm, as from Oryx/nodejs.md at main · microsoft/Oryx (github.com):

“If a yarn.lock file is found in your repo root or if you specify "yarn" in the engines field of package.json, the latest or specified version of yarn will be used instead of npm.”

Which seems to indicate that yarn should be used in this scenario.

 

Question:

Is this an expected outcome? If not, why Oryx used npm instead of yarn in this scenario?
Is there another way to avoid oryx to use npm other than removing the line that defines npm in engines part?"


I investigated the issue and found that yarn never takes the precedence over npm is yarn is specified in `package.json` file or there is `yarn.lock` file. In this PR we resolved the issue.

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -187,11 +187,13 @@ public PlatformDetectorResult Detect(RepositoryContext context)
                 throw new InvalidUsageException("Multiple monorepo package management tools are found, please choose to use either Lerna or Lage.");
             }
 
+            string yarnVersionSpec = null;
             if (ctx.SourceRepo.FileExists(NodeConstants.YarnLockFileName) || packageJson?.engines?.yarn != null)
             {
                 packageManagerCmd = NodeConstants.YarnCommand;
                 configureYarnCache = false;
                 packageInstallerVersionCommand = NodeConstants.YarnVersionCommand;
+                yarnVersionSpec = ctx.SourceRepo.FileExists(NodeConstants.YarnLockFileName) ? "latest" : packageJson?.engines?.yarn?.Value as string;
```
cormacpayne, line 196: Let's just stick with using the value of `engines.yarn` for the version of `yarn` we update to -- since the existence of a `yarn.lock` file is how we determine that `yarn` should be used most of the time, I'm hesitant to have us snap to a dynamic version with `latest` instead of a fixed version. There could be a breaking change introduced in the `latest` version that we're not aware of, and customers are stuck using it unless they update their `engines.yarn` field to snap to something different. I think the proper flow is to use the fixed version that we pre-bake into build images and if the user wants something different, they can specify it in their `engines.yarn` field.

Also, the logic currently has the existence of the `yarn.lock` file take precedence over the `engines.yarn` value if the user has set it, so even if the user wanted to change to something that isn't `latest`, they would be locked to it unless they got rid of their `yarn.lock` file, which wouldn't be feasible for most projects.

src/BuildScriptGenerator/Node/NodePlatform.cs
```diff
@@ -187,11 +187,13 @@ public PlatformDetectorResult Detect(RepositoryContext context)
                 throw new InvalidUsageException("Multiple monorepo package management tools are found, please choose to use either Lerna or Lage.");
             }
 
+            string yarnVersionSpec = null;
```
cormacpayne, line 190: nit: let's assign the value of `yarnVersionSpec` immediately and use it in the following line:


```cs
string yarnVersionSpec = packageJson?.engines?.yarn;
if (ctx.SourceRepo.FileExists(NodeConstants.YarnLockFileName) || yarnVersionSpec != null)
{
    ...
}
```

cormacpayne (changes requested):
We should add a test for this to ensure that the script gets properly generated -- check the `GeneratedScript_HasNpmVersion_SpecifiedInPackageJson` test in the `NodeBuildScriptGenerationTest` class as an example of how we can set up a mock node app that has a `package.json` with `engines.yarn` set (in the above test's example, it has `engines.npm` set, so we would just create a new mock npm app with `engines.yarn` set)

src/BuildScriptGenerator/Node/NodePlatform.cs
waliMSFT: Thank Cormac for the feedback. I have the same concerns that you mentioned for the `yarn` version. 
However, according to the doc -  https://github.com/microsoft/Oryx/blob/main/doc/runtimes/nodejs.md#package-manager, 
`If a yarn.lock file is found in your repo root or if you specify "yarn" in the engines field of package.json, the latest or specified version of yarn will be used instead of npm.`. 
it says to use the `latest` version of yarn if there is `yarn.lock` file is found or take the specified version of yarn if mentioned in the `engines` field of `package.json` file. By this doc, I was convinced to use the latest version of `yarn` if there is `yarn.lock` found and even take the precedence over the yarn version mentioned in the `engines` field of `package.json` file. 
Besides, there is still a scenario that can create bug. if there is `yarn.lock` file is found but no yarn version is not mentioned in the  `engines` field of `package.json` file then according to our existing code the NPM package manager would be used. According to my understanding, in this scenario the latest version of `yarn` should be used.

Based on that, is it still fine to remove the `latest` version part of yarn as you suggested?

src/BuildScriptGenerator/Node/NodePlatform.cs
waliMSFT: sure.

src/BuildScriptGenerator/Node/NodePlatform.cs
cormacpayne: We should update that document to accurately reflect what the behavior is (and I would argue should be):

_If a `yarn.lock` file is found in your repo root, or if you specify a version spec for `yarn` in the `engines` field of your application's `package.json`, a stable or specified version of yarn will be used instead of npm._

This document was introduced almost four years ago, so the behavior may have changed over time without the document being updated, or the behavior was never aligned with the document. As it stands today, the document is incorrect and I think it should be updated to reflect the above behavior, which relies on us having a stable version of yarn that the customer can use, or allow them to specify a version spec that resolves to a version of yarn that works for the customer's application. I personally don't feel it's a good idea to rely on `stable` to pick the version for customers, which can introduce a breaking change at any point. If we know that the stable version we're pre-baking into our images now works, we should continue to snap to it and allow customers to update via the `engines` field if they would like to.

cormacpayne (changes requested):
Let's also update [this document](https://github.com/microsoft/Oryx/blob/main/doc/runtimes/nodejs.md) to reflect the updated behavior when choosing a package manager, more specifically, changing the line we talked about previously to something like this:

_If a yarn.lock file is found in your repo root, or if you specify a version spec for yarn in the engines field of your application's package.json, a stable or specified version of yarn will be used instead of npm._

cormacpayne (approved):
:shipit:

PR #1687, opened 2022-11-16
Pauld/exclude sampleapps from sbom

pauld-msft:
I've noticed that we see the [SBOM generation step fail](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6960522&view=logs&j=e67ff1db-fb5d-597d-bb68-9b394d24d310&t=19b932f7-ddfe-50d0-e9e8-ad26dbc516ed) fairly often during our Oryx-CI jobs, and it additionally seems to run into a number of issues while parsing the sample apps dir. This change will exclude us from running SBOM generation in the sample apps dir. I'm also in talks with the owners of the SBOM process to determine if there is anything going wrong with the signing API that is causing these errors.

Furthermore, according to [the docs](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/secure-supply-chain/ado-sbom-generator), `PackageName` and `PackageVersion` should correlate to _our_ software name and version if it is specified. I've removed these fields, and will let the SBOM step autogenerate them with the build name
image


snehapar9 (approved):

cormacpayne (approved):
:shipit:

PR #1689, opened 2022-11-18
Update Go 1.18 and 1.19 binaries to latest version

cormacpayne:
Updates Go 1.18 and 1.19 supported by dynamic installation to the latest versions released.


build/__golangVersions.sh
```diff
@@ -4,7 +4,7 @@ GOLANG16_VERSION='1.16.7'
 GOLANG16_TAR_SHA256='7fe7a73f55ba3e2285da36f8b085e5c0159e9564ef5f63ee0ed6b818ade8ef04'
 GOLANG17_VERSION='1.17'
 GOLANG17_TAR_SHA256='6bf89fc4f5ad763871cf7eac80a2d594492de7a818303283f1366a7f6a30372d'
-GOLANG18_VERSION='1.18'
-GOLANG18_TAR_SHA256='e85278e98f57cdb150fe8409e6e5df5343ecb13cebf03a5d5ff12bd55a80264f'
-GOLANG19_VERSION='1.19'
-GOLANG19_TAR_SHA256='464b6b66591f6cf055bc5df90a9750bf5fbc9d038722bb84a9d56a2bea974be6'
+GOLANG18_VERSION='1.18.8'
```
pauld-msft, line 7: I double checked and it looks like none of these or the dotnet constant variables are used in the codebase, otherwise we would need to worry about locking in the stretch 18 and 19 versions. So just a heads up @waliMSFT, golang will be very easy to lock in and it will only involve creating the final-stretch constants. Code substitutions elsewhere in the codebase won't be needed.

pauld-msft (dismissed):
:shipit:

waliMSFT (approved):
looks good to me.

doc/supportedPlatformVersions.md
```diff
@@ -284,7 +288,9 @@
 - 1.16.7
 - 1.17
 - 1.18
+- 1.18.8
 - 1.19
+- 1.19.3
```
waliMSFT, line 293: why we don't need to update the patch version of 1.18 for stretch images?

cormacpayne: We're only supporting buster and bullseye for the full image, so the stretch flavor isn't needed in this case.

PR #1692, opened 2022-11-18
Updates sym links for codespaces

samruddhikhandale:
Addresses - https://github.com/devcontainers/images/issues/187

The [universal image](https://github.com/devcontainers/images/tree/main/src/universal) is now used by Codespaces which utilizes the oryx tool (for auto-detection & auto-installation)

It installs java with [Java Feature](https://github.com/devcontainers/features/tree/main/src/java) - which installs it with `sdkman`. Java is first found at `/home/codespace/.java/current` which is sym linked to `/usr/local/sdkman/candidates/java/current`

For Java projects, oryx installs another version of Java and installs it in `/opt/java` (from source) & links to `/home/codespace/.java/current`. Hence, if the user then tries to set a default java version with sdkman, it doesn't work.

This PR attempts to fix that along with https://github.com/devcontainers/images/pull/219

This issue also happens for Node projects --> https://github.com/devcontainers/images/pull/220


cormacpayne (approved):
:shipit:

cormacpayne:
@samruddhikhandale thanks for the context and the change! Feel free to merge whenever and we'll notify you when the updated VSO/Codespaces image(s) are on MCR.

samruddhikhandale:
@cormacpayne Thanks for the review. For the issue I have reported in the PR details, it only needs this PR to be merged. Hence, no rush in updating/releasing the images from our end.

PR #1695, opened 2022-11-21
Update python and node versions

snehapar9:
- [<del>] Tests are included and/or updated for code changes.
- [<del>] Proper license headers are included in each file.


cormacpayne (approved):
:shipit: 

As a side note: we still need to determine whether or not we need these version-specific Dockerfiles since they are generated from the `template.Dockerfile` and will be generated (and I believe overridden) during our pipelines when building the runtime images, so I think we can add these to the `.gitignore`, but would need to double-check.

PR #1697, opened 2022-11-21
snehapar/add support for php 8.2.0

snehapar9:
- [<del>] Proper license headers are included in each file.

[User Story 1680376](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1680376): Support PHP 8.2.0 after release on 11/24


build/constants.yaml
```diff
@@ -167,15 +167,18 @@
...
+    php-fpm-runtime-base-tag: 20221209.2
     composer-version: 2.0.8
     composer-setup-sha384: 55ce33d7678c5a611085589f1f3ddf8b3c52d662cd01d4ba75c0ee0459970c2200a51f492d557530c71c15d8dba01eae
     composer1-9-version: 1.9.3
     composer1-10-version: 1.10.19
     composer2-0-version: 2.0.8
     composer2-2-version: 2.2.9
     composer2-3-version: 2.3.4
+    php82-version: 8.2.0
+    php82-keys: 1198C0117593497A5EC5C199286AF1F9897469DC 39B641343D8C104B2B146DC3F9C39DC0B9698544
```
daniv-msft, line 180: Could you please confirm: is it normal that the second php82-keys value is the same as the second php81-keys value?

doc/runtimes/php.md
```diff
@@ -88,7 +88,7 @@ The following process is applied to determine how to start an app:
 # Version support
 
 The PHP project defines this [release schedule][]. Oryx supports all actively supported
-releases (7.2, 7.3, 7.4, 8.0, 8.1), in addition to 5.6 & 7.0.
+releases (7.2, 7.3, 7.4, 8.0, 8.1, 8.2), in addition to 5.6 & 7.0.
```
daniv-msft, line 91: Should we update this sentence? We're not supporting 7.4 and below, as they're now EOL (even though we're still hosting the images). My suggestion would be to go with: `Oryx supports all [actively supported releases](https://www.php.net/supported-versions.php).`
The link to the PHP website allows us not to keep track of all these in our own doc.

doc/supportedPlatformVersions.md
```diff
@@ -912,6 +915,7 @@
 - 8.1.4
 - 8.1.6
 - 8.1.9
+- 8.2.0
```
daniv-msft, line 918: Just curious about this one: is a consequence of adding this to the 'stretch' versionToBuild, so that even older versions of App Service can still get the new update?

doc/supportedRuntimeVersions.md
```diff
@@ -144,10 +145,10 @@ The information won't show in runtime image tags, but we can always check the de
 - 3.1
 
 ### php
-- 7.4, 8.1
+- 7.4, 8.1, 8.2
```
daniv-msft, line 148: Looks like 7.4 should be dropped now, but 8.0 should be present. Would it please be possible for you to update it?

images/runtime/php-fpm/8.2/8.2.Dockerfile
```diff
@@ -0,0 +1,250 @@
...
+
+# do NOT merge this content with above line because the 
+# above line is shared across all php images
+# Install the Microsoft SQL Server PDO driver on supported versions only.
+#  - https://docs.microsoft.com/en-us/sql/connect/php/installation-tutorial-linux-mac
+#  - https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server
+RUN set -eux \
+	&& apt-get update \
+	&& apt-get install -y --no-install-recommends \
+		gnupg2 \
```
daniv-msft, line 14: I'm just curious: did we have to resolve any case of packages not supported on 8.2 yet, as it was just released? Or were all 8.1 packages also working with 8.2 automatically?

images/runtime/php/8.2/base.bullseye.Dockerfile
```diff
@@ -0,0 +1,146 @@
...
+# deprecated from 5.*, so should be avoided 
+RUN set -eux; \
+    if [[ $PHP_VERSION != 5.* && $PHP_VERSION != 7.0.* ]]; then \
+        echo "pecl/mongodb requires PHP (version >= 7.1.0, version <= 7.99.99)"; \
+        pecl install mongodb && docker-php-ext-enable mongodb; \
+    fi
+
+# https://github.com/microsoft/mysqlnd_azure, Supports  7.2*, 7.3* and 7.4*
+RUN set -eux; \
+    if [[ $PHP_VERSION == 7.2.* || $PHP_VERSION == 7.3.* || $PHP_VERSION == 7.4.* ]]; then \
```
daniv-msft, line 97: Could you please clarify: do we even need this condition in the 8.2 dockerfile? Is there any case where this Dockerfile could be used for non 8.2 scenarios?

build/constants.yaml
snehapar9: From here - https://www.php.net/downloads.php#gpg-8.2, it looks like the second values are the same for php81 and php82.

doc/runtimes/php.md
snehapar9: Updated.

doc/supportedPlatformVersions.md
snehapar9: Yep, since the stretch version tarballs are not in the debian-debian-flavor format, the App Service partners should be able to use this SDK with their older versions as well.
Additionally, the validation tests will fail if the new version is not added to the stretch versionToBuild.txt file since the final stretch version for php changes have not been released yet.

images/runtime/php-fpm/8.2/8.2.Dockerfile
snehapar9: All 8.1 packages worked with 8.2 as well automatically. We did _not_ have to resolve any packages.

images/runtime/php/8.2/base.bullseye.Dockerfile
snehapar9: Not fully sure if we need this condition here, since we dropped support for versions <= 7.4 but I saw this condition in [8.1](https://github.com/microsoft/Oryx/blob/main/images/runtime/php/8.1/base.bullseye.Dockerfile#L96) and [8.0](https://github.com/microsoft/Oryx/blob/main/images/runtime/php/8.0/base.buster.Dockerfile#L96) as well

build/constants.yaml
daniv-msft: Thanks for confirming!

doc/supportedPlatformVersions.md
daniv-msft: OK thanks

images/runtime/php-fpm/8.2/8.2.Dockerfile
daniv-msft: OK that's great. Thanks for confirming

daniv-msft (approved):

PR #1706, opened 2022-11-28
Wali/create final stretch image sdk version constants

waliMSFT:
We are currently using constants like [Oryx/constants.yaml at main · microsoft/Oryx (github.com)](https://github.com/microsoft/Oryx/blob/main/build/constants.yaml#L56-L95) to determine the current SDK version that should be used to pre-bake images.

We want to not have to keep releasing new patch versions for our stretch images, but updating the previous constants means that stretch images will attempt to download the new patch version, like our lts image here: [Oryx/ltsVersions.Dockerfile at main · microsoft/Oryx (github.com)](https://github.com/microsoft/Oryx/blob/main/images/build/Dockerfiles/ltsVersions.Dockerfile#L105-L110)

In order to keep stretch images from depending on the new SDK versions, we should create some new constants in the constants.yaml that coorespond to the _final_ stretch version. This should be the current versions for all languages, except for dotnet 3.1 and 6, which should be 3.1.26 and 6.0.6.


pauld-msft (changes requested):

images/build/Dockerfiles/Dockerfile
```diff
@@ -43,16 +43,16 @@ RUN set -ex \
     && mkdir -p $nugetPacakgesDir \
     && NUGET_PACKAGES="$nugetPacakgesDir" \
     && . $buildDir/__dotNetCoreSdkVersions.sh \
-    && DOTNET_SDK_VER=$DOT_NET_CORE_22_SDK_VERSION $imagesDir/build/installDotNetCore.sh \
+    && DOTNET_SDK_VER=$FINAL_STRETCH_DOT_NET_CORE_22_SDK_VERSION $imagesDir/build/installDotNetCore.sh \
     && DOTNET_SDK_VER=$DOT_NET_CORE_30_SDK_VERSION $imagesDir/build/installDotNetCore.sh \
```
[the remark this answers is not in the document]
waliMSFT: yah we do. probably missed it.

images/build/php/7.4/Dockerfile
```diff
@@ -1,3 +1,3 @@
 # Build PHP 7.4
 FROM oryxdevmcr.azurecr.io/private/oryx/php-build-prereqs AS php74-build
```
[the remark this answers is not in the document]
waliMSFT: yah I think you are right. PhP 7.4 is buster or bullseye image.

pauld-msft (changes requested):
I'm not sure that updating the final stretch versions for python 3.7 and 3.8 will allow us to actually release new versions in the future without testing failures. What we really need to do is go through all of the tests, find places where we are using the current constants (such as https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/VersionInformationTest.cs#L273), and then verify that the images used for the tests are stretch. Then, you will want to update the constant to refer to the final stretch version constant instead. In the link above, this would mean changing `PythonVersions.Python38Version` to `FinalStretchVersions.FinalStretchPython38Version` 

If this is not done for all versions we made a final stretch version for, the tests will break the next time a version is released.

Please take a look at the tests in the following examples:
- https://github.com/microsoft/Oryx/pull/1539/files
- https://github.com/microsoft/Oryx/pull/1627/files

build/__finalStretchVersions.sh
```diff
@@ -11,3 +27,73 @@ FINAL_STRETCH_DOT_NET_CORE_APP_70_RUN_TIME_VERSION='7.0.0'
...
+FINAL_STRETCH_PIP_VERSION='21.2.4'
+FINAL_STRETCH_PYTHON27_VERSION='2.7.18'
+FINAL_STRETCH_PYTHON36_VERSION='3.6.15'
+FINAL_STRETCH_PYTHON37_VERSION='3.7.16'
+FINAL_STRETCH_PYTHON38_VERSION='3.8.16'
+FINAL_STRETCH_PYTHON39_VERSION='3.9.16'
+FINAL_STRETCH_PYTHON310_VERSION='3.10.9'
+FINAL_STRETCH_PYTHON311_VERSION='3.11.1'
+FINAL_STRETCH_GEM_VERSION='3.3.10'
+FINAL_STRETCH_RUBY31_VERSION='3.1.1'
```
pauld-msft, line 87: Stretch has 3.1.3 now

build/__finalStretchVersions.sh
```diff
@@ -11,3 +27,73 @@ FINAL_STRETCH_DOT_NET_CORE_APP_70_RUN_TIME_VERSION='7.0.0'
...
+FINAL_STRETCH_RUBY31_TAR_SHA256='7aefaa6b78b076515d272ec59c4616707a54fc9f2391239737d5f10af7a16caa'
+FINAL_STRETCH_RUBY27_VERSION='2.7.2'
```
pauld-msft, line 89: stretch has 2.7.7 now

build/__finalStretchVersions.sh
```diff
@@ -11,3 +27,73 @@ FINAL_STRETCH_DOT_NET_CORE_APP_70_RUN_TIME_VERSION='7.0.0'
...
+FINAL_STRETCH_RUBY27_TAR_SHA256='6e5706d0d4ee4e1e2f883db9d768586b4d06567debea353c796ec45e8321c3d4'
+FINAL_STRETCH_RUBY26_VERSION='2.6.6'
+FINAL_STRETCH_RUBY26_TAR_SHA256='5db187882b7ac34016cd48d7032e197f07e4968f406b0690e20193b9b424841f'
+FINAL_STRETCH_RUBY25_VERSION='2.5.8'
+FINAL_STRETCH_RUBY25_TAR_SHA256='0391b2ffad3133e274469f9953ebfd0c9f7c186238968cbdeeb0651aa02a4d6d'
+FINAL_STRETCH_RUBY30_VERSION='3.0.3'
```
pauld-msft, line 95: stretch has 3.0.5 now

pauld-msft: see https://github.com/microsoft/Oryx/pull/1748

src/BuildScriptGenerator/FinalStretchVersions.cs
```diff
@@ -4,9 +4,25 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Common
 {
     public static class FinalStretchVersions
     {
+        public const string FinalStretchDotNetCore11SdkVersion = "1.1.14";
+        public const string FinalStretchDotNetCore21SdkVersion = "2.1.818";
+        public const string FinalStretchDotNetCore22SdkVersion = "2.2.207";
+        public const string FinalStretchNetCoreApp10 = "1.0.16";
+        public const string FinalStretchNetCoreApp11 = "1.1.13";
```
pauld-msft, line 11: please update all of these stretch versions to be the final one: https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs#L546-L580

src/BuildScriptGenerator/FinalStretchVersions.cs
```diff
@@ -4,9 +4,25 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Common
...
+        public const string FinalStretchNetCoreApp20 = "2.0.9";
+        public const string FinalStretchNetCoreApp21 = "2.1.30";
+        public const string FinalStretchAspnetCoreApp21 = "2.1.30";
+        public const string FinalStretchNetCoreApp22 = "2.2.8";
+        public const string FinalStretchDotNetCore30SdkVersion = "3.0.103";
+        public const string FinalStretchNetCoreApp30 = "3.0.3";
+        public const string FinalStretchAspnetCoreApp30 = "3.0.3";
         public const string FinalStretchDotNetCore31SdkVersion = "3.1.420";
```
pauld-msft, line 19: This was my bad for missing it, but would you be able to make this configure the default based on the debian flavor? https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreOnDiskVersionProvider.cs#L23

You can just have it accept a parameter of debianFlavor, which you can pull from the `BuildScriptGeneratorOptions` that exists wherever it gets invoked

pauld-msft: also for 3.1, it looks like there are some stretch tests in this file that should get updated https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/BaseImage/BaseImageTest.cs#L43

pauld-msft: Also, just realized that we have updated the stretch version of 3.1.26

@daniv-msft do you know how long we are going to be updating the stretch versions for? The whole purpose of this PR was to lock in a final stretch version so that we can safely stop updating the stretch patch versions, but since we have started uploading them again it makes this whole thing more work for us as a team since we would need to update these values as well whenever we add a new stretch patch version...

pauld-msft: As soon as Antares moves to a newer image we are ok to stop uploading stretch versions, right?

pauld-msft (changes requested):
Would like to hear back from @daniv-msft about when we can expect to be able to lock in the stretch versions.

In the meantime, could you please take another look at all the places that constants like the ones in this [file](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/DotNetCore/DotNetCoreRunTimeVersions.cs) are used? I was able to find a few stragglers in the versions of dotnet that i checked (but i only checked up until v3.1)

daniv-msft:
> Would like to hear back from @daniv-msft about when we can expect to be able to lock in the stretch versions.
> 
> In the meantime, could you please take another look at all the places that constants like the ones in this [file](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/DotNetCore/DotNetCoreRunTimeVersions.cs) are used? I was able to find a few stragglers in the versions of dotnet that i checked (but i only checked up until v3.1)

Yes that's annoying. Just to refresh my memory: we can't complete this work and have final stretch versions until we don't need to build any of them anymore, right?
I took a note to ask our partners about this to know if they had any updates on their rollout that would get rid of Stretch versions.

daniv-msft:
> > Would like to hear back from @daniv-msft about when we can expect to be able to lock in the stretch versions.
> > In the meantime, could you please take another look at all the places that constants like the ones in this [file](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/DotNetCore/DotNetCoreRunTimeVersions.cs) are used? I was able to find a few stragglers in the versions of dotnet that i checked (but i only checked up until v3.1)
> 
> Yes that's annoying. Just to refresh my memory: we can't complete this work and have final stretch versions until we don't need to build any of them anymore, right? I took a note to ask our partners about this to know if they had any updates on their rollout that would get rid of Stretch versions.

(Tagging @pauld-msft and @waliMSFT)

pauld-msft:
> > Would like to hear back from @daniv-msft about when we can expect to be able to lock in the stretch versions.
> > In the meantime, could you please take another look at all the places that constants like the ones in this [file](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/DotNetCore/DotNetCoreRunTimeVersions.cs) are used? I was able to find a few stragglers in the versions of dotnet that i checked (but i only checked up until v3.1)
> 
> Yes that's annoying. Just to refresh my memory: we can't complete this work and have final stretch versions until we don't need to build any of them anymore, right? I took a note to ask our partners about this to know if they had any updates on their rollout that would get rid of Stretch versions.

@daniv - we _could_, but we would either need to add some code into the Generate Constants project that will update these versions based on the most recent stretch version in versionsToBuild.txt, or we would need to manually update these constants as we released new stretch versions

daniv-msft:
> > > Would like to hear back from @daniv-msft about when we can expect to be able to lock in the stretch versions.
> > > In the meantime, could you please take another look at all the places that constants like the ones in this [file](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/DotNetCore/DotNetCoreRunTimeVersions.cs) are used? I was able to find a few stragglers in the versions of dotnet that i checked (but i only checked up until v3.1)
> > 
> > 
> > Yes that's annoying. Just to refresh my memory: we can't complete this work and have final stretch versions until we don't need to build any of them anymore, right? I took a note to ask our partners about this to know if they had any updates on their rollout that would get rid of Stretch versions.
> 
> @daniv - we _could_, but we would either need to add some code into the Generate Constants project that will update these versions based on the most recent stretch version in versionsToBuild.txt, or we would need to manually update these constants as we released new stretch versions

OK thanks. Let's see what our partners say. If I remember correctly, they were in the process of rolling out (that takes a couple of months) so hopefully they should be done in the coming weeks, and we don't need to integrate a workaround.

PR #1707, opened 2022-11-28
Harrli/gunicorn num cores update

harryli0108:
Updating python gunicorn multiworker multithread functionality.
Added two extra environments: PYTHON_GUNICORN_CUSTOM_WORKER_NUM and PYTHON_GUICORN_CUSTOM_THREAD_NUM.
Steps to enable MWMT:
1.	Set PYTHON_ENABLE_GUNICORN_MULTIWORKERS to true;
2.	Set PYTHON_GUNICORN_CUSTOM_WORKER_NUM to the desired number of workers;
3.	Set PYTHON_GUNICORN_CUSTOM_THREAD_NUM to the desired number of threads.

1.	If user doesn’t enable multiworkers, then the default worker would be single worker;
2.	If user does enable multiworkers but doesn’t set custom worker number, then it will use the system suggestion to set number of worker to (2 * numCores) + 1;
3.	If user does enable multiworkers and set custom worker number but doesn’t set custom thread number, then it will use the default single thread.


src/startupscriptgenerator/src/python/scriptgenerator.go
```diff
@@ -282,11 +282,22 @@ func (gen *PythonStartupScriptGenerator) buildGunicornCommandForModule(module st
...
+		// Otherwise, this script will use the recommended setting by Gunicorn.
+		pythonCustomWorkerNum := os.Getenv(consts.PythonGunicornCustomWorkerNum)
+		pythonCustomThreadNum := os.Getenv(consts.PythonGunicornCustomThreadNum)
+		workers := ""
+		if ((pythonCustomWorkerNum != "") && (pythonCustomThreadNum != "")) {
+			workers = pythonCustomWorkerNum
+			args = appendArgs(args, "--threads="+pythonCustomThreadNum)
+		} else {
+			workers = strconv.Itoa((2 * runtime.NumCPU()) + 1)
+		}
```
cormacpayne, line 300: This handles the case that Antares requested (_i.e.,_ setting both values based on the SKU), but from Oryx's point of view, we should be able to handle users who want to provide different combination of gunicorn arguments when running their application. The logic above ignores the cases where the user provides exactly one of the worker or threads arguments, so we should clean it up a little to account for either provided.

```go
if (pythonCustomThreadNum != "") {
    args = appendArgs(args, "--threads="+pythonCustomThreadNum)
}

workers := strconv.Itoa((2 * runtime.NumCPU()) + 1)
if (pythonCustomWorkerNum != "") {
    workers = pythonCustomWorkerNum
}

args = appendArgs(args, "--workers="+workers)
```

cormacpayne (changes requested):
Two additional comments:
- Let's update the [`configuration.md`](https://github.com/microsoft/Oryx/blob/main/doc/configuration.md) document to call out the two new environment variables added in the Python section
- If possible, let's add test coverage to ensure that the run script that we generate with these environment variables set contains the correct arguments
  - The `CanBuildAndRun_NetCore60MvcApp_WithCustomizedRunCommand` test in the .NET Core integration test suite has a good example of checking the contents of a generated run script for specific strings

src/startupscriptgenerator/src/python/scriptgenerator.go
harryli0108: That's a good point. I haven't thought about that corner case. Just added it. I've also added some documentations in configuration.md

doc/configuration.md
```diff
@@ -54,6 +54,9 @@ DISABLE\_PYTHON\_BUILD       | Do not apply Python build even if repo indicates
 VIRTUALENV\_NAME             | Specify Python virtual environment name                        | ""      | "antenv2.7"
 DISABLE\_COLLECTSTATIC       | Disable running `collectstatic` when building Django apps.     | `false` | `true`, `false`
 CUSTOM\_REQUIREMENTSTXT\_PATH| Specify where a requirements.txt is locating. If not set, default is at root of the repo.| ""    | "subdir/requirements.txt"
+PYTHON\_ENABLE\_GUNICORN\_MULTIWORKERS| Enable Gunicorn multi worker multi thread config.     | `false` | `true`, `false`
+PYTHON\_GUNICORN\_CUSTOM\_WORKER\_NUM| Specify Gunicorn multi worker number. If not set, default is (2 * CPU core num) + 1| ""    | "2"
+PYTHON\_GUICORN\_CUSTOM\_THREAD\_NUM| Specify Gunicorn multi thread number. If not set, default is single thread. | ""    | "4"
```
cormacpayne, line 59: nit: typo in the environment variable name: `GUICORN` --> `GUNICORN`

doc/configuration.md
```diff
@@ -54,6 +54,9 @@ DISABLE\_PYTHON\_BUILD       | Do not apply Python build even if repo indicates
 VIRTUALENV\_NAME             | Specify Python virtual environment name                        | ""      | "antenv2.7"
 DISABLE\_COLLECTSTATIC       | Disable running `collectstatic` when building Django apps.     | `false` | `true`, `false`
 CUSTOM\_REQUIREMENTSTXT\_PATH| Specify where a requirements.txt is locating. If not set, default is at root of the repo.| ""    | "subdir/requirements.txt"
+PYTHON\_ENABLE\_GUNICORN\_MULTIWORKERS| Enable Gunicorn multi worker multi thread config.     | `false` | `true`, `false`
+PYTHON\_GUNICORN\_CUSTOM\_WORKER\_NUM| Specify Gunicorn multi worker number. If not set, default is (2 * CPU core num) + 1| ""    | "2"
```
cormacpayne, line 58: nit: for these two environment variables you added, we can set the default value column to `(2 * CPU core num) + 1` and `1`, respectively

cormacpayne, line 58: nit: I think it's worth calling out in the description of the two environment variables that you added that they'll only be used if `PYTHON_ENABLE_GUNICORN_MULTIWORKERS` is set to `true`

src/startupscriptgenerator/src/python/scriptgenerator.go
```diff
@@ -282,12 +282,25 @@ func (gen *PythonStartupScriptGenerator) buildGunicornCommandForModule(module st
 	}
 	
 	pythonEnableGunicornMultiWorkers := common.GetBooleanEnvironmentVariable(consts.PythonEnableGunicornMultiWorkersEnvVarName)
+
 	if pythonEnableGunicornMultiWorkers {
 		// 2N+1 number of workers is recommended by Gunicorn docs.
 		// Where N is the number of CPU threads.
 		// One worker will be reading or writing from the socket while the other worker is processing a request.
```
cormacpayne, line 289: nit: let's move this block comment about the `2N+1` into the `else` statement below where we set `workers` to that calculated value

cormacpayne (dismissed):
Some minor comments, otherwise LGTM

cormacpayne (approved):

PR #1741, opened 2022-12-13
Pauld/fix jupiter rise

pauld-msft:
Fixes the jupiter rise sample app by adding [some packages](https://github.com/python-jsonschema/jsonschema/blob/b4513dc84737097b3ad9d390f42c004eabaa8f16/pyproject.toml#L56-L65) that are requirements of the jsonschema package, but were not being brought in as dependencies. 

Also fix a separate issue that we had not detected, which caused github action builds to fail.


images/build/Dockerfiles/gitHubActions.Dockerfile
```diff
@@ -45,9 +45,9 @@ RUN if [ "${DEBIAN_FLAVOR}" = "bullseye" ]; then \
             libssl1.1 \
             libyaml-dev \
         && rm -rf /var/lib/apt/lists/* \
-        && curl -LO http://security.debian.org/debian-security/pool/updates/main/libx/libxml2/libxml2_2.9.10+dfsg-6.7+deb11u2_amd64.deb \
-        && dpkg -i libxml2_2.9.10+dfsg-6.7+deb11u2_amd64.deb \
-        && rm libxml2_2.9.10+dfsg-6.7+deb11u2_amd64.deb ; \
+        && curl -LO http://security.debian.org/debian-security/pool/updates/main/libx/libxml2/libxml2_2.9.10+dfsg-6.7+deb11u3_amd64.deb \
```
pauld-msft, line 48: Github action builds for bullseye failing because the `u2` no logner exists at the above link. I've updated to the latest security update, detailed in https://www.debian.org/security/2022/dsa-5271

cormacpayne (approved):
:shipit:

william-msft (approved):

PR #1743, opened 2022-12-14
Http request exception handling

pauld-msft:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1510080

Sample exception:
```
root@8c28e900cce6:/# oryx build //app
Operation performed by Microsoft Oryx, https://github.com/Microsoft/Oryx
You can report issues at https://github.com/Microsoft/Oryx/issues

Oryx Version: 0.2.0.0, Commit: e1e2de86aecf7da54ae4fc2c0211ef5e2b566e15, ReleaseTagName:

Build Operation ID: |jXkJAKLVGbk=.36382fa_

Detecting platforms...
Error: Oops... An unexpected error has occurred.
Error: System.AggregateException: Http request to 'https://oryx-cdn-fake.microsoft.io' failed. Please ensure that your network configuration allows traffic to the oryx dependencies, as documented in https://github.com/microsoft/Oryx/blob/main/doc/hosts/appservice.md#network-dependencies
System.AggregateException: One or more errors occurred. (A task was canceled.)
 ---> System.Threading.Tasks.TaskCanceledException: A task was canceled.
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at System.Threading.Tasks.Task`1.get_Result()
   at Microsoft.Oryx.BuildScriptGenerator.ListBlobsHelper.GetAllBlobs(String sdkStorageBaseUrl, String platform, HttpClient httpClient, String oryxSdkStorageAccountAccessToken) in /usr/oryx/src/BuildScriptGenerator/Helpers/ListBlobsHelper.cs:line 24
   at Microsoft.Oryx.BuildScriptGenerator.ListBlobsHelper.GetAllBlobs(String sdkStorageBaseUrl, String platform, HttpClient httpClient, String oryxSdkStorageAccountAccessToken) in /usr/oryx/src/BuildScriptGenerator/Helpers/ListBlobsHelper.cs:line 32

```


snehapar9 (approved):

src/BuildScriptGenerator/Helpers/ListBlobsHelper.cs
```diff
@@ -2,22 +2,41 @@
...
+            }
+
+            // AggregateException is thrown by .Result
+            catch (AggregateException ae)
+            {
+                throw new AggregateException(
+                        $"Http request to '{sdkStorageBaseUrl}' failed. {Constants.NetworkConfigurationHelpText}\n{ae}");
+            }
+
+            if (string.IsNullOrEmpty(blobList))
```
daniv-msft, line 35: I'm just curious: did this happen? Did we sometimes get some empty results from the call? I'm just curious to know if this is a preventive check to make sure this never happens, or if we actually did see that happening sometimes.

src/BuildScriptGenerator/Helpers/ListBlobsHelper.cs
```diff
@@ -2,22 +2,41 @@
...
-                .GetStringAsync(url)
-                .Result;
+            string blobList;
+            try
+            {
+                blobList = httpClient
+                    .GetStringAsync(url)
+                    .Result;
+            }
+
```
daniv-msft, line 27: Nit: Not a big deal, so feel free to ignore if you feel strongly about it.
Most of the time, we don't let any spaces between the try block and the catch one, so that it's clear that all this is a common thing. Looking at the Oryx code (https://github.com/microsoft/Oryx/search?l=C%23&q=catch&type=code), putting them together seems pretty standard in our codebase.
However, again it's not a big deal so if you prefer this (and can convince most of the time to go with this new convention), I'm good with it.

pauld-msft: I only did it like that because otherwise stylecop fails the build because i have to have an empty line above a comment (and I cannot do an inline comment after the line either thanks to stylecop)

I don't really like it, but if you can think of a better place to put the comment, let me know!

src/BuildScriptGenerator/Helpers/ListBlobsHelper.cs
pauld-msft: No, its preventative. Mainly because I need to initialize the `blobList` variable before the try/catch block, and I initialize it to empty. So I just wanted to cover all my bases and basically to a null check before assuming that the http request returned as expected and trying to parse it.

src/BuildScriptGenerator/Helpers/ListBlobsHelper.cs
```diff
@@ -2,22 +2,41 @@
...
+            // AggregateException is thrown by .Result
+            catch (AggregateException ae)
+            {
+                throw new AggregateException(
+                        $"Http request to '{sdkStorageBaseUrl}' failed. {Constants.NetworkConfigurationHelpText}\n{ae}");
```
daniv-msft, line 32: Please consider using Environment.NewLine rather than hardcoding "\n". Especially if in the future we switch between Windows and Linux for some stuff (I know it wouldn't impact here, but you see what I mean), having a consistent way to do things will make sure we don't have surprises on one OS and not in the other.

src/BuildScriptGenerator/SdkStorageVersionProviderBase.cs
```diff
@@ -97,9 +98,25 @@ protected string GetDefaultVersion(string platformName, string sdkStorageBaseUrl
...
+            // AggregateException is thrown by .Result
+            catch (AggregateException ae)
+            {
+                throw new AggregateException(
+                        $"Http request to '{defaultVersionUrl}' failed. {Constants.NetworkConfigurationHelpText}\n{ae}");
+            }
+
+            if (string.IsNullOrEmpty(defaultVersionContent))
+            {
+                throw new InvalidOperationException($"Http request to '{defaultVersionContent}' cannot return an empty result.");
```
daniv-msft, line 118: I believe this one should be `defaultVersionUrl`, and not `defaultVersionContent`

src/BuildScriptGenerator/Helpers/ListBlobsHelper.cs
daniv-msft, line 32: Nit: please consider adding a bit more details to the error message, e.g.:
```
Http request to get the SDKs available to download (URL '{sdkStorageBaseUrl}') failed. ...
```

This might be more satisfying for the customers/customer support folks, as well as giving us a 100% certain way to differentiate between the error thrown here and the one with same text thrown in SdkStorageVersionProviderBase.cs.
Obviously I'd have the same suggestion there as well.

src/BuildScriptGenerator/Helpers/ListBlobsHelper.cs
daniv-msft: Honestly, I'd consider just dropping the comment? It seems pretty natural to me that if you're catching specifically this exception, this is what is thrown in the try {} block.
Alternatively, maybe putting it after the first bracket of the catch block?

src/BuildScriptGenerator/Helpers/ListBlobsHelper.cs
daniv-msft: OK fair enough, thanks!

cormacpayne (approved):
:shipit:

src/BuildScriptGenerator/SdkStorageVersionProviderBase.cs
```diff
@@ -97,9 +98,25 @@ protected string GetDefaultVersion(string platformName, string sdkStorageBaseUrl
...
+                throw new AggregateException(
+                    $"Http request to retrieve the default version from '{defaultVersionUrl}' failed. " +
+                    $"{Constants.NetworkConfigurationHelpText}{Environment.NewLine}{ae}");
+            }
+
+            if (string.IsNullOrEmpty(defaultVersionContent))
+            {
+                throw new InvalidOperationException(
+                    $"Http request to retrieve the default version from '{defaultVersionUrl}' cannot return an empty result.");
+            }
```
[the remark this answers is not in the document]
pauld-msft: Yep makes sense. I _generally_ will try to follow the rule of 3, and so I thought this would be fine as is. Getting close to that point though. If people feel strongly, I'd be happy to spend some more time investing a bit more by extending httpclient.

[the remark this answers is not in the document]
cormacpayne: No need -- just an idea if we do end up reusing this snippet again in the future for any additional calls, but for now it's good as-is 👍

daniv-msft (approved):

PR #1746, opened 2022-12-15
Added a sort step when handling sub directories generating support docs

harryli0108:
Possible reason for the support doc to be reshuffled: 
- C# GetDirectory is not stable in terms of ordering the result. Adding a sorting step could fix the problem.
- Since most of these issues came from github automation upgrading versions, the environment that github action was using. Some language packs preinstalled into the os could be different from Windows language packs (e.g. C.UTF8 and enUS.UTF8 has different alphabetic ordering rules.)


build/tools/SharedCodeGenerator/Program.cs
```diff
@@ -107,8 +107,11 @@ private static void GenerateSupportedPlatformsReadmeFile(string repoDir)
             {
                 sw.WriteLine("# Supported platforms and versions");
                 sw.WriteLine();
-                foreach (var subDirPath in Directory.GetDirectories(platformsDir))
+                string[] subDirs = Directory.GetDirectories(platformsDir);
+                string[] sortedSubDirs = subDirs.OrderBy(n => n).ToArray();
```
daniv-msft, line 111: Nit: feel free to ignore. Could you please clarify why we needed the `.ToArray()`? I would have expected that OrderBy would return an IEnumerable or a similar type, that we could directly use in the foreach.

build/tools/SharedCodeGenerator/Program.cs
```diff
@@ -107,8 +107,11 @@ private static void GenerateSupportedPlatformsReadmeFile(string repoDir)
...
+                foreach (var subDirPath in sortedSubDirs)
                 {
+                    Console.WriteLine(subDirPath);
```
daniv-msft, line 114: Could you clarify if we need to keep the Console.WriteLine? It feels like if we need to keep it, then we should add more info in there to make the log more actionable.

daniv-msft (dismissed):
Looks good to me, only minor comments

build/tools/SharedCodeGenerator/Program.cs
harryli0108: completely missed this one when cleaning up the code.. I'll delete that asap

build/tools/SharedCodeGenerator/Program.cs
harryli0108: that makes sense! Those were for debugging if the `orderby` did anything. Just changed it with the new commit.

snehapar9 (approved):

build/tools/SharedCodeGenerator/Program.cs
```diff
@@ -107,7 +107,7 @@ private static void GenerateSupportedPlatformsReadmeFile(string repoDir)
             {
                 sw.WriteLine("# Supported platforms and versions");
                 sw.WriteLine();
-                foreach (var subDirPath in Directory.GetDirectories(platformsDir))
+                foreach (var subDirPath in Directory.GetDirectories(platformsDir).OrderBy(n => n))
                 {
                     var subDirInfo = new DirectoryInfo(subDirPath);
                     var platformSubDirPaths = Directory.GetDirectories(subDirPath);
```
pauld-msft, line 113: Should we also enforce order for this call to `GetDirectories`?

daniv-msft:
I would have expected the doc file to be impacted by this sorting, and the file to be part of this PR. Were we lucky and the file was already perfectly sorted?

build/tools/SharedCodeGenerator/Program.cs
daniv-msft: Thanks!

build/tools/SharedCodeGenerator/Program.cs
harryli0108: i think it would be ok since the `subDirPath` is sorted, but just to be safe i'll add sorting here as well. thanks!

harryli0108:
> I would have expected the doc file to be impacted by this sorting, and the file to be part of this PR. Were we lucky and the file was already perfectly sorted?

I think the support docs are in the right order now since some of the new merges were not from github action. We should see changes when dependentBot or github action workflow opens pr and updates the dependency. The diff should be just appending new version numbers rather than reshuffling the whole doc.

daniv-msft:
> > I would have expected the doc file to be impacted by this sorting, and the file to be part of this PR. Were we lucky and the file was already perfectly sorted?
> 
> I think the support docs are in the right order now since some of the new merges were not from github action. We should see changes when dependentBot or github action workflow opens pr and updates the dependency. The diff should be just appending new version numbers rather than reshuffling the whole doc.

Sounds good, thanks for clarifying!

daniv-msft (approved):

PR #1753, opened 2022-12-19
Dependency changes

snehapar9:
[User Story 1501523](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1501523): Improve StaticWebApps build times
https://microsoft.sharepoint.com/:w:/t/Oryx350/Ec5r7BUD1F5Gup5U0Y5YOt0BldwKeVYcAgyzyWLj6IyHrg?e=CB1akR


images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -11,38 +11,78 @@ ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
...
+        # Required for mysqlclient
+        default-libmysqlclient-dev \
+        # Required for ts
+        moreutils \
+        rsync \
+        zip \
+        tk-dev \
+        uuid-dev \
+        #TODO : Add these to fix php failures. Check if these can be removed.
+        libargon2-0 \
```
snehapar9, line 42: Added these packages to fix validation pipelines errors even though these are dynamically installed for PHP.

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
```diff
@@ -82,6 +82,19 @@ public DotNetCoreDynamicInstallTest(ITestOutputHelper output) : base(output)
                 appName, runtimeVersion, _imageHelper.GetCliImage(ImageTestHelperConstants.CliBusterRepository));
         }
 
+        [Theory, Trait("category", "cli-bullseye")]
+       // [InlineData(NetCoreApp21WebApp, "2.1")]
+        [InlineData(NetCoreApp31MvcApp, "3.1")]
+       // [InlineData(NetCoreApp50MvcApp, "5.0")]
+        [InlineData(NetCore7PreviewMvcApp, "7.0")]
+        public void BuildsApplication_ByDynamicallyInstallingSDKs_CliBullseye(
```
snehapar9, line 90: Seems like dotnet 2.1 and dotnet 5.0 are not among the supported versions? Happy to remove these instead of commenting after confirming.

tests/Oryx.BuildImage.Tests/Hugo/HugoDynamicInstallationTest.cs
```diff
@@ -115,7 +122,7 @@ public void DynamicInstall_ReInstallsSdk_IfSentinelFileIsNotPresent()
             // Act
             var result = _dockerCli.Run(new DockerRunArguments
             {
-                ImageId = _restrictedPermissionsImageHelper.GetGitHubActionsBuildImage(),
+                ImageId = _imageHelper.GetGitHubActionsBuildImage(),
```
snehapar9, line 125: Not fully sure if we need a test image with restricted permissions for this test. Similar tests for [dotnet](https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs#L172), [node](https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/Node/NodeDynamicInstallationTest.cs#L226) and [python](https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/Python/PythonDynamicInstallationTest.cs#L261) do not use  test image with restricted permissions.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -11,38 +11,78 @@ ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
     PYTHONIOENCODING="UTF-8" \
     LANG="C.UTF-8" \
     LANGUAGE="C.UTF-8" \
-    LC_ALL="C.UTF-8"
-
-ARG IMAGES_DIR="/opt/tmp/images"
-RUN oryx prep --skip-detection --platforms-and-versions nodejs=14 --debug \
-    && echo "$DEBIAN_FLAVOR" \
```
snehapar9, line 18: Removed oryx prep step as I believe it installs nodejs, yarn and python dependencies redundantly.

build/buildBuildImages.sh
```diff
@@ -533,11 +535,14 @@ elif [ "$imageTypeToBuild" == "vso-focal" ]; then
 	buildVsoImage "focal"
 elif [ "$imageTypeToBuild" == "vso-bullseye" ]; then
 	buildVsoImage "bullseye"
-elif [ "$imageTypeToBuild" == "cli" ]; then
+elif [ "$imageTypeToBuild" == "cli-stretch" ]; then
 	buildCliImage
 	buildCliImage "buster"
+	buildCliImage "bullseye"
```
cormacpayne, line 541: nit: should we only call `buildCliImage` if `cli-stretch` is provided?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -11,38 +11,78 @@ ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
...
-    && rm -rf $downloadedFileName
-
-RUN set -ex \
-    # Install Python SDKs
-    # Upgrade system python
-    && PYTHONIOENCODING="UTF-8" \
-    # It's not clear whether these are needed at runtime...
-    && apt-get update \
+    LC_ALL="C.UTF-8" \
+    ORYX_PATHS="/opt/oryx:/opt/nodejs/lts/bin:/opt/python/latest/bin:/opt/yarn/stable/bin"
```
cormacpayne, line 15: nit: just to confirm: this means the Jamstack image will still come pre-baked with node, Python and Yarn, correct?

src/BuildScriptGenerator/DotNetCore/DotNetCorePlatformInstaller.cs
```diff
@@ -47,8 +47,9 @@ public override void InstallPlatformSpecificSkeletonDependencies(StringBuilder s
                 "zlib1g",
                 "libuuid1",
                 "libunwind8");
-
+            stringBuilder.AppendLine("if grep -q cli \"/opt/oryx/.imagetype\"; then");
```
cormacpayne, line 50: This change means we no longer install Python for the Jamstack image, correct?

src/BuildScriptGenerator/Hugo/HugoPlatformInstaller.cs
```diff
@@ -32,6 +32,10 @@ public virtual string GetInstallerScriptSnippet(string version)
             var snippet = new StringBuilder();
             snippet
                 .AppendLine()
+                .AppendLine("IMAGES_DIR=\"/opt/tmp/images\"")
```
cormacpayne, line 35: This change ensures that we don't need to pre-bake Hugo for all images and only install it at run-time for the CLI image, correct?

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -97,7 +113,7 @@ public virtual void InstallPlatformSpecificSkeletonDependencies(StringBuilder st
             var snippet = new StringBuilder();
             snippet
                 .AppendLine()
-                .AppendLine($"if grep -q cli \"/opt/oryx/.imagetype\"; then")
+                .AppendLine($"if grep -q -e cli \"/opt/oryx/.imagetype\" -e jamstack \"/opt/oryx/.imagetype\"; then")
```
cormacpayne, line 116: nit: just to clarify (since I'm not too familiar with `grep`): this is installing skeleton dependencies if the `.imagetype` file contains _either_ `cli` or `jamstack`, correct?

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
cormacpayne: .NET Core 3.1 is EOL as well, so we may be able to scrap that as well? I'd defer to @daniv-msft

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
snehapar9: Yep, that is correct.

src/BuildScriptGenerator/DotNetCore/DotNetCorePlatformInstaller.cs
snehapar9: Yep, correct. We would only install SDKs for versions other than 3.8

src/BuildScriptGenerator/Hugo/HugoPlatformInstaller.cs
snehapar9: Added this line here to run installHugo.sh script here _only_ for jamstack image so hugo and it's dependencies (golang) are installed dynamically.

For the cli image, however, [this](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/Node/NodePlatformInstaller.cs#L43) line runs the installHugo.sh script.

src/BuildScriptGenerator/PlatformInstallerBase.cs
snehapar9: yep, that is correct.

build/buildBuildImages.sh
```diff
@@ -410,7 +410,8 @@ function buildCliImage() {
 
 	if [ -z "$debianFlavor" ] || [ $debianFlavor == "stretch" ] ; then
 		debianFlavor="stretch"
-		builtImageName="$builtImageName:debian-$debianFlavor"
+		#Change buildImage name to fix validation pipeline
+		builtImageName="$builtImageName-$debianFlavor:debian-$debianFlavor"
```
daniv-msft, line 414: Could you please provide more context about the double -$debianFlavor info? I see that we do it in the `else` condition but this seems redundant. Do you have more info?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG PARENT_DEBIAN_FLAVOR
-FROM oryxdevmcr.azurecr.io/public/oryx/build:github-${PARENT_DEBIAN_FLAVOR} AS main
 ARG DEBIAN_FLAVOR
+FROM oryxdevmcr.azurecr.io/public/oryx/cli-${DEBIAN_FLAVOR}:${PARENT_DEBIAN_FLAVOR} AS main
```
daniv-msft, line 3: Just a random question, not a comment on the code: do you know the difference between DEBIAN_FLAVOR and PARENT_DEBIAN_FLAVOR? In which case are these different?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -11,38 +11,77 @@ ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
...
-RUN set -ex \
-    # Install Python SDKs
-    # Upgrade system python
-    && PYTHONIOENCODING="UTF-8" \
-    # It's not clear whether these are needed at runtime...
-    && apt-get update \
+    LC_ALL="C.UTF-8" \
+    ORYX_PATHS="/opt/oryx:/opt/nodejs/lts/bin:/opt/python/latest/bin:/opt/yarn/stable/bin"
+    
+RUN apt-get update \
```
daniv-msft, line 17: Could you please clarify if the `RUN set -ex \` was removed on purpose? It seems still valuable as we add it in most scripts to exit on first failure and print the commands used, but maybe in this case we don't want that anymore?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
daniv-msft: Could you please clarify the consequences of pre-baking these? Should we have any concerns in terms of the size of adding these, or is it negligible and not worth looking into another solution to fix the validation pipeline?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -11,38 +11,77 @@ ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
...
+ARG IMAGES_DIR="/opt/tmp/images"
+ARG BUILD_DIR="/opt/tmp/build"
 
+
+RUN set -ex \
+ && yarnCacheFolder="/usr/local/share/yarn-cache" \
+ && mkdir -p $yarnCacheFolder \
+ && chmod 777 $yarnCacheFolder \
+ && . ${BUILD_DIR}/__nodeVersions.sh \
+ && if [ "${DEBIAN_FLAVOR}" == "bullseye" || "${DEBIAN_FLAVOR}" == "buster" ]; then ${IMAGES_DIR}/installPlatform.sh nodejs ${NODE16_VERSION}; fi \
```
daniv-msft, line 57: I imagine we don't have anything better, but just asking in case we're lucky: do we have any variable that would allow us to use the current default Node version, rather than hardcoding it to Node 16? That would be one less thing to remember to update in the future.

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
daniv-msft: +1 on deleting the obsolete versions at this point. And another +1 to include 3.1 in the versions we delete as well, as we no longer support it.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
william-msft: Ideally we should be keeping `set -ex` since it is helpful for debugging and avoiding false negatives.

william-msft:
@snehapar9 does the con listed in your [doc ](https://microsoft.sharepoint.com/:w:/t/Oryx350/Ec5r7BUD1F5Gup5U0Y5YOt0BldwKeVYcAgyzyWLj6IyHrg?e=CB1akR)still applies?
`Cons – Need to work with SWA team to investigate the known root user issue encountered previously with this change. `

build/buildBuildImages.sh
daniv-msft: Thanks @pauld-msft for the explanation! Yes, I agree I would rather go on the long term with a consistent approach, specifically `oryx/build:cli-debian-buster-version`. It feels like whatever we can simplify in terms of images is a win, so it would go this way. If we rename this, we could even go when step further and rename the CLI image to "skeleton" as we talked about in the past, to remove even further the ambiguity about what is included in there.

Now in terms of short term work, we're not going to do this in this PR. I'm not sure however if the move to `cli-stretch` puts us closer to the long term goal, or if it will make things more complicated later. If it's neutral, then I'm fine with it, if we expect this would just create more work for us later, then I would rather stick with the old format. @pauld-msft @snehapar9 Any thoughts?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
snehapar9: In terms of size, pre-baking these are pretty insignificant.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
snehapar9: Currently, [this](https://github.com/microsoft/Oryx/blob/main/build/buildBuildImages.sh#L232) is the arg for PARENT_DEBIAN_FLAVOR which is essentially just the parent/base image's tag.

daniv-msft: Thanks for your reply! So just to be sure I got it right, `github` is the value passed as `PARENT_DEBIAN_FLAVOR`? If yes then that seems like something else we'll need to rename as some point.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
daniv-msft: Sounds good, thanks for clarifying

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
snehapar9: Yep not sure if we have something as of today to keep track of the default version being used. Briefly chatted with @Cormac about this.
We may be able to add a constant in constants.yml but since the default versions do not change that often, this may be a far-fetched solution.

snehapar9:
> @snehapar9 does the con listed in your [doc ](https://microsoft.sharepoint.com/:w:/t/Oryx350/Ec5r7BUD1F5Gup5U0Y5YOt0BldwKeVYcAgyzyWLj6IyHrg?e=CB1akR)still applies? `Cons – Need to work with SWA team to investigate the known root user issue encountered previously with this change. `
Yep this still applies, there may be some permission issues while installing apt-get packages.
A quick workaround I see for this, would be to pre-install these, since these packages are not significantly that large in size.
We may also be able to do something similar to what we did for the npm packages. Adding @pauld-msft for more context on how we resolved this for npm?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
snehapar9: Updated Dockerfile

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
snehapar9: Removed these for this test.

build/buildBuildImages.sh
cormacpayne: Chatted with @snehapar9 about this offline -- I'm in the boat that we should move to the single repo approach of `oryx/cli:debian-{flavor}-{tag}` since (1) we already have the `oryx/cli` repository registered with MCR and (2) we already follow that pattern with our `oryx/build` images and would make sense to be consistent here.

I don't think this would be a large amount of work, and would probably warrant being done in its own separate PR as to not "bloat" this PR with changes "outside of the dependency changes being made". That being said, it's easy for me to say that as the person not making the changes (:smile:) so I'd make sure we could timebox the effort to make the changes, and if it's not anything too crazy, we could open that PR soon, get it merged, and incorporate the new tag structure in this PR.

Any thoughts?

pauld-msft: I'm on board with the single cli repo approach of `oryx/cli:debian-{flavor}-{tag}` and agree that it would be good to have a separate PR that merges into this one. I agree that it shouldn't be a _ton_ of work, as the areas that need to be changed are mostly highlighted in this pr AFAIK. 

It would be a step in the right direction towards consistency, and I think it would be good to do here instead of waiting to do the slightly larger conversion to `oryx/build:cli-debian-{flavor}-{tag}` at some point in the future, since I don't really see us prioritizing that work soon.

snehapar9: +1. I'm onboard with just using the single cli repo approach for all the debian flavors as well. Would be easier to make this change along with this PR now than re-doing this at some point in the future and having to undo changes in Oryx.yml file.

I do not expect other files outside of the files included in this PR to change.

build/buildBuildImages.sh
cormacpayne: This comment still applies -- I think it makes sense to structure this chain of conditions as follows:

```bash
elif [ "$imageTypeToBuild" == "cli"]; then
    buildCliImage
    buildCliImage "buster"
    buildCliImage "bullseye"
elif [ "$imageTypeToBuild" == "cli-stretch"]; then
    buildCliImage
elif [ "$imageTypeToBuild" == "cli-buster"]; then
    buildCliImage "buster"
elif [ "$imageTypeToBuild" == "cli-bullseye"]; then
    buildCliImage "bullseye"
```

This gives us the ability to build all three flavors of CLI images with just the `cli` option, while also allowing us to build any of the individual flavors with a different option.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
cormacpayne, line 3: Just calling out that we'd need to change this repo if we decide to go with the `oryx/cli:debian-{flavor}-{tag}` format before merging this PR.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -11,38 +11,78 @@ ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
...
+        libargon2-0 \
+        libonig-dev \
+        libedit-dev \
+    && rm -rf /var/lib/apt/lists/* \
+    # This is the folder containing 'links' to benv and build script generator
+    && mkdir -p /opt/oryx
+ARG IMAGES_DIR="/opt/tmp/images"
+ARG BUILD_DIR="/opt/tmp/build"
 
+
```
cormacpayne, line 52: nit: would you mind removing this extra newline?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
cormacpayne: @daniv-msft -- Sneha and I synced about this and it would be do-able if we migrated away from the `defaultVersion.txt` files that we upload to storage and used constants generated via `constants.yaml`, which would allow us to easily reference these values within the Dockerfiles (like we're doing on line 57 above with the `__nodeVersion.sh` file.

However, in my opinion, since we only need to update the default version used for a given platform every 1-2 years, the amount of work needed to migrate to this new model wouldn't necessarily take precedence over the other things we want to take care of in the short term. I think filing a work item for someone in the future to take a look at when they have time would be a nice-to-have.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -55,4 +95,6 @@ RUN set -ex \
     && ln -s $PYTHON38_VERSION latest \
     && ln -s $PYTHON38_VERSION stable \
     && echo "jamstack" > /opt/oryx/.imagetype \
-    && echo "DEBIAN|${DEBIAN_FLAVOR}" | tr '[a-z]' '[A-Z]' > /opt/oryx/.ostype
\ No newline at end of file
+    && echo "DEBIAN|${DEBIAN_FLAVOR}" | tr '[a-z]' '[A-Z]' > /opt/oryx/.ostype
+
+
```
cormacpayne, line 100: nit: same comment here about removing the extra newline

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
snehapar9: From [here](https://github.com/microsoft/Oryx/blob/main/build/buildBuildImages.sh#L240), the value of PARENT_DEBIAN_FLAVOR is github-actions-debian-bullseye for the azfunc-jamstack-debian-bullseye image.

cormacpayne (dismissed):
Some minor comments, otherwise LGTM whenever we make the `oryx/cli` repository change :shipit:

build/buildBuildImages.sh
snehapar9: > Chatted with @snehapar9 about this offline -- I'm in the boat that we should move to the single repo approach of `oryx/cli:debian-{flavor}-{tag}` since (1) we already have the `oryx/cli` repository registered with MCR and (2) we already follow that pattern with our `oryx/build` images and would make sense to be consistent here.
> 
> I don't think this would be a large amount of work, and would probably warrant being done in its own separate PR as to not "bloat" this PR with changes "outside of the dependency changes being made". That being said, it's easy for me to say that as the person not making the changes (😄) so I'd make sure we could timebox the effort to make the changes, and if it's not anything too crazy, we could open that PR soon, get it merged, and incorporate the new tag structure in this PR.
> 
> Any thoughts?
It would actually be easier adding these cli repo changes in the current PR. Opening a new PR adds other redundant changes and conflicts.

tests/Oryx.Tests.Common/ImageTestHelper.cs
```diff
@@ -517,8 +527,10 @@ public static class ImageTestHelperConstants
         public const string PackRepository = "pack";
         public const string CliRepository = "cli";
         public const string CliBusterRepository = "cli-buster";
+        public const string CliBullseyeRepository = "cli-bullseye";
```
snehapar9, line 530: CliBusterRepository and CliBullseyeRepository are referenced in multiple cli tests to get the appropriate cli image name using the GetCliImage method like in [this](https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/Python/PythonDynamicInstallationTest.cs#L59)example.

However, since we would only have the cli repository now, it makes sense to remove CliBusterRepository and CliBullseyeRepository from here but we would have to make changes to all the tests calling this method(has 80+ references). 
If we decide to remove these constants, we would have to modify this method to check for the tag instead the repository which would lead to modifying all those 80+ references. 
So, I would think it would be okay to leave these as-is for the sake of simplicity. Any thoughts?

build/buildBuildImages.sh
snehapar9: Updated buildBuildImages.sh script to incorporate this change.

tests/Oryx.Tests.Common/ImageTestHelper.cs
pauld-msft: @snehapar9 Its quite easy to rename a variable and all references in visual studio by selecting something like `CliBullseyeRepository` and hitting `ctrl + R + R`. I'd probably handle this for bullseye (and other flavors) by:
- deleting current `CliBullseyeTag` 
- `ctrl + R + R` renaming `CliBullseyeRepository` to `CliBullseyeTag`
- replacing the value `cli-bullseye` with `debian-bullseye`
- remove the `_cliBullseyeRepository` private var in this file
- fix the `GetCliImage` and `GetBuildImage` to use the cli bullseye tag

All in all, it will bloat this pr a bit for us reviewers now, but it shouldn't take too long and will make more sense for us in the future IMO

snehapar9: By replacing cli-bullseye by debian-bullseye, you mean the arg we pass to the buildBuildImages.sh script like [here](https://github.com/microsoft/Oryx/blob/snehapar/Install-deps-dynamically/build/buildBuildImages.sh#L546)?
And you probably mean [these](https://github.com/microsoft/Oryx/blob/snehapar/Install-deps-dynamically/tests/Oryx.BuildImage.Tests/Node/NodeDynamicInstallationTest.cs#L101) traits tied to [these](https://github.com/microsoft/Oryx/blob/snehapar/Install-deps-dynamically/vsts/pipelines/validation.yml#L38) as well?

pauld-msft: > By replacing cli-bullseye by debian-bullseye, you mean the arg we pass to the buildBuildImages.sh script like [here](https://github.com/microsoft/Oryx/blob/snehapar/Install-deps-dynamically/build/buildBuildImages.sh#L546)?

@snehapar9 sorry no, was only referring to the the value of `CliBullseyeRepository`, which is currently `cli-bullseye`. 

> And you probably mean [these](https://github.com/microsoft/Oryx/blob/snehapar/Install-deps-dynamically/tests/Oryx.BuildImage.Tests/Node/NodeDynamicInstallationTest.cs#L101) traits tied to [these](https://github.com/microsoft/Oryx/blob/snehapar/Install-deps-dynamically/vsts/pipelines/validation.yml#L38) as well?

I'm not sure that those even need to be switched, as they don't refer to the CLI repository. They are just categories of tests, and `cli-bullseye` as a testing category still makes sense, regardless of which repo the image ends up in.

snehapar9: Okay. That makes sense.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
daniv-msft: Thank you both!

src/BuildScriptGenerator/Hugo/HugoPlatformInstaller.cs
```diff
@@ -32,6 +32,10 @@ public virtual string GetInstallerScriptSnippet(string version)
...
+                .AppendLine("if grep -q jamstack \"/opt/oryx/.imagetype\"; then")
```
daniv-msft, line 36: Nit: Please consider adding a comment explaining the goal of this. Minor comment, feel free to ignore if you have no other changes to push

daniv-msft (dismissed):
Looks good to me!

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -11,32 +11,29 @@ ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
...
-    && rm -rf $downloadedFileName
-
+    LC_ALL="C.UTF-8" \
+    ORYX_PATHS="/opt/oryx:/opt/nodejs/lts/bin:/opt/python/latest/bin:/opt/yarn/stable/bin"
+    
 RUN set -ex \
     # Install Python SDKs
     # Upgrade system python
     && PYTHONIOENCODING="UTF-8" \
+    # It's not clear whether these are needed at runtime...
```
daniv-msft, line 21: Could you please add more info about this? From what I've seen in the past I needed to update apt-get before using it for it to be able to detect and pull the latest packages. However it also came with a significant cost, as it seems that updating it pulls A LOT of data. If you have any more info, it could be valuable so that we update that code later or something.

src/BuildScriptGenerator/Hugo/HugoPlatformInstaller.cs
```diff
@@ -52,6 +58,7 @@ public virtual string GetInstallerScriptSnippet(string version)
 
                 // Write out a sentinel file to indicate downlaod and extraction was successful
                 .AppendLine($"echo > {Path.Combine(versionDirInTemp, SdkStorageConstants.SdkDownloadSentinelFileName)}");
+            InstallGolangToolingAndLanguage(snippet);
```
daniv-msft, line 61: I hope I'm not confused: please feel free to resolve if it is the case. Should we only run `InstallGolangToolingAndLanguage` on the same condition that we `installHugo.sh` above? Are there some cases where we want one and not the other?

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
snehapar9: Seems like this [this](https://github.com/microsoft/Oryx/blob/180b2a86eaa248433d7b7f3defbdf112026356d3/images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile#L44) comment has been around for a while. It is not clear why it was added from the PR.

src/BuildScriptGenerator/Hugo/HugoPlatformInstaller.cs
snehapar9: Just to be clear, in this block, we run installHugo.sh only if the image type is jamstack because installHugo.sh is run in NodePlatform.cs for cli images. However, we install golang irrespective of the image type.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
daniv-msft: OK thanks. One of the things to investigate separately, but I agree it doesn't need to be done as part of this PR.

daniv-msft (approved):

PR #1754, opened 2022-12-19
Pauld/1694164 add image type logs

pauld-msft:
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1694164

Update so we are logging the OS type and the Image type, pulled from the `.ostype` and `.imagetype` files, both to the console, and to application insights.

### DIFF
--- a/build/constants.yaml
+++ b/build/constants.yaml
@@ -244,6 +244,7 @@
     compressed-output-file-name: oryx_output.tar.gz
     build-manifest-file-name: oryx-manifest.toml
     os-type-file-name: .ostype
+    image-type-file-name: .imagetype
     build-commands-file-name: oryx-build-commands.txt
     run-script-generator-dir: /opt/startupcmdgen
     node-app-insights-loader-file-name: applicationinsights/out/Bootstrap/Oryx.js
--- a/images/build/Dockerfiles/gitHubActions.Dockerfile
+++ b/images/build/Dockerfiles/gitHubActions.Dockerfile
@@ -107,8 +107,7 @@ RUN ${IMAGES_DIR}/retry.sh "curl -o /usr/local/share/ca-certificates/verisign.cr
 
 # Install PHP pre-reqs	# Install PHP pre-reqs
 RUN if [ "${DEBIAN_FLAVOR}" = "buster" ] || [ "${DEBIAN_FLAVOR}" = "bullseye" ]; then \
-    set -ex \
-    && apt-get update \
+    apt-get update \
     && apt-get upgrade -y \
     && apt-get install -y \
         $PHPIZE_DEPS \
--- a/images/build/Dockerfiles/gitHubActions.Dockerfile
+++ b/images/build/Dockerfiles/gitHubActions.Dockerfile
@@ -44,7 +44,7 @@ RUN if [ "${DEBIAN_FLAVOR}" = "bullseye" ]; then \
             libcurl4 \
             libssl1.1 \
             libyaml-dev \
-            libxml2-dev \
+            libxml2 \
         && rm -rf /var/lib/apt/lists/* ; \
     elif [ "${DEBIAN_FLAVOR}" = "buster" ]; then \
         apt-get update \
--- a/images/build/Dockerfiles/gitHubActions.Dockerfile
+++ b/images/build/Dockerfiles/gitHubActions.Dockerfile
@@ -44,10 +44,8 @@ RUN if [ "${DEBIAN_FLAVOR}" = "bullseye" ]; then \
             libcurl4 \
             libssl1.1 \
             libyaml-dev \
-        && rm -rf /var/lib/apt/lists/* \
-        && curl -LO http://security.debian.org/debian-security/pool/updates/main/libx/libxml2/libxml2_2.9.10+dfsg-6.7+deb11u3_amd64.deb \
-        && dpkg -i libxml2_2.9.10+dfsg-6.7+deb11u3_amd64.deb \
-        && rm libxml2_2.9.10+dfsg-6.7+deb11u3_amd64.deb ; \
+            libxml2-dev \
+        && rm -rf /var/lib/apt/lists/* ; \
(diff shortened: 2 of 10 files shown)


images/build/Dockerfiles/gitHubActions.Dockerfile
```diff
@@ -44,10 +44,8 @@ RUN if [ "${DEBIAN_FLAVOR}" = "bullseye" ]; then \
             libcurl4 \
             libssl1.1 \
             libyaml-dev \
-        && rm -rf /var/lib/apt/lists/* \
-        && curl -LO http://security.debian.org/debian-security/pool/updates/main/libx/libxml2/libxml2_2.9.10+dfsg-6.7+deb11u3_amd64.deb \
```
pauld-msft, line 48: not sure why we are pulling from this manual source to install. I've updated to `libxml2-dev` which we use elsewhere in our codebase. The benefit of this is that we no longer need to manually bump these 3 lines when they release a new libxml2 version

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -156,6 +158,16 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
...
             }
 
+            if (!string.IsNullOrWhiteSpace(options.DebianFlavor))
+            {
+                buildInfo.AddDefinition("Os Type", options.DebianFlavor);
+            }
+
+            if (!string.IsNullOrWhiteSpace(options.ImageType))
+            {
+                buildInfo.AddDefinition("Image Type", options.ImageType);
```
pauld-msft, line 168: These will be logged in the build output on every run, similar to repository commit

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -209,6 +221,8 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
             {
                 { "scriptPath", buildScriptPath },
                 { "envVars", string.Join(",", GetEnvVarNames(environment)) },
+                { "osType", options.DebianFlavor },
+                { "imageType", options.ImageType },
```
pauld-msft, line 225: We log these props in the `RunBuildScript` event

images/build/Dockerfiles/gitHubActions.Dockerfile
daniv-msft: I looked into the original PR that introduced this (https://github.com/microsoft/Oryx/pull/1483) and couldn't find the reason. Maybe something related to Bullseye? Might be worth double checking with @cormacpayne in case he remembers this PR and asked Stella about that maybe?
I can see that libxml2 as installed before targeted 2.9.10, when the one you changed it to is 2.9.4 I believe. Maybe that's the reason? Also, do you know why we use libxml2-dev and not libxml2?

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -156,6 +158,16 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
                 buildInfo.AddDefinition("Repository Commit", sourceRepoCommitId);
             }
 
+            if (!string.IsNullOrWhiteSpace(options.DebianFlavor))
+            {
+                buildInfo.AddDefinition("Os Type", options.DebianFlavor);
```
daniv-msft, line 163: Nit: please consider replacing by "OS"? Minor comment, feel free to ignore

daniv-msft (approved):
Looks good, thanks

images/build/Dockerfiles/gitHubActions.Dockerfile
pauld-msft: I also looked at that PR and didn't see any indications, but potentially a PHP dep? Yeah @cormacpayne if you remember anything once you are back, let us know!

`when the one you changed it to is 2.9.4 I believe`
Curious where you are seeing this? I thought that would install 2.9.10: https://packages.debian.org/stable/libxml2-dev

As for `libxml2-dev` vs `libxml2`, I installed it as it is generally used a few different places in our code. That said, its 2MB larger than `libxml2` so I'm going to try to swap them and see if we can get through our nightly testing suite.

daniv-msft: Thanks, sounds good. Regarding 2.9.10 vs. 2.9.4: you're right, 2.9.10 will indeed be installed with this.

build/constants.yaml
```diff
@@ -244,6 +244,7 @@
     compressed-output-file-name: oryx_output.tar.gz
     build-manifest-file-name: oryx-manifest.toml
     os-type-file-name: .ostype
+    image-type-file-name: .imagetype
```
william-msft, line 247: @pauld-msft just double checking if we add the `ostype` and `imagetype` in the `oryx-manifest.toml` file? If not, this could be beneficial for our partners. The toml file stores build information meta data to be used between build and runtime images.

william-msft (approved):

build/constants.yaml
pauld-msft: As far as I know, we do not add those to the oryx-manifest.toml. However, we place the ostype and imagetype files in the exact same directory as the manifest file for our partners to consume and they are aware of these files.

images/build/Dockerfiles/gitHubActions.Dockerfile
cormacpayne: @pauld-msft @daniv-msft -- unfortunately I can't find any additional context on this, but it looks like it was a temporary holdover while we couldn't publish bullseye binaries, so I'm assuming it was a library that was bundled with the buster binaries that was available in some layer that wasn't available in the bullseye flavor of the GitHub Actions build image. We should be good now since all of this has been resolved with bullseye binaries 👍

daniv-msft: Thanks @cormacpayne!

PR #1759, opened 2022-12-20
Run npm install as current user

pauld-msft:
See https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1688940 for full details.

Summarizing the ticket, we were seeing permission errors when we ran npm install once we upgraded to node 16. The new version of npm used for node 16 (vs 14) has behavior that causes it to execute npm install commands as whoever owns the directory where it is executing. In ADO pipelines this causes an issue because a non-root user owns the application directory, but the npm cache is located in `/root/.npm/`. So the non-root user that npm install is running as does not have permissions to write to the cache.

This solution should make the application directory be owned by whoever is executing the Oryx script. So in cases where that is root, the root user will then execute npm install. But if a customer is running as a non-root user, we won't be changing the owner to root.
Tested with github actions and ADO pipeline in https://github.com/pauld-msft/create-react-app-sample-testing-root

### DIFF
--- a/src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
@@ -129,6 +129,10 @@ then
 	fi
 fi
 
+# ensure that the current user owns the application source directory,
+# so we execute npm install with the correct permissions.
+chown -R "$(whoami)" $SOURCE_DIR
+
 cd "$SOURCE_DIR"
 
 {{ if CustomBuildCommand | IsNotBlank }}
--- a/src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Node/NodeBashBuildSnippet.sh.tpl
@@ -129,9 +129,13 @@ then
 	fi
 fi
 
-# ensure that the current user owns the application source directory,
-# so we execute npm install with the correct permissions.
-chown -R "$(whoami)" $SOURCE_DIR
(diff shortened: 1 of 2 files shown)


iarna:
Link to the original issue: [Azure/static-web-apps#909](https://github.com/Azure/static-web-apps/issues/909)

Paul and I talked about this weird error and here's what we covered:

1. The install happens as root
2. The error occurs in a run-script triggered by the install
3. We pass --unsafe-perms to npm
4. This only happens in npm >=7 on ADO. Github actions doesn't have this error, nor does npm6.
5. The user on ADO is root, but the source dir is not owned by root.
6. npm, when run as root, will, when running scripts, change it's effective user (EUID) to match the owner of the directory it is running in. We pass --unsafe-perm to disable this behavior.
7. The project being installed has a run-script that calls npm.
8. That run-script fails with EACCESS errors from trying to access the npm cache.

The root cause (no pun intended) for this is:

npm 7 dropped support for --unsafe-perm and now ALWAYS switches to the directory owner before running scripts. There is no longer any facility for changing this behavior. As such, the only solution is to either not run as root, ensure the directory we run from is owned by root (to disable the EUID change) or ensure our cachedir is writable by the owner of the source dir.

iarna (dismissed):
As weird as this is, this is probably the least intrusive solution available to us. (In particular, this allows fully custom build commands that call npm to work.) Or, arguably, not run as root, but I suspect some custom build commands install os level stuff and need root.

I would also remove use of --unsafe-perm from our calls to npm so that it behaves consistently across versions.

william-msft (approved):

snehapar9 (approved):

PR #1766, opened 2022-12-22
Williamhe/1684436 add dotnet runtime automation

william-msft:
- [ ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as
  - Added Runtime functionality for dotnet. Update dotnet SDK and Runtime is autonomous for minor and patch versions
  - Update workflows to be re-usable, using matrix, so it's easier to integrate with other platforms
  - Added retries for checking triggered pipeline, for SDK and BaseImages
  - Example of test invocation with new versions of dotnet: https://github.com/microsoft/Oryx/pull/1761/files
    - Base Image Pipeline invocation: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7114418&view=results
    - SDK pipeline invocation: https://devdiv.visualstudio.com/DevDiv/_build?definitionId=13049
  
https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1641601


build/tools/Automation/checkPipelineStatus.sh
```diff
@@ -0,0 +1,40 @@
+﻿# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+# --------------------------------------------------------------------------------------------
+# This script will check the pipeline status' result
+# for a given a pipeline invocation id.
```
daniv-msft, line 6: Nit: very minor typo: `for a given a pipeline invocation id.` => `for a given pipeline invocation id.`

build/tools/Automation/checkPipelineStatus.sh
```diff
@@ -0,0 +1,40 @@
...
+do
+	echo "retry $retryCount"
+	# store response in status.json for later use
+	az pipelines runs show --id ${pipelineInvocationId} --organization https://devdiv.visualstudio.com/ --project DevDiv > /tmp/status.json
+	result=$( cat /tmp/status.json | jq ".result" | tr -d '"' )
+	echo "pipeline ${pipelineInvocationId} invocation result: $result"
+	if [[ "$result" == "succeeded" ]]; then
+		exit 0
+	fi
+	echo "retrying in $timeoutSeconds seconds..."
```
daniv-msft, line 34: Nit: from the usage, the $timeoutSeconds variable feels more like a delay between retries than an actual timeout where we would stop an operation after some time. Please consider renaming to `retryDelaySeconds` or something of that kind.

build/tools/Automation/checkPipelineStatus.sh
```diff
@@ -0,0 +1,40 @@
...
+
+retryCount=0
+while [ "$retryCount" -le "$maxRetries" ]
+do
+	echo "retry $retryCount"
+	# store response in status.json for later use
+	az pipelines runs show --id ${pipelineInvocationId} --organization https://devdiv.visualstudio.com/ --project DevDiv > /tmp/status.json
+	result=$( cat /tmp/status.json | jq ".result" | tr -d '"' )
+	echo "pipeline ${pipelineInvocationId} invocation result: $result"
+	if [[ "$result" == "succeeded" ]]; then
```
daniv-msft, line 31: Should we have a similar case if we detect that the result corresponds to cases where we know it has failed, and that retrying will never help? In that case, going through another couple of set of retries wouldn't be useful.

build/tools/Automation/checkPipelineStatus.sh
```diff
@@ -0,0 +1,40 @@
...
+# The 'buildNumber' field of status.json is used later in the workflow
+# to update constants.yaml runtime image tag.
+
+pipelineInvocationId=$1
+maxRetries=3
+if [[ -n "${MAX_RETRIES}" ]]; then
+  maxRetries=${MAX_RETRIES}
+fi
+
+timeoutSeconds=10
```
daniv-msft, line 18: I'm just curious: here we're setting our defaults to take 40 seconds (10 seconds per try, 1 initial try + 3 retries). Is that realistic? Looking at the average time to run the pipeline, is it in this range? I would have expected it to be taking a lot more time than that, but it's good if it's closer to 30 seconds.

build/tools/Automation/checkPipelineStatus.sh
```diff
@@ -0,0 +1,40 @@
...
+retryCount=0
+while [ "$retryCount" -le "$maxRetries" ]
+do
+	echo "retry $retryCount"
+	# store response in status.json for later use
+	az pipelines runs show --id ${pipelineInvocationId} --organization https://devdiv.visualstudio.com/ --project DevDiv > /tmp/status.json
+	result=$( cat /tmp/status.json | jq ".result" | tr -d '"' )
+	echo "pipeline ${pipelineInvocationId} invocation result: $result"
+	if [[ "$result" == "succeeded" ]]; then
+		exit 0
```
daniv-msft, line 32: Nit: please consider adding a `success` echo or something. Not a big deal, but when investigating issues it feels good to see that we were successful for sure, as the exit code might not be as visible (not sure if that's the case here).

.github/workflows/automationTemplate.yaml
```diff
@@ -0,0 +1,146 @@
...
+        required: true
+        type: string
+
+  workflow_dispatch:
+    inputs:
+      releaseDate:
+        description: 'The date to check for new releases. Format: yyyy-mm-dd'
+        required: false
+        type: string
+      platformName:
```
daniv-msft, line 22: Could you please clarify why we need these two names (platformName and platformNameCamelCase) to be different? Could we generate one from the other (for example just making platformNameCamelCase to lower case to create platformName)?
If we actually need these names to be very specific (for example because we'll go look for a URL with these names or something), please consider adding that to the comments to make it clearer.

.github/workflows/automationTemplate.yaml
```diff
@@ -0,0 +1,146 @@
...
+          if [ -z "${changes}" ]; then
+              echo "No git changes to commit. Cancelling workflow..."
+              exit 1
+          fi
+          git status
+
+      - name: Commit new ${{ inputs.platformNameCamelCase }} SDK versions
+        run: |
+          git status
+          git config user.name 'William Hernandez'
```
daniv-msft, line 72: Could you please clarify: do we need to set a user for this to work?
If yes, do we need to set a valid user, or can we set something more generic rather than a personal account?

build/tools/Automation/checkPipelineStatus.sh
william-msft: I've updated the default time. Thanks for catching this!

build/tools/Automation/checkPipelineStatus.sh
william-msft: Updated

build/tools/Automation/checkPipelineStatus.sh
william-msft: [This line](https://github.com/microsoft/Oryx/pull/1766/files#diff-0e2923dd4411ce28feba9070d3da26c568fbe188982c47f9eddc59732c644cb5R30) `echo`s the status every iteration before an exit is performed. Which means `result` would have been printed. Is this what you were referring to?

build/tools/Automation/checkPipelineStatus.sh
william-msft: Updated

.github/workflows/automationTemplate.yaml
william-msft: We need a camel case name since for example: We need to map `dotnet` to `DotNetCore` since our Pipelines have camel casing. However, I was able to remove the extra input by [adding this as an environment variable](https://github.com/microsoft/Oryx/pull/1766/commits/0a5663ef08be8fd331346efb40914d734f71657d).

.github/workflows/automationTemplate.yaml
william-msft: Double checking...

william-msft: A user needs to be specified. But I've updated to use Oryx Team instead.

build/tools/Automation/checkPipelineStatus.sh
daniv-msft: Thanks!

build/tools/Automation/checkPipelineStatus.sh
daniv-msft: Yes OK, looks good

.github/workflows/automationTemplate.yaml
```diff
@@ -51,10 +44,16 @@ jobs:
             releaseDate=`date +%F`
             echo "Defaulting to todays date ${releaseDate}"
           fi
+          echo "RELEASE_DATE=${releaseDate}" >> $GITHUB_ENV
 
-          echo "releaseDate=${releaseDate}" >> $GITHUB_ENV
+          declare -A platformsCamelCase
+          echo "platformName: ${{ inputs.platformName }}"
+          echo "platformNameCamelCase: ${platformNamesCamelCase[${{ inputs.platformName }}]}"
```
daniv-msft, line 51: Could you please clarify what this line will echo? Won't `platformNamesCamelCase` be undefined at this point? Or can this variable also be declared somewhere else?
Also, minor comment if we keep the log, there is a small typo: `platformNameCamelCase` => `platformNamesCamelCase`

.github/workflows/automationTemplate.yaml
```diff
@@ -51,10 +44,16 @@ jobs:
...
+          platformNamesCamelCase=( ["dotnet"]="DotNetCore" )
```
daniv-msft, line 52: I'm wondering if we should make the platform name even more descriptive. Here, I don't believe that anything indicates to us that the `DotNetCore` name must actually correspond to what we use in the base pipeline, and so someone could expect `dotnet` to be `DotNet`.
Please consider renaming to something like `platformNamesAsInBasePipeline` or something like that.

.github/workflows/automationTemplate.yaml
william-msft: Updated

.github/workflows/automationTemplate.yaml
william-msft: Agreed & updated.

daniv-msft (approved):

PR #1777, opened 2023-01-03
Make detection requirements clearer

pamelafox:

To me, the current wording makes it sound as if *all* of the detection requirements must be met. I reworded it to clarify "any" of the detection requirements must be met. 

Also changed to unordered list instead of ordered list but perhaps it should be ordered if thats the detection sequence?

### DIFF
--- a/doc/runtimes/python.md
+++ b/doc/runtimes/python.md
@@ -41,19 +41,19 @@ gunicorn (a Python package) is also included.
 
 # Detect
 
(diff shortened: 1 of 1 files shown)


pauld-msft:
Apologies, accidentally kicked off wrong job which I cancelled. Kicked off correct one now, which will allow merging once it passes. Thanks for submitting

cormacpayne (approved):
:shipit:

PR #1786, opened 2023-01-04
Williamhe/1578887 update workflow exiting

william-msft:
  - Cancel SDK & Runtime automation when no changes are detected.
  - Fix bug that opens PRs that contain spacing changes: https://github.com/microsoft/Oryx/pull/1787/files


.github/workflows/automationTemplate.yaml
```diff
@@ -53,28 +53,18 @@ jobs:
...
-          git status
+          set -x
           git config user.name 'Oryx Team'
           git config user.email 'oryx@microsoft.com'
+          
           git add -A
-          git commit -m "Add new SDK versions"
+          git commit -m "Add new SDK versions" | true
+          gitCommitExitCode=${PIPESTATUS[0]}
+          [[ $gitCommitExitCode = 0 ]] && cancelWorkflow=false || cancelWorkflow=true
```
william-msft, line 65: checking the `git commit` exit status is replacing the following check: https://github.com/microsoft/Oryx/pull/1786/files#diff-3ed663dcfb84882e18b67fce2ab60dcb67f0b57ac69932ca286dcea41b5d80a5L56-L65

.github/workflows/automationTemplate.yaml
```diff
@@ -85,46 +75,85 @@ jobs:
       
       - name: Install jq
         run: |
+          if [ ${{ env.CANCEL_WORKFLOW }} = true ]; then
```
william-msft, line 78: Tried using github [expressions ](https://docs.github.com/en/actions/learn-github-actions/expressions) but was unable to when referencing the environment variable: 
https://github.com/microsoft/Oryx/pull/1786/commits/e27d8f55d64176059d5fd5e789d103faff58e837#diff-3ed663dcfb84882e18b67fce2ab60dcb67f0b57ac69932ca286dcea41b5d80a5R83

william-msft: Tried using github [expressions ](https://docs.github.com/en/actions/learn-github-actions/expressions)but was unable to when referencing the environment variable:
https://github.com/microsoft/Oryx/commit/e27d8f55d64176059d5fd5e789d103faff58e837#diff-3ed663dcfb84882e18b67fce2ab60dcb67f0b57ac69932ca286dcea41b5d80a5R83

A startup failure would occur:
image
https://github.com/microsoft/Oryx/actions/runs/3840383559

.github/workflows/automationTemplate.yaml
daniv-msft, line 65: Hope I'm not misunderstanding the code: should this line be an `if / else` block? It feels a bit confusing to make this in a one liner that looks like a condition.

william-msft: This one is a bash ternary operator but I can update to be more explicit.

william-msft: Updated 😃

.github/workflows/automationTemplate.yaml
```diff
@@ -85,46 +75,85 @@ jobs:
...
+            echo "No new ${{ inputs.platformName }} SDK versions detected. Canceling workflow..."
```
daniv-msft, line 85: Nit: please consider moving this echo to the previous step where we actually detected that no SDK versions existed.
Else people might not expect to find this information in the "Install JQ" step. Also, if for any reasons in the future we add new reasons for the flow to be canceled, then this message would become misleading.

.github/workflows/automationTemplate.yaml
daniv-msft: OK my bad, I didn't know about this structure in bash. Thanks for updating!

.github/workflows/automationTemplate.yaml
william-msft: Makes sense! I've updated too.

daniv-msft: Thanks!

daniv-msft (dismissed):
Looks good to me!

.github/workflows/automationTemplate.yaml
cormacpayne: I was able to do this in a few places for the Container Apps GitHub Action -- [example](https://github.com/Azure/container-apps-deploy-action/blob/main/action.yml#L227)

Was the following not working for you?

```
if: ${{ env.CANCEL_WORKFLOW != true }}
```

william-msft: No, unfortunately the if statement is [ignored completely](https://github.com/microsoft/Oryx/pull/1786/commits/61af04ddf2e0d12af05968ebaa815424bbb3645b#diff-3ed663dcfb84882e18b67fce2ab60dcb67f0b57ac69932ca286dcea41b5d80a5R83). But I'm going to try similar to what you're doing in your example:
`if: ${{ env.CANCEL_WORKFLOW != '' }}`

build/constants.yaml
```diff
@@ -26,14 +26,14 @@
...
   outputs:
-    - type: csharp
-      directory: src/BuildScriptGenerator.Common
-      namespace: Microsoft.Oryx.BuildScriptGenerator.Common
-    - type: shell
-      directory: build
-      file-name-prefix: __
-    - type: go
-      directory: src/startupscriptgenerator/src/common/consts
+  - type: csharp
```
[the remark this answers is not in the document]
william-msft: Resetting to what the `YamlDotNet` module indentation (shown in green) applies

.github/workflows/automationTemplate.yaml
```diff
@@ -53,77 +53,85 @@ jobs:
...
           buildNumber=$(cat /tmp/status.json | jq ".buildNumber" | tr -d '"')
-          echo "buildNumber: $buildNumber"
           platform=${{ inputs.platformName }}
           [[ "$platform" == "dotnet" ]] && platform="dot-net-core" || platform=$platform
-          echo "platform: $platform"
           yq -i "(.[] | select(.name==\"$platform-run-time-versions\") | .constants.$platform-runtime-base-tag) = $buildNumber" build/constants.yaml
+          
+          # run automation again to preserve constants.yaml YamlDotNet indentation
+          # since YamlDoNet and yq do not share indentation spacing
+          dotnet run --project build/tools/Automation/Automation.csproj ${RELEASE_DATE} ${{ github.workspace }}
```
william-msft, line 129: This will prevent spacing changes PRs for example: https://github.com/microsoft/Oryx/pull/1787/files

.github/workflows/automationTemplate.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
           git add -A
           git commit -m "Add new SDK versions" | true
           gitCommitExitCode=${PIPESTATUS[0]}
-          if [[ $gitCommitExitCode = 0 ]]; then
+          if [[ $gitCommitExitCode != 0 ]]; then
```
daniv-msft, line 65: Should this condition be inverted? I expected the gitCommitExitCode to be >= 1 if there was nothing to commit, in which case we would want to cancel the workflow.

daniv-msft: Looking at the rest of the PR, I'm pretty sure this was done for testing purposed. :)

.github/workflows/automationTemplate.yaml
daniv-msft: I see that yq has an --indent parameter to control the number of spaces used for indentation. Would this be an option here?
https://mikefarah.gitbook.io/yq/usage/output-format

.github/workflows/automationTemplate.yaml
william-msft: AH yes! Excellent catch :)

.github/workflows/automationTemplate.yaml
william-msft: I've tried `--indent 0` however the minimum seems to be `--indent 2` which is the default. Unfortunately, YamlDotNet does not have indentation support. So this would be our simplest workaround.

.github/workflows/automationTemplate.yaml
william-msft: Updated

daniv-msft: Thanks

.github/workflows/automationTemplate.yaml
daniv-msft: OK makes sense. Thanks for considering this!

daniv-msft (approved):

PR #1791, opened 2023-01-05
Pauld/1685848 update dependencies

pauld-msft:
This pull request updates the Oryx dependencies to have them all fall under these rules (as defined in https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1685848):

"All open-source components used must have a release date no older than 18 months (for still-maintained components) and no older than 48 months (for components with no later version)."


tests/Oryx.Integration.Tests/Python/PythonDynamicInstallationTest.cs
```diff
@@ -46,33 +48,37 @@ public async Task CanBuildAndRunPython39AppAsync()
...
         [Fact]
         [Trait("category", "python-3.7")]
         [Trait("build-image", "github-actions-debian-bullseye")]
         public async Task CanBuildAndRunPython37App_UsingGitHubActionsBuildImage_AndDynamicRuntimeInstallationAsync()
         {
-            // Temporarily skip - Bug #1410367
-            // await CanBuildAndRunPythonApp_UsingGitHubActionsBuildImage_AndDynamicRuntimeInstallationAsync("PythonVersions.Python37Version", ImageTestHelperConstants.GitHubActionsBullseye);
+            // Temporarily skip - Bug #1266781
+            // await CanBuildAndRunPythonApp_UsingGitHubActionsBuildImage_AndDynamicRuntimeInstallationAsync(PythonVersions.Python37Version, ImageTestHelperConstants.GitHubActionsBullseye);
+            await Task.FromResult(true);
```
daniv-msft, line 61: Nit: would it be better to use Task.CompletedTask instead as this method is just expecting to return a Task? Not a big deal though, so feel free to ignore.

daniv-msft:
Looks good to me! Please let me know once you move out of Draft mode and I can sign off

tests/Oryx.Integration.Tests/Python/PythonDynamicInstallationTest.cs
pauld-msft: I added this because we were receiving warnings that we were not using an `await` in the async function. So just returning the Task.CompletedTask wouldn't satisfy that requirement

Directory.Build.props
```diff
@@ -13,19 +13,30 @@ See https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?v
   </PropertyGroup>
```
cormacpayne, line 13: General comment: were you able to find a nice way to list the top-level dependencies in our repository? I know we chatted about it previously and I thought there was a way to do it in Azure DevOps, but if you were able to find a way to do so, it would be great to note that somewhere for the person who does the SDL next on our team.

src/CommonFiles/Oryx.ruleset
```diff
@@ -123,7 +123,7 @@
 
   <Rules AnalyzerId="Microsoft.CodeAnalysis.CSharp.Features" RuleNamespace="Microsoft.CodeAnalysis.CSharp.Features">
     <!-- Requires that unused using directives are removed. -->
-    <Rule Id="IDE0005" Action="Error" />
+    <Rule Id="IDE0005" Action="Info" />
```
cormacpayne, line 126: nit: was this causing too many errors with the changes made in this PR?

cormacpayne (dismissed):
Minor comments, otherwise :shipit:

Directory.Build.props
pauld-msft: Unfortunately it seems they removed whatever let us do that in the past :( I needed to manually go through using the Nuget package manager and check the release date of each of the packages, and update any that were older than 18 months

src/CommonFiles/Oryx.ruleset
pauld-msft: ah good catch, I was originally trying to keep errors from being thrown about the unused directives in the test projects, which ended up being solved by making the assets private for the analyzers: https://github.com/microsoft/Oryx/pull/1791/files#diff-9da24614831c308827a1ae533ffea392c97638c261dd42bd0f5226baa136d16eR29-R31

so I'll revert this

tests/Oryx.Integration.Tests/Python/PythonDynamicInstallationTest.cs
daniv-msft: OK thanks for clarifying

daniv-msft (dismissed):

cormacpayne (approved):
:shipit:

Issue #1763, opened 2022-12-21
Custom startup for Azure web app deployment


bryanchen-d:
Hi, I want to deploy a python web app using zip deploy into an azure web service, and I have a custom startup.sh. As I can see from the deployment log, it calls oryx build and run using gunicorn, how should I configure it to run with my own startup script? Like, could it be specified in 'appsvc.yaml'?

Thanks.

cormacpayne:
@bryanchen-d Hey Bryan, apologies for the delayed response -- for context: your application source that you provide as a part of zip deployment will be built within one of Oryx's build images and the build artifacts will be sent to one of Oryx's run images that will host your application. In both of these images, Oryx provides mechanisms to allow the user to run additional commands when needed.

For the build stage, you can check out our [`configuration.md` file](https://github.com/microsoft/Oryx/blob/main/doc/configuration.md) to see the environment variables that you can set to run additional commands, such as `PRE_BUILD_COMMAND` or `PRE_BUILD_SCRIPT_PATH`.

For the run stage, you can use an [`appsvc.yaml` file ](https://github.com/microsoft/Oryx/blob/main/README.md#support-for-build-configuration-file) to define the set of commands that you want to run, including executing a `startup.sh` script that you have.

Please feel free to let us know if this doesn't handle your scenario or doesn't work for you and we can assist anyway we can. Thanks!

bryanchen-d:
> For the run stage, you can use an [`appsvc.yaml` file ](https://github.com/microsoft/Oryx/blob/main/README.md#support-for-build-configuration-file) to define the set of commands that you want to run, including executing a `startup.sh` script that you have.
> 
This link doesn't show how to specify a custom startup.sh script, for `appsvc.yaml` file, it seems to only support two settings, and they are all for the "build" stage:
```
version: 1

pre-build: | 
    apt install curl
  
post-build: | 
    python manage.py makemigrations 
    python manage.py migrate
```
Do I miss anything?

cormacpayne:
@bryanchen-d Sorry, it looks like our documentation around this file wasn't update when we added support for the `run` property -- you should be able to do the following in your `appsvc.yaml` file:

```yaml
version: 1

run: ./startup.sh
```

(you may need to include the full path to your `startup.sh` script in the run command)

bryanchen-d:
[...]

PR #1793, opened 2023-01-05
Update information for appsvc.yaml file

cormacpayne:
Updates the information in our `README.md` file around the `appsvc.yaml` file in accordance with the document produced previously by Jason Freeberg [here](https://github.com/JasonFreeberg/appsvc.yaml-examples).


daniv-msft:
Could you please share more context about this change? Did we receive an ask about this?
Also, if I remember correctly (cc @waliMSFT in case you remember more), we implemented some of the appsvc format, but not all of it. I might misremember though (the codebase would be more reliable than my memory :) )

daniv-msft:
> Could you please share more context about this change? Did we receive an ask about this? Also, if I remember correctly (cc @waliMSFT in case you remember more), we implemented some of the appsvc format, but not all of it. I might misremember though (the codebase would be more reliable than my memory :) )

OK, I found the context in https://github.com/microsoft/Oryx/issues/1763. I'm still not sure that we support all the appsvc properties.

cormacpayne:
@daniv-msft Sorry, forgot to add context pointing to the issue that was opened -- I thought we supported all four of those properties, but it looks like we don't support the `build` property (which I think is OK for the time being since we'd want to control the build command ran 99% of the time).

The `run` property was introduced by Wali as you mentioned above, and the `pre-build` and `post-build` properties have been supported in the [`DefaultBuildScriptGenerator.cs`](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs#L337-L356) for just over a year now by [this PR](https://github.com/microsoft/Oryx/pull/1174).

Let me update the documentation to remove the snippet about the `build` property being supported.

william-msft (approved):

william-msft:
I re-ran `Oryx-Validation` since there seemed to be transient errors.

daniv-msft (approved):

PR #1796, opened 2023-01-06
Upload defaultversion.txt for stretch as well

pauld-msft:
Currently, we only upload a `defaultVersions.stretch.txt` whenever we make updates to the defaultVersions files, but stretch actually uses `defaultVersions.txt` so that it maintains backwards compatibility. This makes it so that we upload both `defaultVersions.stretch.txt` and `defaultVersions.txt` for stretch, instead of only `defaultVersions.stretch.txt`


cormacpayne (approved):
:shipit:

william-msft (approved):

cormacpayne (approved):
:shipit:

Issue #1709, opened 2022-11-30
Ruby version detection fails


olbrich:
### Bug Report

- At what date and time did you most recently experience the problem?
2022-11-30 at 8:25am EST

- Where did you experience the problem? E.g. Azure Web Apps, Azure Functions, Azure Container Registry, or offline use.
Running Github action to deploy static site

- What happened?
Ruby version detection failed.
The ruby version in the Gemfile was specified as '~> 3.1' which resulted in a Gemfile.lock containing a ruby version of `ruby 3.1.1p18`.

Oryx attempted to run and returned with this error:
```
Detecting platforms...
Error: Platform 'ruby' version '3.1.1p18' is unsupported. Supported versions: 2.5.8, 2.6.6, 2.7.1, 2.7.2, 2.7.6, 3.0.0, 3.0.3, 3.1.1, 2.5.8, 2.6.6, 2.7.1, 2.7.2, 2.7.6, 3.0.0, 3.0.3, 3.1.1
```

Also, note that it is conventional to indicate which version of ruby is used in a project by including a `.ruby-version` file in the root of the project containing the ruby version or a string like `ruby-3.1.1`. This mechanism of specifying the version appears to be ignored by Oryx.

A reasonable approach to detecting ruby versions would be to:
1. Properly parse the `Gemfile.lock` entry
2. Fall back to the contents of `.ruby-version`
3. Also look for a `.tool-versions` file (from asdf)
4. Use the `Gemfile` specification last

- What did you expect or want to happen?
It should have detected `3.1.1` as the requested ruby version.

- How can we reproduce it?
- Create a ruby app and generate a gemfile specifying a ruby version of `~> 3.1`.
- `bundle update --ruby`
- attempt to deploy the site

olbrich:
So yes, it did attempt to detect the version from the Gemfile.lock that was generated on my local machine. The problem is that the version put there by bundler always has the patch version applied and Oryx does not appear to know how to interpret that part of the version string so you will never get a correct match. I expect that Oryx should have interpreted `3.1.1p18` to map to `3.1.1`.

pauld-msft:
[...]

PR #1801, opened 2023-01-09
Pauld/fix ruby gemfile lock version detection

pauld-msft:

See discussion at https://github.com/microsoft/Oryx/issues/1709. In short, ruby adds a patch version to the end of the gemfile.lock file version. Since `2.3.1p112` is equal to `2.3.1`, and Oryx only knows about version is `2.3.1`, we should resolve `2.3.1p112` as `2.3.1`. This also ensures that we can continue to detect rc and preview versions correctly.

### DIFF
--- a/src/Detector/Ruby/RubyDetector.cs
+++ b/src/Detector/Ruby/RubyDetector.cs
@@ -5,6 +5,7 @@
 
 using System;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Oryx.Common.Extensions;
@@ -200,7 +201,12 @@ namespace Microsoft.Oryx.Detector.Ruby
                     // Make sure it's in valid format.
                     if (rubyVersionLine.Length == 2)
                     {
-                        return rubyVersionLine[1];
+                        var fullVersion = rubyVersionLine[1];
+                        var parsedVersionMatches = Regex.Match(fullVersion, @"^(.*?)(?:p[0-9]+)*$");
+                        if (parsedVersionMatches.Success)
+                        {
+                            return parsedVersionMatches.Groups[1].Value;
+                        }
                     }
                 }
             }
--- a/src/Detector/Ruby/RubyDetector.cs
+++ b/src/Detector/Ruby/RubyDetector.cs
@@ -202,6 +202,11 @@ namespace Microsoft.Oryx.Detector.Ruby
                     if (rubyVersionLine.Length == 2)
                     {
                         var fullVersion = rubyVersionLine[1];
+
+                        // Parse the ruby version to remove patch versioning from it.
+                        // At times, ruby will add a patch version such as p112 to the end of the semver version,
+                        // which causes Oryx to not find the correct version. This should take a version like
+                        // 2.3.1p112 and convert it to 2.3.1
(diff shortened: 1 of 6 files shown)


src/Detector/Ruby/RubyDetector.cs
```diff
@@ -200,7 +201,17 @@ private string GetVersionFromGemFileLock(DetectorContext context)
...
+                        var fullVersion = rubyVersionLine[1];
+
+                        // Parse the ruby version to remove patch versioning from it.
+                        // At times, ruby will add a patch version such as p112 to the end of the semver version,
+                        // which causes Oryx to not find the correct version. This should take a version like
+                        // 2.3.1p112 and convert it to 2.3.1
+                        var parsedVersionMatches = Regex.Match(fullVersion, @"^(.*?)(?:p[0-9]+)*$");
+                        if (parsedVersionMatches.Success)
+                        {
+                            return parsedVersionMatches.Groups[1].Value;
```
cormacpayne, line 213: Just to confirm: this will also return regular Ruby versions, such as `3.1.1`, correct?

cormacpayne (approved):

daniv-msft (approved):

src/Detector/Ruby/RubyDetector.cs
pauld-msft: Yes! I've verified this with [these tests](https://github.com/microsoft/Oryx/pull/1801/files#diff-d46858e5eb1e1cbbb6b5e806a5e5f15843f228e3c19e16514e6289165e707402R176-R179).

Cases:
`2.3.1` -> `2.3.1` (unchanged)
`2.3.1p112` -> `2.3.1`
`2.3.1.rc1` -> `2.3.1.rc1` (unchanged)
`2.3.1-preview1` -> `2.3.1-preview1` (unchanged)

snehapar9 (commented):
For my own understanding, could you please clarify [this](https://github.com/microsoft/Oryx/issues/1709#issuecomment-1372519555)?

Since 3.2.0.rc1 != 3.2.0, how are we shortening the rc version?

pauld-msft:
> For my own understanding, could you please clarify [this](https://github.com/microsoft/Oryx/issues/1709#issuecomment-1372519555)?

@snehapar9 So as far as I can tell, [ruby has 3 different kinds of releases](https://www.ruby-lang.org/en/downloads/releases/):
- normal release (e.g. `3.0.0` or `3.0.0p123` when written to the `gemfile.lock` file)
- RC ([release candidate](https://en.wikipedia.org/wiki/Software_release_life_cycle#Release_candidate)) release (e.g. `3.0.0-rc1` or `3.0.0.rc1`)
- Preview release (e.g. `3.0.0-preview1`)

That comment was stating that `3.0.0` and `3.0.0p123` are the exact same release, while `3.0.0` and `3.0.0.rc1` are _not_. The rc version is a separate ruby release that may or may not be the exact same, but regardless we do not want to treat them as identical.

> Since 3.2.0.rc1 != 3.2.0, how are we shortening the rc version?

We aren't! See https://github.com/microsoft/Oryx/pull/1801#discussion_r1065077335. We want to keep ruby version detection the same _unless_ we detect a patch in the `gemfile.lock` version, like `3.0.0p123`, which we would just shorten to `3.0.0` since those are the exact same ruby release.

cormacpayne (approved):
:shipit:

PR #1802, opened 2023-01-09
Fix or skip failing tests after default version update

cormacpayne:
The default version for a variety of platforms was updated recently, causing test failures due to certain applications being built against a new version that the app isn't configured to build with in its current state.

This PR attempts to fix the tests that it can and skips the tests that need more investigation that shouldn't block the Validation or Nightly pipelines for other PRs.

### DIFF
--- a/tests/Oryx.BuildImage.Tests/CommandTests/OryxCommandTest.cs
+++ b/tests/Oryx.BuildImage.Tests/CommandTests/OryxCommandTest.cs
@@ -140,10 +140,10 @@ namespace Microsoft.Oryx.BuildImage.Tests
                 () =>
                 {
                     Assert.Contains("#!" + expectedBashPath, result.StdOut);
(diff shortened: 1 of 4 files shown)


snehapar9 (approved):

daniv-msft (approved):

daniv-msft:
Looks good to me. Only question I have is: were these failing tests part of the nightly, and that's why they didn't block the merge?

cormacpayne:
> Looks good to me. Only question I have is: were these failing tests part of the nightly, and that's why they didn't block the merge?

@daniv-msft -- The tests weren't failing previously since they are targeting the `defaultVersion.txt` file in our production storage account, and until the changes I made to update the default versions were published to production, the tests would continue to succeed. It's one of those "chicken or egg" problems we have with testing against the production storage account unfortunately 😞

PR #1816, opened 2023-01-12
Upgrade Cli to Dotnet 7

harryli0108:
Updating csproj file to dotnet 6. Also updating related pipeline files to dotnet 6.


snehapar9 (dismissed):

src/BuildServer/Services/ArtifactBuilders/ArtifactBuilder.cs
```diff
@@ -48,7 +48,9 @@ public bool Build(Build build)
             };
             try
             {
+#pragma warning disable CA1416 // Validate platform compatibility
```
daniv-msft, line 51: Could you please confirm: do we know on which platform this code would be invalid?

src/Detector/Detector.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
```
daniv-msft, line 4: Wow. Didn't know we still had some of this.

daniv-msft: Ah no it's .NET Standard. I thought I read .NET Core and I was like "how did we not upgrade this guy".

daniv-msft:
Looking at the files changed, it seems that the changes are quite straightforward. Were the blockers mentioned previously by Stella already handled separately?

harryli0108:
> Looking at the files changed, it seems that the changes are quite straightforward. Were the blockers mentioned previously by Stella already handled separately?

There are some dependency issues when I'm trying to build the solution. Still trying to see if there are ways to mitigate them. I saw the validation pipeline failed. It was caused by an error that wasn't present when I tried to build locally. I'm also looking into it right now.

src/Detector/Detector.csproj
harryli0108: in this case, should we upgrade it to net6 or keep it as dotnet standard? I think this one is causing some unseen issues in detector.

cormacpayne: We should keep it as `net6.0` -- Dan thought the old target framework was .NET Core 2.0, which ended support four years ago 😄

cormacpayne:
@harryli0108 I'm assuming we need to update some of these ".NET packages" to be in line with the target framework we're snapping to now. For example, some of our `.csproj` files are referencing a `3.1.x` version of Microsoft.Extensions.Logging, which would be compatible with .NET Core 3.1, and we should should be referencing version `6.0.0` of this package (although, looking at this version's [NuGet page](https://www.nuget.org/packages/Microsoft.Extensions.Logging/6.0.0#supportedframeworks-body-tab), it doesn't look like it supports `net6.0`, so we may have to reference version `7.0.0`, which _does_ support `net6.0` -- I would try `6.0.0` first though).

Similarly, we should bump the references to `NLog.Extensions.Logging` up to its latest version, `5.2.1`, which does support `net6.0`.

src/BuildScriptGenerator/BuildScriptGenerator.csproj
```diff
@@ -21,16 +21,16 @@
   <Import Project="$(MSBuildThisFileDirectory)\..\CommonFiles\AssemblyVersion.proj" />
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Http.Polly" Version="3.1.32" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.32" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.32" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="3.1.32" />
-    <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.32" />
+    <PackageReference Include="Microsoft.Extensions.Http.Polly" Version="6.0.10" />
```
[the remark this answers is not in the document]
harryli0108: Thanks for pointing out! I forgot to update that. Just updated it in new commit.

src/BuildScriptGenerator.Common/BuildScriptGenerator.Common.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
```
daniv-msft, line 4: Just to be sure I get it right: so basically at this point we're moving to .NET 7, and not .NET 6? Sounds good to me, but I just want to be sure that's the case.

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -270,15 +338,15 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
                 }
 
                 // Not using IConsole.WriteErrorLine intentionally, to keep the child's error stream intact
```
daniv-msft, line 340: Nit: should we update the `IConsole.WriteErrorLine` part of the comment at this point?

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -270,15 +338,15 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
...
-                console.Error.WriteLine(line);
+                console.WriteLine(line);
```
daniv-msft, line 341: I'm wondering if moving from console.Error.WriteLine to console.WriteLine can have some side effects, as we won't be sending data to stderr anymore. Is there any equivalent we could use? If yes, why not using it?

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
harryli0108: thanks for pointing out. that makes sense! I'll update and add more comments later on.

daniv-msft:
@harryli0108 As this PR is pretty massive, I'm wondering if we should consider how we'll test these changes. Should we have a bug bash? Do you have any opinions on what would give us a good certitude that the changes don't break a hidden scenario?

src/BuildScriptGeneratorCli/Commands/BuildCommandBaseBinder.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Oryx.BuildScriptGeneratorCli.Commands
+{
+    public abstract class BuildCommandBaseBinder<T> : CommandBaseBinder<T>
+        where T : BuildCommandBaseProperty
+    {
+#pragma warning disable SA1401 // Fields should be private
```
daniv-msft, line 14: Nit: Please consider adding to the comment why we don't make these fields private, so that future devs reading this code don't need to dig too much to understand why.

src/BuildScriptGeneratorCli/Program.cs
```diff
@@ -3,33 +3,46 @@
...
-        [Option(CommandOptionType.NoValue, Description = "Print version information.")]
         public bool Version { get; set; }
 
+        internal static async Task<int> Main(string[] args)
+        {
+            var rootCommand = new RootCommand();
+            rootCommand.Name = "oryx";
+            rootCommand.Description = "Generates and runs build scripts for multiple platforms.";
+
+            // var versionOption = new Option<bool>(aliases: new[] { "-v", "--version" }, "Print version information.");
```
daniv-msft, line 29: Nit: looks like we might be able to remove this one? Similar comment below.

src/BuildScriptGenerator.Common/BuildScriptGenerator.Common.csproj
harryli0108: yep, there's some dependency compatibility issue with packages. ww've updated system packages to version 7.0.0, which some of them have complaints about using net6.0.

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
harryli0108: There is a `console.Error.Write`. I think this was some early changes that i kept the old method. I'll update them in the next commit.

daniv-msft: Thanks!

src/BuildScriptGeneratorCli/Commands/BuildCommandBaseBinder.cs
harryli0108: I'll definitely add more docs and comments to explain it for maintainability.

src/BuildScriptGenerator.Common/BuildScriptGenerator.Common.csproj
daniv-msft: Thanks! So to be 100% sure: in practice will we have any projects relying on .NET 6 or will they all move to .NET 7?

harryli0108: For the sake of compatability, i would say moving to net7 since most of the packages we're using already released ver 7. But i would like to hear more opinions from team just to make sure we're making the right move.

src/BuildScriptGenerator/BuildScriptGenerator.csproj
```diff
@@ -21,16 +21,16 @@
   <Import Project="$(MSBuildThisFileDirectory)\..\CommonFiles\AssemblyVersion.proj" />
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Http.Polly" Version="3.1.32" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" Version="3.1.32" />
-    <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="3.1.32" />
-    <PackageReference Include="Microsoft.Extensions.Options" Version="3.1.32" />
-    <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.32" />
+    <PackageReference Include="Microsoft.Extensions.Http.Polly" Version="7.0.2" />
```
cormacpayne, line 24: nit: we might want to downgrade this to `7.0.0 to be consistent with the other `Microsof.Extensions.Http.*` references

cormacpayne, line 24: I see this across a few `.csproj` files: would you mind ensuring that the package references are in alphabetical order?

src/BuildScriptGeneratorCli/Commands/Binders/BuildCommandBaseBinder.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
```
cormacpayne, line 1: For all new files, please make sure that they have the Microsoft license header (you can pull this from any other `.cs` file in the repository)

src/BuildScriptGeneratorCli/Commands/Binders/BuildCommandBaseBinder.cs
```diff
@@ -0,0 +1,54 @@
...
+            Option<string> platformVersion,
+            Option<bool> package,
+            Option<string> osRequirements,
+            Option<string> appType,
+            Option<string> buildCommandFile,
+            Option<bool> compressDestinationDir,
+            Option<string[]> property,
+            Option<string> dynamicInstallRootDir,
+            Option<string> logPath,
+            Option<bool> debugMod)
```
cormacpayne, line 39: nit: `debugMod` --> `debugMode`

This applies for a few other files

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -82,18 +88,85 @@ public string LanguageVersion
...
-            OptionTemplates.ManifestDir,
-            CommandOptionType.SingleValue,
-            Description = "The path to a directory into which the build manifest file should be written.")]
         public string ManifestDir { get; set; }
 
+        public static Command Export()
+        {
+            var logOption = new Option<string>(OptionTemplates.Log, OptionTemplates.LogDescription);
+            var debugOption = new Option<bool>(OptionTemplates.Debug, OptionTemplates.DebugDescription);
+            var sourceDirArgument = new Argument<string>("SourceDir", "The source directory.");
```
cormacpayne, line 99: nit: I think it would be worth introducing a new `ArgumentTemplates` file (similar to `OptionTemplates`) that can store the hard-coded strings referenced on this line

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -82,18 +88,85 @@ public string LanguageVersion
...
+            var platformOption = new Option<string>(OptionTemplates.Platform, OptionTemplates.PlatformDescription);
+            var platformVersionOption = new Option<string>(OptionTemplates.PlatformVersion, OptionTemplates.PlatformVersionDescription);
+            var packageOption = new Option<bool>(OptionTemplates.Package, OptionTemplates.PackageDescription);
+            var osReqOption = new Option<string>(OptionTemplates.OsRequirements, OptionTemplates.OsRequirementsDescription);
+            var appTypeOption = new Option<string>(OptionTemplates.AppType, OptionTemplates.AppTypeDescription);
+            var buildCommandFileNameOption = new Option<string>(OptionTemplates.BuildCommandsFileName, OptionTemplates.BuildCommandsFileNameDescription);
+            var compressDestDirOption = new Option<bool>(OptionTemplates.CompressDestinationDir, OptionTemplates.CompressDestinationDirDescription);
+            var propertyOption = new Option<string[]>(aliases: new[] { "-p", OptionTemplates.Property }, OptionTemplates.PropertyDescription);
```
cormacpayne, line 107: nit: it may be worth to change `OptionTemplates.Property` to `OptionTemplates.PropertyAliases` and have it return a `string[]` that is the array of values accepted as you have on this line

```cs
public string[] PropertyAliases = new string[] { "-p", "--property" };
```

This comment applies for the other alias values we have.

src/BuildScriptGeneratorCli/Commands/BuildpackDetectCommand.cs
```diff
@@ -4,35 +4,77 @@
...
-        [Option("--platform-dir <dir>", CommandOptionType.SingleValue, Description = "Platform directory path.")]
         public string PlatformDir { get; set; }
 
-        [Option("--plan-path <path>", CommandOptionType.SingleValue, Description = "Build plan TOML path.")]
         public string PlanPath { get; set; }
 
+        public static Command Export()
+        {
+            var sourceDirArgument = new Argument<string>("SourceDir", "The source directory.");
+            var platformDirOption = new Option<string>("--platform-dir", "Platform directory path.");
```
cormacpayne, line 47: nit: we should add `--platform-dir` and `--plan-path` to the `OptionTemplates` file

src/BuildScriptGeneratorCli/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
```
cormacpayne, line 1: Do we need this file?

src/BuildScriptGeneratorCli/readme.md
```diff
@@ -0,0 +1,18 @@
+# Oryx Cli Structure
+## Background
+Due to the transition to net7, we found that the past cli package isn't compatible with the new framework. Hence, we've decided to move to `System.CommandLine` package to develop our cli. For developers, the structure of the cli codes are different then the past.
```
cormacpayne, line 3: nit: for `.md` files, I always like to cap the line length at 120 characters to make it easier to read in GitHub reviews :smile:

cormacpayne, line 3: nit: `net7` --> `.NET 7`, `cli` --> `CLI`

src/BuildScriptGeneratorCli/readme.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Structure
+- Program.cs (Define root command `oryx` and add subcommands)
+  - Subcommand.cs (Define subcommands, register its options and arguments, and define the execute action)
+
+### 1. Command Files
+Now, the command folder will not handle option and argument inputs directly. This will be done in the binder. The command file will only define the subcommands, register its options and arguments, and also define the execute action. The command (with all its options and executes) will then be exported to be added to the root command.
```
cormacpayne, line 10: nit: "also define the execute action" --> "also define the `Execute` method"

src/BuildScriptGeneratorCli/readme.md
```diff
@@ -0,0 +1,18 @@
...
+
+**Please be aware, developer still need to define the properties used in the execute method in command file!** The property file is a intermediate step in order to make the binder's result accessable to the command file.
```
cormacpayne, line 12: nit: "developer still" --> "the developer will still", "in the execute method" --> in the `Execute` method"

src/BuildScriptGeneratorCli/readme.md
```diff
@@ -0,0 +1,18 @@
...
+
+### 2. Binders
+With the transition, we've updated command folder with binder folder. Each subcommand will have its own custom binder defined. This is where `System.CommandLine` package actually handles user inputs in each option. We define the expected user intput type (e.g. `bool` or `string`) for options and arguments and set property.
+
+### 3. Properties
+Oryx now has a new folder in the cli project that contains all the properties the cli commmands need to execute. We still keeps different properties getter and setter in each command class since most of the commands execute method is using them. The property file is a easy way to pass the values from parsed options and arguments to the execute.
```
cormacpayne, line 18: nit: `cli` --> `CLI`, "execute method" --> "`Execute` method"

src/Detector/DotNetCore/DotnetCoreDetector.cs
```diff
@@ -94,7 +94,7 @@ internal static string DetermineRuntimeVersion(string targetFramework)
             // Support .NET moniker aliases
             // Ex: "472" => "4.72"
             //     "48" => "4.8"
-            if (targetFramework.IndexOf('.') == -1)
+            if (targetFramework.Contains('.') == false)
```
cormacpayne, line 97: nit: any reason for this change? If so, we should consider removing the `== false`:

```cs
if (!targetFramework('.'))
```

tests/BuildScriptGeneratorCli.Tests/TestConsole.cs
```diff
@@ -88,6 +89,10 @@ public ConsoleColor BackgroundColor
             set => throw new NotImplementedException();
         }
 
+        IStandardStreamWriter IStandardOut.Out => throw new NotImplementedException();
```
cormacpayne, line 92: We'll need to sync up on implementing this for the tests to run properly

cormacpayne (changes requested):

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
harryli0108: For this one, i think hardcoded argument would be ok because unlike multiple options, there is only one argument throughout all the commands.

tests/Oryx.BuildImage.Tests/CommandTests/OryxCommandTest.cs
```diff
@@ -89,7 +89,7 @@ public void BuildImage_Build_UsesCwd_WhenNoSourceDirGiven()
             RunAsserts(
                 () =>
                 {
-                    Assert.Contains("Error: Could not detect", result.StdErr);
+                    Assert.Contains("Required argument missing for command: 'build'.", result.StdErr);
```
cormacpayne, line 92: I don't think this is the error we want to be checking for -- based on the test, if there's no value provided for the `sourceDir` argument, we used the current working directory, and the error we checked for previously was that we couldn't detect a platform from the `cwd`, but now we're throwing an error saying that an argument is missing for `oryx build`, which seems to imply that we can't run `oryx build` without the `sourceDir` argument. We should double check this since it would be a breaking change if users couldn't run this command without explicitly providing a path to build.

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
```diff
@@ -737,7 +737,7 @@ public void DotnetFails_ToInstallUnsupportedSdk(string runtimeVersion, string im
                 () =>
                 {
                     Assert.False(result.IsSuccess);
-                    Assert.Contains($"Error: Platform '{DotNetCoreConstants.PlatformName}' version '{runtimeVersion}' is unsupported.", result.StdErr);
+                    Assert.Contains($"Debug mode enabled\nPlatform '{DotNetCoreConstants.PlatformName}' version '{runtimeVersion}' is unsupported.", result.StdErr);
```
cormacpayne, line 740: We should figure out why `Debug mode enabled` is being displayed without us providing the `--debug` flag in the command above (as far as I can tell we aren't providing the flag, but it may be provided under-the-hood from one of our helper methods called on `ShellScriptBuilder`)

This comment applies to all of the other tests that have this same issue.

cormacpayne (changes requested):

src/BuildScriptGeneratorCli/Program.cs
```diff
@@ -55,23 +77,17 @@ internal static string GetMetadataValue(string name)
             return null;
         }
 
-        internal int OnExecute(CommandLineApplication app, IConsole console)
+        internal static int OnExecute(IConsole console, bool setVersion)
         {
-            if (this.Version)
+            if (setVersion)
```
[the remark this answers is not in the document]
harryli0108: this one is there to make sure the version and commit info doesn't show up if `--info` is not present. for some reason the info (our customized version option) page would still show up if i only type `oryx` or `oryx +any-subcommand`.

tests/BuildScriptGeneratorCli.Tests/ProgramTest.cs
```diff
@@ -1,44 +0,0 @@
...
-
-using McMaster.Extensions.CommandLineUtils;
-using Xunit;
-
-namespace Microsoft.Oryx.BuildScriptGeneratorCli.Tests
-{
-    public class ProgramTest
-    {
-        [Fact]
-        public void OnExecute_ShowsHelp_AndExits_WithSuccessExitCode()
```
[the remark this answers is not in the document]
harryli0108: we could reuse this test for testing the new `--info` option output. I'll add it back. Thanks!

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreDynamicInstallTest.cs
harryli0108: yea it seems like --debug was provided in `[ShellScriptBuilder](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator.Common/ShellScriptBuilder.cs#L62-L65)`. But it's interesting why it didn't show up in the past.

harryli0108: Found the reason why debug message went into error stream. I changed console.WriteLine to console.Error.WriteLine for outputing debug enable message. Shoudl be fixed now.

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -82,18 +89,86 @@ public string LanguageVersion
...
+            var dynamicInstallRootDirOption = new Option<string>(OptionTemplates.DynamicInstallRootDir, OptionTemplates.DynamicInstallRootDirDescription);
+
+            // Hiding Language Option because it is obselete
+            var languageOption = new Option<string>(OptionTemplates.Language, OptionTemplates.LanguageDescription);
+            languageOption.IsHidden = true;
+
+            // LanguageVer Option is obselete
+            var languageVerOption = new Option<string>(OptionTemplates.LanguageVersion, OptionTemplates.LanguageVersionDescription);
+            languageVerOption.IsHidden = true;
+            var intermediateDirOption = new Option<string>(OptionTemplates.IntermediateDir, OptionTemplates.IntermediateDirDescription);
```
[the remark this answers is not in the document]
cormacpayne: nit: this comment still applies

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -79,18 +90,89 @@ public string LanguageVersion
...
+            command.SetHandler(
+                (prop) =>
+                {
+                    // InvocationContext provided in SetHandler
+                    var buildCommand = new BuildCommand(prop);
+                    var returnCode = buildCommand.OnExecute(console);
+                    return Task.FromResult(returnCode);
+                },
+                new BuildCommandBinder(
+                    languageOption,
```
cormacpayne, line 156: One suggestion that would be nice and would help to prevent future incidents: when you're initializing an object and providing a large amount of arguments to the constructor (or even with a method that has a lot of arguments), it's best to specify the argument name before each value you provide so it's clear which argument each value is being mapped to.

This is especially useful if we were to introduce a new argument to this constructor in the future and it wasn't necessarily tacked on to the end, there's a possibility the value gets mapped to the wrong argument.

I would do this for each of the `Binder` objects that you're initializing for the `SetHandler` call in each command class.

src/BuildScriptGeneratorCli/Commands/BuildpackDetectCommand.cs
```diff
@@ -4,35 +4,81 @@
...
+                name: "SourceDir",
+                description: "The source directory.",
+                getDefaultValue: () => Directory.GetCurrentDirectory());
+            var platformDirOption = new Option<string>(OptionTemplates.PlatformDir, "Platform directory path.");
+            var planPathOption = new Option<string>(OptionTemplates.PlanPath, "Build plan TOML path.");
+            var logFilePathOption = new Option<string>(OptionTemplates.Log, OptionTemplates.LogDescription);
+            var debugMode = new Option<bool>(OptionTemplates.Debug, OptionTemplates.DebugDescription);
+
+            var command = new Command("buildpack-detect", "Determine whether Oryx can be applied as a buildpack to an app in the current " +
+        "working directory.")
```
cormacpayne, line 57: nit: when moving a string over to the next line, let's make sure it's idented past the previous lined:

```cs
var command = new Command("...", "..." +
    "working directory.)
```

src/BuildScriptGeneratorCli/readme.md
```diff
@@ -0,0 +1,11 @@
+# Oryx CLI Structure
+### 1. Command Files
+Option and argument inputs are handled through binders. The command file only define the subcommands and register options and arguments, and export them to be added to root command in `Program.cs`.
```
cormacpayne, line 3: nit: comment still applies about capping `.md` file lines to 120 characters where we can for readability

tests/Oryx.BuildImage.Tests/CommandTests/OryxCommandTest.cs
```diff
@@ -89,7 +89,7 @@ public void BuildImage_Build_UsesCwd_WhenNoSourceDirGiven()
             RunAsserts(
                 () =>
                 {
-                    Assert.Contains("Error: Could not detect", result.StdErr);
```
cormacpayne, line 92: Just confirming: with the new command line library, it's not writing an `Error` prefix?

cormacpayne (dismissed):

tests/Oryx.BuildImage.Tests/CommandTests/OryxCommandTest.cs
harryli0108: i think the new console doesn't provide Error or Warning tag anymore even using `console.Error.WriteLine`

src/BuildScriptGeneratorCli/Commands/Binders/CommandBaseBinder.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Oryx.BuildScriptGeneratorCli.Commands
+{
+    public abstract class CommandBaseBinder<T> : BinderBase<T>
+        where T : CommandBaseProperty
+    {
+#pragma warning disable SA1401 // Fields should be private
```
daniv-msft, line 19: Could you please clarify why we're not making these fields directly private? What would be the issue if we did so?

src/BuildScriptGeneratorCli/Commands/Binders/DetectCommandBinder.cs
```diff
@@ -0,0 +1,41 @@
...
+            this.sourceDirArgument = sourceDirArgument;
+            this.outputFormatOption = outputFormatOption;
+        }
+
+        protected override DetectCommandProperty GetBoundValue(BindingContext bindingContext) =>
+            new DetectCommandProperty
+            {
+                SourceDir = bindingContext.ParseResult.GetValueForArgument(this.sourceDirArgument),
+                OutputFormat = bindingContext.ParseResult.GetValueForOption(this.outputFormatOption),
+                LogFilePath = bindingContext.ParseResult.GetValueForOption(this.logPath),
```
daniv-msft, line 37: Nit: I notice that some of the parameters have different names from the fields we assign them to (logPath vs. logFilePath, outputFormatOption vs. outputFormat). Please consider renaming one of them to be consistent.

src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
```diff
@@ -4,31 +4,110 @@
...
+                    platformVersion: platformVersionOption,
+                    package: packageOption,
+                    osRequirements: osReqOption,
+                    appType: appTypeOption,
+                    buildCommandFile: buildCommandFileNameOption,
+                    compressDestinationDir: compressDestDirOption,
+                    property: propertyOption,
+                    dynamicInstallRootDir: dynamicInstallRootDirOption,
+                    logPath: logOption,
+                    debugMode: debugOption));
```
daniv-msft, line 107: Similar comment as above: please consider aligning the names so that we're consistent everywhere. That way, it will reduce the uncertainty for devs working with this code late, as well as making searches in the code easier.

daniv-msft (changes requested):
Looks good to me overall, and I just added a couple of comments. Once these are done, I'm happy to sign-off! Thanks for making this huge change

src/BuildScriptGeneratorCli/Commands/Binders/CommandBaseBinder.cs
harryli0108: So because it is used later in other binders that inherits from `CommandBaseBinder`, it would be more convenient to go with `protected` than making these private and use getter.

src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
harryli0108: I've just changed most of the alignment issue across different binders. For the alignments between parameter names and actual option names in single command file, I kept the naming convention so that it is obvious which type (option or argument) it is, which helps with the readability a bit of the code imo.

src/BuildScriptGeneratorCli/Commands/Binders/CommandBaseBinder.cs
pauld-msft: Could you define each field as a property instead, like
`protected Argument<string> SourceDir { get; set; }`
or alternatively be set as static and readonly if you would like to keep it as a field?
`protected static readonly Argument<string> sourceDir;`

harryli0108: that makes sense. just pushed those changes and waiting for validations. thanks for the suggestion!

src/BuildScriptGeneratorCli/Commands/Binders/BuildScriptCommandBinder.cs
```diff
@@ -33,25 +33,25 @@ public class BuildScriptCommandBinder : BuildCommandBaseBinder<BuildScriptComman
...
-                CompressDestinationDir = bindingContext.ParseResult.GetValueForOption(this.compressDestinationDir),
-                Properties = bindingContext.ParseResult.GetValueForOption(this.property),
-                DynamicInstallRootDir = bindingContext.ParseResult.GetValueForOption(this.dynamicInstallRootDir),
-                LogFilePath = bindingContext.ParseResult.GetValueForOption(this.logPath),
-                DebugMode = bindingContext.ParseResult.GetValueForOption(this.debugMode),
+                OutputPath = bindingContext.ParseResult.GetValueForOption(this.buildScriptOutput),
+                SourceDir = bindingContext.ParseResult.GetValueForArgument(this.SourceDir),
+                PlatformName = bindingContext.ParseResult.GetValueForOption(this.Platform),
+                PlatformVersion = bindingContext.ParseResult.GetValueForOption(this.PlatformVersion),
+                ShouldPackage = bindingContext.ParseResult.GetValueForOption(this.Package),
```
daniv-msft, line 46: Could you please clarify if you see any issues with reconciling the names together?
OutputPath vs. buildScriptOutput
PlatformName vs Platform
etc.

src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
daniv-msft: OK, I had another comment along these lines that I added tonight. Please don't hesitate to let me know if it falls in the case you're mentioning here. Getting them aligned would have been great, but if it would reduce the readability overall I'm good with keeping as is.

harryli0108: Yes, so i kept the type of the parameter into the variable name (e.g. option or argument). I think it'll be a reminder of what type of parameter is used to whoever is changing it.

src/BuildScriptGeneratorCli/Commands/Binders/BuildScriptCommandBinder.cs
harryli0108: Sorry i forgot to change the outputPath name. This one is aligned now.
For PlatformName and Platform, the property name used in `BuildScriptCommandBinder` is PlatformName, but in `BuildCommandBaseBinder` where it was inherited from, it uses Platform as the variable name. In that case, i kept the different names here.

daniv-msft: Thanks for your reply. Is it normal that `this.SourceDir` and others are capitalized now? I thought that here we would be referring to the class member that are lowercase. Am I missing something?
Similarly, could you please clarify if for the other names we still should keep them as non-reconciled? I'm thinking about ShouldPackage vs. Package, BuildCommandsFileName vs. BuildCommandFile, Properties vs. Property, LogFilePath vs. LogPath.
Even if they are called that way in another class, would it be expensive to update them?

harryli0108: Yes, so for instance, `SourceDir` at the start of line 43 is refering to the properties from the property file. Here, I kept the naming convention the same as we have in command files. Later in the line, `this.SourceDir` refers to the arguments defined from its inherited file `BuildCommandBaseBinder`. Because we changed ways of defining options and arguments in base classes from protected fields to protected properties, we need to capitalize them. But for `outputPath`, because it is a private field defined in this level, it needs to be in small letters.

And for changing some of the others, because lots of them are shared by multiple inherited docs, some of the names might change from one to another in different files and also a bit time consuming to change. I would prefer to keep them as it is for now.

pauld-msft: For the sake of consistency, I would prefer to update all of the Binders to use properties so that we don't run into any confusion in the future for why the different options are capitalized differently. You'll notice that even now, we are seeing a warning about converting to readonly, which can be an indication that something is a bit off.

image

You can use `ctrl+R+R` when renaming these, which should find all references to each field across all files (_though, I'll note that all of these are private methods, so it would be fishy if there were references to them outside of a single file_), and update to whatever the new name is:
image
Then you can just add a `{ get; set; }` and call it a day

images/receiveGpgKeys.sh
```diff
@@ -16,7 +16,7 @@ fi
 for keyID in $KEY_IDS; do
     for i in {1..10}; do
         echo "Try # $i"
-        gpg --batch --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys "$keyID" || \
+        gpg --batch --keyserver hkps://keyserver.ubuntu.com:443 --recv-keys "$keyID" || \
```
pauld-msft, line 19: out of curiousity, did this resolve the issues you were seeing with the gpg keys when building images?

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
```diff
@@ -266,7 +349,7 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
                     return;
                 }
 
-                // Not using IConsole.WriteErrorLine intentionally, to keep the child's error stream intact
+                // Not using IConsole.WriteLine intentionally, to keep the child's error stream intact
```
pauld-msft, line 352: nit: I think we can revert the change in this comment, since we are still not using `WriteErrorLine` to preserve the error stream

src/BuildScriptGeneratorCli/Commands/OptionTemplates.cs
```diff
@@ -9,26 +9,58 @@ namespace Microsoft.Oryx.BuildScriptGeneratorCli
...
-        public const string ManifestDir = "--manifest-dir <directory-path>";
-        public const string BuildCommandsFileName = "--buildcommands-file <file-name>";
-        public const string AppType = "--apptype <application-type>";
-        public const string CompressDestinationDir = "--compress-destination-dir";
-        public const string DynamicInstallRootDir = "--dynamic-install-root-dir <dir-path>";
-
-        public const string EventName = "--event-name <event-name>";
-        public const string ProcessingTime = "--processing-time <processing-time>";
-        public const string CallerId = "--caller-id <caller-id>";
+        public static readonly string LanguageVersion = "--language-version";
```
pauld-msft, line 25: Nit: Is there a reason why some of these `string` fields are const, and some are static readonly? Lets try to opt for consistency wherever we can

src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
```diff
@@ -4,31 +4,110 @@
...
+            var packageOption = new Option<bool>(OptionTemplates.Package, OptionTemplates.PackageDescription);
+            var osReqOption = new Option<string>(OptionTemplates.OsRequirements, OptionTemplates.OsRequirementsDescription);
+            var appTypeOption = new Option<string>(OptionTemplates.AppType, OptionTemplates.AppTypeDescription);
+            var buildCommandFileNameOption = new Option<string>(OptionTemplates.BuildCommandsFileName, OptionTemplates.BuildCommandsFileNameDescription);
+            var compressDestDirOption = new Option<bool>(OptionTemplates.CompressDestinationDir, OptionTemplates.CompressDestinationDirDescription);
+            var propertyOption = new Option<string[]>(OptionTemplates.Property, OptionTemplates.PropertyDescription);
+            var dynamicInstallRootDirOption = new Option<string>(OptionTemplates.DynamicInstallRootDir, OptionTemplates.DynamicInstallRootDirDescription);
+            var buildScriptOutputOption = new Option<string>(
+                name: "--output",
+                description: "The path that the build script will be written to. " +
```
pauld-msft, line 68: Nit: Move these name and description strings to OptionTemplates

pauld-msft: Same comment for SourceDir, and all other option names and descriptions.

src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
```diff
@@ -4,31 +4,110 @@
...
+                          "If not specified, the result will be written to STDOUT.");
+
+            var command = new Command("build-script", "Generate build script to standard output.")
```
pauld-msft, line 71: Nit: Is there a reason we changed this from a constant to hard-coded? I'd prefer to put these back into constants in this file or a general constants file that has all command names

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
pauld-msft: Though actually, this makes it seems as though `console.Error.WriteLine != console.WriteErrorLine` even though we are replacing the latter with the former later on in this code. Do you know how this affects the logging?

src/BuildScriptGeneratorCli/Commands/BuildpackDetectCommand.cs
```diff
@@ -4,35 +4,81 @@
...
-        [Option("--plan-path <path>", CommandOptionType.SingleValue, Description = "Build plan TOML path.")]
         public string PlanPath { get; set; }
 
+        public static Command Export(IConsole console)
+        {
+            var sourceDirArgument = new Argument<string>(
+                name: "SourceDir",
+                description: "The source directory.",
+                getDefaultValue: () => Directory.GetCurrentDirectory());
+            var platformDirOption = new Option<string>(OptionTemplates.PlatformDir, "Platform directory path.");
```
pauld-msft, line 51: Nit: Same comment here about putting the option description into OptionTemplates as well, like we do with other descriptions

src/BuildScriptGeneratorCli/Commands/ExecCommand.cs
```diff
@@ -15,23 +17,65 @@
...
+using Microsoft.Oryx.BuildScriptGeneratorCli.Commands;
 using Microsoft.Oryx.BuildScriptGeneratorCli.Options;
 
 namespace Microsoft.Oryx.BuildScriptGeneratorCli
 {
-    [Command(Name, Description = "Execute an arbitrary command in the default shell, in an environment " +
-        "with the best-matching platform tool versions.")]
     internal class ExecCommand : CommandBase
     {
         public const string Name = "exec";
```
pauld-msft, line 27: Nit: Per https://github.com/microsoft/Oryx/pull/1816/files#r1129788587 I would lean towards keeping this and using it in the Command creation.

src/BuildScriptGeneratorCli/Commands/RunScriptCommand.cs
```diff
@@ -4,48 +4,85 @@
...
 namespace Microsoft.Oryx.BuildScriptGeneratorCli
 {
-    [Command(
-        Name,
-        Description = "Generate startup script for an app.",
-        UnrecognizedArgumentHandling = UnrecognizedArgumentHandling.StopParsingAndCollect,
-        AllowArgumentSeparator = true)]
     internal class RunScriptCommand : CommandBase
     {
         public const string Name = "run-script";
```
pauld-msft, line 21: Nit: Per https://github.com/microsoft/Oryx/pull/1816/files#r1129788587 I would lean towards keeping this and using it in the Command creation. Otherwise, should probably be cleaned up

images/receiveGpgKeys.sh
harryli0108: yes! this one solved the issue for me.

src/BuildScriptGeneratorCli/Commands/BuildCommand.cs
pauld-msft: Ok, yeah, so looks like WriteErrorLine is an Oryx defined method that prepends `Error:` to Error.WriteLine with `console.Error.WriteLine("Error: " + message)`:
https://github.com/microsoft/Oryx/blob/harrli/cliDotnetUpgrade/src/BuildScriptGeneratorCli/IConsoleExtensions.cs

Can we revert all of the changes from `console.WriteErrorLine -> console.Error.WriteLine` so we get the Error prepended again?

tests/Oryx.BuildImage.Tests/CommandTests/OryxCommandTest.cs
pauld-msft: https://github.com/microsoft/Oryx/pull/1816/files#r1129817709

src/BuildScriptGeneratorCli/Commands/BuildScriptCommand.cs
```diff
@@ -31,44 +34,43 @@ public BuildScriptCommand(BuildScriptCommandProperty input)
...
+            var packageOption = new Option<bool>(OptionArgumentTemplates.Package, OptionArgumentTemplates.PackageDescription);
+            var osReqOption = new Option<string>(OptionArgumentTemplates.OsRequirements, OptionArgumentTemplates.OsRequirementsDescription);
+            var appTypeOption = new Option<string>(OptionArgumentTemplates.AppType, OptionArgumentTemplates.AppTypeDescription);
+            var buildCommandFileNameOption = new Option<string>(OptionArgumentTemplates.BuildCommandsFileName, OptionArgumentTemplates.BuildCommandsFileNameDescription);
+            var compressDestDirOption = new Option<bool>(OptionArgumentTemplates.CompressDestinationDir, OptionArgumentTemplates.CompressDestinationDirDescription);
+            var propertyOption = new Option<string[]>(OptionArgumentTemplates.Property, OptionArgumentTemplates.PropertyDescription);
+            var dynamicInstallRootDirOption = new Option<string>(OptionArgumentTemplates.DynamicInstallRootDir, OptionArgumentTemplates.DynamicInstallRootDirDescription);
             var buildScriptOutputOption = new Option<string>(
                 name: "--output",
```
pauld-msft, line 70: Nit: update to be `aliases: OptionArgumentTemplates.Output,`?

src/BuildScriptGeneratorCli/Commands/DetectCommand.cs
```diff
@@ -45,17 +48,15 @@ public static Command Export(IConsole console)
         {
             var sourceDirArgument = new Argument<string>(
                 name: "SourceDir",
```
pauld-msft, line 50: Nit: `OptionArgumentTemplates.SourceDir`

pauld-msft (commented):
A few minor nitpicks things left for me, happy to approve once those are cleared up! Thanks

pauld-msft (dismissed):
Looks good to me! As Dan mentioned in standup, lets wait until we have a bug bash / some kind of testing which will increase our confidence before merging in this huge PR. Thanks for quickly handling all the updates!

william-msft (dismissed):

harryli0108:
Merged test acr pipeline into this pr.

daniv-msft (approved):

PR #1821, opened 2023-01-14
Add back PHP 8.2 as runtime version in constants.yaml

daniv-msft:
Just adding back PHP 8.2 to the runtime versions in constants.yaml, as it was removed by mistake recently.


daniv-msft:
Parallel change: https://github.com/microsoft/Oryx/pull/1819/files

PR #1832, opened 2023-01-19
Williamhe/1717548 fix python 310 311 failing

william-msft:
  - When `regex` is used in python 3.10 or 3.11 it fails due to a python include library not being found. This PR will set the path, when dynamic installation occurs, to be used later down the lifecycle of building the project
    - Example: 
image
    - Please note the incorrect intermediate `gcc` path: `-I/opt/python/3.10.9/include/python3.10` 
    - In the case of python 3.10.9, the path needed to be updated to  `/tmp/oryx/platforms/python/3.10.9/include/python3.10/` which this PR addresses.
  - Related issue/ICM
    - https://github.com/microsoft/Oryx/issues/1767
    - https://portal.microsofticm.com/imp/v3/incidents/details/359062821/home 
  - One popular solution was to `apt-get install python3.10-dev` which will not work for our Debian solutions since this is only supported Debian bookworm and sid:
    -  https://packages.debian.org/search?keywords=python3.10-dev
image
    - We first check if `pip install` fails without the global option flags. If it fails, we retry but with the global option flags. Since passing `--global-option` sometimes fails due to [PEP517](https://github.com/pypa/pip/issues/7991); this is case-by-case depending on the module pip is installing.
    - Using symbolic link does not work as tested in this [PR](https://github.com/microsoft/Oryx/pull/1836) since python virtual environment does not recognize symlinks outside the virtual environment.

~- [] Proper license headers are included in each file.~


cormacpayne (approved):
LGTM once both pipelines succeed

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -195,6 +195,17 @@ public virtual void InstallPlatformSpecificSkeletonDependencies(StringBuilder st
                 .AppendLine($"  fi")
                 .AppendLine("fi")
 
+                // Required for python 3.10 and 3.11
+                .AppendLine($"PIP_GCC_FLAGS=\"\"")
+                .AppendLine($"if [ \"$platformName\" = \"python\" ]; then")
+                .AppendLine($"  if [[ '{version}' == 3.10* ]] || [[ '{version}' == 3.11* ]]; then")
```
daniv-msft, line 201: Just to be sure I get it: do we expect the issue to not happen in the next Python version? Or do we just not know?
If we expect it should happen in 3.12, then having a logic that targets "greater or equal" versions would reduce the risk that we just forget about this case when 3.12 arrives. That said, if we reasonably believe that this might not be an issue in 3.12, please feel free to ignore.

william-msft: We'd have to support this in Oryx to be able to test. Since the library that would be responsible for addressing header files is only supported for 3.10 and 3.11, I figured that would be the versions we have this work around for:

- https://packages.debian.org/sid/python3.10-dev
- https://packages.debian.org/sid/python3.11-dev

daniv-msft: OK seems fair, thanks.

PR #1836, opened 2023-01-21
Williamhe/1717548 fix python310 311 pythonh error

william-msft:
  - When `regex` is used in python 3.10 or 3.11 it fails due to a python include library not being found. This PR will set a symbolic link from the dynamic installed directory to the directories where python is expected to be installed.
    - Example: 
image
  - Related issue/ICM
    - https://github.com/microsoft/Oryx/issues/1767
    - https://portal.microsofticm.com/imp/v3/incidents/details/359062821/home 
  - One popular solution was to `apt-get install python3.10-dev` which will not work for our Debian solutions since this is only supported Debian bookworm and sid:
    -  https://packages.debian.org/search?keywords=python3.10-dev
image

~- [] Proper license headers are included in each file.~


william-msft:
Closing this PR in favor of: https://github.com/microsoft/Oryx/pull/1832

PR #1843, opened 2023-01-25
Handle unsupported runtime platforms in 'oryx dockerfile'

cormacpayne:
Currently when running the `oryx dockerfile` command against an application that targets a platform without a supporting set of runtime images, the Dockerfile generated has empty values for the runtime, which is causing some users issues when using `azure/container-app-deploy-action` or the `AzureContainerAppsRC` task.

Relevant issue filed: https://github.com/Azure/container-apps-deploy-pipelines-task/issues/9

This PR now checks for the build and runtime images provided/discovered by the `oryx dockerfile` command and ensures that they are valid before trying to populate the Dockerfile template with them, and if not valid, throws meaningful exceptions to let the user know why the command failed (and in the case of runtimes, point them to [our doc](https://github.com/microsoft/Oryx/blob/main/doc/supportedRuntimeVersions.md) with the supported runtimes).

### DIFF
--- a/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
+++ b/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
@@ -266,45 +266,44 @@ namespace Microsoft.Oryx.BuildScriptGenerator
         {
             if (string.IsNullOrEmpty(dockerfileProperties.BuildImageName))
             {
-                var message = "Provided build image name parsed from the provided --build-image argument is empty. " +
-                    "Please provide a valid value for --build-image, or remove the argument and use the default " +
-                    "'cli:stable' build image from mcr.microsoft.com/oryx.";
-                var exc = new InvalidDockerfileImageException(message);
-                this.logger.LogError(exc, message);
-                throw exc;
+                this.LogInvalidDockerfileImageException(
+                    "Provided build image name parsed from the provided --build-image argument is empty. " +
+                    "Please provide a valid value for --build-image, or remove the argument to use the default " +
+                    "'cli:stable' build image from mcr.microsoft.com/oryx.");
             }
 
             if (string.IsNullOrEmpty(dockerfileProperties.BuildImageTag))
             {
-                var message = "Provided build image tag parsed from the provided --build-image argument is empty. " +
-                    "Please provide a valid value for --build-image, or remove the argument and use the default " +
-                    "'cli:stable' build image from mcr.microsoft.com/oryx.";
-                var exc = new InvalidDockerfileImageException(message);
-                this.logger.LogError(exc, message);
-                throw exc;
+                this.LogInvalidDockerfileImageException(
+                    "Provided build image tag parsed from the provided --build-image argument is empty. " +
+                    "Please provide a valid value for --build-image, or remove the argument to use the default " +
+                    "'cli:stable' build image from mcr.microsoft.com/oryx.");
             }
 
             if (string.IsNullOrEmpty(dockerfileProperties.RuntimeImageName))
             {
-                var message = "Either the value provided to the --runtime-platform argument is empty, or the platform " +
+                this.LogInvalidDockerfileImageException(
+                    "Either the value provided to the --runtime-platform argument is empty, or the platform " +
                     "discovered by Oryx does not have any available or supported runtime images. Please view the following " +
                     "document for more information on runtimes supported by Oryx: " +
-                    "https://github.com/microsoft/Oryx/blob/main/doc/supportedRuntimeVersions.md";
-                var exc = new InvalidDockerfileImageException(message);
-                this.logger.LogError(exc, message);
-                throw exc;
+                    "https://github.com/microsoft/Oryx/blob/main/doc/supportedRuntimeVersions.md");
             }
 
             if (string.IsNullOrEmpty(dockerfileProperties.RuntimeImageTag))
             {
-                var message = "Either the value provided to the --runtime-platform-version argument is empty, or the version " +
+                this.LogInvalidDockerfileImageException(
+                    "Either the value provided to the --runtime-platform-version argument is empty, or the version " +
                     "discovered by Oryx is not available or supported for the given platform. Please view the following " +
                     "document for more information on runtime versions supported by Oryx: " +
-                    "https://github.com/microsoft/Oryx/blob/main/doc/supportedRuntimeVersions.md";
-                var exc = new InvalidDockerfileImageException(message);
-                this.logger.LogError(exc, message);
-                throw exc;
+                    "https://github.com/microsoft/Oryx/blob/main/doc/supportedRuntimeVersions.md");
             }
+        }
+
+        private void LogInvalidDockerfileImageException(string message)
+        {
+            var exc = new InvalidDockerfileImageException(message);
+            this.logger.LogError(exc, message);
+            throw exc;
         }
     }
 }
--- a/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
+++ b/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
@@ -266,7 +266,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator
         {
             if (string.IsNullOrEmpty(dockerfileProperties.BuildImageName))
             {
-                this.LogInvalidDockerfileImageException(
+                this.LogAndThrowInvalidDockerfileImageException(
                     "Provided build image name parsed from the provided --build-image argument is empty. " +
                     "Please provide a valid value for --build-image, or remove the argument to use the default " +
                     "'cli:stable' build image from mcr.microsoft.com/oryx.");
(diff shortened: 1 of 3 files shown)


src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
```diff
@@ -255,5 +261,50 @@ private string ConvertToRuntimeVersion(string platformName, string platformVersi
         {
             return this.platformDetector.GetCompatiblePlatforms(ctx);
         }
+
+        private void ValidateDockerfileImages(DockerfileProperties dockerfileProperties)
+        {
+            if (string.IsNullOrEmpty(dockerfileProperties.BuildImageName))
+            {
+                var message = "Provided build image name parsed from the provided --build-image argument is empty. " +
+                    "Please provide a valid value for --build-image, or remove the argument and use the default " +
```
william-msft, line 270: Should this read "...remove the argument **to** use the default..." instead of "...remove the argument **and** use the default..."

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
```diff
@@ -255,5 +261,50 @@ private string ConvertToRuntimeVersion(string platformName, string platformVersi
...
+            }
+
+            if (string.IsNullOrEmpty(dockerfileProperties.BuildImageTag))
+            {
+                var message = "Provided build image tag parsed from the provided --build-image argument is empty. " +
+                    "Please provide a valid value for --build-image, or remove the argument and use the default " +
+                    "'cli:stable' build image from mcr.microsoft.com/oryx.";
+                var exc = new InvalidDockerfileImageException(message);
+                this.logger.LogError(exc, message);
+                throw exc;
```
william-msft, line 284: Nit: Wondering if it's worth moving this block into:
```
if (string.IsNullOrEmpty(message))
{
  var exc = new InvalidDockerfileImageException(message);
  this.logger.LogError(exc, message);
  throw exc;
}

**Review by william-msft — DISMISSED** [review]

**Review by william-msft — DISMISSED** [review]

**Comment by daniv-msft** [comment]
Could you please clarify if this will have any impact on the scenarios where we legitimately support Build but not a Runtime for a language? I didn't review the code yet, so I might update this comment if the answer is in it.

File: src/BuildScriptGenerator/DefaultDockerfileGenerator.cs [hunk 3]
```diff
@@ -255,5 +261,49 @@ private string ConvertToRuntimeVersion(string platformName, string platformVersi
         {
             return this.platformDetector.GetCompatiblePlatforms(ctx);
         }
+
+        private void ValidateDockerfileImages(DockerfileProperties dockerfileProperties)
+        {
+            if (string.IsNullOrEmpty(dockerfileProperties.BuildImageName))
+            {
+                this.LogInvalidDockerfileImageException(
+                    "Provided build image name parsed from the provided --build-image argument is empty. " +
```
daniv-msft, line 270: Nit: should we mention something about the format expected? I imagine that it's kind of implied from the `cli:stable` example, but I don't know if some customers could be confused about this. If we have a meaningful comment in the help of the --build-image argument, please feel free to ignore.

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
```diff
@@ -255,5 +261,49 @@ private string ConvertToRuntimeVersion(string platformName, string platformVersi
...
+                    "'cli:stable' build image from mcr.microsoft.com/oryx.");
+            }
+
+            if (string.IsNullOrEmpty(dockerfileProperties.RuntimeImageName))
+            {
+                this.LogInvalidDockerfileImageException(
+                    "Either the value provided to the --runtime-platform argument is empty, or the platform " +
+                    "discovered by Oryx does not have any available or supported runtime images. Please view the following " +
+                    "document for more information on runtimes supported by Oryx: " +
+                    "https://github.com/microsoft/Oryx/blob/main/doc/supportedRuntimeVersions.md");
```
daniv-msft, line 289: Nit: please consider using an aka.ms link instead, so that if in the future the location of this doc changes, all the older images don't redirect to a wrong URL.

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
```diff
@@ -255,5 +261,49 @@ private string ConvertToRuntimeVersion(string platformName, string platformVersi
...
+            }
+
+            if (string.IsNullOrEmpty(dockerfileProperties.RuntimeImageTag))
+            {
+                this.LogInvalidDockerfileImageException(
```
daniv-msft, line 294: Could you please confirm that we should see these exceptions in our logs, and be able to measure how many customers got it?

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
```diff
@@ -255,5 +261,49 @@ private string ConvertToRuntimeVersion(string platformName, string platformVersi
...
+                    "Either the value provided to the --runtime-platform-version argument is empty, or the version " +
+                    "discovered by Oryx is not available or supported for the given platform. Please view the following " +
+                    "document for more information on runtime versions supported by Oryx: " +
+                    "https://github.com/microsoft/Oryx/blob/main/doc/supportedRuntimeVersions.md");
+            }
+        }
+
+        private void LogInvalidDockerfileImageException(string message)
```
daniv-msft, line 302: Please consider renaming to `LogAndThrowInvalidDockerfileImageException` or something: I thought this was just logging and that the flow would continue.

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
daniv-msft: OK, I see below that we do log these before throwing.

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
cormacpayne: The help for the `--build-image` argument is currently the following:

> _The mcr.microsoft.com/oryx build image to use in the dockerfile, provided in the format '<image>:<tag>'. If no value is provided, the 'cli:stable' Oryx image will be used._

I'll skip this comment since the documentation is already in place to specify the format, but I don't mind calling it out again in the exception message if necessary.

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
cormacpayne: Good call -- https://aka.ms/oryx-runtime-images

src/BuildScriptGenerator/DefaultDockerfileGenerator.cs
daniv-msft: OK sounds good, thanks!

daniv-msft (approved):
Looks good to me! The only open question I have is my remaining comment on the potential impact for scenarios where no runtime is legitimate.

cormacpayne:
> Could you please clarify if this will have any impact on the scenarios where we legitimately support Build but not a Runtime for a language? I didn't review the code yet, so I might update this comment if the answer is in it.

@daniv-msft this change only impacts users calling the `oryx dockerfile` command, with the intention that the Dockerfile they're attempting to produce from that command can both build _and_ run their application, and if the application they've provided doesn't have a corresponding runtime, the Dockerfile we are producing isn't meaningful (_i.e._, it wouldn't build).

So in this case, rather than producing an unusable Dockerfile, we're throwing a meaningful exception on why we couldn't produce a Dockerfile that could both build _and_ run their application.

I think that answers your question, but please let me know otherwise!

daniv-msft:
> > Could you please clarify if this will have any impact on the scenarios where we legitimately support Build but not a Runtime for a language? I didn't review the code yet, so I might update this comment if the answer is in it.
> 
> @daniv-msft this change only impacts users calling the `oryx dockerfile` command, with the intention that the Dockerfile they're attempting to produce from that command can both build _and_ run their application, and if the application they've provided doesn't have a corresponding runtime, the Dockerfile we are producing isn't meaningful (_i.e._, it wouldn't build).
> 
> So in this case, rather than producing an unusable Dockerfile, we're throwing a meaningful exception on why we couldn't produce a Dockerfile that could both build _and_ run their application.
> 
> I think that answers your question, but please let me know otherwise!

Thanks, it makes sense. The only extra question I would have is that, if I remember correctly, in the Github Action/ADO Task scenarios we only call `oryx dockerfile` as a way to detect the platform to use, but we don't actually use the resulting Dockerfile. Is that correct or did I misunderstand the use here?

cormacpayne:
> Thanks, it makes sense. The only extra question I would have is that, if I remember correctly, in the Github Action/ADO Task scenarios we only call `oryx dockerfile` as a way to detect the platform to use, but we don't actually use the resulting Dockerfile. Is that correct or did I misunderstand the use here?

That's correct -- the Dockerfile we produce has the following first line:

```
ARG RUNTIME=<platform>:<version>
```

This is what we append to `mcr.microsoft.com/oryx/` to denote the runtime image we're running the built binaries in for this image. When a user provides an application that didn't have a supported runtime, the first line we'd produce in the Dockerfile would appear as the following:

```
ARG RUNTIME=:
```

As you can imagine, this caused some issues 😅 Now we'll be failing on this command and giving users an error that is actionable.

PR #1744, opened 2022-12-14
Snehapar/upgrade python versions

snehapar9:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as \#\<issue\_number\>.
[User Story 1696265](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1696265): [Due 2023-01-03] Update python versions with security fixes
- [x] Tests are included and/or updated for code changes.
- [<del>] Proper license headers are included in each file.

PR #1850, opened 2023-02-01
Revert "Snehapar/upgrade python versions"

william-msft:
Reverts microsoft/Oryx#1744

Our partners are still seeing Python.h errors so we are reverting the offending versions:
image

We will also be manually removing the offending versions from our storage accounts.


tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
```diff
@@ -946,7 +946,7 @@ public void BuildPythonApps_AndHasLzmaModule(string appName)
             var appOutputDir = "/tmp/app1-output";
             var script = new ShellScriptBuilder()
                 .AddDefaultTestEnvironmentVariables()
-                .AddBuildCommand($"{appDir} -o {appOutputDir} --platform python --platform-version {PythonVersions.Python310Version}")
+                .AddBuildCommand($"{appDir} -o {appOutputDir} --platform python --platform-version 3.10.8")
```
snehapar9, line 949: Think we can keep this change to use {PythonVersions.Python310Version} as-is since we are rolling back to version 3.10.8 anyway.

william-msft: Updated

snehapar9 (dismissed):

tests/Oryx.BuildImage.Tests/Python/PythonDynamicInstallationTest.cs
```diff
@@ -197,8 +197,8 @@ public void PipelineTestInvocationCliBuster()
         }
 
         [Theory, Trait("category", "jamstack")]
-        [InlineData(PythonVersions.Python310Version)]
-        [InlineData(PythonVersions.Python311Version)]
+        [InlineData("3.10.8")]
```
snehapar9, line 200: Same Comment here. We should be okay leaving PythonVersions.Python310Version and PythonVersions.Python311Version as-is.

william-msft: Updated

snehapar9 (approved):

PR #1858, opened 2023-02-03
Add 'oryx telemetry' command to send generic events to App Insights

cormacpayne:
Introduces `oryx telemetry` command that will allow us to send events to App Insights from sources such as our GitHub Action and ADO Task.

### DIFF
--- a/src/BuildScriptGeneratorCli/Commands/TelemetryCommand.cs
+++ b/src/BuildScriptGeneratorCli/Commands/TelemetryCommand.cs
@@ -1,0 +1,81 @@
+﻿// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using McMaster.Extensions.CommandLineUtils;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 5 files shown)


src/BuildScriptGeneratorCli/Commands/TelemetryCommand.cs
```diff
@@ -0,0 +1,81 @@
...
+        [Option(
+            OptionTemplates.ProcessingTime,
+            CommandOptionType.SingleValue,
+            Description = "The processing time of the event, in milliseconds")]
+        public double ProcessingTime { get; set; }
+
+        [Option(
+            OptionTemplates.Property,
+            CommandOptionType.MultipleValue,
+            Description = "Additional information that should be logged with the event, in the form --property KEY=VALUE.")]
```
daniv-msft, line 35: So just curious: in practice this would be used as `oryx telemetry XYZ --property A=1 --property B=2 --property C=3`?

cormacpayne: That's correct -- it gives us a little more flexibility when logging events to provide `customDimensions` without needing to represent them as their own arguments for the command.

daniv-msft: Thanks!

daniv-msft (approved):

PR #1865, opened 2023-02-07
Williamhe/Python.h error workaround

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1735613
  - Add LIBRARY_PATH and CPLUS_INCLUDE_PATH environment variables for Python.h error workaround
  - Added more modules to Django sample app requirements.txt which caused failures
  - Plan is to roll-out 3.11.1 first
   - Tested with locally built image, pushed to my private ACR repo against the failing versions: https://github.com/william-msft/acr-test/actions/runs/4116911704/jobs/7107565340
   - Added an E2E test for python 3.11.1
   - Failing versions: 
image


images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -37,6 +37,18 @@ RUN set -ex \
         swig \
         tk-dev \
         uuid-dev \
+        python3-dev \
```
william-msft, line 40: Adding additional libraries to support all optional python modules.  
image
https://devguide.python.org/getting-started/setup-building/index.html#install-dependencies

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -97,7 +100,7 @@ public virtual void InstallPlatformSpecificSkeletonDependencies(StringBuilder st
             var snippet = new StringBuilder();
             snippet
                 .AppendLine()
-                .AppendLine($"if grep -q cli \"/opt/oryx/.imagetype\"; then")
+                .AppendLine($"if grep -q -e cli \"/opt/oryx/.imagetype\"; then")
```
snehapar9, line 103: We need to add -e githubactions \"/opt/oryx/.imagetype\ if we are enabling dynamic installation for github actions images as well. If we are not, we can remove -e. Essentially, -e option is used if we need to match multiple patterns. -e is added for each pattern.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -37,6 +37,18 @@ RUN set -ex \
...
+        lcov \
+        pkg-config \
+        libgdbm-dev \
+        libgdbm-compat-dev \
+        liblzma-dev \
+        libreadline6-dev \
+        lzma \
+        lzma-dev \
+        zlib1g-dev \
     && rm -rf /var/lib/apt/lists/*
```
snehapar9, line 52: Double confirming if we need to add these for other images as well considering the app services team only uses the github-actions image?

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -195,17 +198,6 @@ public virtual void InstallPlatformSpecificSkeletonDependencies(StringBuilder st
...
-                .AppendLine($"PIP_GCC_FLAGS=\"\"")
-                .AppendLine($"if [ \"$platformName\" = \"python\" ]; then")
-                .AppendLine($"  if [[ '{version}' == 3.10* ]] || [[ '{version}' == 3.11* ]]; then")
-                .AppendLine($"    echo \"Enabling --global-option to pip for gcc modules...\"")
-                .AppendLine($"    pythonMajorMinorVersion={version.Substring(0, version.LastIndexOf('.'))}")
-                .AppendLine($"    dynamicInstallationPythonIncludeDirectory={versionDirInTemp}/include/python$pythonMajorMinorVersion")
-                .AppendLine($"    PIP_GCC_FLAGS=\"--global-option=build_ext --global-option=-I$dynamicInstallationPythonIncludeDirectory\"")
-                .AppendLine($"  fi")
-                .AppendLine("fi")
-
```
snehapar9, line 208: Can we keep this for cli and add a condition to skip this if the image is github-actions?

images/build/Dockerfiles/ltsVersions.Dockerfile
```diff
@@ -37,6 +37,19 @@ RUN LANG="C.UTF-8" \
...
+        gdb \
+        lcov \
+        pkg-config \
+        libgdbm-dev \
+        libgdbm-compat-dev \
+        liblzma-dev \
+        libreadline6-dev \
+        lzma \
+        lzma-dev \
+        zlib1g-dev \
```
snehapar9, line 52: Same comment as above.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
cormacpayne: Would you mind adding a comment with this link above the lines where we're installing the Python 3 dependencies?

images/build/benv.sh
```diff
@@ -217,6 +217,8 @@ benv-resolve() {
     # successfully compiled and linked.
     export LIBRARY_PATH="$platformDir/lib:$LIBRARY_PATH"
     export LD_LIBRARY_PATH="$platformDir/lib:$LD_LIBRARY_PATH"
+    export C_INCLUDE_PATH="$platformDir/include/python$majorAndMinorParts"
```
cormacpayne, line 220: Would you mind including a comment above these two environment variables that elaborate on their purpose, similar to what we're doing above with `LIBRARY_PATH` and `LD_LIBRARY_PATH`?

cormacpayne (dismissed):
:shipit:

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -53,7 +53,10 @@ public static void InstallPythonToolingAndLanguage(StringBuilder stringBuilder)
                 "swig",
                 "tk-dev",
                 "unixodbc-dev",
-                "uuid-dev");
+                "uuid-dev",
+                "libffi-dev",
+                "python3-dev",
+                "python3-cffi");
```
pauld-msft, line 59: don't we also want to dynamically install the same packages when we build python on the CLI image?

src/BuildScriptGenerator/PlatformInstallerBase.cs
pauld-msft: yah i don't think we want to dynamicall install packages for github actions as well, so removing `-e` should be the way to go here

src/BuildScriptGenerator/PlatformInstallerBase.cs
pauld-msft: @snehapar9 Why do we want this for the CLI image? This entire section is no longer needed since we added the environment variables to `benv.sh`, right?

snehapar9: Okay. That makes sense.

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
william-msft: This depends on whether we don't want to add the 99MB to the other images. If don't add them, we'll have to continue adding this case-by-case. Did you have concerns this would increase the Jamstack image download performance?

william-msft: This depends on whether we don't want to add the 99MB to the other images. If don't add them, we'll have to continue adding this case-by-case. Did you have concerns this would increase the other image download performance?

snehapar9: Yep, I thought this may increase the size of the other images that App service does not consume but since it is only ~100 MB, it _should not_ increase the time to pull down the image significantly.

src/BuildScriptGenerator/PlatformInstallerBase.cs
william-msft: Good catch! I've removed `-e`

images/build/benv.sh
william-msft: Yup! Updated

build/__pythonVersions.sh
```diff
@@ -5,7 +5,7 @@ PIP_VERSION='21.2.4'
 PYTHON27_VERSION='2.7.18'
 PYTHON36_VERSION='3.6.15'
 PYTHON37_VERSION='3.7.15'
-PYTHON38_VERSION='3.8.16'
+PYTHON38_VERSION='3.8.15'
```
william-msft, line 8: Change after running `./build/generateConstants.sh` since the script was not ran during revert

snehapar9 (dismissed):

platforms/python/versions/bullseye/versionsToBuild.txt
```diff
@@ -6,6 +6,7 @@
 3.8.6, E3FF2839C048B25C084DEBE9B26995E310250568,
 3.8.12, E3FF2839C048B25C084DEBE9B26995E310250568,
 3.8.15, E3FF2839C048B25C084DEBE9B26995E310250568,
+3.8.16, E3FF2839C048B25C084DEBE9B26995E310250568,
```
pauld-msft, line 9: Curious why we are adding 3.8.16 back? Weren't we just going to release 3.11.1?

tests/Oryx.Integration.Tests/Python/PythonEndToEndTests.cs
```diff
@@ -66,6 +66,52 @@ public async Task CanBuildAndRun_Tweeter3AppAsync()
                 });
         }
 
+        [Fact]
+        [Trait("category", "python-3.11")]
```
pauld-msft, line 70: For the sake of getting something out quickly i'm fine with leaving this for a later date, but could we perhaps parameterize this testing method and invoke the django regex tests for the other python versions that had this same bug? E.g. adding some parameters to this method to have it invoked in another method like

```
[Fact]
[Trait("category", "python-3.10")]
[Trait("build-image", "github-actions-debian-bullseye")]
public async Task CanBuildAndRun_DjangoRegex_310()
{
    await this.CanBuildAndRun_DjangoRegex(PythonVersions.Python310Version, "3.10");
}
```

pauld-msft (approved):

platforms/python/versions/bullseye/versionsToBuild.txt
william-msft: 3.8.16 was not causing issue for our parnters:
image

tests/Oryx.Integration.Tests/Python/PythonEndToEndTests.cs
william-msft: Sure thing!

PR #1869, opened 2023-02-07
Create extension methods for telemetry client

snehapar9:
[User Story 1612862](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1612862): Create a new env variable for Connection String for AppInsights


tests/BuildScriptGenerator.Tests/DefaultBuildScriptGeneratorTest.cs
```diff
@@ -635,7 +635,8 @@ public void Checkers_DontFailTheBuild_WhenTheyThrow()
                     Options.Create(commonOptions)),
                 checkers,
                 NullLogger<DefaultBuildScriptGenerator>.Instance,
-                new DefaultStandardOutputWriter());
+                new DefaultStandardOutputWriter(),
+                null);
```
cormacpayne, line 639: We'll run into issues if we provide `null` wherever a method needs `TelemetryClient` -- it's probably easiest to create a `TestTelemetryClient` class in the `Oryx.Tests.Common` project that inherits from `TelemetryClient` and then we override its logging methods that we use throughout the repo (specifically within the methods of the extensions class) to do nothing so that we're not passing in a `null` object, but we don't actually do anything with the client.

src/BuildScriptGeneratorCli/ServiceProviderBuilder.cs
```diff
@@ -26,22 +23,22 @@ internal class ServiceProviderBuilder
...
             this.serviceCollection = new ServiceCollection();
+            var connectionString = string.Empty;
             this.serviceCollection
                 .AddBuildScriptGeneratorServices()
                 .AddCliServices(console)
                 .AddLogging(builder =>
                 {
+                    builder.AddApplicationInsights(
+                         configureTelemetryConfiguration: (config) => config.ConnectionString = connectionString,
+                         configureApplicationInsightsLoggerOptions: (options) => { options.FlushOnDispose });
```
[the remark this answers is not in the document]
snehapar9: Looks like the default value of `options.FlushOnDispose` is true.

src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
```diff
@@ -22,21 +22,19 @@
...
     <PackageReference Include="Microsoft.Extensions.Logging" Version="3.1.32" />
+    <PackageReference Include="Microsoft.Extensions.Logging.ApplicationInsights" Version="2.21.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="3.1.32" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="3.1.32" />
     <PackageReference Include="Microsoft.VisualStudioEng.MicroBuild.Core" Version="1.0.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="NLog" Version="5.1.1" />
-    <PackageReference Include="NLog.Extensions.Logging" Version="5.2.1" />
```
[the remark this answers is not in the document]
snehapar9: Great Catch! Updated.

src/BuildScriptGenerator.Common/Extensions/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System.Collections.Generic;
```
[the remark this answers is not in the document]
snehapar9: Updated.

tests/BuildScriptGenerator.Tests/DefaultBuildScriptGeneratorTest.cs
snehapar9: We are unable to clear a test class that inherits from TelemetryClient because TelemetryClient is a sealed class.
I created a mock class that returns a telemetry client object instead.

src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
```diff
@@ -22,32 +22,28 @@
...
+    <PackageReference Include="Moq" Version="4.18.4" />
```
cormacpayne, line 38: Do you know why `Moq` was added to this project? It should only be needed in test projects

tests/BuildScriptGenerator.Tests/DefaultBuildScriptGeneratorTest.cs
```diff
@@ -617,7 +620,9 @@ public void Checkers_DontFailTheBuild_WhenTheyThrow()
             commonOptions = commonOptions ?? new BuildScriptGeneratorOptions();
             commonOptions.SourceDir = "/app";
             commonOptions.DestinationDir = "/output";
-
+            var telemetryClientMock = new Mock<TelemetryClientMock>();
```
cormacpayne, line 623: Is there a reason we can't just do `Mock<TelemetryClient>` here?

images/build/Dockerfiles/full.Dockerfile
```diff
@@ -21,7 +21,8 @@ RUN ./build.sh golang     /opt/startupcmdgen/golang
 FROM buildpack-deps:${DEBIAN_FLAVOR}-curl
 ARG DEBIAN_FLAVOR
 ARG SDK_STORAGE_BASE_URL_VALUE="https://oryx-cdn.microsoft.io"
-ARG AI_KEY
+ARG AI_CONNECTION_STRING
+ENV ORYX_AI_CONNECTION_STRING=${AI_CONNECTION_STRING}
```
cormacpayne, line 25: There's a lingering `ORYX_AI_INSTRUMENTATION_KEY="${AI_KEY}"` environment variable being set at the bottom of this file

src/BuildScriptGeneratorCli/BuildScriptGeneratorCli.csproj
snehapar9: That is a good point. Removed it.

src/BuildScriptGenerator.Common/Extensions/TelemetryClientExtensions.cs
```diff
@@ -0,0 +1,71 @@
+﻿// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
```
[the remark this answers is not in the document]
snehapar9: Updated.

daniv-msft:
Just to be 100% sure, we do not expect this to have any impact on the existing telemetry, right? Basically only the authentication change, but the telemetry format or anything isn't going to change, right?

images/build/Dockerfiles/full.Dockerfile
```diff
@@ -21,7 +21,8 @@ RUN ./build.sh golang     /opt/startupcmdgen/golang
 FROM buildpack-deps:${DEBIAN_FLAVOR}-curl
 ARG DEBIAN_FLAVOR
 ARG SDK_STORAGE_BASE_URL_VALUE="https://oryx-cdn.microsoft.io"
-ARG AI_KEY
+ARG AI_CONNECTION_STRING
```
daniv-msft, line 24: I couldn't find any preexisting reference to `AI_CONNECTION_STRING` or `ORYX_AI_CONNECTION_STRING`. Could you please clarify why we wouldn't directly call the arg ORYX_AI_CONNECTION_STRING? Is it more consistent with the rest of similar arguments?

tests/BuildScriptGenerator.Tests/DefaultBuildScriptGeneratorTest.cs
snehapar9: Syced with @Cormac offline. Since, `TelemetryClient `is a sealed class, I created a static helper class `TelemetryClientHelper` that would return a telemetry client object when the `GetTelemetryClient` method is called.

src/BuildScriptGeneratorCli/Commands/DockerfileCommand.cs
```diff
@@ -90,37 +90,33 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
...
+                SourceRepo = sourceRepo,
+            };
 
-                var dockerfile = serviceProvider.GetRequiredService<IDockerfileGenerator>().GenerateDockerfile(ctx);
-                if (string.IsNullOrEmpty(dockerfile))
+            var dockerfile = serviceProvider.GetRequiredService<IDockerfileGenerator>().GenerateDockerfile(ctx);
+            if (string.IsNullOrEmpty(dockerfile))
+            {
+                exitCode = ProcessConstants.ExitFailure;
+                console.WriteErrorLine("Couldn't generate dockerfile.");
```
daniv-msft, line 105: Nit: please consider to directly return from this `if`, so that we can avoid the extra nesting coming from the `else` block below.

src/BuildScriptGeneratorCli/ServiceProviderBuilder.cs
```diff
@@ -26,22 +23,31 @@ internal class ServiceProviderBuilder
 
         public ServiceProviderBuilder(string logFilePath = null, IConsole console = null)
         {
-            LogManager.Configuration = BuildNLogConfiguration(logFilePath);
-            LogManager.ReconfigExistingLoggers();
+            var disableTelemetryEnvVariableValue = Environment.GetEnvironmentVariable(
+               LoggingConstants.OryxDisableTelemetryEnvironmentVariableName);
+            _ = bool.TryParse(disableTelemetryEnvVariableValue, out bool disableTelemetry);
+            var config = new TelemetryConfiguration();
+            var aiKey = disableTelemetry ? string.Empty : Environment.GetEnvironmentVariable(
```
daniv-msft, line 30: Would it be possible to use `null` if the telemetry is disabled, rather than `string.Empty`? It feels like `null` would better represent what we mean, but maybe there are some challenges coming with it

daniv-msft: Actually do we even need this extra variable? could we just replace this with:
```
if (!disableTelemetry)
{
    config.ConnectionString = Environment.GetEnvironmentVariable(LoggingConstants.ApplicationInsightsConnectionStringKeyEnvironmentVariableName);
}
```

snehapar9:
> Just to be 100% sure, we do not expect this to have any impact on the existing telemetry, right? Basically only the authentication change, but the telemetry format or anything isn't going to change, right?

Yep, that is correct. We should not see any impact on telemetry format. I verified by spinning up a container with a new image with these changes and was able to see the same telemetry we see now.

src/BuildScriptGeneratorCli/ServiceProviderBuilder.cs
snehapar9: With the new changes I checked in, looks like we need an additional variable for aiKey to check if we need to send telemetry to AI.
The method `AddApplicationInsights` throws an error if Connection String is empty or null.

snehapar9: > Would it be possible to use `null` if the telemetry is disabled, rather than `string.Empty`? It feels like `null` would better represent what we mean, but maybe there are some challenges coming with it

Makes sense! I do not see any issue if we use null instead of string.Empty here.

images/build/Dockerfiles/full.Dockerfile
snehapar9: Used `ORYX_AI_CONNECTION_STRING` and `ORYX_AI_CONNECTION_STRING` in the same manner `ORYX_AI_INSTRUMENTATION_KEY` and `AI_KEY` are used throughout the repo now just to be consistent.

src/BuildScriptGeneratorCli/ServiceProviderBuilder.cs
```diff
@@ -65,24 +86,6 @@ public IServiceProvider Build()
         private static LoggingConfiguration BuildNLogConfiguration([CanBeNull] string logPath)
         {
             var config = new LoggingConfiguration();
```
snehapar9, line 88: Added this method back here so we can continue to use NLog to write telemetry to a .txt file. 
Spent some time experimenting with Serilog and Telemetry Channel (can be found [here](https://learn.microsoft.com/en-us/azure/azure-monitor/app/telemetry-channels#configuration-in-code-for-aspnet-applications)). The latter proved to be not that useful because it only provides a temporary storage for logs that were not sent to AI.
With Serilog however, we need to make some tweaks to make sure the log format doesn't change.
Since we need a third party package anyway for this use case, I think it is best we stick to NLog to write log telemetry to a text file. I was also able to verify that there is no change to the log format if continue to use NLog.

src/BuildScriptGeneratorCli/ServiceProviderBuilder.cs
```diff
@@ -29,19 +31,38 @@ public ServiceProviderBuilder(string logFilePath = null, IConsole console = null
...
             this.serviceCollection = new ServiceCollection();
             this.serviceCollection
                 .AddBuildScriptGeneratorServices()
                 .AddCliServices(console)
                 .AddLogging(builder =>
                 {
+                    if (!string.IsNullOrWhitespace(aiKey))
+                    {
+                        builder.AddApplicationInsights(
+                            configureTelemetryConfiguration: (c) => c.ConnectionString = aiKey,
```
snehapar9, line 54: We need to add this here to correctly send telemetry to AI.

src/BuildScriptGeneratorCli/ServiceProviderBuilder.cs
```diff
@@ -29,19 +31,38 @@ public ServiceProviderBuilder(string logFilePath = null, IConsole console = null
...
+
             this.serviceCollection = new ServiceCollection();
             this.serviceCollection
                 .AddBuildScriptGeneratorServices()
                 .AddCliServices(console)
                 .AddLogging(builder =>
                 {
+                    if (!string.IsNullOrWhiteSpace(aiKey))
+                    {
+                        builder.AddApplicationInsights(
```
cormacpayne, line 53: If `aiKey` is `null`, then we don't run `AddApplicationInsights` -- do you know if this affects the `ILogger` object we try to retrieve via dependency injection?

tests/Oryx.Tests.Common/TestConstants.cs
```diff
@@ -12,5 +14,8 @@ public static class TestConstants
         // Xunit trait values
         public const string Category = "Category";
         public const string Release = "Release";
+
+        //AI key 
+        public const string AiKey = "InstrumentationKey=test";
```
cormacpayne, line 19: nit: I'd rename this to `AppInsightsConnectionString` so we're more clear about what this string is

tests/Oryx.Tests.Common/TelemetryClientHelper.cs
```diff
@@ -0,0 +1,23 @@
...
+// --------------------------------------------------------------------------------------------
+
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.Extensibility;
+using Microsoft.Oryx.Tests.Common;
+using System;
+
+public static class TelemetryClientHelper
+{
+    public static string aiKey = TestConstants.AiKey;
```
cormacpayne, line 13: nit: I'd personally remove this `aiKey` private variable since we can directly access` TestConstants.AiKey`

tests/Oryx.Tests.Common/TelemetryClientHelper.cs
```diff
@@ -0,0 +1,23 @@
...
+    public static TelemetryClient GetTelemetryClient()
+    {
+        var telemetryConfig = new TelemetryConfiguration();
+        if (!string.IsNullOrWhiteSpace(aiKey))
+        {
+            telemetryConfig.ConnectionString = aiKey;
+        }
+        return new TelemetryClient(telemetryConfig);                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
+    }
```
cormacpayne, line 22: I think we should make this method a little more robust when it comes to where the connection string comes from -- specifically, it would be great if we could first check to see if an environment variable was set that we could use, and then fall back to `TestConstants.AiKey` if it wasn't set.

You'd want to add another value to `TestConstants` that is the environment variable we're reading, something like 

```cs
public const string AppInsightsConnectionStringEnvironmentVariable = "TEST_AI_CONNECTION_STRING";
```

Then we could have this method work as so:

```cs
public static TelemetryClient GetTelemetryClient()
{
    var connectionString = Environment.GetEnvironmentVariable(TestConstants.AppInsightsConnectionStringEnvironmentVariable)
        ?? TestConstants.AppInsightsConnectionString;
    var telemetryConfig = new TelemetryConfig()
    {
        ConnectionString = connectionString,
    };
    return new TelemetryClient(telemetryConfig);
}
```

cormacpayne (commented):

tests/Oryx.Tests.Common/TelemetryClientHelper.cs
snehapar9: Okay. That makes sense.

src/BuildScriptGeneratorCli/Commands/DockerfileCommand.cs
```diff
@@ -90,37 +90,33 @@ internal override int Execute(IServiceProvider serviceProvider, IConsole console
 
             int exitCode;
             var logger = serviceProvider.GetRequiredService<ILogger<DockerfileCommand>>();
-            using (var timedEvent = logger.LogTimedEvent("DockerFileCommand", buildEventProps))
```
pauld-msft, line 93: Why are we not replacing this timed event?

pauld-msft: Doesn't this mean that we would no longer be logging the dockerfile command to AI, and we would lose all of the buildEventProps data we have created here?

tests/BuildScriptGenerator.Tests/Hugo/HugoPlatformTest.cs
```diff
@@ -76,13 +80,12 @@ private BuildScriptGeneratorContext CreateContext()
...
             buildScriptGeneratorOptions = buildScriptGeneratorOptions ?? new BuildScriptGeneratorOptions();
             hugoScriptGeneratorOptions = hugoScriptGeneratorOptions ?? new HugoScriptGeneratorOptions();
-
             return new HugoPlatform(
                 Options.Create(buildScriptGeneratorOptions),
                 Options.Create(hugoScriptGeneratorOptions),
                 NullLogger<HugoPlatform>.Instance,
                 new HugoPlatformInstaller(Options.Create(buildScriptGeneratorOptions), NullLoggerFactory.Instance),
-                detector);
+                detector, TelemetryClientHelper.GetTelemetryClient());
```
pauld-msft, line 88: nit: newline here

tests/BuildScriptGenerator.Tests/Ruby/RubyPlatformTests.cs
```diff
@@ -372,7 +376,7 @@ public void DoesNotHaveRubyInstallScript_IfDynamicInstallNotEnabled_AndRubyVersi
                 versionProvider,
                 NullLogger<TestRubyPlatform>.Instance,
                 detector,
-                rubyInstaller);
+                rubyInstaller, TelemetryClientHelper.GetTelemetryClient());
```
pauld-msft, line 379: nit: newline

tests/Oryx.Integration.Tests/BuildConfigurationFile/BuildConfigurationReadYaml.cs
```diff
@@ -146,7 +148,7 @@ public void UsingBuildConfigurationFile_PostBuildOnly_Invalid_ReturnsTrue()
                     Options.Create(commonOptions)),
                 checkers,
                 NullLogger<DefaultBuildScriptGenerator>.Instance,
-                new DefaultStandardOutputWriter());
+                new DefaultStandardOutputWriter(), TelemetryClientHelper.GetTelemetryClient());
```
pauld-msft, line 151: nit: newline

src/BuildScriptGeneratorCli/Commands/DockerfileCommand.cs
snehapar9: We removed this timed event from DockerfileCommand because the dockerfile command is being logged [here](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGeneratorCli/Commands/CommandBase.cs#L85) anyway.
So, this is actually redundant.

tests/Oryx.RuntimeImage.Tests/Node/NodeRuntimeImageCanRunWhenAppInsightsModuleNotFound.cs
```diff
@@ -37,12 +37,12 @@ public async Task GeneratesScript_CanRun_AppInsightsModule_NotFoundAsync(string
             var volume = DockerVolume.CreateMirror(hostDir);
             var appDir = volume.ContainerDir;
             var imageName = _imageHelper.GetRuntimeImage("node", nodeVersion);
-            var aIKey = ExtVarNames.UserAppInsightsKeyEnv;
+            var aIKey = ExtVarNames.UserAppInsightsConnectionStringEnv;
```
pauld-msft, line 40: nit: this is no longer an `aIKey`, its now an aiConnectionString

src/BuildScriptGeneratorCli/Commands/DockerfileCommand.cs
pauld-msft: aren't we adding a bunch of valuable info with the [buildEventProps](https://github.com/microsoft/Oryx/blob/f9f754e9571c12b724e224ba9f14c6260abd4d8f/src/BuildScriptGeneratorCli/Commands/DockerfileCommand.cs#L82-L89) variable in this method that is currently not being used? 

If we are going to not log that info, then we should delete the buildEventProps object, but i personally feel like it is useful to have the inputs logged?

snehapar9: That is a valid point. So, we see more value logging these (would probably show up twice) instead of removing?
@Cormac, since we chatted about this before, do we agree on adding this back instead of removing it?

src/BuildScriptGeneratorCli/Commands/CommandBase.cs
```diff
@@ -252,10 +257,6 @@ private void DisposeServiceProvider()
             {
                 disposable.Dispose();
             }
-
-            // Sends queued messages to Application Insights
-            NLog.LogManager.Flush(LoggingConstants.FlushTimeout);
-            NLog.LogManager.Shutdown();
```
snehapar9, line 258: Do we need to add this back if we use NLog for writing logs to text file?

src/BuildScriptGeneratorCli/ServiceProviderBuilder.cs
snehapar9: Do not think this would affect the `ILogger` object. My understanding is that if the `aiKey` is null, the logs collected from ILogger would _not_ be submitted to AI because `AddApplicationInsights` would not run.
I quickly tested by assigning a null value to `aiKey` and spun up a container with this new image, was able to still successfully build the application. Any reason why you think this may affect the `ILogger` object?

cormacpayne: I wasn't sure if the `ILogger` that we were pulling from dependency injection was set up with the `.AddLogging` or `.AddApplicationInsights` call -- if it's set up with the latter, then my concern would be `aiKey` having a `null` or empty value and we avoid setting up the dependency injection for `ILogger` since `.AddApplicationInsights` is not called, and when we request an `ILogger` in a class constructor, there would be an exception thrown because DI doesn't know how to provide an object of that type.

If you tested it locally without providing a value for `aiKey` and no exceptions were thrown, then it seems like `ILogger` is set up in the `.AddLogging` call, which we're making no matter what, so we should be good here.

src/BuildScriptGeneratorCli/Commands/DockerfileCommand.cs
cormacpayne: @snehapar9 @pauld-msft we have an existing call in [`DefaultDockerfileGenerator`](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs#L51) that logs a timed event for "GenerateDockerfile", and as a part of it, we include most of the same information in the [`buildEventProps`](https://github.com/microsoft/Oryx/blob/main/src/BuildScriptGenerator/DefaultDockerfileGenerator.cs#L159-L166) that we were previously logging here. We lose out on the arguments provided to the command, but I think we'd generally be in favor of logging what the ultimate combination of build and runtime platforms and tags were generated for the Dockerfile that what was provided by the user. We also pull the platform arguments during this "GenerateDockerfile" timed event, so if we ever felt like we needed to log those, we can easily do so, but I personally don't think we need to at this moment.

cormacpayne (dismissed):
LGTM pending any additional comments from @pauld-msft :shipit:

pauld-msft (dismissed):
LGTM (once we remove un-used build props https://github.com/microsoft/Oryx/blob/f9f754e9571c12b724e224ba9f14c6260abd4d8f/src/BuildScriptGeneratorCli/Commands/DockerfileCommand.cs#L82-L89)

daniv-msft (dismissed):

daniv-msft (approved):

PR #1878, opened 2023-02-10
Wali/oryx private environment setup 1675864

waliMSFT:
-  This change will introduce the private environment (storage account) for Oryx. 
-  The new sdk staging account url: `https://oryxsdksstaging.blob.core.windows.net`
-  The associated Env Variable is required to access the private account should carry the sasToken: `ORYX-SDK-STAGING-PRIVATE-SAS-TOKEN`
-  Tests are included and/or updated for code changes.
-  Please contact with me @waliMSFT if you have any concerns.


build/__functions.sh
```diff
@@ -90,6 +90,41 @@ function shouldStageRuntimeVersion()
 			return 0
 		fi
 		;;
+	'python'
```
cormacpayne, line 93: Does there need to be a `)` after each string we're checking `$platformName` against? We do that above with

```bash
'dotnet'|'dotnetcore')
```

I would think we would want to do `python)`, `node)`, etc. for the values we're adding here.

images/runtime/dotnetcore/template.Dockerfile
```diff
@@ -41,6 +41,6 @@ RUN echo $USER_DOTNET_AI_VERSION && ln -s /opt/startupcmdgen/startupcmdgen /usr/
     && apt-get install unzip -y \ 
     && apt-get upgrade --assume-yes \
     && mkdir -p /DotNetCoreAgent \
-    && curl -o /DotNetCoreAgent/appinsights.zip "https://oryxsdksdev.blob.core.windows.net/appinsights-agent/DotNetCoreAgent.$USER_DOTNET_AI_VERSION.zip" \
+    && curl -o /DotNetCoreAgent/appinsights.zip "https://oryxsdksstaging.blob.core.windows.net/appinsights-agent/DotNetCoreAgent.$USER_DOTNET_AI_VERSION.zip" \
```
cormacpayne, line 44: Is this `oryxsdksstaging` value one that we can pass in as an `ARG` to this Dockerfile? That way we can set the value with `--build-arg` when we call `docker build` on this file and minimize the blast radius if we ever needed to change the name of the storage account in the future (or which storage account we wanted to use here if we needed to make it more dynamic)

cormacpayne: I see this `oryxsdksstaging` in a few other Dockerfiles in this PR, so this comment would apply to those files as well.

vsts/scripts/copyAllBlobsToProdStorageAccount.sh
```diff
@@ -81,6 +81,12 @@ if [ "$1" = $SANDBOX_SDK_STORAGE_BASE_URL ]; then
 elif [ "$1" = $DEV_SDK_STORAGE_BASE_URL ]; then
     sourceSdk=$DEV_SDK_STORAGE_BASE_URL
     sasToken=$DEV_STORAGE_SAS_TOKEN
+elif [ "$1" = $PRIVATE_STAGING_SDK_STORAGE_BASE_URL ]; then
+    sourceSdk=$PRIVATE_STAGING_SDK_STORAGE_BASE_URL
+    sasToken=$SDK_STAGING_PRIVATE_STORAGE_SAS_TOKEN
+elif [ "$1" = $PRIVATE_STAGING_SDK_STORAGE_BASE_URL ]; then
+    sourceSdk=$PRIVATE_STAGING_SDK_STORAGE_BASE_URL
+    sasToken=$SDK_STAGING_PRIVATE_STORAGE_SAS_TOKEN
```
cormacpayne, line 89: Duplicate code?

cormacpayne (commented):
Some initial comments to take a look at

build/constants.yaml
```diff
@@ -456,6 +458,20 @@
...
+      -
+    php-staging-runtime-versions:
+      -
+    hugo-staging-runtime-versions:
+      -
+    ruby-staging-runtime-versions:
+      -
+    golang-staging-runtime-versions:
+      -    
   outputs:
```
snehapar9, line 475: Could you please clarify why these constants do not have values assigned?

src/BuildScriptGenerator/StagingRuntimeConstants.cs
```diff
@@ -7,5 +7,12 @@ namespace Microsoft.Oryx.BuildScriptGenerator.Common
...
     {
         public static readonly List<string> DotnetcoreStagingRuntimeVersions = new List<string> { };
+        public static readonly List<string> PythonStagingRuntimeVersions = new List<string> { };
+        public static readonly List<string> NodeStagingRuntimeVersions = new List<string> { };
+        public static readonly List<string> JavaStagingRuntimeVersions = new List<string> { };
+        public static readonly List<string> PhpStagingRuntimeVersions = new List<string> { };
+        public static readonly List<string> HugoStagingRuntimeVersions = new List<string> { };
+        public static readonly List<string> RubyStagingRuntimeVersions = new List<string> { };
+        public static readonly List<string> GolangStagingRuntimeVersions = new List<string> { };
     }
```
snehapar9, line 17: Could you please clarify why we need empty lists here?

images/runtime/dotnetcore/template.Dockerfile
pauld-msft: Have you been able to run the validation and nightly tests on this and verify that this exists in `oryxsdksstaging`? Also yeah, agree w/ cormac about parametrizing

build/__functions.sh
pauld-msft: Yep. @waliMSFT, a good way that you can test a lot of these changes is by running the scripts to build runtime images and runtime bases locally for each language. e.g.
- `./build/buildRunTimeImages.sh dotnet buster`
- `./build/buildRunTimeImageBases.sh python bullseye`

images/__common.sh
```diff
@@ -12,12 +12,13 @@ function downloadFileAndVerifyChecksum() {
...
+    local sasToken="$5"
     local headersFile="/tmp/headers.txt"
 
     echo "Downloading $platformName version '$version' from '$sdkStorageAccountUrl'..."
     request="curl 
         -D $headersFile 
-        -SL $sdkStorageAccountUrl/$platformName/$downloadableFileName 
+        -SL $sdkStorageAccountUrl/$platformName/$downloadableFileName$sasToken
         --output $downloadedFileName"
     $__CURRENT_DIR/retry.sh "$request"
```
pauld-msft, line 23: `retry.sh` here has an `echo` whenever it fails to download, which could expose the `sasToken` when running in our pipelines. We should make sure that doesn't happen by removing the echo or adding some logic to only echo it in some cases

images/build/installDotNetCore.sh
```diff
@@ -10,10 +10,13 @@ CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 source $CURRENT_DIR/../__common.sh
 
 sdkStorageAccountUrl="$ORYX_SDK_STORAGE_BASE_URL"
+sasToken=""
 if [ -z "$sdkStorageAccountUrl" ]; then
-    sdkStorageAccountUrl=$DEV_SDK_STORAGE_BASE_URL
+    sdkStorageAccountUrl=$PRIVATE_STAGING_SDK_STORAGE_BASE_URL
+fi
+if [ sdkStorageAccountUrl = $PRIVATE_STAGING_SDK_STORAGE_BASE_URL ]; then
```
pauld-msft, line 17:
```suggestion
if [ $sdkStorageAccountUrl = $PRIVATE_STAGING_SDK_STORAGE_BASE_URL ]; then
```

images/installPlatform.sh
```diff
@@ -45,10 +45,13 @@ debianFlavor=$DEBIAN_FLAVOR
 fileName="$PLATFORM_NAME-$VERSION.tar.gz"
 
 sdkStorageAccountUrl="$ORYX_SDK_STORAGE_BASE_URL"
+sasToken=""
 if [ -z "$sdkStorageAccountUrl" ]; then
-  sdkStorageAccountUrl=$DEV_SDK_STORAGE_BASE_URL
+  sdkStorageAccountUrl=$PRIVATE_STAGING_SDK_STORAGE_BASE_URL
+fi
+if [ sdkStorageAccountUrl = $PRIVATE_STAGING_SDK_STORAGE_BASE_URL ]; then
```
pauld-msft, line 52:
```suggestion
if [ $sdkStorageAccountUrl = $PRIVATE_STAGING_SDK_STORAGE_BASE_URL ]; then
```

tests/Oryx.Integration.Tests/DevStorageAccountSanityTest.cs
```diff
@@ -23,7 +23,7 @@ public class DevStorageAccountSanityTest : StorageAccountSanityTestBase
                 output,
```
pauld-msft, line 23: I can't remember... are we using the dev storage account at all after all of this? If so, I would keep this test and create a new one called StagingStorageAccountSanityTest. If not, then maybe we should just rename this?  

I also would like to point out that we still have the default set to `SdkStorageConstants.DevSdkStorageBaseUrl` in this constructor.

vsts/scripts/publishSdksFromDevToProdStorageAccount.sh
```diff
@@ -42,11 +42,11 @@ function copyBlob() {
         echo "Blob '$blobName' exists in Prod storage container '$platformName'. Overwriting it..."
         if [ $dryRun == "False" ]; then
             "$azCopyDir/azcopy" copy \
-                "$DEV_SDK_STORAGE_BASE_URL/$platformName/$blobName$DEV_STORAGE_SAS_TOKEN" \
+                "$PRIVATE_STAGING_SDK_STORAGE_BASE_URL/$platformName/$blobName$SDK_STAGING_PRIVATE_STORAGE_SAS_TOKEN" \
```
pauld-msft, line 45: This file/script is no longer named accurately

pauld-msft (changes requested):
Left some comments as well.  

Would also love to see some examples of staging the runtime images working locally for each language (the image tags should be sent to a file like `artifacts/drop/images/runtime-images-acr.txt`)

build/constants.yaml
william-msft: I have the same question 🙂 My guess is to have them populated later? If so, we should be able to populate these when the value is available

images/__common.sh
william-msft: That's a great point Paul!

william-msft:
@waliMSFT gentle reminder to merge with `main`'s latest changes

build/__functions.sh
```diff
@@ -90,9 +90,60 @@ function shouldStageRuntimeVersion()
...
+	'ruby')
+		if [[ " ${RUBY_STAGING_RUNTIME_VERSIONS[*]} " =~ " ${platformRuntimeVersion} " ]]; then
+			return 0
+		fi
+		;;
+	'golang')
+		if [[ " ${GOLANG_STAGING_RUNTIME_VERSIONS[*]} " =~ " ${platformRuntimeVersion} " ]]; then
+			return 0
+		fi
+		;;
```
william-msft, line 127: We can clean this a bit:

```
declare -A PLATFORM_RUNTIME_VERSIONS=(
  ['dotnet']="${DOTNETCORE_STAGING_RUNTIME_VERSIONS[*]}"
  ['dotnetcore']="${DOTNETCORE_STAGING_RUNTIME_VERSIONS[*]}"
  ['python']="${PYTHON_STAGING_RUNTIME_VERSIONS[*]}"
  ['node']="${NODE_STAGING_RUNTIME_VERSIONS[*]}"
  ['java']="${JAVA_STAGING_RUNTIME_VERSIONS[*]}"
  ['php']="${PHP_STAGING_RUNTIME_VERSIONS[*]}"
  ['hugo']="${HUGO_STAGING_RUNTIME_VERSIONS[*]}"
  ['ruby']="${RUBY_STAGING_RUNTIME_VERSIONS[*]}"
  ['golang']="${GOLANG_STAGING_RUNTIME_VERSIONS[*]}"
)

if [[ " ${PLATFORM_RUNTIME_VERSIONS[$platformName]} " =~ " ${platformRuntimeVersion} " ]]; then
  return 0
else
  echo "Error message..."
  return 1
fi
```

src/BuildScriptGenerator/StagingRuntimeConstants.cs
snehapar9: Just figured these empty lists were propogated by the following constants defined in constants.yaml file. 

```
constants:
      -
    python-staging-runtime-versions:
      -
    node-staging-runtime-versions:
      -
    java-staging-runtime-versions:
      -
    php-staging-runtime-versions:
      -
    hugo-staging-runtime-versions:
      -
    ruby-staging-runtime-versions:
      -
    golang-staging-runtime-versions:
      -    
```

images/build/Dockerfiles/Dockerfile
```diff
@@ -110,5 +112,6 @@ RUN set -ex \
     && cd /opt/php \
     && ln -s 5.6 5 \
     && echo "full" > /opt/oryx/.imagetype \
-    && echo "DEBIAN|${DEBIAN_FLAVOR}" | tr '[a-z]' '[A-Z]' > /opt/oryx/.ostype
+    && echo "DEBIAN|${DEBIAN_FLAVOR}" | tr '[a-z]' '[A-Z]' > /opt/oryx/.ostype \
+    && export ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN=""
```
snehapar9, line 116: Was just curious if we are going to go ahead and try the alternative that Paul mentioned so we can get rid of this 
`export ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN=""` here?

pauld-msft: Feel free to watch the recording from yesterday for the more in depth explanation, but at a high level the changes I suggested are:
1. Update the `ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN` env var in the Dockerfile `RUN` scripts to be `ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN_PATH`. Update their value to be the path to the secret file instead of the token itself.
2. Update the scripts that use `ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN` to instead pull the token out the path env var into a temporary variable (or directly into the url)
3. remove the `&& export ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN=""` lines from the dockerfile, since we no longer save the token to an env var

build/buildBuildImages.sh
```diff
@@ -36,28 +36,32 @@ labelContent="git_commit=$GIT_COMMIT, build_number=$BUILD_NUMBER, release_tag_na
...
-		echo "Error: Unsupported flag $1" >&2
-		exit 1
-		;;
-		*) # preserve positional arguments
-		PARAMS="$PARAMS $1"
-		shift
-		;;
-	esac
+  case "$1" in
+    -token|--sas-token)
```
pauld-msft, line 40: - [ ] As discussed in the meeting yesterday, lets remove this and implement the logic to detect if being run locally, and if so running the script to load the token into a local variable w/o printing it

pauld-msft:
- [ ] As @cormacpayne mentioned in the meeting, lets add tests to make sure we don't propagate the token in any of the images

build/constants.yaml
snehapar9: Can we assign a `string.Empty` to these constants instead of leaving them empty?

images/__common.sh
william-msft: @waliMSFT did we get an answer for this?

cormacpayne: +1 to the above point

tests/Oryx.Integration.Tests/Database/DatabaseTestsBase.cs
```diff
@@ -81,6 +84,25 @@ protected DatabaseTestsBase(ITestOutputHelper outputHelper, Fixtures.DbContainer
...
+        {
+            List<EnvironmentVariable> envVariableList = _dbFixture.GetCredentialsAsEnvVars();
+            var testStorageAccountUrl = Environment.GetEnvironmentVariable(SdkStorageConstants.TestingSdkStorageUrlKeyName);
+            var sdkStorageUrl = string.IsNullOrEmpty(testStorageAccountUrl) ? SdkStorageConstants.PrivateStagingSdkStorageBaseUrl : testStorageAccountUrl;
+            
+            envVariableList.Add(new EnvironmentVariable(SdkStorageConstants.SdkStorageBaseUrlKeyName, sdkStorageUrl));
+
+            if (sdkStorageUrl == SdkStorageConstants.PrivateStagingSdkStorageBaseUrl)
+            {
+                string stagingStorageSasToken = Environment.GetEnvironmentVariable(SdkStorageConstants.PrivateStagingStorageSasTokenKey) != null
```
cormacpayne, line 97: nit: same comment from the other file about using `??` in this case where we're only checking if the environment variable was set or not

tests/Oryx.Integration.Tests/Database/DatabaseTestsBase.cs
```diff
@@ -93,5 +115,12 @@ protected void RunAsserts(Action action, string message)
                 throw;
             }
         }
+
+        protected string GetKeyvaultSecretValue(string keyvaultUri, string secretName)
```
cormacpayne, line 119: nit: same comment about Key Vault naming convention

tests/Oryx.Integration.Tests/DotNetCore/DotNetCorePreRunCommandOrScriptTest.cs
```diff
@@ -38,17 +38,16 @@ public async Task CanBuildAndRun_NetCore31MvcApp_UsingPreRunCommand_WithDynamicI
...
                 $"--platform dotnet --platform-version {runtimeVersion} -o {appOutputDir}")
                .ToString();
 
             // split run script to test pre-run command before running the app.
             var runtimeImageScript = new ShellScriptBuilder()
-                .AddDefaultTestEnvironmentVariables()
                 .SetEnvironmentVariable(FilePaths.PreRunCommandEnvVarName,
-                    $"\"touch {appOutputDir}/_test_file.txt\ntouch {appOutputDir}/_test_file_2.txt\"")
+                    $"\"touch {appOutputDir}/_test_file.txt\ntouch {appOutputDir}/_test_file_2.txt\"",
+                    true)
```
cormacpayne, line 50: nit: let's move this `true` to the end of the previous line to be consistent with what we're doing below

tests/Oryx.Integration.Tests/Fixtures/MySqlDbContainerFixture.cs
```diff
@@ -56,7 +56,7 @@ protected override void InsertSampleData()
         {
             const string sqlFile = "/tmp/mysql_setup.sql";
             var sqlQuery = GetSampleDataInsertionSql();
-            var dbSetupScript = new ShellScriptBuilder()
+            var dbSetupScript = new ShellScriptBuilder(null, false)
```
cormacpayne, line 59: nit: why does this constructor need the additional arguments now?

tests/Oryx.Integration.Tests/Node/NodeSqlServerIntegrationTests.cs
```diff
@@ -80,5 +83,30 @@ private async Task Run_NodeApp_MicrosoftSqlServerDBAsync(string imageTag)
                 });
         }
 
+        protected string GetKeyvaultSecretValue(string keyvaultUri, string secretName)
```
cormacpayne, line 86: nit: same comment about Key Vault naming convention

tests/Oryx.Integration.Tests/Node/NodeSqlServerIntegrationTests.cs
```diff
@@ -80,5 +83,30 @@ private async Task Run_NodeApp_MicrosoftSqlServerDBAsync(string imageTag)
...
+        {
+            var client = new SecretClient(new Uri(keyvaultUri), new DefaultAzureCredential());
+            var sasToken = client.GetSecret(secretName).Value.Value;
+            return sasToken;
+        }
```
cormacpayne, line 91: nit: let's separate the end of one method and the start of the next with a new line

tests/Oryx.Integration.Tests/Node/NodeSqlServerIntegrationTests.cs
```diff
@@ -80,5 +83,30 @@ private async Task Run_NodeApp_MicrosoftSqlServerDBAsync(string imageTag)
...
+        private List<EnvironmentVariable> getEnvironmentVariableList()
```
cormacpayne, line 92: nit: same comment in this file about using Pascal casing for method naming

build/__sdkStorageConstants.sh
```diff
@@ -3,6 +3,10 @@
 ENABLE_DYNAMIC_INSTALL_KEY='ENABLE_DYNAMIC_INSTALL'
 SDK_STORAGE_BASE_URL_KEY_NAME='ORYX_SDK_STORAGE_BASE_URL'
 TESTING_SDK_STORAGE_URL_KEY_NAME='ORYX_TEST_SDK_STORAGE_URL'
+PRIVATE_STAGING_STORAGE_SAS_TOKEN_KEY='ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN'
+ORYX_KEYVAULT_URI='https://oryx.vault.azure.net'
+STAGING_STORAGE_SAS_TOKEN_KEYVAULT_SECRET_NAME='ORYX-SDK-STAGING-PRIVATE-SAS-TOKEN'
+PRIVATE_STAGING_SDK_STORAGE_BASE_URL='https://oryxsdksstaging.blob.core.windows.net'
 DEV_SDK_STORAGE_BASE_URL='https://oryxsdksdev.blob.core.windows.net'
```
cormacpayne, line 10: Do we need to maintain this `DEV_SDK_STORAGE_BASE_URL` anymore in the repo if we're treating the staging storage account as the new dev?

cormacpayne (commented):
Mostly nit comments, but a few things worth calling out before merging

pauld-msft:
> > * [x]   As @cormacpayne mentioned in the meeting, lets add tests to make sure we don't propagate the token in any of the images
> 
> @pauld-msft I have added a work item regarding this. https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1818074 @daniv-msft

@waliMSFT - so these tests are supposed to be an automated way of ensuring that the token is secure (at least in the final images). Until we have these in place, or some manual verification, I'm not sure that we would be able to be certain the token is not exposed. So is the idea to just consider the storage account as public until then, or are you planning on manually verifying that they don't get passed through?

images/build/Dockerfiles/Dockerfile
pauld-msft: @waliMSFT - re-opening this conversation, as I cannot find these changes in the code

pauld-msft:
> * [x]  After the updates to ensure we _NEVER_ are handling the SAS token directly or are surfacing it in the pipeline, we need to rotate the SAS token in the key vault. As a rule of thumb, we never should have this token in a command line command, as commands are saved to the local disk in a history file. If this accidentally happens, or we accidentally expose it some other way, this token needs to be immediately rotated.

@waliMSFT can you just confirm that you rotated the token, since this task was checked off? Did you happen to keep track of the steps that you did to rotate it? I think I remember us discussing that it would be a good exercise to do it and keep track of the steps so that it is easy to do in the future

src/BuildScriptGenerator.Common/ShellScriptBuilder.cs
```diff
@@ -161,23 +178,38 @@ public ShellScriptBuilder AddStringDoesNotExistInFileCheck(string searchString,
...
         }
 
         public override string ToString()
         {
             return this.scriptBuilder.ToString();
         }
 
+        protected string GetKeyvaultSecretValue(string keyvaultUri, string secretName)
+        {
+            var client = new SecretClient(new Uri(keyvaultUri), new DefaultAzureCredential());
```
snehapar9, line 211: Nit: May need to rename this as well from `keyvaultUri` to `keyVaultUri`

src/BuildScriptGenerator/SdkStorageVersionProviderBase.cs
```diff
@@ -150,14 +149,15 @@ protected string GetPlatformBinariesStorageBaseUrl()
...
             {
-                throw new InvalidOperationException(
-                    $"Environment variable '{SdkStorageConstants.SdkStorageBaseUrlKeyName}' is required.");
+                platformBinariesStorageBaseUrl = Environment.GetEnvironmentVariable(SdkStorageConstants.TestingSdkStorageUrlKeyName);
+
+                // platformBinariesStorageBaseUrl = SdkStorageConstants.PrivateStagingSdkStorageBaseUrl;
             }
 
+            this.logger.LogDebug("Using the Sdk storage url {sdkStorageUrl}.", platformBinariesStorageBaseUrl);
+
```
snehapar9, line 158: Nit: Remove new line

tests/Oryx.BuildImage.Tests/BaseImage/BaseImageTest.cs
```diff
@@ -82,7 +82,6 @@ public BaseImageTest(ITestOutputHelper output) : base(output)
             var appOutputDir = "/tmp/output";
             var manifestFile = $"{appOutputDir}/{FilePaths.BuildManifestFileName}";
             var script = new ShellScriptBuilder()
-                .AddDefaultTestEnvironmentVariables()
```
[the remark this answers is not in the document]
snehapar9: +1. Is the intention to use the prod storage account here or the newly created private storage account?
Some tests may fail during validation/nightly pipelines if we take the route of using the prod storage account for these tests since version specific tarballs may not be released to prod yet.

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreSampleAppsTest.cs
```diff
@@ -938,7 +931,7 @@ public void DotNetCore_Muxer_ChoosesAppropriateSDKVersion(string sdkversion)
             var script = new ShellScriptBuilder()
                 .AddCommand($"mkdir -p {appDir} && cd {appDir}")
                 .AddCommand($"dotnet new globaljson --sdk-version {sdkversion}")
-                .SetEnvironmentVariable("PATH", $"{flattenedDotNetInstallDir}:$PATH")
+                .SetEnvironmentVariable("PATH", $"{flattenedDotNetInstallDir}:$PATH", true)
```
snehapar9, line 934: @waliMSFT do you mind clarifying why this `true` param was added to this `SetEnvironnmentVariable` method and what it means?

tests/Oryx.Integration.Tests/Fixtures/MySqlDbContainerFixture.cs
snehapar9: +1

src/BuildScriptGenerator.Common/ShellScriptBuilder.cs
waliMSFT: done

src/BuildScriptGenerator.Common/ShellScriptBuilder.cs
```diff
@@ -74,9 +81,19 @@ public ShellScriptBuilder AddScriptCommand(string argumentsString)
...
 
-        public ShellScriptBuilder SetEnvironmentVariable(string name, string value)
+        public ShellScriptBuilder SetEnvironmentVariable(string name, string value, bool isVariableSubstitutionNeeded = false)
         {
-            return this.Append($"export {name}={value}");
+            if (isVariableSubstitutionNeeded)
+            {
+                // linux does not allow variable substitution in a single quoted string.
+                // So, if we need to do variable substitution it has to be inside a double quoted string.
+                // e.g. to add a new path in an existing env variable, variable substitution is required.
```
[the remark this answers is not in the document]
waliMSFT: I need to test it. I will update here when done. Thanks for the idea Paul.

tests/Oryx.BuildImage.Tests/DotNetCore/DotNetCoreSampleAppsTest.cs
waliMSFT: yah sure. I actually added few lines of comment in the definition of the `SetEnvironnmentVariable` method ([url](https://github.com/microsoft/Oryx/pull/1878/files/3840c7f9d114fa8a6d0b706eca5842985f933c4a#diff-553be68e282a55d647d81b70373bbc10beeb087b9e673a1531a89bda3889ec79:~:text=//%20linux%20does%20not,substitution%20is%20required.)). 

The reason to add the boolean value is we need to use two different kinds of quotations: single and double. `', "`. The reason:
1) when the sas-token is set as ENV variable we need to enquote the token value with a single quote. Otherwise, the Linux system cannot handle the special characters within the token value.
2) Except one special case this single quoted value works fine. In a situation where substitution is used to construct a string (for example: `"{flattenedDotNetInstallDir}:$PATH" where flattenedDotNetInstallDir is a path to a directory`) Linux system doesn't allow the variable substitution inside a single quote. Variable substitution is only allowed inside a double quoted string value.

src/BuildScriptGenerator/SdkStorageVersionProviderBase.cs
```diff
@@ -150,14 +149,13 @@ protected string GetPlatformBinariesStorageBaseUrl()
         {
             var platformBinariesStorageBaseUrl = this.commonOptions.OryxSdkStorageBaseUrl;
 
-            this.logger.LogDebug("Using the Sdk storage url {sdkStorageUrl}.", platformBinariesStorageBaseUrl);
-
             if (string.IsNullOrEmpty(platformBinariesStorageBaseUrl))
             {
-                throw new InvalidOperationException(
-                    $"Environment variable '{SdkStorageConstants.SdkStorageBaseUrlKeyName}' is required.");
+                platformBinariesStorageBaseUrl = Environment.GetEnvironmentVariable(SdkStorageConstants.TestingSdkStorageUrlKeyName);
```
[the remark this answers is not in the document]
waliMSFT: yah, makes sense. I will update that.

vsts/pipelines/copyAllBlobsToProd.yml
```diff
@@ -40,8 +40,10 @@ jobs:
         displayName: 'Copy all blobs from a source storage account to the prod storage account'
         env:
           DEV_STORAGE_SAS_TOKEN: $(DEV-STORAGE-SAS-TOKEN)
+          ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN: $(ORYX-SDK-STAGING-PRIVATE-SAS-TOKEN)
           SANDBOX_STORAGE_SAS_TOKEN: $(SANDBOX-STORAGE-SAS-TOKEN)
           PROD_STORAGE_SAS_TOKEN: $(PROD-STORAGE-SAS-TOKEN)
+          ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN: $(ORYX-SDK-STAGING-PRIVATE-SAS-TOKEN)
```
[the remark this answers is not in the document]
waliMSFT: Thanks for the catch!

waliMSFT:
> > * [x]   After the updates to ensure we _NEVER_ are handling the SAS token directly or are surfacing it in the pipeline, we need to rotate the SAS token in the key vault. As a rule of thumb, we never should have this token in a command line command, as commands are saved to the local disk in a history file. If this accidentally happens, or we accidentally expose it some other way, this token needs to be immediately rotated.
> 
> @waliMSFT can you just confirm that you rotated the token, since this task was checked off? Did you happen to keep track of the steps that you did to rotate it? I think I remember us discussing that it would be a good exercise to do it and keep track of the steps so that it is easy to do in the future

I have updated the token twice so far. In last 5-6 weeks I did not update it though.

tests/Oryx.BuildImage.Tests/BaseImage/BaseImageTest.cs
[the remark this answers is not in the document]
waliMSFT: I have explained the reason here:
https://github.com/microsoft/Oryx/pull/1878#discussion_r1191447904

Please let me know your thoughts. thanks

tests/Oryx.Integration.Tests/Fixtures/MySqlDbContainerFixture.cs
snehapar9: @waliMSFT do you mind pointing to this method that uses the staging account as default? Just my two cents that it's more intuitive to make this change in this method - `AddDefaultTestEnvironmentVariables` than introduce new arguments to this constructor - `ShellScriptBuilder` to determine if we want to use the staging account or not.

william-msft:
> > > * [x]    As @cormacpayne mentioned in the meeting, lets add tests to make sure we don't propagate the token in any of the images
> > 
> > 
> > @pauld-msft I have added a work item regarding this. https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1818074 @daniv-msft
> 
> @waliMSFT - so these tests are supposed to be an automated way of ensuring that the token is secure (at least in the final images). Until we have these in place, or some manual verification, I'm not sure that we would be able to be certain the token is not exposed. So is the idea to just consider the storage account as public until then, or are you planning on manually verifying that they don't get passed through?

+1
@waliMSFT could we at least provide some automated tests that will test the core functionality?

Do we have a testing plan for this? Either in the description or in a linked document will help us identify gaps in our testing plan before releasing to production. This is especially important if don't rely on any automated tests. Debugging this work in production in the midst of an expected pre-release will not be fun.

images/__common.sh
william-msft: Re-opening. Sorry having the token in the URL still makes me a bit concerned about the secrets showing up in the logs later since it's part of the URL. Like what happens if the base URL is ever wrong, will the SassToken still appear as `***` I image the SassToken eventually has to decrypt before storing in the string. 

Have we considered passing this as a header? Example:
curl -D headers.txt -SL "{sdkStorageBaseUrl}/{platformName}/{platformName}-{version}.tar.gz" --header "Authorization: Bearer $ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN"

pauld-msft: Throughout the build script we are printing out the token. 

This can be seen when you run `export BUILDKIT_PROGRESS=plain` in git bash before executing the `buildBuildImages.sh` script. This will make sure that the full log output doesn't get hidden in your git bash terminal, and you can see what the actual output is, similar to how the full output of `docker build` commands are printed in the Azure Pipeline.

The token primarily is printed due to the `set -ex` found throughout all of the Oryx script that are run during the docker builds.

Technically, it isn't getting leaked since the Azure Pipeline is masking the token, but relying on this functionality does not seem (to me) like good practice.

waliMSFT: @william-msft @pauld-msft  masking the SasToken does not depend on the URL. As we are retrieving the KeyVault secret from the azure library, Azure considers it as secret and does not expose the secret. I thought this gives us the efficient way of handling the secrets and Azure is taking care of it. 
In the local builds we do not retrieve the token from keyVault if the account url does not point to the Staging one. 
Besides, in the local builds and the tests when the sasToken is retrieved from the KeyVault I avoided printing the steps and  the token in the logs by adding `+x` and `-x`  at the beginning and end respectively of the method.
https://github.com/microsoft/Oryx/pull/1878/files#:~:text=function%20retrieveSastokenFromKeyvault(),%7D

images/runtime/python/3.11/bullseye.Dockerfile
```diff
@@ -49,11 +49,15 @@ RUN true
...
+RUN --mount=type=secret,id=oryx_sdk_storage_account_access_token \
+    set -ex \
+    && export ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN="$(cat /run/secrets/oryx_sdk_storage_account_access_token)" \
+    && ${BUILD_DIR}/buildPythonSdkByVersion.sh $PYTHON_VERSION $DEBIAN_FLAVOR \
+    && export ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN=""
 
 RUN set -ex \
  && cd /opt/python/ \
- && ln -s 3.11.1 3.11 \
+ && ln -s 3.11.0 3.11 \
```
daniv-msft, line 60: Could you please clarify if you remember why this change from 3.11.1 to 3.11.0 was made?

waliMSFT: Dan, I did not make this change. Assuming it happened during a merge.

images/__common.sh
pauld-msft: @waliMSFT and I caught up on this offline, and he has addressed the issue with [this commit](https://github.com/microsoft/Oryx/pull/1878/commits/4ff61b4b3c591e08fb895c344fee2090999001f3) that disables the debugging flag `-x` wherever we are using this token.

william-msft: nit: would be nice if we could toggle this on via flag or something, to help us debug in the future.

waliMSFT: @william-msft thanks for the suggestion. For now, I will keep it as it is. an

> > > > * [ ]     As @cormacpayne mentioned in the meeting, lets add tests to make sure we don't propagate the token in any of the images
> > > 
> > > 
> > > @pauld-msft I have added a work item regarding this. https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1818074 @daniv-msft
> > 
> > 
> > @waliMSFT - so these tests are supposed to be an automated way of ensuring that the token is secure (at least in the final images). Until we have these in place, or some manual verification, I'm not sure that we would be able to be certain the token is not exposed. So is the idea to just consider the storage account as public until then, or are you planning on manually verifying that they don't get passed through?
> 
> +1 @waliMSFT could we at least provide some automated tests that will test the core functionality?
> 
> Do we have a testing plan for this? Either in the description or in a linked document will help us identify gaps in our testing plan before releasing to production. This is especially important if don't rely on any automated tests. Debugging this work in production in the midst of an expected pre-release will not be fun.

We discussed this issue recently and as we discussed I believe we do not need additional tests as functionality-wise or workflow-wise nothing changed. The main changes in this PR is the staging-url and passing the token with the URL to access the resources. For this, we have the tests to check if the account is working fine or not. 
https://github.com/microsoft/Oryx/blob/6c59cde0b3e22ecbdff11b1a9a194cb8a1b61a9d/tests/Oryx.Integration.Tests/StagingStorageAccountSanityTest.cs

Here I have just pointed to the new Staging account to test.

src/BuildScriptGenerator.Common/ShellScriptBuilder.cs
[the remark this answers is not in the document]
waliMSFT: @pauld-msft I could not test it. So, created a work item  to test it later. 
https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1831372

images/__common.sh
waliMSFT: > nit: would be nice if we could toggle this on via flag or something, to help us debug in the future.

Hi William, thanks for the suggestion. I have created a work item:
https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1831398

tests/Oryx.BuildImage.Tests/BaseImage/BaseImageTest.cs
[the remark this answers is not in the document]
pauld-msft: I like it. Thanks for making the change

tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
```diff
@@ -356,5 +364,12 @@ private string GetDefaultVersion(string debianFlavor, params string[] platformPa
 
             return defaultVersion;
         }
+
+        private string GetKeyvaultSecretValue(string keyvaultUri, string secretName)
```
pauld-msft, line 368: We have 4 different implementations of the `GetKeyVaultSecretValue` method. Lets pull them all into a shared  `GetKeyVaultSecretValue` method that is defined in the `Oryx.Tests.Common` project and can be used in each location.

tests/Oryx.Integration.Tests/Database/DatabaseTestsBase.cs
```diff
@@ -81,6 +84,24 @@ protected DatabaseTestsBase(ITestOutputHelper outputHelper, Fixtures.DbContainer
                 });
         }
 
+        private List<EnvironmentVariable> GetEnvironmentVariableList()
```
pauld-msft, line 87: We have 2 different methods called `GetEnvironmentVariableList` that are returning a list of environment variables with essentially the same logic. Can we convert these into a single method that extends an env var list that performs the required logic?

E.g. something along the lines of

```csharp
public List<EnvironmentVariable> AddStorageAccountEnvironmentVariables(this List<EnvironmentVariable> envVarList) 
{
  var testStorageAccountUrl = Environment.GetEnvironmentVariable(SdkStorageConstants.TestingSdkStorageUrlKeyName);
  var sdkStorageUrl = string.IsNullOrEmpty(testStorageAccountUrl) ? SdkStorageConstants.PrivateStagingSdkStorageBaseUrl : testStorageAccountUrl;

  envVarList.Add(new EnvironmentVariable(SdkStorageConstants.SdkStorageBaseUrlKeyName, sdkStorageUrl));

  if (sdkStorageUrl == SdkStorageConstants.PrivateStagingSdkStorageBaseUrl)
  {
      string stagingStorageSasToken = Environment.GetEnvironmentVariable(SdkStorageConstants.PrivateStagingStorageSasTokenKey) ??
      this.GetKeyVaultSecretValue(SdkStorageConstants.OryxKeyvaultUri, SdkStorageConstants.StagingStorageSasTokenKeyvaultSecretName);
      envVarList.Add(new EnvironmentVariable(SdkStorageConstants.PrivateStagingStorageSasTokenKey, stagingStorageSasToken));
  }

  return envVarList;
}
```

so then we could do the following above:

`_dbFixture.GetCredentialsAsEnvVars().AddStorageAccountEnvironmentVariables()`

and the following later on:

`SqlServerDbTestHelper.GetEnvironmentVariables().AddStorageAccountEnvironmentVariables()`

pauld-msft (changes requested):
A few more (fairly minor) changes. Getting close!

images/runtime/dotnetcore/7.0/base.buster.staging.Dockerfile
```diff
@@ -0,0 +1,64 @@
+# dotnet tools are currently available as part of SDK so we need to create them in an sdk image
```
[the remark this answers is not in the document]
waliMSFT: yah.. it was for testing.

tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
waliMSFT: I was thinking the same initially. Thanks for pointing it out.

build/buildBuildImages.sh
```diff
@@ -69,6 +69,9 @@ if [ -z "$sdkStorageAccountUrl" ]; then
 	sdkStorageAccountUrl=$PROD_SDK_CDN_STORAGE_BASE_URL
 fi
 
+# checking and retrieving token for the `oryxsdksstaging` account.
+retrieveSastokenFromKeyvault $sdkStorageAccountUrl
```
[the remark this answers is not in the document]
waliMSFT: a common work item which will address the issues: https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1848573

tests/Oryx.Integration.Tests/StorageAccountSanityTestBase.cs
waliMSFT: a common work item which will address the issues: https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1848573

tests/Oryx.Integration.Tests/Database/DatabaseTestsBase.cs
waliMSFT: a common work item which will address the issues: https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1848573

pauld-msft (approved):
Approving for merge, but we should make sure to do the work items linked in this PRs comments, and https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1848573

PR #1880, opened 2023-02-13
Update Node PHP and GO Versions

harryli0108:
Part of this PR is at #1889. This one now only updates php version.
This PR is updating:
- PHP
  - current -> 8.0.27
  - current -> 8.1.15
  - current -> 8.2.2


images/runtime/php/template.base.Dockerfile
```diff
@@ -27,7 +27,8 @@ RUN { \
 } >> /etc/apache2/apache2.conf
 
 # Install common PHP extensions
-RUN apt-get update \
+RUN apt-mark hold msodbcsql18 odbcinst1debian2 odbcinst unixodbc unixodbc-dev \
```
cormacpayne, line 30: What's the purpose of this `apt-mark hold` command? It may be worth adding a comment above this line to clarify why this is being done.

images/runtime/php/runbase.Dockerfile
```diff
@@ -45,7 +45,6 @@ RUN set -eux; \
 		libtidy-dev \
 		libkrb5-dev \
 		libxslt-dev \
-		unixodbc-dev \
```
cormacpayne, line 48: Just to confirm: we're removing `unixodbc-dev` from being installed in the base image, but will later install a specific version of the package as a part of the PHP version's Dockerfile?

harryli0108: That is correct. But this will be reverted once the package source fixes the faulty package.

images/runtime/php/template.base.Dockerfile
harryli0108: This one is to mark specific packages as hold. This will prevent those packages from being upgraded when `apt-get upgrade` runs. I'll add comments to explain this. This will be reverted once the package source fixes the issue as well.

cormacpayne (approved):
:shipit:

PR #1881, opened 2023-02-13
Fix dockerfile command

pauld-msft:
Found a bug with the dockerfile command where it was unable to detect the os type and image type from the .ostype and .imagetype files. Noticed that this would also be the case for the other 2 commands that accept the build command options, exec and prepare env.


cormacpayne (approved):
:shipit:

PR #1888, opened 2023-02-16
Williamhe/1732771 refactor automation

william-msft:
  - Re-factor DotNet automation which includes
    - Getting all new DotNet versions
      - Before we would match the release date to the current date
      - We also added semantic version range; supporting a minimum version
    - Updated to use dependency injection
    - Separated deserialization classes into their own Models folder


build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -33,148 +31,142 @@ namespace Microsoft.Oryx.Automation
...
-        ///
-        /// This is accomplished by:
-        ///     - Checking release meta data url if there's a new release
-        ///     - If new release, then store release information into PlatformConstants
-        ///        Otherwise don't store anything
-        /// </Summary>
-        /// <param name="dateTarget">yyyy-mm-dd format string that defaults to today's date</param>
-        /// <returns>PlatformConstants used later to update constants.yaml</returns>
-        public override async Task<List<PlatformConstant>> GetPlatformConstantsAsync(string dateTarget)
+        public async Task RunAsync(string oryxRootPath)
```
cormacpayne, line 49: We should make `oryxRootPath` a private variable in this class that we set after it's been provided to `RunAsync` so that every method is able to easily reference it without us having to pass it around in each method. This value should only ever be provided once on application startup (since it's an argument to the app), so there shouldn't be any concern about having it overridden or changed during the lifespan of it executing.

build/tools/Automation/DotNet/Models/FileObj.cs
```diff
@@ -0,0 +1,23 @@
+﻿// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+using Newtonsoft.Json;
+
+namespace Microsoft.Oryx.Automation.DotNet.Models
+{
+    public class FileObj
```
cormacpayne, line 9: nit: I'm typically opposed to including `Obj` in the name of a class since it's already implied that it'll be an object once instantiated, but if this makes more sense for you, then I'm OK keeping it this way

(this comment applies for all class with the `Obj` suffix)

build/tools/Automation/DotNet/Models/FileObj.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        [JsonProperty(PropertyName = "name")]
+        public string Name { get; set; } = string.Empty;
+
+        [JsonProperty(PropertyName = "rid")]
+        public string Rid { get; set; } = string.Empty;
```
cormacpayne, line 15: nit: what does `rid` mean in this context?

build/tools/Automation/Models/YamlConstants.cs
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+using System.Collections.Generic;
+
+namespace Microsoft.Oryx.Automation.Models
+{
+    /// <Summary>
+    /// This is used to deserialize Constants.ConstantsYaml file
+    /// </Summary>
+    public class YamlConstants
```
cormacpayne, line 12: I would avoid having the `Constants` suffix on this class name if the class isn't encapsulating static string constants (I know this is focused on the `constants.yaml` file, so maybe a better name is `ConstantsYamlFile`)

build/tools/Automation/Program.cs
```diff
@@ -2,180 +2,50 @@
...
+            string platform = args[0].ToLower();
+            switch (platform)
+            {
+                case "dotnet":
+                    var dotNet = serviceProvider.GetRequiredService<DotNet.DotNet>();
+                    await dotNet.RunAsync(oryxRootPath);
+                    break;
+
+                default:
+                    Console.WriteLine($"Unsupported platform: {args[0]}");
```
cormacpayne, line 45: nit: we should use `platform` rather than `args[0]` here since it's a variable we already defined

build/tools/Automation/Services/VersionService.cs
```diff
@@ -0,0 +1,64 @@
...
+                && SemanticVersion.TryParse(maxVersion, out var maxSemanticVersion)
+                && semanticVersion > maxSemanticVersion)
+            {
+                return false;
+            }
+
+            // Check if the version is in the list of exception versions (if specified)
+            if (exceptionVersions != null)
+            {
+                var exceptionVersionList = exceptionVersions.Split(',');
```
cormacpayne, line 47: Any reason the `exceptionVersions` argument is a string of comma-separated versions and isn't just passed in as an array of versions?

build/tools/Automation/Services/YamlFileReaderService.cs
```diff
@@ -0,0 +1,27 @@
...
+using YamlDotNet.Serialization;
+using YamlDotNet.Serialization.NamingConventions;
+
+namespace Microsoft.Oryx.Automation.Services
+{
+    public class YamlFileReaderService : IYamlFileReaderService
+    {
+        public async Task<List<YamlConstants>> ReadConstantsYamlFileAsync(string filePath)
+        {
+            string fileContents = await File.ReadAllTextAsync(filePath);
```
cormacpayne, line 18: We should have some error handling when dealing with I/O

build/tools/Automation/Telemetry/LoggerImpl.cs
```diff
@@ -0,0 +1,20 @@
+﻿// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Oryx.Automation.Telemetry
+{
+    public class LoggerImpl : ILogger
```
cormacpayne, line 9: Similar to the `HttpClientImpl` comment -- what is the purpose of this class?

tests/AutomationTests/DotNetTests.cs
```diff
@@ -0,0 +1,39 @@
+using Microsoft.Oryx.Automation.Client;
```
cormacpayne, line 1: Please add a license header

william-msft: Updated

build/tools/Automation/DotNet/Models/FileObj.cs
william-msft: So `File` belongs to belongs to `System.IO.File`. I'll just have to specify the namespace to differentiate. What are you thoughts on that?

build/tools/Automation/Services/VersionService.cs
william-msft: Originally I was thinking of storing the `exceptionVersions` as an environment variable string, and passing this to this method to avoid having to submit a PR every time we want to skip a version. But also a part of me is also thinking of storing the `exceptionVersions` as an array in a `Constants.cs` file. We'll just have to submit  PR every time we want to skip.

build/tools/Automation/Client/HttpClientImpl.cs
```diff
@@ -0,0 +1,66 @@
...
+// --------------------------------------------------------------------------------------------
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading.Tasks;
+using System.Xml.Linq;
+
+namespace Microsoft.Oryx.Automation.Client
+{
+    public class HttpClientImpl : IHttpClient
```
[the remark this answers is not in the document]
william-msft: Updated using an extension class and have added documentation. Thank you for helping me with creating this an extension 😄

build/tools/Automation/Client/HttpClientImpl.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        private readonly HttpClient httpClient;
```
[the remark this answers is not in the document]
william-msft: Updated

build/tools/Automation/Services/YamlFileReaderService.cs
william-msft: Updated

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -33,148 +31,142 @@ namespace Microsoft.Oryx.Automation
...
+        {
+            List<VersionObj> versionObjs = new List<VersionObj>();
+            string url = Constants.OryxSdkStorageBaseUrl + "/dotnet?restype=container&comp=list&include=metadata";
+            HashSet<string> oryxSdkVersions = await this.httpClient.GetOryxSdkVersionsAsync(url);
 
-            // releaseIndex contains release meta data
-            var releasesIndex = releaseNotes == null ? new List<ReleaseNote>() : releaseNotes.ReleasesIndex;
-            List<PlatformConstant> platformConstants = new List<PlatformConstant>();
+            // Deserialize release meta data
+            var response = await this.httpClient.GetDataAsync(Constants.ReleasesIndexJsonUrl);
```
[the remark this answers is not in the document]
william-msft: Updated.

build/tools/Automation/Services/VersionService.cs
cormacpayne: Independent of how we actually provide the versions-to-skip to this program (I agree it should be an environment variable that we can set in the _Variables_ section of our repository), this question was more about the actual `exceptionVersions` argument to this method being a comma-separated `string` rather than a `string[]` to make it easier to work with.

build/tools/Automation/DotNet/Models/FileObj.cs
cormacpayne: I think if we set up the classes correctly (_i.e._, if we contain all of the I/O calls within their own `FileService` class), then I would be optimistic we wouldn't use a `File` object in the same class that has a `using System.IO` reference.

If that's not the case, then we could defer to using the `Metadata` suffix instead of `Obj`, which should make sense in the context of these classes you've created.

build/tools/Automation/Extensions/HttpClientExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+                Console.WriteLine("Making GET request to: " + url);
+                HttpResponseMessage response = await httpClient.GetAsync(url);
+                Console.WriteLine($"Response received.: {response.StatusCode}");
+                if (!response.IsSuccessStatusCode)
+                {
+                    return null;
+                }
+
+                string responseContent = await response.Content.ReadAsStringAsync();
+                return responseContent;
```
cormacpayne, line 39: nit: let's combine these two lines together

```cs
return await response.Content.ReadAsStringAsync();
```

build/tools/Automation/Extensions/HttpClientExtensions.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// <returns>A Task HashSet<string> representing the asynchronous operation. The result of
+        /// the task is a HashSet<string> containing the ORYX SDK versions retrieved from the
+        /// response content.
+        /// </returns>
+        public static async Task<HashSet<string>> GetOryxSdkVersionsAsync(this HttpClient httpClient, string url)
+        {
+            HashSet<string> versions = new HashSet<string>();
+            var response = await httpClient.GetDataAsync(url);
+
+            XDocument xmlDoc = XDocument.Parse(response);
```
cormacpayne, line 68: We still need to handle the case where `response` is `null` here -- we should be able to do a quick `null` check and return `versions` if it is

build/tools/Automation/Program.cs
```diff
@@ -2,180 +2,46 @@
...
-            bool match = datesMatch == 0;
-            Console.WriteLine($"[DatesMatch]  releasedDate: {releasedDate} targetDate: {targetDate} " +
-                $"datesMatch: {datesMatch} match: {match}");
-            return match;
-        }
+            var serviceProvider = new ServiceCollection()
+                .AddHttpClient()
+                .AddSingleton<IVersionService, VersionService>()
+                .AddSingleton<IYamlFileReaderService, YamlFileReaderService>()
+                .AddScoped<DotNet.DotNet>()
```
cormacpayne, line 28: Similar comment in the Python PR, but it would be nice to rename these platform processing classes to something like `DotNetProcessor` or `DotNetAutomator` so that we don't need the redundant namespace suffix and the class name more accurately describes what it's responsible for.

tests/AutomationTests/DotNetTests.cs
```diff
@@ -0,0 +1,42 @@
+// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+/*using Microsoft.Oryx.Automation.Client;
```
cormacpayne, line 5: Are we planning on adding tests before this PR is merged?

cormacpayne (commented):
Some additional comments to take a look at -- feel free to ping me offline if you'd like to sync about any of them

build/tools/Automation/Services/VersionService.cs
william-msft: Updated to use List<string>

build/tools/Automation/DotNet/Constants.cs
```diff
@@ -1,11 +1,10 @@
...
+﻿// --------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
 // --------------------------------------------------------------------------------------------
-
-namespace Microsoft.Oryx.Automation
+namespace Microsoft.Oryx.Automation.DotNet
 {
-    internal static class Constants
+    public class Constants
```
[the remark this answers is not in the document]
william-msft: Updated

build/tools/Automation/Program.cs
william-msft: Updated

tests/AutomationTests/DotNetTests.cs
william-msft: Yes, working on updating this

build/tools/Automation/Extensions/HttpClientExtensions.cs
william-msft: Updated

build/tools/Automation/DotNet/Constants.cs
```diff
@@ -17,9 +16,11 @@ internal static class Constants
         public const string DotNetAspCoreName = "aspnet-core";
         public const string DotNetSdkKey = "dot-net-core-sdk-versions";
         public const string DotNetRuntimeKey = "dot-net-core-run-time-versions";
-        public const string DotNetReleasesMetaDataUrl = "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json";
         public const string DotNetLinuxTarFileRegex = ".*-linux-x64.tar.gz";
-
-        // TODO: add constants for other platforms
+        public const string ReleasesIndexJsonUrl = "https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json";
+        public const string OryxSdkStorageBaseUrl = "https://oryx-cdn.microsoft.io";
```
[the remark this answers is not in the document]
william-msft: I've added a Constant.cs under Automation/Constants.cs

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -33,148 +31,142 @@ namespace Microsoft.Oryx.Automation
...
+            var releaseNotes = JsonConvert.DeserializeObject<ReleaseNotes>(response);
+            var releasesIndex = releaseNotes == null ? new List<ReleaseNote>() : releaseNotes.ReleaseIndexes;
             foreach (var releaseIndex in releasesIndex)
             {
-                var dateReleased = releaseIndex.LatestReleaseDate;
-                if (!DatesMatch(dateTarget, dateReleased) || this.prodSdkVersions.Contains(releaseIndex.LatestSdk))
+                string latestVersion = releaseIndex.LatestSdk;
+                if (!this.versionService.IsVersionWithinRange(latestVersion, minVersion: Constants.MinSdkVersion) ||
```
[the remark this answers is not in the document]
william-msft: Updated

build/tools/Automation/Client/HttpClientImpl.cs
```diff
@@ -0,0 +1,66 @@
...
+            }
+        }
+
+        public async Task<HashSet<string>> GetOryxSdkVersionsAsync(string url)
+        {
+            HashSet<string> versions = new HashSet<string>();
+            HttpClientImpl httpClientImpl = new HttpClientImpl();
+            var response = await httpClientImpl.GetDataAsync(url);
+
+            XDocument xmlDoc = XDocument.Parse(response);
```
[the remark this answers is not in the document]
william-msft: Updated

build/tools/Automation/Telemetry/LoggerImpl.cs
william-msft: I've converted Logger to an extensions class. I'm thinking of adding the telemetry code in the python PR

build/tools/Automation/DotNet/DotNet.cs
william-msft: Updated to use via RunAsync

build/tools/Automation/DotNet/DotNet.cs
```diff
@@ -33,148 +31,142 @@ namespace Microsoft.Oryx.Automation
...
+                    string sha = this.GetSha(release.Sdk.Files);
+                    VersionObj versionObj = new VersionObj
                     {
                         Version = sdkVersion,
                         Sha = sha,
-                        PlatformName = Constants.DotNetName,
                         VersionType = Constants.SdkName,
                     };
-                    platformConstants.Add(platformConstant);
+                    versionObjs.Add(versionObj);
```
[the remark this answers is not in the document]
william-msft: I've updated the the logic a little here so we don't need to check for `null`

build/tools/Automation/Automation.csproj
```diff
@@ -3,13 +3,23 @@
...
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.1</TargetFramework>
-    <Nullable>enable</Nullable>
     <DisableDocRequirement>true</DisableDocRequirement>
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
+    <PackageReference Include="LoggingAdvanced.Console" Version="0.4.1" />
+    <PackageReference Include="Microsoft.AspNet.SignalR.Client" Version="2.4.3" />
```
cormacpayne, line 11: Are either of these packages used in the `Automation` project?

build/tools/Automation/DotNet/DotNetAutomator.cs
```diff
@@ -0,0 +1,274 @@
...
+
+                if (versionType.Equals(DotNetConstants.SdkName))
+                {
+                    ConstantsYamlFile dotNetYamlConstant = dotnetYamlConstants[DotNetConstants.DotNetSdkKey];
+                    dotNetYamlConstant.Constants[dotNetConstantKey] = version;
+
+                    // update versionsToBuild.txt
+                    this.fileService.UpdateVersionsToBuildTxt(
+                        DotNetConstants.DotNetName,
+                        $"\n{dotNetVersion.Version}, {dotNetVersion.Sha},");
```
cormacpayne, line 210: nit: do we need to provide the newline character as a prefix to the second argument in `UpdateVersionsToBuildTxt` or can we automatically set it inside of the method when the argument value is used?

build/tools/Automation/DotNet/DotNetAutomator.cs
```diff
@@ -0,0 +1,274 @@
...
+
+        private string GenerateDotNetConstantKey(DotNetVersion dotNetVersion)
+        {
+            string[] splitVersion = dotNetVersion.Version.Split('.');
+            string majorVersion = splitVersion[0];
+            string minorVersion = splitVersion[1];
+
+            // prevent duplicate majorMinor where 11.0 and 1.10 both will generate a 110 key.
+            int majorVersionInt = int.Parse(majorVersion);
+            string majorMinor = majorVersionInt < 10 ? $"{majorVersion}{minorVersion}" : $"{majorVersion}_{minorVersion}";
```
cormacpayne, line 241: I may have missed the previous comment resolution, but is there a reason we can't just always use the `{majorVersion}_{minorVersion}` pattern as a key?

build/tools/Automation/Extensions/HttpClientExtensions.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// the task is the response content as a string, or null if the response status code is not
+        /// successful.
+        /// </returns>
+        public static async Task<string> GetDataAsync(this HttpClient httpClient, string url)
+        {
+            try
+            {
+                Console.WriteLine("Making GET request to: " + url);
+                HttpResponseMessage response = await httpClient.GetAsync(url);
+                Console.WriteLine($"Response received.: {response.StatusCode}");
```
cormacpayne, line 32: nit: would you mind removing the `.` character before the `:` character on this line?

```
Response received: {response.StatusCode}
```

build/tools/Automation/Program.cs
```diff
@@ -2,180 +2,48 @@
...
-            public Dictionary<string, object> Constants { get; set; } = new Dictionary<string, object>();
-
-            public List<object> Outputs { get; set; } = new List<object>();
+            string oryxRootPath = args[1];
+            string platform = args[0].ToLower();
+            var serviceProvider = new ServiceCollection()
+                .AddHttpClient()
+                .AddSingleton<IVersionService, VersionService>()
+                .AddSingleton<IFileService>(new FileService(oryxRootPath))
+                .AddSingleton<IYamlFileService>(new YamlFileService(oryxRootPath))
```
cormacpayne, line 31: nit: would you make sure that the mappings here are in lexicographical order based?

```cs
.AddSingleton<IFileService>(...)
.AddSingleton<IVersionService, VersionService>()
.AddSingleton<IYamlFileService>(...)
```

build/tools/Automation/Services/IYamlFileService.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Threading.Tasks;
+using Microsoft.Oryx.Automation.Models;
+
+namespace Microsoft.Oryx.Automation.Services
+{
+    public interface IYamlFileService
+    {
+        Task<List<ConstantsYamlFile>> ReadConstantsYamlFileAsync(string filePath);
+
+        public void WriteConstantsYamlFile(string filePath, List<ConstantsYamlFile> yamlConstants);
```
cormacpayne, line 15: Let's remove `public` here

cormacpayne (commented):
A few additional comments on top of our `IHttpService` discussion, otherwise it LGTM :shipit:

build/tools/Automation/DotNet/DotNetAutomator.cs
william-msft: Doesn't seem to need it. But I've added an `Environment.NewLine` after the `line` that is provided in the `FileService`

build/tools/Automation/Program.cs
william-msft: Updated

build/tools/Automation/Services/IYamlFileService.cs
william-msft: Ah yes, removed.

build/tools/Automation/DotNet/DotNetAutomator.cs
william-msft: None that I could see. YAML allows `_` and underscore also is not part of semantic versioning.

william-msft: Side note: I've also updated this for PythonAutomator

build/tools/Automation/Extensions/HttpClientExtensions.cs
william-msft: Remove

build/tools/Automation/Services/IHttpService.cs
```diff
@@ -0,0 +1,16 @@
...
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Microsoft.Oryx.Automation.Services
+{
+    public interface IHttpService
+    {
+        public Task<string> GetDataAsync(string url);
```
cormacpayne, line 12: nit: the `public` access modifier isn't needed for interface methods (prior to C# 8.0, it would've caused a compilation error) since it's the default value when no access modifier is presented, but it's OK to keep this as-is (unless there are other changes that need to be made, then it would be a nice clean up change to make)

tests/AutomationTests/Services/YamlFileServiceTests.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        private readonly YamlFileService yamlFileService;
+        private readonly string oryxRootPath;
+        private readonly string testConstantsYamlFilePath;
+
+        public YamlFileServiceTests()
+        {
+            this.oryxRootPath = Directory.GetCurrentDirectory();
+            this.testConstantsYamlFilePath = Path.Combine(this.oryxRootPath, "test-constants.yaml");
+            this.yamlFileService = new YamlFileService(this.oryxRootPath);
```
cormacpayne, line 25: Since we're writing to files in this test suite, we should create a temp directory for the files we're creating so there's no potential crossover of tests targeting the same files or paths. The easiest way to do this would be to generate a random GUID that acts as the intermediate folder between the root path that we're using and any files that we'd be writing to.

tests/AutomationTests/Services/YamlFileServiceTests.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        [Fact]
+        public async Task ReadConstantsYamlFileAsync_ReturnsExpectedYamlContents()
+        {
+            // Arrange
+            string yamlContents = "- name: test\r\n  constants:\r\n    key1: value1\r\n    key2: value2";
+            File.WriteAllText(this.testConstantsYamlFilePath, yamlContents);
```
cormacpayne, line 33: For the tests in this file where we're writing the contents to a new file, we should handle the case where an unexpected error is thrown but we still delete the file we generated. In this case, if the `ReadConstantsYamlFileAsync` method below fails for an unexpected error, we never hit the `File.Delete` call at the bottom and the file we wrote to lingers around (which isn't a _huge_ deal since these run on separate container agents, but locally it may linger).

I suggest wrapping the test with a `try-finally`:

```cs
try
{
    // Arrange
   ...

    // Act
    ...

    // Assert
    ...
}
finally
{
    // Clean up
    File.Delete(this.testConstantsYamlFilePath);
}
```

cormacpayne (dismissed):
A couple of minor comments with the tests, but otherwise LGTM :shipit:

build/tools/Automation/Constants.cs
```diff
@@ -1,25 +1,14 @@
-// --------------------------------------------------------------------------------------------
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT license.
-// --------------------------------------------------------------------------------------------
```
daniv-msft, line 4: Could you please clarify if there are any reasons to remove the MIT license header here? Looks like it should still be valid in that case.

build/tools/Automation/DotNet/DotNetAutomator.cs
```diff
@@ -0,0 +1,267 @@
...
+            }
+        }
+
+        public async Task<List<DotNetVersion>> GetNewDotNetVersionsAsync()
+        {
+            List<DotNetVersion> dotNetVersions = new List<DotNetVersion>();
+
+            // Deserialize release metadata
+            var response = await this.httpService.GetDataAsync(DotNetConstants.ReleasesIndexJsonUrl);
+            var releaseNotes = response == null ? null : JsonConvert.DeserializeObject<ReleaseNotes>(response);
```
daniv-msft, line 91: I'm wondering if in the case where `response == null`, we should just throw an exception/exit the flow early.
Could you please clarify if there are any advantages to continue if we didn't get any reply there?

daniv-msft: Also, I'm assuming that this would mostly happen if the .NET team changes the path where they store the release index, right? If that's the case, how would we detect that an issue happened?

build/tools/Automation/DotNet/DotNetAutomator.cs
```diff
@@ -0,0 +1,267 @@
...
+                }
+
+                // Get the actual release information from releases.json
+                string releasesJsonUrl = releaseIndex.ReleasesJsonUrl;
+                response = await this.httpService.GetDataAsync(releasesJsonUrl);
+                var releasesJson = JsonConvert.DeserializeObject<ReleasesJson>(response);
+                var releases = releasesJson == null ? new List<Release>() : releasesJson.Releases;
+                foreach (var release in releases)
+                {
+                    // Check again since the "release" is separate from "releaseIndex".
```
daniv-msft, line 111: This would only happen if the .NET team had messed up versions (something like the .NET 7 release index containing .NET 6 releases), right? If it's the case, then please consider throwing an exception as this should never happen, and if it does we should let them know.

build/tools/Automation/DotNet/DotNetAutomator.cs
```diff
@@ -0,0 +1,267 @@
...
+                    var dotNetCoreVersion = this.GetDotNetVersion(release, DotNetConstants.DotNetCoreName);
+                    dotNetVersions.Add(dotNetCoreVersion);
+
+                    // Runtime (aspnetcore)
+                    var dotNetAspCoreVersion = this.GetDotNetVersion(release, DotNetConstants.DotNetAspCoreName);
+                    dotNetVersions.Add(dotNetAspCoreVersion);
+                }
+            }
+
+            return dotNetVersions = dotNetVersions.OrderBy(v => v.Version).ToList();
```
daniv-msft, line 131: Nit: please consider removing the re-assignment on `dotNetVersions` as we're about to return anyway.

build/tools/Automation/Services/IHttpService.cs
william-msft: Updated. I might have copied the signature from the `HttpService`

tests/AutomationTests/Services/YamlFileServiceTests.cs
william-msft: Updated to use GUID

tests/AutomationTests/Services/YamlFileServiceTests.cs
william-msft: Great idea! Especially since we only have 10GB disk space in our agents.

build/tools/Automation/DotNet/DotNetAutomator.cs
william-msft: We need to check our storage account twice because on how .NET releases information is structured. Both release index and releases can have versions that we should check for in our storage account. 

Here is an example flow:
1. GET https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/releases-index.json
a. Check storage account once to make sure the latest version is not already in storage account
b. This is the release index
3. Get releases URL from `"releases.json"`
4. GET https://dotnetcli.blob.core.windows.net/dotnet/release-metadata/7.0/releases.json
a. Check storage account again. Since since the `"releases"` elements has all `7.0` versions

build/tools/Automation/DotNet/DotNetAutomator.cs
william-msft: Thanks for catching this one!

build/tools/Automation/Constants.cs
william-msft: Sorry, I accidentally deleted. But have added back

build/tools/Automation/DotNet/DotNetAutomator.cs
william-msft: These should never be empty. So it would make sense to throw an  exception for when they are. I've updated the code for this.

build/tools/Automation/DotNet/DotNetAutomator.cs
daniv-msft: OK fair enough. I was mostly focusing on validating that the version is correct, but it's true that we also need to check if things already exist.

daniv-msft (dismissed):

daniv-msft (dismissed):

daniv-msft (dismissed):

daniv-msft (approved):

PR #1895, opened 2023-02-18
Respect Debian flavor for CLI image tag in 'oryx dockerfile'

cormacpayne:
Currently the `oryx dockerfile` command generates a Dockerfile that uses the `mcr.microsoft.com/oryx/cli:stable` image as the default image that the build layer takes place in.

This `oryx/cli:stable` image is no longer being published and doesn't respect the potential Debian flavor used by the runtime image, so this PR aims to first determine what the runtime image is and find the corresponding Debian flavor to use with `oryx/cli:debian-{flavor}-stable` as the build layer image.

The `constants.yaml` file was updated to now have a dictionary of fully-supported runtime images per platform that maps a version to the associated Debian flavor. The `SharedCodeGenerator` tool was also updated to now support generating code snippets for dictionaries in the `constants.yaml` file.


pauld-msft:
Do we even need https://github.com/microsoft/Oryx/blob/main/tests/Oryx.Tests.Common/ImageTestHelper.cs#L438-L498 anymore? Could we replace it with the constant dictionaries created here, so that we condense the number of places that need to be updated when we convert debian flavors?

build/constants.yaml
```diff
@@ -152,12 +147,11 @@
...
     python311-version: 3.11.0
     runtime-versions:
-    - 2.7
-    - 3.6
-    - 3.8
-    - 3.9
-    - 3.10
-    - 3.11
+      3.7: bullseye
+      3.8: bullseye
```
william-msft, line 151: Double checking if these should be `buster`? Since 3.9 is `buster`.

cormacpayne: According to the [`generateDockerfiles.sh` script for Python](https://github.com/microsoft/Oryx/blob/main/images/runtime/python/generateDockerfiles.sh#L24-L25), all of the runtime versions are based on `bullseye` _except_ for 3.9, which is based on `buster`.

cormacpayne:
> Do we even need https://github.com/microsoft/Oryx/blob/main/tests/Oryx.Tests.Common/ImageTestHelper.cs#L438-L498 anymore? Could we replace it with the constant dictionaries created here, so that we condense the number of places that need to be updated when we convert debian flavors?

@pauld-msft -- this is a good point, we shouldn't need to hard-code these values for the tests anymore. I updated the `constants.yaml` file again to include the `dynamic` runtime and include the OS prefix (`debian`) when declaring the flavor.

william-msft (dismissed):

william-msft (approved):

PR #1896, opened 2023-02-21
Williamhe/1732771 add python automation

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as [\#\<issue\_number\>.](https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1732771)


build/tools/Automation/Program.cs
```diff
@@ -29,6 +29,7 @@ private static async Task Main(string[] args)
                 .AddSingleton<IVersionService, VersionService>()
                 .AddSingleton<IYamlFileReaderService, YamlFileReaderService>()
                 .AddScoped<DotNet.DotNet>()
+                .AddScoped<Python.Python>()
```
cormacpayne, line 32: nit: how do you feel about renaming the `DotNet` and `Python` classes to something like `DotNetAutomator` or `DotNetProcessor` so that we can drop the namespace referenced as a prefix and just do `.AddScoped<DotNetAutomator>` ? (this also clears up what the responsibility of these classes)

build/tools/Automation/Python/Constants.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
```
cormacpayne, line 1: nit: all of the new files add should have a license header

build/tools/Automation/Python/Constants.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace Oryx.Microsoft.Automation.Python
+{
+    public class Constants
+    {
+        public const string ConstantsYaml = "constants.yaml";
```
cormacpayne, line 7: Can we reuse the `ConstantsYaml` property defined in the `Automation/Constants.cs` file instead of redefining it here?

build/tools/Automation/Python/Constants.cs
```diff
@@ -0,0 +1,17 @@
...
+        public const string PythonName = "python";
+        public const string VersionsToBuildTxt = "versionsToBuild.txt";
+        public const string YamlPythonKey = "python-versions";
+        public static readonly Dictionary<string, string> VersionGpgKeys = new Dictionary<string, string>()
+        {
+            { "3.10", "A035C8C19219BA821ECEA86B64E628F8D684696D" },
+            { "3.11", "A035C8C19219BA821ECEA86B64E628F8D684696D" },
```
cormacpayne, line 14: Are these values that we'll need to update for every new release of Python 3.x or are they placeholders?

build/tools/Automation/Python/Python.cs
```diff
@@ -0,0 +1,159 @@
...
+namespace Microsoft.Oryx.Automation.Python
+{
+    public class Python
+    {
+        private readonly IHttpClient httpClient;
+        private readonly ILogger logger;
+        private readonly IVersionService versionService;
+        private readonly IYamlFileReaderService yamlFileReaderService;
+
+        public Python(IHttpClient httpClient, ILogger logger, IVersionService versionService, IYamlFileReaderService yamlFileReaderService)
```
cormacpayne, line 25: nit: would you mind moving each argument to a different line?

```cs
public Python(
    IHttpClient httpClient,
    ILogger logger,
    IVersionService versionService,
    IYamlFileReaderService yamlFileReaderService)
```

build/tools/Automation/Python/Python.cs
```diff
@@ -0,0 +1,159 @@
...
+        {
+            this.httpClient = httpClient;
+            this.logger = logger;
+            this.versionService = versionService;
+            this.yamlFileReaderService = yamlFileReaderService;
+        }
+
+        public async Task RunAsync(string oryxRootPath)
```
cormacpayne, line 33: I made a similar comment in the refactor PR about not passing the `oryxRootPath` variable around, but thinking about it a little more, I think the correct route would be the following:

`IYamlFileService` accepts the `oryxRootPath` as a constructor parameter that we'll provide when setting up DI for that interface

```cs
 .AddSingleton<IYamlFileReaderService>(new YamlFileReaderService(oryxRootPath))
```

Then every call in the `YamlFileReaderService` will be in the scope of that provided root path, so we don't need to worry about providing it again as a part of the method signature for every single call.

Similarly, we should introduce another service interface that's responsible for reading/writing the `versionsToBuild.txt` files so that we can roll-out a similar pattern that avoids having to pass around a path.

build/tools/Automation/Python/Python.cs
```diff
@@ -0,0 +1,159 @@
...
+                string constantsYamlAbsolutePath = Path.Combine(oryxRootPath, "build", Constants.ConstantsYaml);
+                List<YamlConstants> yamlConstantsObjs = await this.yamlFileReaderService.ReadConstantsYamlFileAsync(constantsYamlAbsolutePath);
+
+                this.UpdateOryxConstantsForNewVersions(versionObjs, yamlConstantsObjs, oryxRootPath);
+            }
+        }
+
+        public async Task<List<VersionObj>> GetNewVersionObjsAsync()
+        {
+            var url = "https://www.python.org/api/v2/downloads/release/";
```
cormacpayne, line 48: Should we add this to the Python `Constants.cs` class?

build/tools/Automation/Python/Constants.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+
+namespace Oryx.Microsoft.Automation.Python
+{
+    public class Constants
```
cormacpayne, line 5: Are we able to rename this class to `PythonConstants.cs` so that we don't have to use a namespace prefix every time we want to reference a constant in this project? (similar comment for .NET)

build/tools/Automation/Python/Python.cs
```diff
@@ -0,0 +1,159 @@
...
+            url = "https://oryx-cdn.microsoft.io/python?restype=container&comp=list&include=metadata";
+            HashSet<string> oryxSdkVersions = await this.httpClient.GetOryxSdkVersionsAsync(url);
+
+            var versionObjs = new List<VersionObj>();
+            foreach (var release in releases)
+            {
+                string newVersion = release.Name.Replace("Python", string.Empty).Trim();
+
+                if (!release.PreRelease &&
+                    this.versionService.IsVersionWithinRange(newVersion, minVersion: "3.10.9") &&
```
cormacpayne, line 61: Any reason that we're hard-coding this version of Python here?

build/tools/Automation/Python/Python.cs
```diff
@@ -0,0 +1,159 @@
...
+                    !oryxSdkVersions.Contains(newVersion))
+                {
+                    Console.WriteLine($"newVersion {newVersion}");
```
cormacpayne, line 64: Are we able to add more context to this message that will be written to the console?

build/tools/Automation/Python/Python.cs
```diff
@@ -0,0 +1,159 @@
...
+                this.UpdateVersionsToBuildTxt(versionObj, oryxRootPath);
+            }
+
+            var serializer = new SerializerBuilder()
+                .WithNamingConvention(UnderscoredNamingConvention.Instance)
+                .Build();
+
+            var constantsYamlAbsolutePath = Path.Combine(oryxRootPath, "build", Constants.ConstantsYaml);
+            var stringResult = serializer.Serialize(yamlConstants);
+            File.WriteAllText(constantsYamlAbsolutePath, stringResult);
```
cormacpayne, line 96: I would expect all of this functionality to occur within the `IYamlFileReaderService` interface, although it may then need to be changed to something like `IYamlFileService` so it handles _both_ reading and writing

build/tools/Automation/Python/Python.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        private string GeneratePythonConstantKey(VersionObj versionObj)
+        {
+            string[] splitVersion = versionObj.Version.Split('.');
+            string majorVersion = splitVersion[0];
+            string minorVersion = splitVersion[1];
+
+            // prevent duplicate majorMinor where 11.0 and 1.10 both will generate a 110 key.
+            int majorVersionInt = int.Parse(majorVersion);
+            string majorMinor = majorVersionInt < 10 ? $"{majorVersion}{minorVersion}" : $"{majorVersion}Dot{minorVersion}";
```
cormacpayne, line 129: Does the key need to follow a specific format here? Any reason we can't just do `$"{majorVersion}_{minorVersion}"` ?

build/tools/Automation/Python/Python.cs
```diff
@@ -0,0 +1,159 @@
...
+                var versionsToBuildTxtAbsolutePath = Path.Combine(
+                    oryxRootPath, "platforms", Constants.PythonName, "versions", debianFlavor, Constants.VersionsToBuildTxt);
+                string line = $"\n{platformConstant.Version}, {platformConstant.GpgKey},";
+                File.AppendAllText(versionsToBuildTxtAbsolutePath, line);
+
+                // sort
+                Console.WriteLine($"[UpdateVersionsToBuildTxt] Updating {versionsToBuildTxtAbsolutePath}...");
+                var contents = File.ReadAllLines(versionsToBuildTxtAbsolutePath);
+                Array.Sort(contents);
+                File.WriteAllLines(versionsToBuildTxtAbsolutePath, contents.Distinct());
```
cormacpayne, line 155: Similar to the `IYamlFileService` comment I made above, this is where I would expect to make some call out to an `IFileService` interface to handle these types of read/write operations

william-msft: I've updated this into a `FileService`

build/tools/Automation/Python/Python.cs
william-msft: Updated to use `_`, since no specific format is needed.

build/tools/Automation/Python/Python.cs
william-msft: Yup! I've update to YamlFileService to include both read and write

build/tools/Automation/Python/Python.cs
william-msft: Updated the parameter with `this.pythonMinReleaseVersion` which contains the environment variable min version for python

build/tools/Automation/Python/Constants.cs
william-msft: Updated

build/tools/Automation/Python/Python.cs
william-msft: Yes, updated

build/tools/Automation/Python/Python.cs
william-msft: Updated to `YamlFileSerivce` and have introduced a `FileService`. Now `oryxRootPath` does not need to be passed to `RunAsync`

build/tools/Automation/Python/Python.cs
william-msft: Updated

build/tools/Automation/Python/Constants.cs
william-msft: Every couple minor versions we'll have to update these. For example, 3.10.x and 3.11.x have the same key.

This observation is from here:
https://www.python.org/downloads/
image

Are you thinking we should store these differently?

build/tools/Automation/Python/Constants.cs
william-msft: Yup! Updated here too

build/tools/Automation/Python/Constants.cs
william-msft: Updated.

build/tools/Automation/Program.cs
william-msft: Makes sense to me; I got this updated.

.github/workflows/automationTemplate.yaml
```diff
@@ -1,6 +1,9 @@
 name: Oryx-SDK-Runtime-Automation-Template
 
 on:
+  push:
+    branches:
+      - 'williamhe/**'
```
william-msft, line 6: Need this for a little longer to help me rollout new versions to sandbox storage account

daniv-msft (approved):
Looks good to me (apart from the `branches` field of course, but you're on it)! Thanks

PR #1902, opened 2023-02-23
Auto assign issues

harryli0108:
This PR is to add features that assign new/reopened github issues to current primary oncall engineer.


oncall/customRotation.txt
```diff
@@ -0,0 +1,4 @@
+#Enter a custom shift info in the following format:
+#{Github username} {Cover start date} {Cover end date}
+#Please make sure the date is in the format YYYY-MM-DD
+harryli0108 1999-02-01 1999-02-08
```
snehapar9, line 4: Probably missed this in your demo, but do we need to add all our schedules here? and curious why the start date is 1999?

harryli0108: this one is only used when say you're having days off during you oncall, and you need someone to cover your rotation. The reason why it has an entry in 1999 is because that one is an example to show the format of each entry without accidentally assigning issues to the wrong person.

snehapar9: Thanks for clarifying!

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+autoAssign() {
+  echo "Current oncall is $1"
+  response=`curl -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $2" https://api.github.com/repos/microsoft/Oryx/issues | jq '.[]| select(.assignee == null or .assignee == "")' | jq '.number'`
```
cormacpayne, line 4: Is this getting all issues in our repository that don't have an assignee?

Since we know that this workflow is being triggered _only_ when an issue has been (re)opened, we should be able to assign the current on-call _only_ to the corresponding issue.

From the workflow file itself, we should be able to use [`${{ github.event.issue.number }}`](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#issues) to get the issue number, which we can pass to this script and use in this `autoAssign` function one time rather than iterating over each unassigned issue.

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
cormacpayne, line 1: nit: we should create a `scripts` folder at the root of this repository that this `.sh` file lives in -- we can then rename it to `oncallAutoAssignIssue.sh`

oncall/customRotation.txt
```diff
@@ -0,0 +1,4 @@
+#Enter a custom shift info in the following format:
```
cormacpayne, line 1: I'm in the boat that we shouldn't have this file in our repository as it will force us to have to create a PR any time we want to make custom changes to the rotation -- ideally we would be able to move this to a [repository variable](https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository) that we can modify whenever we want independent of a PR.

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,51 @@
...
+    # Check if the current date is within the date range
+    if [ "$today" -ge "$startDate" ] && [ "$today" -le "$endDate" ]; then
+      echo "Custom rotation found"
+      autoAssign $name $token
+      exit 0
+    fi
+  fi
+done < "$file"
+
+oncallList=("waliMSFT" "cormacpayne" "harryli0108" "snehapar9" "pauld-msft" "william-msft")
```
cormacpayne, line 42: Similar comment to the one left below for `customRotation.txt` -- I think it would be great if we could have the `oncallList` and `anchorDate` as repository variables that are able to be modified at any instance in case the rotation drastically changes and/or we need to add/remove people. I just want to minimize the number of PRs we need to make to support this automated workflow moving forward.

oncall/customRotation.txt
cormacpayne, line 4: nit: I would make this line a comment and call out that it's an example:

```
# Example: harryli0108 1999-02-01 1999-02-08
```

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,51 @@
...
+  else
+    # Extract the name, start date, and end date from the line
+    name=$(echo $line | cut -d ' ' -f1)
+    startDate=$(echo $line | cut -d ' ' -f 2)
+    startDate=$(date -d $startDate +%s)
+    endDate=$(echo $line | cut -d ' ' -f 3)
+    endDate=$(date -d $endDate +%s)
+    # Check if the current date is within the date range
+    if [ "$today" -ge "$startDate" ] && [ "$today" -le "$endDate" ]; then
+      echo "Custom rotation found"
```
cormacpayne, line 35: I'd update this comment to call out that you found a user in the custom rotation that fits in the current window.

oncall/customRotation.txt
harryli0108: I took a look on the repo variable, and seems like it's viable only if either the repo is owned by us or the organization is owned by us. I believe we currently don't have access to change anything in the setting page.
Actually, if we have access to that page, we could also add tags to different issue templates for future automation use.

oncall/autoIssueAssign.sh
harryli0108: I think this one is getting the first 30 open issues.
I'm not very familiar with the webhook, but i'll look into it a bit more. Thanks for the suggestion!

oncall/autoIssueAssign.sh
cormacpayne: This comment still applies -- let's create a `scripts` folder and move this script there, renaming it to `oncallAutoAssignIssue.sh`

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
+autoAssign() {
+  curl -X POST "https://api.github.com/repos/microsoft/Oryx/issues/$3/assignees" -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $2" -d "{\"assignees\": ["\"$1\""]}"
```
cormacpayne, line 4: nit: let's assign the arguments to this function to variables so it's clear what each argument represents when provided to the `POST` request here

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,75 @@
...
+    shift
+    ;;
+  esac
+done
+eval set -- "$ONCALLS"
+
+today=$(date +%s)
+
+oncallArrLen=${#ONCALLS[@]}
+# achor Date is Feb 21 2022 0:00
```
cormacpayne, line 64: nit: typo: `achor` --> `anchor`

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,75 @@
...
+    --custom) # Custom shift flag
+    customRotation=$2
+    shift 2
+    ;;
+    --)
+    shift
+    break
+    ;;
+    *)
+    ONCALLS="$ONCALLS $1"
```
cormacpayne, line 54: I checked the Variables for the repository and it looks like the values provided to this list are our Microsoft aliases and not our GitHub aliases

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,75 @@
...
+  input=$1
+  today=$2
+  token=$3
+  issue=$4
+  currentOncall=$5
+  if [ -z "$input" ]; then
+    return 1
+  fi
+
+  IFS="," read -ra arr <<< "$1"
```
cormacpayne, line 17: nit: it may be worth putting a comment above this line that highlights the format of a line in the custom on-call rotation list

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,75 @@
...
+  startDate=$(date -d $startDate +%s)
+  endDate=$(date -d $endDate +%s)
+  if [ "$today" -ge "$startDate" ] && [ "$today" -le "$endDate" ]; then
+    echo "Custom shift found"
+    autoAssign $name $token $issue
+    return 0
+  else
+    autoAssign $currentOncall $token $issue
+    return 0
+  fi
```
cormacpayne, line 30: I think we should handle the case where multiple entries are provided in the custom on-call rotation list -- currently we're just checking the first line of the Variable.

I would suggest following something like the following pseudocode:

```
assignCustomOncall()
  customRotation = parseVariable()
  for each entry in customRotation
    if entry is in dateRange
      assignIssue()
      return
```

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
+autoAssign() {
```
cormacpayne, line 3: nit: let's rename this function to `assignIssue`

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,75 @@
...
+  curl -X POST "https://api.github.com/repos/microsoft/Oryx/issues/$3/assignees" -H "Accept: application/vnd.github+json" -H "Authorization: Bearer $2" -d "{\"assignees\": ["\"$1\""]}"
+}
+
+checkCustomOncall() {
```
cormacpayne, line 7: nit: let's rename this function to `tryAssignIssueToCustomOncall`

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,75 @@
...
+anchorDate=1677002400
+d=$(((today - anchorDate)/60/60/24/7))
+pos=`echo "$d%$oncallArrLen" | bc`
+currentOncall=`echo ${ONCALLS[$pos]}`
+
+if [[ -v customRotation ]]; then #check if custom rotation argument is present or not
+  checkCustomOncall $customRotation $today $authToken $issueNum $currentOncall
+else
+  autoAssign $currentOncall $authToken $issueNum
+fi
```
cormacpayne, line 74: We should update `checkCustomOncall` (or `assignIssueToCustomOncall` if you decide to rename it) to _exit_ `0` if it was able to assign the issue to one of the on-calls in the custom rotation, otherwise it should do nothing and we should call `autoAssign` (or `assignIssue`) separately.

```
if [[ -v customRotation ]]; then
  assignIssueToCustomOncall $customRotation $today $authToken $issueNum $currentOncall
fi

assignIssue $currentOncall $authToken $issueNum
```

oncall/autoIssueAssign.sh
cormacpayne, line 4: We should check the response status of this `POST` request since we're just shooting the request off and then assuming it worked based on the calls to this function down below. There may be an issue with the token being expired or an invalid user being provided from one of our Variables, so it would be best to check if the response of this `POST` was a `201`, log an error if not, and return an appropriate value from this function to signal to the caller if the request was successful or not.

cormacpayne (commented):
Some minor comments that just need to be cleaned up before we can roll this out

oncall/readme.md
```diff
@@ -0,0 +1,13 @@
+# Auto Issue Assign Instruction
+## Define Variables
+Make sure to define correct repository variables in repo setting page. Enter github username in `ONCALL_LIST`. Custom oncall info is in `CUSTOM_ONCALL_ROTATION` with the following format:
+
+`"{github-username},{start-date},{end-date}"`
```
pauld-msft, line 5: Are these start/end dates inclusive or exclusive? Trying to understand what we should put if there are 2 custom shifts back-to-back (for example)

oncall/autoIssueAssign.sh
```diff
@@ -0,0 +1,75 @@
...
+  case "$1" in
+    --token) #Github token flag
+    authToken=$2
+    shift 2
+    ;;
+    --issue) # Issue number flag
+    issueNum=$2
+    shift 2
+    ;;
+    --custom) # Custom shift flag
```
pauld-msft, line 45: Hey, so in the demo video there was a custom shift flag. Did you decide to do away with that in favor of a single specified custom shift?

If so, I would lean towards going back to the file approach and parsing through every line of the file, as cormac suggests. This would allow us to plan a few custom shifts in advance, especially if they are back-to-back without having any gap time

scripts/readme.md
```diff
@@ -0,0 +1,13 @@
+# Auto Issue Assign Instruction
+## Define Variables
+Make sure to define correct repository variables in repo setting page. Enter github username in `ONCALL_LIST`. Custom oncall info is in `CUSTOM_ONCALL_ROTATION` with the following format:
+
+`{github-username},{start-date},{end-date}`
+
+**Do not put space between commas!**
+
+Add new entries by starting a new line.
```
pauld-msft, line 9: here you mention that we should add new entries by starting a new line, but in the code it looks like new entries are determined through splitting the array by `;`? Do you mind fixing this readme to have the correct format, and also include an example of a multiple date entry?

`{github-username-1},{start-date},{end-date};{github-username-2},{start-date},{end-date}`

Also, I'm still unclear from this readme as to whether these dates are inclusive or exclusive, as i mentioned in https://github.com/microsoft/Oryx/pull/1902#discussion_r1124513804

scripts/oncallAutoAssignIssue.sh
```diff
@@ -0,0 +1,97 @@
...
+  fi
+}
+
+tryAssignIssueToCustomOncall() {
+  today=$1
+  token=$2
+  issue=$3
+  input="$CUSTOM_ONCALL_ROTATION"
+  input="${input//$'\n'/''}"
+  input="${input//$'\r'/';'}"
```
cormacpayne, line 31: We should make a comment here about converting the CRLF to `;` for easier string splitting later on.

I know @pauld-msft made a comment about why we're splitting on the `;` character when our documentation says to split on new lines, so documenting that we first do a conversion to separate the entries by the `;` character should clear this up.

scripts/oncallAutoAssignIssue.sh
```diff
@@ -0,0 +1,97 @@
...
+  if [ -z "$input" ]; then
+    return 1
+  fi
+  # Custom oncall rotation defined in repo variable. Using format:
+  # {github-username},{start-date},{end-date}
+
+  # Devide each entry by ";"
```
cormacpayne, line 38: nit: typo, `devide` --> `divide`

Same comment down below in the next in-line comment you've made

cormacpayne (commented):
Two minor comments, otherwise this should be good to merge

oncall/readme.md
harryli0108: Yes so the dates are inclusive date. In a back-to-back scenario, if say first person has a end date on March 6 and the second person put March 6 as the start date, it'll assign issues to whichever is on the top of the list.

scripts/readme.md
harryli0108: thanks! I missed that earlier on.

scripts/oncallAutoAssignIssue.sh
pauld-msft: ah yep i missed this. 

Also, seems a little strange that we are expecting the environment variable to have newline characters in it? Is there a reason we do it this way instead of just using `;` as a separator?

Also, maybe I'm missing something, but are we sure that we can expect `\r` as a newline character for all cases? Wouldn't this make a Unix file with `LF` line endings (instead of Windows `CRLF`) not be split properly?

harryli0108: unfortunately yes. so upon checking on friday, it seems like it uses CRLF instead of just `\n` or `\r`. this should be a weird behavior from how github treated new lines. but i think we could try and match for `\n\r` first and replace it to `;`. Otherwise, replace `\n` or `\r` with `;`.

cormacpayne: > Also, seems a little strange that we are expecting the environment variable to have newline characters in it? Is there a reason we do it this way instead of just using ; as a separator?

The main reason was for "easy-of-use"; someone looking to add a new entry to the custom on-call rotation would edit the `CUSTOM_ONCALL_ROTATION` variable in the repository's settings and just add their entry to the next line and be done. Similarly, someone looking to remove outdated entries could quickly sift through the entries since there would be one on each line. Having the semicolons between each entry wouldn't make it impossible to read, but it may make it just a tad bit more difficult to differentiate between the entries.

> Also, maybe I'm missing something, but are we sure that we can expect \r as a newline character for all cases? Wouldn't this make a Unix file with LF line endings (instead of Windows CRLF) not be split properly? I guess I'm just not sure how we are getting the newline characters into the env var, and would prefer an OS agnostic way of doing that.

We shouldn't have to worry about different line endings based on the OS since this entry is set within our repository's settings for the `CUSTOM_ONCALL_ROTATION` variable, so as long as GitHub doesn't change how they're resolving new lines in their own Variables fields, we should be OK to stick with this. If they were to change it for some reason, I'd be OK with switching to the `;` separator for our entries so that we're agnostic to any changes they make on their end.

cormacpayne (approved):
:shipit:

scripts/oncallAutoAssignIssue.sh
pauld-msft: Thanks both for the clarifications!

pauld-msft (approved):
:shipit:

PR #1907, opened 2023-02-27
Fix spacing

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as: since YamlDoNet in Automation expects a particular indentation

### DIFF
--- a/build/constants.yaml
+++ b/build/constants.yaml
@@ -430,7 +430,7 @@
 - name: staging-runtime-constants
   constants:
     dotnetcore-staging-runtime-versions:
-    -
+    - 
   outputs:
   - type: csharp
     directory: src/BuildScriptGenerator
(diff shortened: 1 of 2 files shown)


build/constants.yaml
```diff
@@ -430,7 +430,7 @@
 - name: staging-runtime-constants
   constants:
     dotnetcore-staging-runtime-versions:
-    -
+    -
```
daniv-msft, line 433: Could you please provide more context? Was the version without the space not a valid YAML?

william-msft: It's a valid YAML. But since Automation checks changes to `constants.yaml` for detecting if new versions are available. Having a space discrepancy in `constants.yaml` leads Automation to think there are new versions.

william-msft: YamlDotNet library, in Automation, expects there to be a space after the `-`

daniv-msft: Thanks for clarifying!

daniv-msft (approved):

PR #1916, opened 2023-03-01
Williamhe/fix python platform binary pipeline

william-msft:
  - Fixes [Oryx-PlatformBinary-Python](https://devdiv.visualstudio.com/DevDiv/_build?definitionId=13047&_a=summary)
  - Moving comments to the top. Since comments in the middle of `apt-get` is not allowed. 


platforms/python/prereqs/build.sh
```diff
@@ -31,8 +33,6 @@ PYTHON_GET_PIP_URL="https://github.com/pypa/get-pip/raw/3cb8888cc2869620f57d5d2d
         libbluetooth-dev \
         tk-dev \
         uuid-dev \
-        # Adding additional python packages to support all optional python modules:
-        # https://devguide.python.org/getting-started/setup-building/index.html#install-dependencies
```
pauld-msft, line 35: do you know why this failed, while https://github.com/microsoft/Oryx/blob/main/images/build/Dockerfiles/gitHubActions.Dockerfile#L6-L52 does not cause issues?

william-msft: Yes, you are referencing a Dockerfile. So `#` comments: https://docs.docker.com/engine/reference/builder/
Are ignored during docker build process.

daniv-msft (approved):

PR #1917, opened 2023-03-02
Update cli.Dockerfile

snehapar9:
This PR removes the redundant `DEBIAN_FLAVOR` ARG [here](https://github.com/microsoft/Oryx/blob/main/images/build/Dockerfiles/cli.Dockerfile#L1).

### DIFF
--- a/images/build/Dockerfiles/cli.Dockerfile
+++ b/images/build/Dockerfiles/cli.Dockerfile
@@ -1,5 +1,3 @@
-ARG DEBIAN_FLAVOR
-
 # Use the curl flavor of buildpack-deps as the base image, which is lighter than the standard flavor; more information here: https://hub.docker.com/_/buildpack-deps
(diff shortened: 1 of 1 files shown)


images/build/Dockerfiles/cli.Dockerfile
```diff
@@ -1,5 +1,3 @@
-ARG DEBIAN_FLAVOR
```
pauld-msft, line 1: Isn't this DEBIAN_FLAVOR used in the actual FROM command?
`FROM buildpack-deps:${DEBIAN_FLAVOR}-curl as main`

Also, i'm not sure that it is redundant, based on https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact ? Maybe I'm not understanding correctly though

snehapar9: Yep just realized this is being used in the FROM command. Apologies!

PR #1922, opened 2023-03-07
Fix auto assign issue in usual scenario

harryli0108:
Fix the issue when the shell script parses the hardcoded oncall list. Change the method of feeding the oncall list from positional arguments to an environment variable in the format:
`{github-username-1},{github-username-2},...`


pauld-msft (approved):
:shipit:

PR #1929, opened 2023-03-15
disable oryx pip upgrade changes

waliMSFT:
There appears to be a bug with pip that causes issues on occasion when we specify --upgrade and --target flags.

for more context:
https://teams.microsoft.com/_?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47#/conversations/unknown?threadId=19:87805dd03a82426e90204b154fa0844b@thread.tacv2&messageId=1677613976504&replyChainId=1677613976504&ctx=channel

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -125,9 +125,9 @@ fi
         echo
         echo Running pip install...
         START_TIME=$SECONDS
-        InstallCommand="$python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" --upgrade | ts $TS_FMT"
+        InstallCommand="$python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" {{ oryxPipUpgradeCommand }} | ts $TS_FMT"
         printf %s " , $InstallCommand" >> "$COMMAND_MANIFEST_FILE"
-        output=$( ( $python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" --upgrade | ts $TS_FMT; exit ${PIPESTATUS[0]} ) 2>&1; exit ${PIPESTATUS[0]} )
+        output=$( ( $python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" {{ oryxPipUpgradeCommand }} | ts $TS_FMT; exit ${PIPESTATUS[0]} ) 2>&1; exit ${PIPESTATUS[0]} )
         pipInstallExitCode=${PIPESTATUS[0]}
 
         ELAPSED_TIME=$(($SECONDS - $START_TIME))
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -125,9 +125,9 @@ fi
         echo
         echo Running pip install...
         START_TIME=$SECONDS
-        InstallCommand="$python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" {{ pipUpgradeFlag }} | ts $TS_FMT"
+        InstallCommand="$python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" {{ PipUpgradeFlag }} | ts $TS_FMT"
         printf %s " , $InstallCommand" >> "$COMMAND_MANIFEST_FILE"
-        output=$( ( $python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" {{ pipUpgradeFlag }} | ts $TS_FMT; exit ${PIPESTATUS[0]} ) 2>&1; exit ${PIPESTATUS[0]} )
+        output=$( ( $python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" {{ PipUpgradeFlag }} | ts $TS_FMT; exit ${PIPESTATUS[0]} ) 2>&1; exit ${PIPESTATUS[0]} )
         pipInstallExitCode=${PIPESTATUS[0]}
 
         ELAPSED_TIME=$(($SECONDS - $START_TIME))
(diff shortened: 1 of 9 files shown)


src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -275,7 +282,8 @@ public PlatformDetectorResult Detect(RepositoryContext context)
                 pythonVersion: pythonVersion,
                 pythonBuildCommandsFileName: pythonBuildCommandsFile,
                 pythonPackageWheelProperty: pythonPackageWheelType,
-                customRequirementsTxtPath: customRequirementsTxtPath);
+                customRequirementsTxtPath: customRequirementsTxtPath,
+                oryxPipUpgradeCommand:pipUpgrade);
```
cormacpayne, line 286: nit: space between the `:` and the `pipUpgrade`  variable

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -125,9 +125,9 @@ fi
         echo
         echo Running pip install...
         START_TIME=$SECONDS
-        InstallCommand="$python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" --upgrade | ts $TS_FMT"
+        InstallCommand="$python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" {{ oryxPipUpgradeCommand }} | ts $TS_FMT"
```
cormacpayne, line 128: The two `oryxPipUpgradeCommand` references that you've added in this file need to be changed to match the property name in the `PythonBashBuildSnippetProperties`, which is currently `OryxPipCommand` (or `PipUpgradeFlag` if we change the property name), so that they are properly rendered when the script is generated

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -125,9 +125,9 @@ fi
...
         printf %s " , $InstallCommand" >> "$COMMAND_MANIFEST_FILE"
-        output=$( ( $python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" --upgrade | ts $TS_FMT; exit ${PIPESTATUS[0]} ) 2>&1; exit ${PIPESTATUS[0]} )
+        output=$( ( $python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" {{ oryxPipUpgradeCommand }} | ts $TS_FMT; exit ${PIPESTATUS[0]} ) 2>&1; exit ${PIPESTATUS[0]} )
         pipInstallExitCode=${PIPESTATUS[0]}
 
         ELAPSED_TIME=$(($SECONDS - $START_TIME))
```
cormacpayne, line 133: nit: there are also references to the `--upgrade` flag in `pip install` commands down below on lines 146 and 148 -- do we need to use this same logic there?

cormacpayne (changes requested):

src/BuildScriptGenerator/Python/PythonPlatform.cs
```diff
@@ -180,6 +185,8 @@ public PlatformDetectorResult Detect(RepositoryContext context)
                 Path.Combine(context.SourceRepo.RootPath, pythonBuildCommandsFile) :
                 Path.Combine(this.commonOptions.ManifestDir, pythonBuildCommandsFile);
             manifestFileProperties[nameof(pythonBuildCommandsFile)] = pythonBuildCommandsFile;
+            //if OryxDisablePipUpgrade is ture then we do not upgrade pip hence is set to the empty string.
+            var pipUpgrade = this.commonOptions.OryxDisablePipUpgrade? string.Empty : pipUpgradeCommand;
```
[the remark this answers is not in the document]
william-msft: @waliMSFT looks like the pipeline is catching these issues:
image
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7473483&view=logs&j=b2b771d4-9387-5666-709b-f98dda7c9c3e&t=62e5562e-7a14-50b4-101c-2eef3301ddfa&l=48

building the solution locally will help solve

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
waliMSFT: sorry Cormac, I have missed this comment to reply earlier.
We do not need to use the same logic here. The reason we added the logic is because `pip` does not go well along with `target` and `upgrade` together.

pauld-msft (changes requested):
Hey @waliMSFT, would you be able to add some tests for this feature? It would be great to have a few tests that are able to at least confirm that the build script contains the pip install with the --upgrade flag normally, but does not contain that when we pass the env to disable?

It would probably be best to add tests at least here to show that it works with just the build snippet props:
- https://github.com/microsoft/Oryx/blob/main/tests/BuildScriptGenerator.Tests/Python/PythonBashBuildSnippetTest.cs

And optionally here to show that it works when we pass in the environment variable to the build image:
- https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs

If you don't do the second test suggested here, could you at least show a quick example of the TeradataSQL project that the customer shared with us working once you pass in the correct env var?

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
cormacpayne: This comment still applies -- the value of `PipUpgradeFlag` in the `PythonBashBuildSnippetProperties` won't be rendered in this template file if the property name in the template file does not match.

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
```diff
@@ -1639,6 +1639,47 @@ public void JamSpell_CanBe_Installed_In_The_BuildImage(string tagName)
                 result.GetDebugInfo());
         }
 
+        [Theory]
+        [InlineData("true")]
+        [InlineData("false")]
+        private void DisablePipUpgradeFlag(string pipUpgradeFlag)
+        {
+            // Arrange
+
```
pauld-msft, line 1648: nit, remove empty line here

tests/BuildScriptGenerator.Tests/Python/PythonBashBuildSnippetTest.cs
```diff
@@ -186,5 +186,61 @@ public void GeneratedSnippet_ContainsPackageWheelType_When_PackageCommandAndPack
             Assert.Contains("Creating universal package wheel", text);
             Assert.Contains("setup.py sdist --formats=gztar,zip,tar bdist_wheel --universal", text);
         }
+
+        [Fact]
+        public void GeneratedSnippet_DisablePipUpgradeFlag_IfPipUpgradeFlag_IsNull()
```
pauld-msft, line 191: Technically the PipUpgradeFlag is empty, not null, so lets update this method name.

pauld-msft (commented):
Thanks for adding the tests! Left a few more comments, but otherwise looks pretty good to me

waliMSFT:
> 

thanks for suggesting to add the test. It actually caught the issue that was happening locally. I needed to fix it for future purposes..

pauld-msft (approved):
LGTM. I think you will still Cormac to approve since he requested changes as well

cormacpayne (approved):
:shipit:

PR #1931, opened 2023-03-15
New Python Sdks 2023-03-15

william-msft:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


build/constants.yaml
```diff
@@ -145,8 +145,8 @@
     python37-version: 3.7.15
     python38-version: 3.8.16
     python39-version: 3.9.15
-    python310-version: 3.10.8
-    python311-version: 3.11.0
+    python310-version: 3.10.9
```
daniv-msft, line 148: @william-msft Would you know why 3.10.9 is mentioned here when 3.10.10 seems to be available as well?

william-msft: This is likely due because 3.10.9 being ingested last when retrieving from the releases. One quick fix would be to bump up the `PYTHON_MIN_RELEASE_VERSION`

daniv-msft (approved):

PR #1934, opened 2023-03-16
Add builder code and release pipelines

pauld-msft:
Work item: https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1756476

This pull request has a few different main goals:
1. Create a new `cli-builder` image that prebakes the apt-get packages instead of installing them at runtime. This was copied over from the `corm/builder-skeleton-image` branch. The image can be found in `images/build/Dockerfiles/cliBuilder.Dockerfile`, and a number of the changes in this PR are related to adding it to our build, test, and release pipelines.
2. Move the builder code over from the `oryx-builder` repo to here. Much of the code in the `/builders` directory is copied over, with the exception of:
- `builders/README.md`
- `builders/base/buildBaseBuilder.sh`
- `builders/base/builder/template.builder.toml`
- `builders/base/buildpack/template.buildpack.toml`
- first 3 lines of `builders/base/stack/Dockerfile`
- `builders/container-apps/buildCappsBuilder.sh`
- first 2 lines of `builders/container-apps/Dockerfile`
3. Add the builders to our CI pipelines. The order that we need to build these images is:
`oryx/cli:builder-{buildNum}` -> `oryx/builder:{buildNum}` -> `oryx/builder:capps-{buildNum}`


builders/base/buildBaseBuilder.sh
```diff
@@ -0,0 +1,147 @@
...
+buildPackImage="$destinationFqdn/$destinationRepo:buildpack-$builderImageVersion"
+echo
+echo "Packaging buildpack image: $buildPackImage"
+echo
+pack buildpack package $buildPackImage --config $SCRIPT_DIR/packaged-buildpack/package.toml
+echo "$buildPackImage" >> $ACR_BUILDER_IMAGES_ARTIFACTS_FILE
+echo "-------------------------------------------------"
+
+# replace image tags with their MCR equivalent
+mcrBaseImage=$(replaceRepoWithBuilderMCR "$baseImage")
```
pauld-msft, line 126: So this is probably the most confusing part of this whole release pipeline for me. Maybe @cormacpayne can help confirm that this is actually needed...

Basically, when we build this base builder image, we build it using `stack` images that exist in the oryxdevmcr registry. 
_(This is so that in the future, if we would like to add a test suite or something for these images, we can do so. It also allows us to not perform any builds in the `Release` stage of the CI pipeline, as that stage currently just pulls and retags and pushes images for the most part.)_
Currently the builder image was created using `stack` images that exist in the MCR repo, as noted [here](https://github.com/microsoft/Oryx/pull/1934/files#diff-6757db7154ebae953a9876f3831c460cc305817fb5270bbfeff4f4cff8455b26R67). So what these `replaceRepoWith*MCR` methods aim to do is retag the stack images with the MCR repo so they can be used locally with that tag, since _eventually_ we will push those images to MCR anyways in the Release section of the CI pipelines.

My question is whether this is even needed? Is there an issue with just using oryxdevmcr images as the `stack` as long as the final images end up in MCR?

build/buildBuildImages.sh
```diff
@@ -438,6 +438,40 @@ function buildCliImage() {
...
+	buildBuildScriptGeneratorImage
+	
+	local debianFlavor=$1
+	local builtImageRepo="$ACR_CLI_BUILD_IMAGE_REPO"
+	local devImageRepo="$DEVBOX_CLI_BUILD_IMAGE_REPO"
+
+	if [ -z "$debianFlavor" ]; then
+		debianFlavor="stretch"
+	fi
+	imageTag="builder-debian-$debianFlavor"
```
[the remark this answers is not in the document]
pauld-msft: I did this for consistency with the rest of the script, but I can try to update to be a more generic OS type parameter

build/buildBuildImages.sh
```diff
@@ -438,6 +438,40 @@ function buildCliImage() {
...
+		--build-arg DEBIAN_FLAVOR=$debianFlavor \
+		--label com.microsoft.oryx="$labelContent" \
+		-f "$BUILD_IMAGES_CLI_BUILDER_DOCKERFILE" \
+		.
+
+	createImageNameWithReleaseTag $builtImageName
+
+	echo
+	echo "$builtImageName image history"
+	docker history $builtImageName
```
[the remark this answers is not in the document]
pauld-msft: This is done in all of the other functions for building images, so i wanted it to be consistent. I've used it to make sure the right set of images was being built in the pipeline jobs before

builders/base/README.md
```diff
@@ -0,0 +1,86 @@
...
+### Steps for releasing new builder
+
+#### Prerequisites
+
+You must have access to the `oryxprodmcr` ACR instance in the Oryx production subscription account and have logged in
+locally to `docker` with this ACR instance's credentials.
+
+#### Build CLI image
+
+_Note_: The CLI Builder image used has its OS packages pre-baked, since the builder **CAN NOT** run as a root user and
```
daniv-msft, line 12: Nit: it might be worth completing the sentence to say that we cannot do that because the builder will be running in an unprivileged environment. I don't believe it's a fundamental limitation of Buildpack or something, more just the place where we'll run.

pauld-msft: So actually the CNB docs state that the CNB_USER can not be a root user, https://buildpacks.io/docs/operator-guide/create-a-stack/#specification. I've added a link to that doc to the README

`
The CNB_USER_ID is the UID of the user as which the detect and build steps are run. The given user MUST NOT be a root user and have it’s home directly writeable. CNB_GROUP_ID is the primary GID of the above user.
`

builders/container-apps/startup-script.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Extract app code to CNB_APP_DIR directory.
+echo "Found app source at '$COMPRESSED_APP_LOCATION'. Extracting to $CNB_APP_DIR"
+mkdir -p $CNB_APP_DIR
+cd $CNB_APP_DIR
+tar -xzf "$COMPRESSED_APP_LOCATION"
+
+# Detecting runtime stack and setting correct oryx runtime image.
+runtime_stack=$(oryx dockerfile . | head -n 1 | sed 's/ARG RUNTIME=//')
+export CNB_RUN_IMAGE="mcr.microsoft.com/oryx/$runtime_stack"
```
cormacpayne, line 22: nit: any thoughts on a fallback plan we should have if this `oryx dockerfile` command fails or isn't able to detect a suitable platform for the given application?

cormacpayne (approved):

builders/base/buildBaseBuilder.sh
```diff
@@ -0,0 +1,150 @@
...
+    IFS=':' read -ra SPLIT_IMAGE_NAME <<< "$imageName"
+    local repo="${SPLIT_IMAGE_NAME[0]}"
+    local tag="${SPLIT_IMAGE_NAME[1]}"
+    local newImage="$newRepo:$tag"
+    docker tag "$imageName" "$newImage"
+    echo $newImage
+}
+
+if [ -z $cliBuilderImage ]; then
+    cliBuilderImage="$destinationFqdn/public/oryx/cli:builder-debian-buster-$builderImageVersion"
```
daniv-msft, line 72: Could you please clarify why we default to the buster version? Any reasons not to use Bullseye?

builders/base/builder/builder.toml
```diff
@@ -0,0 +1,16 @@
+[[buildpacks]]
```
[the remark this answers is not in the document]
pauld-msft: Good point. I'll go ahead and add in some gitignores to ignore the actual builder.toml files, since you're right that theyd get generated whenever someone runs the scripts locally

builders/container-apps/startup-script.sh
pauld-msft: yeah this actually brings up a few good points. First, I know that we are planning on allowing customers to be able to manually override the runtime platform and version with some parameters that will be passed in, so that update should help customers not be blocked by oryx's failures to detect a language. As far as a default / fallback, I'm not quite sure. Node is probably the highest usage? So maybe that? Or we just hard fail and tell them to override the platform and version manually

builders/base/buildBaseBuilder.sh
pauld-msft: I defaulted to buster since thats what the current builder is based on. I think that when @cormacpayne originally made it, we didn't have a bullseye CLI image? 

Shouldn't be incredibly difficult to swap out the base OS, probably just a few find and replaces (as long as the tests still pass w/ bullseye). Do you think that would be worthwhile to do?

builders/base/README.md
```diff
@@ -0,0 +1,87 @@
+## Oryx Builder
+
+### Steps for releasing new builder
+
+#### Prerequisites
+
+You must have access to the `oryxprodmcr` ACR instance in the Oryx production subscription account and have logged in
```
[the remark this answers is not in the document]
pauld-msft: Kept them around, but clarified its for manual pushes, and added some section above for the offiical release with Oryx-CI and how to locally test

builders/base/buildpack/bin/template.build
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+set -eo pipefail
```
william-msft, line 2: Nit: would be nice to add `xu` --> `set -eoux pipefail` to all the bash script. Makes bash debugging easier: https://gist.github.com/mohanpedala/1e2ff5661761d3abd0385e8223e16425

images/build/Dockerfiles/cliBuilder.Dockerfile
```diff
@@ -0,0 +1,157 @@
...
+        libc6 \
+        libgcc1 \
+        libgssapi-krb5-2 \
+        libstdc++6 \
+        zlib1g \
+        # For .NET Core 1.1 (I believe we can remove this, .NET Core 1.1 has been 'End of Support" for three years)
+        libuuid1 \
+        libunwind8 \
+        rsync \
+        libgdiplus \
```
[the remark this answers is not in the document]
pauld-msft: Honestly, I've got no idea. @cormacpayne any thoughts?

builders/base/buildpack/bin/template.build
pauld-msft: so the `u` actually throws errors whenever we add the __Variables file, since not all of the environment variables are necessarily set locally (some are pipeline variables, like BUILD_BUILDNUMBER). Added the others to most scripts, besides the startup script which contains secrets so we don't want to be printing that out

builders/base/buildBaseBuilder.sh
cormacpayne: Missed this comment when reviewing the PR yesterday -- this is an interesting point because components of the builder depend on specific versions of published images, and it was easy to point to the MCR version of each image when changes were made manually.

As long as the images we're referencing from the `oryxdevmcr` registry are baked along with the builder and there's no additional pulling that needs to be done for these images, then we should be OK. We run into issues when the builder is on MCR and references an `oryxdevmcr` that isn't baked with it and we attempt to pull it down, which causes an authentication issue since `oryxdevmcr` is private (or rather doesn't have anonymous pull enabled).

If it's required that each image referenced amongst the different components must be publicly accessible, then we need to ensure the order of operations is correct here, ideally:
- Build and publish stacks to MCR
- Build and publish buildpack to MCR
  - AFAIK, buildpack doesn't reference specific images, but an `id` that is universal for the stack images
- Update the builder to reference the MCR buildpack and stacks
- Build and publish the builder to MCR

builders/container-apps/startup-script.sh
cormacpayne: I'm in the boat that we should fail fast and notify the user to the best of our ability that we were unable to detect a suitable platform version for their application.

Although now that I'm thinking about it, that's something that we should be doing in the `detect` script of the actual buildpack, so the question then becomes: is there any difference in detection logic between `oryx detect` and `oryx dockerfile`? I'd assume no since they _should_ be using the same library, but I wonder if there's a difference between "we've detected a platform" (`oryx detect`) and "we've detected a specific platform version to use" (`oryx dockerfile`).

builders/base/buildBaseBuilder.sh
cormacpayne: That's correct -- there was no bullseye flavor of the CLI image published out of our repository when the builder was first iterated on. If possible, I think we should migrate to bullseye since it's the only stable version of Debian released (after buster moved to LTS a few months ago).

images/build/Dockerfiles/cliBuilder.Dockerfile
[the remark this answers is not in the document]
cormacpayne: I assume this comment is about the .NET Core 1.1 comment, which applies to the two libraries below it (according to [this original commit](https://github.com/microsoft/Oryx/commit/03e7f76590d343c8d34907519b47b3f3bee8c405)) -- I would like to think we're OK to remove them since we don't support .NET Core 1.1 anymore, but those two packages are installed as a part of other build Dockerfiles (such as GitHub Actions), so it may be worth taking them out of those Dockerfiles as well and seeing if any tests fail? 🤔

builders/container-apps/startup-script.sh
pauld-msft: So afaik, the `detect` script of the buildpack is purely for detecting which buildpacks to use, in case there are more than one buildpack, correct? So in our case where we only have a single buildpack, there really isn't any need for the `detect` phase to have any logic since it will ALWAYS choose the oryx buildpack that we made, right?

I've tried to implement it so that we are able to set the run image that is used within the `detect` phase of buildpack, but that didn't work which is why I needed to set it before we even run the buildpack phases.

But yeah, I agree about failing fast. Made https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1775134 to finalize the capps builder (since its current state was just made for the demo)

cormacpayne: > So afaik, the detect script of the buildpack is purely for detecting which buildpacks to use, in case there are more than one buildpack, correct?

This is correct, but even though we only have a single buildpack in our builder, I think it would make sense to have that phase "fail" if the application provided by the user wouldn't be successfully built by our corresponding `build` script. I think it just comes down to semantics since we're OK either way, it's just a matter of where we want the "failed message" to pop up for the user. I'm honestly not sure if we have the ability to customize the message seen by the user if no compatible buildpack is found for the provided application source, so I think we're good with what you have now.

builders/base/buildBaseBuilder.sh
pauld-msft: Ok, I'll work on migrating to bullseye. Thanks for the input!

images/build/Dockerfiles/cliBuilder.Dockerfile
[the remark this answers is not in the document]
pauld-msft: Ok, I'll do some testing around with the removal of these. Thanks!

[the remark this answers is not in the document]
pauld-msft: Looks like we still need libunwind2 according to https://learn.microsoft.com/en-us/dotnet/core/install/linux-ubuntu-2204#dependencies, but i'm trying to remove libuuid1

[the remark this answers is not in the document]
william-msft: Sounds good! Thanks for updating Paul!

builders/base/buildBaseBuilder.sh
pauld-msft: Thanks for your comment! Yeah i feel like
`If it's required that each image referenced amongst the different components must be publicly accessible` 
is the crux of the question. I haven't found anything quite yet in CNB docs about whether the various stack images are bundled in with the builder image, or pulled dynamically during runtime. Going to try and do a wee bit of testing to see if I can figure it out

pauld-msft: Ok testing complete here. Ran:
`pack build pauldtestacr.azurecr.io/testing/container-app:1234 --path ./tests/SampleApps/DotNetCore/NetCore6PreviewMvcApp --builder oryxdevmcr.azurecr.io/public/oryx/builder:Oryx-Validation.20230317.7 --run-image mcr.microsoft.com/oryx/dotnetcore:6.0`

That command uses the image that this CI pipeline produced. Specifically, the stack images that were specified were only tagged with `mcr.microsoft.com` locally in the validation job, and they were never actually pushed to MCR. So if the builder was installing them dynamically during runtime, I should have seen it fail to pull the non-existant images from `mcr.microsoft.com`. Since that didn't happen, I'm concluding that the stack is pre-baked, and I can safely remove the re-tagging logic without ill effects. Does that sound right?

cormacpayne: That sounds right to me -- my original assumption was that the builder is bundled with all of the images it needs to perform the application build, so if the builder image was built with context to all of these component images, they should be bundled with it (also based on image sizes it made sense the builder included everything else)

images/build/Dockerfiles/cliBuilder.Dockerfile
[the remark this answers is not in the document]
pauld-msft: Had to move around some other packages (for some reason we were only installing PHP deps for buster), but it seems like this was safely removed. Thanks for bringing it up!

william-msft (approved):

images/build/Dockerfiles/cliBuilder.Dockerfile
```diff
@@ -0,0 +1,154 @@
+ARG DEBIAN_FLAVOR
```
pauld-msft, line 1: Diff from current CLI image:
image
image

PR #1939, opened 2023-03-22
added ORYX-SDK-STAGING-PRIVATE-SAS-TOKEN as env variable

waliMSFT:


pauld-msft (changes requested):
I think you will need to add an additional condition to the script in order to parse the SAS TOKEN: https://github.com/microsoft/Oryx/blob/main/vsts/scripts/copySdksFromProdToStorageAccount.sh#L84-L85

Also, preference to rename the env var from `ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN` to `STAGING_SDK_STORAGE_SAS_TOKEN`, so it matches with other names here, and gives more info about which storage account it relates to.

PR #1941, opened 2023-03-22
Williamhe/update automation base tag 

william-msft:
- [x]  Minor update on how the runtime-base-tag in constants.yaml for Automation is updated. DotNet has a [specific format](https://github.com/microsoft/Oryx/blob/main/build/constants.yaml#L63) `dot-net-core-run-time-versions`. However all other platforms follow a different format, `${platform}-versions` [for example](https://github.com/microsoft/Oryx/blob/main/build/constants.yaml#L139). This PR addresses the difference so the correct base-tag can be updated.


daniv-msft (approved):
Looks good to me, thanks

PR #1889, opened 2023-02-17
Update Go and Node Versions

harryli0108:
This PR is updating:

- Node:
  - current -> 14.21.2
  - current -> 16.19.0
  - current -> 18.14.0
- GO
  - current -> 1.18.10
  - current -> 1.19.5

PR #1955, opened 2023-04-13
Wali/support new version of php node go 1786787

waliMSFT:
version to support:

PHP | 8.0.28 
PHP | 8.1.17 
PHP | 8.2.4 
NODE | 12.22.12 
NODE | 14.21.3
NODE | 16.20.0 
NODE | 18.15.0
GO | 1.19.7


build/constants.yaml
```diff
@@ -297,7 +297,7 @@
       directory: build
       file-name-prefix: __
     - type: csharp
-      directory: src/BuildScriptGenerator/Node
+      directory: src/BuildScriptGenerator/Node\
```
snehapar9, line 300: Do we need this additional "\" here?

tests/Oryx.BuildImage.Tests/Node/NodeDynamicInstallationTest.cs
```diff
@@ -28,9 +28,9 @@ public NodeDynamicInstallationTest(ITestOutputHelper output) : base(output)
             {
                 var data = new TheoryData<string, string>();
                 var imageTestHelper = new ImageTestHelper();
-                data.Add("12.22.11", imageTestHelper.GetGitHubActionsBuildImage());
-                data.Add("14.19.1", imageTestHelper.GetGitHubActionsBuildImage());
-                data.Add("16.14.2", imageTestHelper.GetGitHubActionsBuildImage());
+                data.Add("12.22.12", imageTestHelper.GetGitHubActionsBuildImage());
```
snehapar9, line 31: We could add a new constant `final-stretch-node12-version` [here](https://github.com/microsoft/Oryx/blob/main/build/constants.yaml#L114) and run the ./build/generateConstants.sh script. That way, we would be able to use the constant for node 12 instead of hardcoding to be consistent with what we are doing for node 14 and node 16.

tests/Oryx.RuntimeImage.Tests/Php-fpm/PhpFpmImageTest.cs
```diff
@@ -178,7 +178,6 @@ public void Redis_IsInstalled(string imageTag)
         [InlineData("8.2-fpm")]
         [InlineData("8.1-fpm")]
         [InlineData("8.0-fpm")]
-        [InlineData("7.4-fpm")]
```
snehapar9, line 181: Any reason why this test was removed?

tests/Oryx.RuntimeImage.Tests/Php/PhpImageTest.cs
```diff
@@ -276,7 +276,6 @@ public void Redis_IsInstalled(string imageTag)
         [InlineData("8.2")]
         [InlineData("8.1")]
         [InlineData("8.0")]
-        [InlineData("7.4")]
```
snehapar9, line 279: Same comment as above.

build/constants.yaml
waliMSFT: no, it is a typo. thanks for the catch.

tests/Oryx.BuildImage.Tests/Node/NodeDynamicInstallationTest.cs
waliMSFT: yah, I considered that. Actually, the versions of node 14,16 for stretch was partially hard coded, I replaced them with the constants. For version I did not add the additional constant as it was not there and for stretch, we will not consider the new versions further.

tests/Oryx.RuntimeImage.Tests/Php-fpm/PhpFpmImageTest.cs
waliMSFT: in this test it actually checks if the `pdo_sqlsrv` is installed on the image. But `pdo_sqlsrv` requires php version above 8.0. hence, I removed the installation of `pdo_sqlsrv` on the image. Therefore, testing `pdo_sqlsrv` in php version below 8.0 is pointless. 
I believe App service team does not use php version lower than 8.0 anymore. For that reason, we are now considering not to create images with php lower than 8.0. Dan will probably create a work item for that.

tests/Oryx.RuntimeImage.Tests/Php/PhpImageTest.cs
waliMSFT: please have a look on the above reply. https://github.com/microsoft/Oryx/pull/1955#discussion_r1168159697

build/constants.yaml
```diff
@@ -110,8 +110,8 @@
     final-stretch-dot-net-70-sdk-version: 7.0.100
     final-stretch-dot-net-core-app-70-run-time-version: 7.0.0
     final-stretch-aspnet-core-app-70-run-time-version: 7.0.0
-    final-stretch-node14-version: 14.19.1
-    final-stretch-node16-version: 16.14.2
+    final-stretch-node14-version: 14.21.3
+    final-stretch-node16-version: 16.20.0
```
cormacpayne, line 114: Why do we need to update these stretch versions?

waliMSFT: actually, I updated the versions and ran all the pipelines before remembering that we don't need to update the stretch versions. However, as Paul suggested "app service still has not integrated our changes to dynamically install SDKs from the storage accounts corresponding to the debian version", so thought to keep the updated stretch version (as we are running fast towards the deadline). 
If you suggest to revert the changes, I will do asap.

pauld-msft: As far as I know app service is still using old images from August. Those old images are pulling the stretch versions of languages, so we still need to upload these versions. Since we are uploading new stretch versions, the `final-stretch` version also needs to get updated since it is used in tests and such where we use a stretch image and dynamically install something like node 16 (without specifying the exact version), and expect it to install the final stretch version: https://github.com/microsoft/Oryx/blob/main/tests/Oryx.BuildImage.Tests/Node/NodeDynamicInstallationTest.cs#L167

waliMSFT:
> Do we also have to update the GO version that is used for the runtime images?

I have followed the previous GO version update PR. for example : https://github.com/microsoft/Oryx/pull/1889
The version is not updated in the dockerfils where we pull the golang image. 
But, yah it makes sense to me to update the version in the Dockerfile as you mentioned [pauld-msft](https://github.com/pauld-msft)

platforms/golang/versions/focal-scm/versionsToBuild.txt
```diff
@@ -8,4 +8,5 @@
 1.18.10, 5e05400e4c79ef5394424c0eff5b9141cb782da25f64f79d54c98af0a37f8d49,
 1.19, 464b6b66591f6cf055bc5df90a9750bf5fbc9d038722bb84a9d56a2bea974be6,
 1.19.3, 74b9640724fd4e6bb0ed2a1bc44ae813a03f1e72a4c76253e2d5c015494430ba,
-1.19.5, 36519702ae2fd573c9869461990ae550c8c0d955cd28d2827a6b159fda81ff95,
\ No newline at end of file
+1.19.5, 36519702ae2fd573c9869461990ae550c8c0d955cd28d2827a6b159fda81ff95,
+1.19.7, 7a75720c9b066ae1750f6bcc7052aba70fa3813f4223199ee2a2315fd3eb533d,
```
pauld-msft, line 12: Since app service still uses old images, we should probably add go for stretch as well, right?

waliMSFT: we did not update Golang stretch version since 1.18. that is why I did not add the new version.

pauld-msft: checked the email. looks like they are able to get golang 19.5, so i think your change here is fine, and its not needed for stretch... 

side note: i wonder if app service updated their build images to a more recent version. that would be nice as it would unblock some of the work to deprecate stretch

build/__nodeVersions.sh
```diff
@@ -1,16 +1,16 @@
...
 YARN_MINOR_VERSION='1.17'
 YARN_MAJOR_VERSION='1'
 NODE6_VERSION='6.17.1'
 NODE8_VERSION='8.17.0'
 NODE10_VERSION='10.23.0'
-NODE12_VERSION='12.22.11'
-NODE14_VERSION='14.21.2'
-NODE16_VERSION='16.19.0'
-NODE18_VERSION='18.14.0'
+NODE12_VERSION='12.22.12'
```
pauld-msft, line 10: We no longer create node 12 runtime versions... hopefully that is fine with them...

waliMSFT: we will keep the stretch changes as we are running out of time to meet the deadline.

pauld-msft (approved):
:shipit:

PR #1957, opened 2023-04-18
Support extensibility model for build and run scripts

cormacpayne:
This PR adds support for the `oryx-config.yaml` extensibility configuration file to allow the caller of Oryx (or Oryx++) more control over commands and steps that are executed within the build and run scripts (as well as future control over the runnable application image produced by Oryx++).

For the initial implementation, the `oryx-config.yaml` file will take the following shape:

```yaml
base-os: '<base OS for the image>'
platform: '<build and runtime platform for the application>'
platform-version: '<version of the runtime platform to use>'
env:
  - name: '<name of the environment variable to export>'
    value: '<value of the environment variable to export>'
pre-build:
  - description: '<description of first pre-build step>'
    scripts:
      - '<list of custom commands to run>'
    http-get:
      url: '<URL to make HTTP GET request against>'
      file-name: '<name of the file that the request should be saved to>'
      headers:
        - '<list of headers to send with the request>'
```

In the future, we can expand on this file to include more control over OS-specific commands and steps that can be executed.

TODO:
- Tests to verify correct parsing of YAML file, if provided, and that the contents are correct translated to steps in the build script that is generated


src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -151,6 +151,12 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
 		fmt.Printf("Setting up Environment Variables for Application Insights is done.. \n")
 	}
 
+	extensibleCommands := common.ParseExtensibleConfigFile(filepath.Join(gen.AppPath, consts.ExtensibleConfigurationFileName))
+	if extensibleCommands!= "" {
```
pauld-msft, line 155:
```suggestion
	if extensibleCommands != "" {
```

src/startupscriptgenerator/src/node/scriptgenerator.go
```diff
@@ -133,6 +133,12 @@ func (gen *NodeStartupScriptGenerator) GenerateEntrypointScript() string {
 		scriptBuilder.WriteString("echo \"Done.\"\n")
 	}
 
+	extensibleCommands := common.ParseExtensibleConfigFile(filepath.Join(gen.SourcePath, consts.ExtensibleConfigurationFileName))
+	if extensibleCommands!= "" {
```
pauld-msft, line 137:
```suggestion
	if extensibleCommands != "" {
```

src/startupscriptgenerator/src/php/scriptgenerator.go
```diff
@@ -50,6 +51,12 @@ func (gen *PhpStartupScriptGenerator) GenerateEntrypointScript() string {
 	scriptBuilder.WriteString("   export APACHE_DOCUMENT_ROOT='" + gen.SourcePath + "'\n")
 	scriptBuilder.WriteString("fi\n\n")
 
+	extensibleCommands := common.ParseExtensibleConfigFile(filepath.Join(gen.SourcePath, consts.ExtensibleConfigurationFileName))
+	if extensibleCommands!= "" {
```
pauld-msft, line 55: nit: same spacing suggestion for all other occurrences of this below

src/startupscriptgenerator/src/common/parseExtensibleConfig.go
```diff
@@ -0,0 +1,49 @@
...
+
+type EnvironmentVariable struct {
+    Name  string `yaml:"name"`
+    Value string `yaml:"value"`
+}
+
+type ExtensibleConfigFile struct {
+    BaseOs               string                `yaml:"base-os"`
+    Platform             string                `yaml:"platform"`
+    PlatformVersion      string                `yaml:"platform-version"`
```
pauld-msft, line 22: Just wanted to confirm that at this point, we don't expect for platform, platform-version, or base-os to be respected or required here?

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
```diff
@@ -151,6 +151,12 @@ func (gen *DotnetCoreStartupScriptGenerator) GenerateEntrypointScript(scriptBuil
 		fmt.Printf("Setting up Environment Variables for Application Insights is done.. \n")
 	}
 
+	extensibleCommands := common.ParseExtensibleConfigFile(filepath.Join(gen.AppPath, consts.ExtensibleConfigurationFileName))
```
pauld-msft, line 154: What is the reasoning for this to be here, vs earlier in the `GenerateEntrypointScript` function? It seems like since it is here, the user could not use this config file to configure things like dynamically installed languages (line 137), or the PORT (line 115). That _should_ be fine since they could just configure those with application settings, right?

tests/Oryx.BuildImage.Tests/ExtensibilityModelBuildScriptTest.cs
```diff
@@ -0,0 +1,316 @@
...
+pre-build: 
+  - description: 'Test description'
+    scripts:
+      - 'echo ""Hello, world!""'
+      - 'echo ""Hello, foobar!""'
+    http-get:
+      url: 'testurl'
+      file-name: 'samplefile'
+      headers:
+        - 'header1'
```
pauld-msft, line 265: maybe add another header here to test that case? Also, don't we support multiple pre-build commands, like 
```
base-os: debian
env: 
  - name: FOO
    value: BAR
pre-build: 
  - description: 'Test description'
    scripts:
      - 'echo ""Hello, world!""'
      - 'echo ""Hello, foobar!""'
    http-get:
      url: 'testurl'
      file-name: 'samplefile'
      headers:
        - 'header1'
        - 'header2'
  - description: 'Test description 2'
    scripts:
      - 'echo ""Hello, lorem!""'
      - 'echo ""Hello, ipsum!""'
    http-get:
      url: 'testurl2'
      file-name: 'samplefile2'
      headers:
        - 'header3'
        - 'header4'
```

src/BuildScriptGenerator/BaseBashBuildScriptProperties.cs
```diff
@@ -19,89 +19,104 @@ public class BaseBashBuildScriptProperties
...
+        /// </summary>
+        public string PreBuildCommand { get; set; }
 
         /// <summary>
-        /// Gets or sets a list of properties for the build. Those properties are stored in a
-        /// manifest file that can be used when running the app.
+        /// Gets or sets the set of commands configured by the extensibility model.
         /// </summary>
-        public IDictionary<string, string> BuildProperties { get; set; }
+        public string ExtensibleConfigurationCommands { get; set; }
```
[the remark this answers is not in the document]
cormacpayne: Sorry, I re-ordered the properties in this file based on the order they appear in the `.tpl` file -- you're correct, those three properties are the only ones actually being added to this file.

src/BuildScriptGenerator/BuildConfigurationFIle.cs
```diff
@@ -1,15 +1,17 @@
...
+// --------------------------------------------------------------------------------------------
+
 using YamlDotNet.Serialization;
 using YamlDotNet.Serialization.NamingConventions;
 
 namespace Microsoft.Oryx.BuildScriptGenerator
 {
     /// <summary>
-    /// BuildConfigurationFIle class to be used for YAML deserialization.
+    /// BuildConfigurationFile class to be used for YAML deserialization.
```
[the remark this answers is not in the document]
cormacpayne: Huh, the file name is changed locally for me, but it doesn't seem that Git is recognizing it. Let me make sure this change gets propagated.

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -427,5 +382,112 @@ private void LogScriptIfGiven(string type, string scriptPath)
...
+                this.logger.LogDebug("Found BuildConfigurationFile");
+                this.writer.WriteLine(Environment.NewLine + "Found BuildConfigurationFile");
+                try
+                {
+                    BuildConfigurationFile buildConfigFile = BuildConfigurationFile.Create(context.SourceRepo.ReadFile(FilePaths.AppSvcFileName));
+                    if (!string.IsNullOrEmpty(buildConfigFile.Prebuild))
+                    {
+                        this.cliOptions.PreBuildCommand = buildConfigFile.Prebuild.Replace("\r\n", ";").Replace("\n", ";");
+                        this.cliOptions.PreBuildScriptPath = null;
+                        this.logger.LogDebug("Overriding the pre-build commands with the BuildConfigurationFile section");
```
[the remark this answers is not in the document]
cormacpayne: Covered during the design meeting :shipit: Let me know if this question still applies.

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
```diff
@@ -427,5 +382,112 @@ private void LogScriptIfGiven(string type, string scriptPath)
...
+        private string ProcessExtensibleConfigurationFile(BuildScriptGeneratorContext context)
+        {
+            var filePath = Path.Combine(context.SourceRepo.RootPath, FilePaths.ExtensibleConfigurationFileName);
+            this.logger.LogDebug($"Checking for path to {FilePaths.ExtensibleConfigurationFileName}: '{filePath}'");
+            this.writer.WriteLine($"Checking for path to {FilePaths.ExtensibleConfigurationFileName}: '{filePath}'");
+
+            if (File.Exists(filePath))
+            {
+                this.logger.LogDebug("Found extensible configuration file.");
+                this.writer.WriteLine($"{Environment.NewLine}Found extensible configuration file, {filePath}");
```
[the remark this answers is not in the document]
cormacpayne: The logger will be writing to Application Insights and the writer will be writing to the console output that the user sees. In general, the writer can be more verbose in describing files, directories, commands specific to the user, etc. since it's _only_ being written to the console output and not stored anywhere, whereas with the logger, since it's being written/stored to AI, we need to be cautious of PII that could be sent there.

src/BuildScriptGenerator/Extensibility/ExtensibleEnvironmentVariable.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// </summary>
+        /// <returns>A build script snippet for the current environment variable.</returns>
+        public string GetBuildScriptSnippet()
+        {
+            if (string.IsNullOrEmpty(this.Name))
+            {
+                return string.Empty;
+            }
+
+            var result = $"export {this.Name}";
```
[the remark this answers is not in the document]
cormacpayne: Thinking about it a little more, you're probably right that there isn't a need for exporting an environment variable without a value. It's useful in the case where you have an existing _local_ variable that you would then want to expose as an environment variable, but I don't think any user would actually be able to leverage that in a meaningful way here (that they wouldn't already be able to do with the custom bash commands they provide). I'll update this to check for both name and value.

src/BuildScriptGenerator/Extensibility/ExtensibleHttpGet.cs
```diff
@@ -0,0 +1,65 @@
...
+        [YamlMember(Alias = "headers", ApplyNamingConventions = false)]
+        public string[] Headers { get; set; }
+
+        /// <summary>
+        /// Gets a build script snippet for the current HTTP GET request.
+        /// </summary>
+        /// <param name="properties">The <see cref="ExtensibleConfigurationProperties"/> containing properties used by
+        /// the extensibility model.</param>
+        /// <returns>A build script snippet for the current HTTP GET request.</returns>
+        public string GetBuildScriptSnippet(ExtensibleConfigurationProperties properties)
```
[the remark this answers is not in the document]
cormacpayne: Mentioned during the design meeting, but this was originally introduced when the `oryx-config.yaml` file had additional properties (such as `platform`) that would leverage the values in this properties file, but after later discussions, we decided to move from this and I forgot to cleanup the remaining artifacts. This properties class and references to it are removed now.

src/BuildScriptGenerator/Extensibility/ExtensibleHttpGet.cs
```diff
@@ -0,0 +1,65 @@
...
+                return string.Empty;
+            }
+
+            var result = "curl";
+            if (!string.IsNullOrEmpty(this.FileName))
+            {
+                result = $"{result} -o {this.FileName}";
+            }
+
+            if (this.Headers?.Any() == true)
```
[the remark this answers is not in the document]
cormacpayne: When `this.Headers` is `null`, the call to `.Any()` will actually throw an `ArgumentNullException`, so we add the null-conditional operator to first resolve the operation as `null` in this case, and the `== true` allows us to do a check against this "nullable boolean" to see if it is (1) not `null` and (2) contains any values.

I think there's an argument to be made that this resolves to just `this.Headers != null`, but I didn't do enough testing with how the YAML parser we're using will resolve the property if different combinations of values are provided to an object property. I'll leave it as-is for clarity unless a change is desired 🙂

src/startupscriptgenerator/src/common/parseExtensibleConfig.go
cormacpayne: Not at this current moment, but it _could_ be something that we use in the future for the runtime script, so I'm just keeping them here for now. I don't have any gripes with removing them if you feel that it would be better for keeping the file clean though.

src/startupscriptgenerator/src/common/parseExtensibleConfig.go
```diff
@@ -0,0 +1,49 @@
...
+    fileContent, err := os.ReadFile(sourcePath)
+    if err != nil {
+        return ""
+    }
+
+    configFile := ExtensibleConfigFile{}
+    yamlErr := yaml.Unmarshal(fileContent, &configFile)
+
+    if yamlErr != nil {
+        return ""
```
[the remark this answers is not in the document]
cormacpayne: Good idea, let me print out the error for the user.

src/startupscriptgenerator/src/dotnetcore/scriptgenerator.go
cormacpayne: This was also discussed during the design meeting, but feel free to let me know if you still have any questions around this topic.

tests/Oryx.BuildImage.Tests/ExtensibilityModelBuildScriptTest.cs
cormacpayne: Good idea, I'll add a test to handle multiple headers and a test to handle multiple pre-build commands.

src/BuildScriptGenerator/DefaultBuildScriptGenerator.cs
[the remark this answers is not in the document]
pauld-msft: Makes sense, thanks for the explanation

src/BuildScriptGenerator/Extensibility/ExtensibleHttpGet.cs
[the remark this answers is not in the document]
pauld-msft: Oof yeah you are absolutely right about the conditional not evaluating the `null` as `false`.  This makes sense, thanks

src/startupscriptgenerator/src/common/parseExtensibleConfig.go
pauld-msft: Nope, I'm all good with keeping them here for now

daniv-msft:
Sorry I didn't have a chance to review the PR yet, but two high-level comments:
1. When we surface this in the UI, we need to make sure we rename it to "build-config.yaml" or something of that kind (I believe that build-config.yaml is what we used in K4Apps discussion), as "Oryx" won't make sense for customers in the context of what we're building this for.
2. The YAML parsing logic will likely need to be moved outside of the Oryx image. @pauld-msft and I were talking about the ARM API update call, and I don't believe anymore we'll be able to sell passing the YAML as a big data string in the payload. That said, we would still need the YAML parsing logic but it would more likely be in the CLI. We should sync about this next week.

build/testStartupScriptGenerators.sh
```diff
@@ -25,7 +25,7 @@ declare -r MODULE_TO_TEST="..."
 declare -r CONTAINER_NAME="oryxtests_$RANDOM"
 
 echo "Running tests in golang docker image..."
-docker run -v $GEN_DIR:$GEN_DIR_CONTAINER_RO:ro --name $CONTAINER_NAME golang:1.15-stretch bash -c \
+docker run -v $GEN_DIR:$GEN_DIR_CONTAINER_RO:ro --name $CONTAINER_NAME golang:1.19-bullseye bash -c \
 	"cp -rf $GEN_DIR_CONTAINER_RO/* $GEN_DIR_CONTAINER && \
```
snehapar9, line 29: test

images/build/Dockerfiles/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM oryxdevmcr.azurecr.io/public/oryx/build:lts-versions-debian-stretch AS main
 # Build run script generators (to be used by the `oryx run-script` command)
 ###
 # DisableDockerDetector "Below image not yet supported in the Docker Hub mirror"
-FROM golang:1.15-stretch as startupScriptGens
+FROM golang:1.18-stretch as startupScriptGens
```
snehapar9, line 7: test2

build/testStartupScriptGenerators.sh
```diff
@@ -25,7 +25,7 @@ declare -r MODULE_TO_TEST="..."
 declare -r CONTAINER_NAME="oryxtests_$RANDOM"
 
 echo "Running tests in golang docker image..."
-docker run -v $GEN_DIR:$GEN_DIR_CONTAINER_RO:ro --name $CONTAINER_NAME golang:1.15-stretch bash -c \
+docker run -v $GEN_DIR:$GEN_DIR_CONTAINER_RO:ro --name $CONTAINER_NAME golang:1.19-bullseye bash -c \
```
snehapar9, line 28: testing

cormacpayne: Test received!

images/build/Dockerfiles/Dockerfile
william-msft, line 7: Is the discrepancy between golang 1.18 and [1.19](https://github.com/microsoft/Oryx/pull/1957/files#diff-fd77fd00771c313a727f3c931eb29e02bda69ae692c91b5677e560aef0ef17a0R28) intentional?

PR #1967, opened 2023-04-24
Fix Debian Stretch Build Image Failure

harryli0108:
Debian recently moved stretch to archive, resulting in build image fail across pipelines. This PR is out to change the repo from deb.debian.org and other usual debian repos to archive debian repo. 

PHP prerequisite install shell script is also modified similarly to creating buster repo list.

### DIFF
--- a/images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
+++ b/images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
@@ -14,6 +14,13 @@ ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
     LC_ALL="C.UTF-8" \
     ORYX_PATHS="/opt/oryx:/opt/nodejs/lts/bin:/opt/python/latest/bin:/opt/yarn/stable/bin"
     
+# stretch was removed from security.debian.org and deb.debian.org, so update the sources to point to the archived mirror
+RUN if [ "${DEBIAN_FLAVOR}" = "stretch" ]; then \
+        sed -i 's/^deb http:\/\/deb.debian.org\/debian stretch-updates/# deb http:\/\/deb.debian.org\/debian stretch-updates/g' /etc/apt/sources.list  \
+        && sed -i 's/^deb http:\/\/security.debian.org\/debian-security stretch/deb http:\/\/archive.debian.org\/debian-security stretch/g' /etc/apt/sources.list \
+        && sed -i 's/^deb http:\/\/deb.debian.org\/debian stretch/deb http:\/\/archive.debian.org\/debian stretch/g' /etc/apt/sources.list ; \
+    fi
+
 RUN set -ex \
     # Install Python SDKs
     # Upgrade system python
--- a/images/build/php/prereqs/installPrereqs.sh
+++ b/images/build/php/prereqs/installPrereqs.sh
@@ -33,7 +33,9 @@ apt-get update \
     --no-install-recommends && rm -rf /var/lib/apt/lists/*
 
 ##<argon2>##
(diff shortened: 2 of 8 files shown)


images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
```diff
@@ -14,6 +14,13 @@ ENV DEBIAN_FLAVOR=$DEBIAN_FLAVOR \
     LC_ALL="C.UTF-8" \
     ORYX_PATHS="/opt/oryx:/opt/nodejs/lts/bin:/opt/python/latest/bin:/opt/yarn/stable/bin"
     
+# stretch was removed from security.debian.org and deb.debian.org, so update the sources to point to the archived mirror
```
daniv-msft, line 17: Could you please clarify how you see this changing in the future? For example, can we assume that we'll be able to remove all references to Stretch from our code (I believe we only keep them while our partners are deploying changes) before the archived version disappears?

daniv-msft (dismissed):

images/build/php/prereqs/installPrereqs.sh
```diff
@@ -33,7 +33,10 @@ apt-get update \
     --no-install-recommends && rm -rf /var/lib/apt/lists/*
 
 ##<argon2>##
-sed -e 's/stretch/buster/g' /etc/apt/sources.list > /etc/apt/sources.list.d/buster.list;
+sed -e 's/# deb http:\/\/archive.debian.org\/debian-security stretch/deb http:\/\/security.debian.org\/debian-security stretch/g' \
```
pauld-msft, line 36: Does the commented out line `# deb http://archive.debian.org/debian-security stretch` exist at this point for it to replace back with the security version? I don't see that link getting created above

harryli0108: this reverting step is mainly for populating the `buster.list`. It'll revert the changes to what the `sources.list` originally looks like so that the `buster.list` could be created correctly. That line will be changed to `deb http://security.debian.org/debian-security buster`.

pauld-msft: oh, I see. I think that this line should just instead revert the `stretch-updates` line, e.g something like:

`s/# deb http:\/\/deb.debian.org\/debian stretch-updates/deb http:\/\/deb.debian.org\/debian stretch-updates/g`

pauld-msft: since we have 2 lines in this block that are updating `debian-security` source

images/build/Dockerfiles/azureFunctions.JamStack.Dockerfile
harryli0108: I think the archive would remain for about 2-3 years, and then be expired (I think I saw this info somewhere in Debian distro webpage, but I couldn't find the link for now...). I think we should be stable for now, but imo it would be best for us to move from stretch in the meantime.

images/build/php/prereqs/installPrereqs.sh
pauld-msft: and original changes are to:
```
# deb http://deb.debian.org/debian stretch-updates             (comment this out, as stretch-updates didn't seem to exist)
deb http://archive.debian.org/debian-security stretch          (archive source)
deb http://archive.debian.org/debian stretch                   (archive source)
```

harryli0108: makes sense! Just pushed a new change removing the duplicate line.

pauld-msft (approved):
Thanks for making this fix :shipit:

PR #1969, opened 2023-04-26
Wali/update php 8.1 8.2 node 18 to their latest versions

waliMSFT:
Upgrading the php and node versions:
image


daniv-msft (approved):
Looks good to me! Thanks

PR #1991, opened 2023-05-10
Update to fix version sorting

william-msft:
This is for Automation. This PR will fix version sorting, to sort by sematic version, to get the latest version in constants.yaml. Related story https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1816926

This was not an issue before since we were always getting a single version. This bug showed up after attempting to release multiple versions.

### DIFF
--- a/build/tools/Automation/DotNet/DotNetAutomator.cs
+++ b/build/tools/Automation/DotNet/DotNetAutomator.cs
@@ -82,6 +82,15 @@ namespace Microsoft.Oryx.Automation.DotNet
             }
         }
 
+        /// <summary>
+        /// Retrieves a list of new .NET versions from the .NET release website,
+        /// and filters the list based on specified criteria.
+        /// The resulting list includes SDK versions, .NET Core runtime versions,
+        /// and ASP.NET Core runtime versions, and is sorted by semantic version number.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.
+        /// The task result contains a list of DotNetVersion objects representing the new
+        /// .NET versions that meet the specified criteria.</returns>
         public async Task<List<DotNetVersion>> GetNewDotNetVersionsAsync()
         {
             List<DotNetVersion> dotNetVersions = new List<DotNetVersion>();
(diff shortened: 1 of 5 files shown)


build/tools/Automation/DotNet/DotNetAutomator.cs
```diff
@@ -118,7 +127,7 @@ public async Task<List<DotNetVersion>> GetNewDotNetVersionsAsync()
                 }
             }
 
-            return dotNetVersions.OrderBy(v => v.Version).ToList();
+            return dotNetVersions.OrderBy(v => new Version(v.Version)).ToList();
```
cormacpayne, line 130: In my opinion, it may be better if we make both `DotNetVersion` and `PythonVersion` implement the `IComparable` interface so that they both implement `CompareTo` and can inherently be sorted without needing to rely on a specific property or make these types of conversions each time we need to.

The change in the `DotNetVersion` class would be the following:

```cs
public class DotNetVersion : IComparable
{
    // ...

    public int CompareTo(object obj)
    {
        var objDotNetVersion = obj as DotNetVersion;
        if (objDotNetVersion == null)
        {
            return 1;
        }

        var thisVersion = new Version(this.Version);
        var objVersion = new Version(objDotNetVersion.Version);
        return thisVersion.CompareTo(objVersion);
   }
}
```

The above line in your PR would then just resolve to `return dotNetVersions.Order().ToList();`

cormacpayne (dismissed):
One potential future-proof comment if you'd like to make the change, otherwise LGTM :shipit:

build/tools/Automation/DotNet/DotNetAutomator.cs
william-msft: This makes sense, thanks!

build/tools/Automation/Python/Models/PythonVersion.cs
```diff
@@ -1,9 +1,24 @@
-﻿namespace Microsoft.Oryx.Automation.Python.Models
```
cormacpayne, line 1: One minor comment I'm seeing from this change: would you quickly confirm that all of the `Automation` project `.cs` files have the license header?

cormacpayne (approved):
:shipit:

cormacpayne (dismissed):
:shipit:

cormacpayne (approved):
:shipit:

PR #1998, opened 2023-05-16
New DotNetCore Sdks 2023-05-16

william-msft:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #2002, opened 2023-05-23
Fix build on local machines with some package updates

pauld-msft:
`Microsoft.CodeAnalysis.Compilers` 4.4.0 -> 4.6.0
`Microsoft.Net.Compilers.Toolset` 4.4.0 -> 4.6.0


william-msft (dismissed):

Directory.Build.props
```diff
@@ -29,11 +29,11 @@ See https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?v
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.CodeAnalysis.Compilers" Version="4.4.0" >
+    <PackageReference Include="Microsoft.CodeAnalysis.Compilers" Version="4.6.0" >
```
waliMSFT, line 32: @pauld do you know why we were forced to update the version?

daniv-msft (dismissed):

Directory.Build.props
pauld-msft: Hey Wali, unfortunately I did not have time to try and investigate much around the reason that we had to update these. I suspect that it was due to a dotnet version update, but I can't be sure right now

build/constants.yaml
```diff
@@ -107,9 +107,9 @@
     final-stretch-dot-net-60-sdk-version: 6.0.301
     final-stretch-dot-net-core-app-60-run-time-version: 6.0.6
     final-stretch-aspnet-core-app-60-run-time-version: 6.0.6
-    final-stretch-dot-net-70-sdk-version: 7.0.100
-    final-stretch-dot-net-core-app-70-run-time-version: 7.0.0
-    final-stretch-aspnet-core-app-70-run-time-version: 7.0.0
+    final-stretch-dot-net-70-sdk-version: 7.0.203
```
waliMSFT, line 110: is this stretch version update related to the build issue?

pauld-msft: That was my guess, but perhaps not the case. Looking into it

pauld-msft: Ok so it was to do with the incompatible versions, since some new ones were added with https://github.com/microsoft/Oryx/pull/1998.

I've merged these changes into there, so I will close this PR and we can merge in that one

pauld-msft:
Closing in favor of https://github.com/microsoft/Oryx/pull/1998, where we have the changes that will allow the validation/nightly jobs to pass

PR #2009, opened 2023-05-31
Updating Hugo detection logic and bumping up version

harryli0108:
Due to new Hugo versions changed the config name convention from `config.toml` etc. to `hugo.toml` etc., this PR updates the detection logic and make sure the new naming convention is also supported while not affecting the old ones. This PR also added a new modified sample app and a test case in order to make sure new naming convention is correctly supported.

And because this change starts from Hugo version 0.110.0, this PR also upgrades the current hugo version to the latest `0.112.5`.


build/constants.yaml
```diff
@@ -346,15 +346,23 @@
...
+    version: 0.112.5
     platform-name: hugo
     installed-hugo-versions-dir: /opt/hugo
     installation-url-format: https://github.com/gohugoio/hugo/releases/download/v#VERSION#/#TAR_FILE#
     tar-file-name-format: hugo_extended_#VERSION#_Linux-64bit.tar.gz
-    toml-file-name: config.toml
-    yaml-file-name: config.yaml
-    yml-file-name: config.yml
-    json-file-name: config.json
+    toml-file-name:
```
cormacpayne, line 354: nit: since these values are now lists, can we pluralize the name of the variables? _i.e._, `toml-file-name` ==> `toml-file-names`

src/Detector/Hugo/HugoDetector.cs
```diff
@@ -66,31 +66,43 @@ private bool IsHugoApp(ISourceRepo sourceRepo, out string appDirectory)
             appDirectory = string.Empty;
 
             // Search for config.toml
-            if (sourceRepo.FileExists(HugoConstants.TomlFileName)
-                && this.IsHugoTomlFile(sourceRepo, HugoConstants.TomlFileName))
+            foreach (string tomlFileName in HugoConstants.TomlFileName)
             {
-                return true;
+                if (sourceRepo.FileExists(tomlFileName)
+                && this.IsHugoTomlFile(sourceRepo, tomlFileName))
```
cormacpayne, line 72: nit: let's indent the second line of each if-statement so that it follows the previous format

```cs
if (sourceRepo.FileExists(tomlFileName)
    && this.IsHugoTomlFile(sourceRepo, tomlFileName))
```

This comment applies for the other instances of this below.

src/Detector/Hugo/HugoDetector.cs
```diff
@@ -66,31 +66,43 @@ private bool IsHugoApp(ISourceRepo sourceRepo, out string appDirectory)
             appDirectory = string.Empty;
 
             // Search for config.toml
```
cormacpayne, line 68: nit: let's update the comment before each if-statement so it accurately reflects what we're searching for (_i.e._, "Search for supported .toml file").

This comment applies for the other instances of this below.

src/Detector/Hugo/HugoDetector.cs
```diff
@@ -66,31 +66,43 @@ private bool IsHugoApp(ISourceRepo sourceRepo, out string appDirectory)
...
             }
 
             // Search for config.json
-            if (sourceRepo.FileExists(HugoConstants.JsonFileName)
-                && this.IsHugoYamlFile(sourceRepo, HugoConstants.JsonFileName))
+            foreach (string jsonFileName in HugoConstants.JsonFileName)
             {
-                return true;
+                if (sourceRepo.FileExists(jsonFileName)
+                && this.IsHugoYamlFile(sourceRepo, jsonFileName))
```
cormacpayne, line 102: This looks like an issue from before, but let's change this call to `this.IsHugeJsonFile`

src/Detector/Hugo/HugoDetector.cs
pauld-msft, line 68: nit: update these comments

src/Detector/Hugo/HugoDetector.cs
```diff
@@ -155,23 +167,25 @@ private bool IsHugoApp(ISourceRepo sourceRepo, out string appDirectory)
...
             TomlTable tomlTable = null;
-
-            var relativeFilePath = Path.Combine(subPaths);
-
-            try
-            {
-                tomlTable = ParserHelper.ParseTomlFile(sourceRepo, relativeFilePath);
-            }
-            catch (FailedToParseFileException ex)
+            foreach (string subPath in subPaths)
```
cormacpayne, line 170: I'm not sure I understand the need for the addition of this `foreach` block -- would you mind elaborating?

build/constants.yaml
pauld-msft, line 354: nit: `name` -> `names`

doc/runtimes/hugo.md
```diff
@@ -16,6 +16,8 @@ The Hugo toolset is run when one of following conditions met:
 2. Recursively look up config directory: `config/**/*.yaml`, `config/**/*.toml`,
    `config/**/*.yml` or `config/**/*.json` configuration files in sub-directories.
 
+**Note**: Starting from [Hugo version `0.110.0`](https://gohugo.io/getting-started/configuration/#hugotoml-vs-configtoml), configuration file name has changed from `config.toml` etc. to `hugo.toml` etc. Please note the old naming convention is still supported.
```
cormacpayne, line 19: We could probably scrap this line in favor of updating the (1.) bullet above to mention that we look for the `hugo.*` equivalent files.

cormacpayne, line 19: nit: let's keep the maximum length of a line in Markdown to 120 characters

cormacpayne (changes requested):
A few comments to take a look at

src/Detector/Hugo/HugoDetector.cs
pauld-msft, line 170: Aren't these multiple sub paths, like and example of this object could be `[ 'C:', 'repos', 'samplehugoapp', 'hugo.toml' ]`? So why are we looping through it here instead of combining it first?

Maybe I'm misreading something though. Also, as far as i can tell, we never pass in more than 2 params, so doesn't seems like its really even being used.

pauld-msft (commented):
Few minor things. Overall, looks good to me, so happy to approve once a few minor things are resolved. 

Also, thanks for adding the test! 

Edit: also noticed builds are failing, and cormac had some good comments to look at :)

src/Detector/Hugo/HugoDetector.cs
harryli0108: Thanks for the suggestion! `IsHugoTomlFile` is being called from [this line](https://github.com/microsoft/Oryx/blob/main/src/Detector/Hugo/HugoDetector.cs#LL70C17-L70C17). And seems like the object in the past is just `config.toml`. But as you said, we actually never pass more than 2 parameters (in fact, many of the constants are also not being used in many detector's platform constant files). I'll try removing it and see if it breaks anything here.

build/constants.yaml
cormacpayne: image

doc/runtimes/hugo.md
```diff
@@ -12,7 +12,7 @@ details on components and configuration of build images.
 
 The Hugo toolset is run when one of following conditions met:
 
-1. `config.yaml`, `config.toml`, `config.yml` or `config.json` configuration files in root of repo.
+1. `config.yaml`, `config.toml`, `config.yml`, `config.json` or `hugo.*` equivalent configuration files in root of repo.
```
daniv-msft, line 15: Could you please clarify why supporting "hugo.*"? Looking quickly at the doc I could only see "hugo.toml" being used. Am I missing something?

tests/Detector.Tests/Hugo/HugoDetectorTest.cs
```diff
@@ -78,8 +79,8 @@ public void IsHugoApp_ReturnsTrue_ForAllSupportedConfigurationKeys_InTomlFile(st
         }
 
         [Theory]
-        [InlineData(HugoConstants.JsonFileName)]
-        [InlineData(HugoConstants.ConfigFolderName, HugoConstants.JsonFileName)]
+        [InlineData("config.json")]
```
daniv-msft, line 82: Just curious: we can't directly use `HugoConstants.JsonFileNames.First()` in these InlineData fields to remain consistent?

daniv-msft (dismissed):
Looks good to me! Thanks for fixing this

waliMSFT (commented):
Looks good to me as well.

doc/runtimes/hugo.md
harryli0108: That makes sense! Updating it right now

doc/runtimes/hugo.md
```diff
@@ -12,7 +12,7 @@ details on components and configuration of build images.
 
 The Hugo toolset is run when one of following conditions met:
 
-1. `config.yaml`, `config.toml`, `config.yml` or `config.json` configuration files in root of repo.
+1. `config/hugo.yaml`, `config/hugo.toml`, `config/hugo.yml` or `config/hugo.json` configuration files in root of repo.
```
cormacpayne, line 15: I actually think we should word this line differently as users may interpret the `config` part of each file type as the directory name, especially because our second bullet below mentions we search the `config` directory recursively.

I'd be in favor of writing this as something like the following:

_1. Any supported `config.*` files at the root of the repository: `config.toml`, `config.yaml`, `config.yml`, `config.json`_
_2. Any supported `hugo.*` files at the root of the repository: `hugo.toml`, `hugo.yaml`, `hugo.yml`, `hugo.json`_
_3. Recursively look up supported files in the `config` directory: `config/**/*.toml`, `config/**/*.toml`
  `config/**/*.yml`, `config/**/*.json`_

tests/Detector.Tests/Hugo/HugoDetectorTest.cs
```diff
@@ -22,8 +23,8 @@ public HugoDetectorTest(TestTempDirTestFixture testFixture)
         }
 
         [Theory]
-        [InlineData(HugoConstants.TomlFileName)]
-        [InlineData(HugoConstants.ConfigFolderName, HugoConstants.TomlFileName)]
+        [InlineData("config.toml")]
+        [InlineData(HugoConstants.ConfigFolderName, "config.toml")]
         public void IsHugoApp_ReturnsTrue_ForAppWithConfigTomlFile(params string[] subPaths)
```
cormacpayne, line 28: Let's introduce the same set of tests that we have in this file but for the `hugo.*` files that we now support to ensure that we're detecting them properly.

tests/Detector.Tests/Hugo/HugoDetectorTest.cs
```diff
@@ -63,7 +64,7 @@ public void IsHugoApp_ReturnsTrue_ForAllSupportedConfigurationKeys_InTomlFile(st
         {
             // Arrange
             var appDir = CreateAppDir();
-            WriteFile($"{configurationKeyName}=\"test\"", appDir, HugoConstants.TomlFileName);
+            WriteFile($"{configurationKeyName}=\"test\"", appDir, HugoConstants.TomlFileNames.First());
```
cormacpayne, line 67: We should extract out the name of the file to be an additional argument of this test to ensure that the support configuration keys work for all versions of `.toml` file.

This comment also applies for the other tests below where we've made the same change.

cormacpayne (changes requested):
A few comments we should take care of before merging

cormacpayne (approved):
:shipit:

jpconstantineau:
Has this been released yet?  
I am having an issue that this PR solves. (wrong/old hugo version) 
How do I get my github action to use this fix?  Will it once it get released?

harryli0108:
@jpconstantineau Hi! Sorry for the delay here. We have a release inbound today that includes this fix.
In order to use this in GH Action, depending on if you're using Azure Web Apps task or the new Container App task. With the Web Apps one, I think you might need to wait for the App service team to include this fix in their new release. But with Container Apps task, you can either specify the oryx build image used in your Dockerfile, or you might need to wait for our team to release the next builder.

Meanwhile, I would encourage you to post this as an issue if you continue to experience it since comments in closed PRs could be accidentally ignored.

jpconstantineau:
Thanks. It's web app indeed.  I'll wait for the release here first.  Hopefully, the GitHub action doesn't need updating itself but simply picks up the latest release (or the one the web app team points to through config)

Issue #2003, opened 2023-05-23
Build failing due to missing GDAL libraries


kdhoermann:
### Bug Report

GeoDjango build fails due to missing GDAL libraries.  Given the required libraries were requested in #1049 and merged in #1056, why are they not available in the build or runtime containers, or what am I doing wrong?  

- At what date and time did you most recently experience the problem?

Today; May 23, 2023.

- Where did you experience the problem? E.g. Azure Web Apps, Azure Functions, Azure Container Registry, or offline use.

Azure Web Apps

- If your repo is publicly available please share its URL:

Issue can be reproduced by this forked quickstart with django.contrib.gis added to installed apps:
[https://github.com/kdhoermann/msdocs-python-django-webapp-quickstart](url)

- What happened?

GDAL libraries are not found when the app is started.

- What did you expect or want to happen?

The libraries should be available, per #1056.

- How can we reproduce it?

Attempt to deploy the above repo.

- Do you have log output? Please include between the backticks:

```
remote: Deploy Async        
remote: Updating branch 'main'.        
remote: Updating submodules.        
remote: Preparing deployment for commit id 'dede7a777c'.        
remote: PreDeployment: context.CleanOutputPath False        
remote: PreDeployment: context.OutputPath /home/site/wwwroot        
remote: Repository path is /home/site/repository        
remote: Running oryx build...        
remote: .............        
remote: Operation performed by Microsoft Oryx, https://github.com/Microsoft/Oryx        
remote: You can report issues at https://github.com/Microsoft/Oryx/issues        
remote: 
remote: Oryx Version: 0.2.20230210.1, Commit: a49c8f6b8abbe95b4356552c4c884dea7fd0d86e, ReleaseTagName: 20230210.1        
remote: 
remote: Build Operation ID: 95602423a8e5d60c        
remote: Repository Commit : dede7a777c2db23d0c9fc88bd2cf77fa43cd6588        
remote: OS Type           : bullseye        
remote: Image Type        : githubactions        
remote: 
remote: Detecting platforms...        
remote: ..        
remote: Detected following platforms:        
remote:   python: 3.11.0        
remote: Version '3.11.0' of platform 'python' is not installed. Generating script to install it...        
remote: 
remote: Using intermediate directory '/tmp/8db5bb164e6f7c4'.        
remote: 
remote: Copying files to the intermediate directory...        
remote: Done in 1 sec(s).        
remote: 
remote: Source directory     : /tmp/8db5bb164e6f7c4        
[...]

PR #2011, opened 2023-06-02
Add GDAL libraries to GitHub Actions images for GeoDjango scenarios

cormacpayne:
Resolves issue https://github.com/microsoft/Oryx/issues/2003

The GDAL libraries used for GeoDjango scenarios in Python were previously added to (and still exist in) our LTS images, which are no longer consumed by App Service. This PR aims to bake these libraries into the GitHub Actions image to re-enable this scenario for customers.

A snippet was added to the `settings.py` file in `django-app` that will fail any test that tries to build the app within an image that doesn't have the GDAL libraries installed in it.

I confirmed locally that the updated `github-actions-debian-bullseye` build image runs successfully against this app, but assume other tests will fail due to their lack of the GDAL library (specifically, the `debian-stretch` images).

### DIFF
--- a/images/build/Dockerfiles/gitHubActions.Dockerfile
+++ b/images/build/Dockerfiles/gitHubActions.Dockerfile
@@ -54,6 +54,11 @@ RUN apt-get update \
         lzma \
         lzma-dev \
         zlib1g-dev \
+        # GIS libraries for GeoDjango (https://docs.djangoproject.com/en/3.2/ref/contrib/gis/install/geolibs/)
+        binutils \
+        libproj-dev \
+        gdal-bin \
+        libgdal-dev \
     && rm -rf /var/lib/apt/lists/* \
     # This is the folder containing 'links' to benv and build script generator
     && mkdir -p /opt/oryx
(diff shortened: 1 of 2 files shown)


pauld-msft:
>  but assume other tests will fail due to their lack of the GDAL library (specifically, the debian-stretch images)
I'm guessing we will either remove these failing tests, or update them to use a new image?

waliMSFT:
> > but assume other tests will fail due to their lack of the GDAL library (specifically, the debian-stretch images)
> 
> I'm guessing we will either remove these failing tests, or update them to use a new image?

@pauld-msft wondering why removing these tests is an option when we can update them (if updating is not costly)?

pauld-msft:
> > > but assume other tests will fail due to their lack of the GDAL library (specifically, the debian-stretch images)
> > 
> > 
> > I'm guessing we will either remove these failing tests, or update them to use a new image?
> 
> @pauld-msft wondering why removing these tests is an option when we can update them (if updating is not costly)?

@cormacpayne and I caught up offline, and the plan would have been to update them to use newer images. I think this PR may get abandoned however, becuase the packages would increase image sizes by ~400-500MBs. But yep you are right that we should try to maintain all the tests we can especially if they are easy to maintain. We should only remove or temporarily disable as a last resort.

daniv-msft:
@cormacpayne Should we close this PR so that we don't merge it by mistake? Unfortunately the size impact of tis library is huge on the resulting image.

PR #2016, opened 2023-06-08
Updated nginx to pull stable version

pauld-msft:
Updated the nginx installation to use the official nginx stable version, as documented in https://docs.nginx.com/nginx/admin-guide/installing-nginx/installing-nginx-open-source/#installing-prebuilt-debian-packages

Sizes (old on top, new below):
image


cormacpayne (dismissed):
:shipit:

daniv-msft (dismissed):
Looks good to me! Thanks

images/runtime/php-fpm/nginx_conf/nginx.conf
```diff
@@ -0,0 +1,83 @@
...
+        access_log /var/log/nginx/access.log;
+        error_log /var/log/nginx/error.log;
+
+        ##
+        # Gzip Settings
+        ##
+
+        gzip on;
+
+        # gzip_vary on;
```
william-msft, line 48: nit: we should remove the commented code if we don't plan on using it

william-msft (approved):

cormacpayne (approved):

images/runtime/php-fpm/template.base.Dockerfile
waliMSFT, line 1: Do we need the `nginx` only for the php-fpm (not for php)?

pauld-msft: Correct, PHP uses Apache

PR #2017, opened 2023-06-09
Wali/support php 8.2.6 and 8.1.19   1831694

waliMSFT:
- updated php versions: 8.2.6 and 8.1.19 (partners asked for these versions)
- updated php versions: 8.2.7 and 8.1.20 (latest versions)


images/runtime/php-fpm/8.1/base.bullseye.Dockerfile
```diff
@@ -0,0 +1,137 @@
+FROM oryxdevmcr.azurecr.io/private/oryx/php-fpm-8.1
```
pauld-msft, line 1: These should be gitignored now, so we can safely remove these `base.*.Dockerfiles` from the git index with `git rm --cached <file>`

images/runtime/php/8.2/base.bullseye.Dockerfile
```diff
@@ -1,148 +0,0 @@
-FROM oryxdevmcr.azurecr.io/private/oryx/php-8.2
```
snehapar9, line 1: Curious why we are deleting base.*.Dockerfile?

snehapar9 (approved):
LGTM. Approved with one clarification request.

images/runtime/php/8.2/base.bullseye.Dockerfile
cormacpayne: The comment from @pauld-msft above led to this removal: we previously added these auto-generated Dockerfiles to the `.gitignore`, but if the files were already a part of the repository when the changes to the `.gitignore` were made, Git doesn't remove the files. The command Paul mentioned will remove files that are already a part of the repository that match the `.gitignore` changes.

cormacpayne (approved):
LGTM :shipit:

harryli0108:
Just one quick question. Why did we stopped using the `base.bullseye.Dockerfile`?
Other than that, all looks good to me as well. Thanks Wali!

images/runtime/php/8.2/base.bullseye.Dockerfile
snehapar9: Thanks for clarifying @cormacpayne!

william-msft (approved):

PR #2022, opened 2023-06-16
added env variable NGINX_CONF_FILE

waliMSFT:
Added support for a `NGINX_CONF_FILE` environment variable
The goal of this environment variable is to allow customers to modify their `nginx.conf file`, and then to point Oryx to where we should load it.

Suggested Steps to implement:
- When this environment variable is set in the runtime image, then perform these actions:
cp ${NGINX_CONF_FILE} /etc/nginx/nginx.conf
service nginx reload
=> This will move the file from its existing location to the one expected by NGINX, and will restart NGINX.
- Update the documentation to reference that environment variable

Steps followed:
- in the php-fpm runtime-image it looks for the env variable `NGINX_CONF_FILE`. The value of the `NGINX_CONF_FILE` should be a customized nginx configuration file.
- if the env variable is found then it copied the customized nginx configuration file to the `/etc/nginx/nginx.conf`
- it follows reloading the nginx.
- added an integration test to testify the above steps.


snehapar9:
@waliMSFT, seems like the pipelines are failing while building Oryx for some minor errors.
The following changes should get us past the build failures.

1. Please change [this](https://github.com/microsoft/Oryx/blob/wali/env_variable_for_NGINX_conf_file_1784748/src/BuildScriptGenerator/Php/PhpPlatform.cs#L158) block of code to the following piece of code 
```
  var props = new PhpBashBuildSnippetProperties
            {
                ComposerFileExists = composerFileExists,
                NginxConfFile = nginxConfFile,
            };
```
2. Please change [this](https://github.com/microsoft/Oryx/blob/wali/env_variable_for_NGINX_conf_file_1784748/src/BuildScriptGenerator/Options/BuildScriptGeneratorOptions.cs#L96) line to `public string NginxConfFile { get; set; }`

cormacpayne (commented):
Beside the one minor comment left, let's make sure we add a test that verifies that when the environment variable is set, the PHP script we generate has the resulting set of commands we expect to run.

waliMSFT:
> Beside the one minor comment left, let's make sure we add a test that verifies that when the environment variable is set, the PHP script we generate has the resulting set of commands we expect to run.

@cormacpayne Adding test in the 2nd phase of the PR was my plan. Now added the test. Thanks for the suggestrion.

waliMSFT:
> @waliMSFT, seems like the pipelines are failing while building Oryx for some minor errors. The following changes should get us past the build failures.
> 
> 1. Please change [this](https://github.com/microsoft/Oryx/blob/wali/env_variable_for_NGINX_conf_file_1784748/src/BuildScriptGenerator/Php/PhpPlatform.cs#L158) block of code to the following piece of code
> 
> ```
>   var props = new PhpBashBuildSnippetProperties
>             {
>                 ComposerFileExists = composerFileExists,
>                 NginxConfFile = nginxConfFile,
>             };
> ```
> 
> 2. Please change [this](https://github.com/microsoft/Oryx/blob/wali/env_variable_for_NGINX_conf_file_1784748/src/BuildScriptGenerator/Options/BuildScriptGeneratorOptions.cs#L96) line to `public string NginxConfFile { get; set; }`

Already fixed the issues. Thanks

harryli0108:
Could the nginx config file location be defaulted to the current working directory if someone only provided a file name or a relative file path? I'm not sure if this is a possible corner case

waliMSFT:
> a file name or a relative file path?

@harryli0108 I believe our partners will provide the file name with the path under the env variable `NGINX_CONF_FILE`. Our partners suggested what they want with this env variable and how the Oryx team can achieve it.

src/BuildScriptGenerator/Php/PhpBashBuildSnippet.sh.tpl
```diff
@@ -12,3 +12,8 @@ php $composer install --ignore-platform-reqs --no-interaction
 {{ else }}
 echo "No 'composer.json' file found; not running 'composer install'."
 {{ end }}
+
+{{ if NginxConfFile | IsNotBlank }}
+    cp {{ NginxConfFile }} /etc/nginx/nginx.conf
+    service nginx reload
```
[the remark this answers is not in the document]
waliMSFT: obliged.

waliMSFT:
> @waliMSFT, seems like the pipelines are failing while building Oryx for some minor errors. The following changes should get us past the build failures.
> 
> 1. Please change [this](https://github.com/microsoft/Oryx/blob/wali/env_variable_for_NGINX_conf_file_1784748/src/BuildScriptGenerator/Php/PhpPlatform.cs#L158) block of code to the following piece of code
> 
> ```
>   var props = new PhpBashBuildSnippetProperties
>             {
>                 ComposerFileExists = composerFileExists,
>                 NginxConfFile = nginxConfFile,
>             };
> ```
> 
> 2. Please change [this](https://github.com/microsoft/Oryx/blob/wali/env_variable_for_NGINX_conf_file_1784748/src/BuildScriptGenerator/Options/BuildScriptGeneratorOptions.cs#L96) line to `public string NginxConfFile { get; set; }`

Those issues were taken care of in the final version of the PR.

src/BuildScriptGenerator/Php/PhpBashBuildSnippet.sh.tpl
```diff
@@ -12,3 +12,9 @@ php $composer install --ignore-platform-reqs --no-interaction
 {{ else }}
 echo "No 'composer.json' file found; not running 'composer install'."
 {{ end }}
+
+{{ if NginxConfFile | IsNotBlank }}
+echo "Using the custom NGINX configuration from {{ NginxConfFile }} and followed by the `service nginx reload` command."
```
pauld-msft, line 17: Does NGINX even exist within our build images? I thought it was only in the [runtime images for php-fpm](https://github.com/microsoft/Oryx/blob/main/images/runtime/php-fpm/template.base.Dockerfile#L27)... I think that this change should possibly be in the [startup script generator for php](https://github.com/microsoft/Oryx/blob/main/src/startupscriptgenerator/src/php/scriptgenerator.go)?

pauld-msft (changes requested):
Would like the build image vs runtime image question to be answered before merging. Thanks.

src/BuildScriptGenerator/Php/PhpBashBuildSnippet.sh.tpl
waliMSFT: you are right. It should be on the run-time image. I will make the change.

waliMSFT: @pauld-msft made the changes on the runtime images. Also added an integration test. Your feedback is appreciated. Thanks again.

tests/Oryx.Integration.Tests/Php/PhpFpmConfigTest.cs
```diff
@@ -73,6 +73,59 @@ public PhpFpmConfigTest(ITestOutputHelper output, TestTempDirTestFixture fixture
                 failureOutputText);
         }
 
+        [Fact]
```
pauld-msft, line 76:
```suggestion
        [Fact, Trait("category", "php-8.0")]
```
otherwise this test will not run with the testing suite

tests/SampleApps/php/php-fpm-config/NGINX_DUMMY_CUSTOM_CONFIG_FILE.txt
```diff
@@ -0,0 +1,49 @@
+user www-data;
```
pauld-msft, line 1: just so that it more closely represents an expected use case, can we rename this file with a `.conf` extension?

tests/Oryx.Integration.Tests/Php/PhpFpmConfigTest.cs
waliMSFT: Thanks for pointing it.

cormacpayne (approved):
:shipit:

tests/Oryx.Integration.Tests/Php/PhpFpmConfigTest.cs
```diff
@@ -73,6 +73,59 @@ public PhpFpmConfigTest(ITestOutputHelper output, TestTempDirTestFixture fixture
...
+            var buildScript = new ShellScriptBuilder()
+               .AddCommand($"oryx build {appDir} -i /tmp/int -o {appOutputDir} " +
+               $"--platform php --platform-version {phpVersion}")
+               .ToString();
+
+            var runScript = new ShellScriptBuilder()
+                .AddCommand($"oryx create-script -appPath {appOutputDir} -output {RunScriptPath} -bindPort {ContainerPort}")
+                .AddCommand("mkdir -p /home/site/wwwroot")
+                .AddCommand($"cp -rf {appOutputDir}/* /home/site/wwwroot")
+                .AddCommand("cat " + RunScriptPath)
```
pauld-msft, line 100: nit: not sure if this cat statement is needed

pauld-msft (approved):
:shipit:

PR #2028, opened 2023-06-29
Williamhe/1843028 add gpg keys for more python versions

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as: Python Automation needs GPG keys for the versions we want to support. Since we want to support older ones, we need to add their GPG keys for Automation to use. [Versions we want Automation to release](https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1843028):
* Python 3.11.4 (released 06-06)
* Python 3.10.12 (released 06-06)
* Python 3.9.17 (released 06-06)
* Python 3.8.17 (released 06-06)
* Python 3.7.17 (released 06-06)


build/tools/Automation/Python/PythonConstants.cs
```diff
@@ -19,6 +19,9 @@ public class PythonConstants
         public const string PythonSuffixUrl = "/python?restype=container&comp=list&include=metadata";
         public static readonly Dictionary<string, string> VersionGpgKeys = new Dictionary<string, string>()
         {
+            { "3.7", "E3FF2839C048B25C084DEBE9B26995E310250568" },
+            { "3.8", "E3FF2839C048B25C084DEBE9B26995E310250568" },
+            { "3.9", "E3FF2839C048B25C084DEBE9B26995E310250568" },
```
william-msft, line 24: GPG keys' [source](https://github.com/docker-library/repo-info/blob/master/repos/python/remote/3.7.17-bullseye.md)

build/tools/Automation/Python/PythonConstants.cs
```diff
@@ -19,6 +19,9 @@ public class PythonConstants
         public const string PythonSuffixUrl = "/python?restype=container&comp=list&include=metadata";
         public static readonly Dictionary<string, string> VersionGpgKeys = new Dictionary<string, string>()
         {
+            { "3.7", "E3FF2839C048B25C084DEBE9B26995E310250568" },
```
snehapar9, line 22: Would this file be updated every time we add support for a new minor version?

william-msft: That's correct. There was no reliable and consistent way found of obtaining the new GPG keys.

snehapar9 (approved):

build/tools/Automation/Python/PythonConstants.cs
snehapar9: Sounds good! Thank you for clarifying!

harryli0108:
LGTM!
(something not related to the change you've made and also nit): Meanwhile, i saw that the way of us getting the gpg key was actually [taking string and split by dot and get major/minor versions](https://github.com/microsoft/Oryx/blob/main/build/tools/Automation/Python/PythonAutomator.cs#L138-L145). Is it possible to use Versions rather than String?

william-msft:
> LGTM! (something not related to the change you've made and also nit): Meanwhile, i saw that the way of us getting the gpg key was actually [taking string and split by dot and get major/minor versions](https://github.com/microsoft/Oryx/blob/main/build/tools/Automation/Python/PythonAutomator.cs#L138-L145). Is it possible to use Versions rather than String?

That's a good point! I can update in a separate PR. Thanks!

PR #2029, opened 2023-06-29
Revert nginx upgrade

pauld-msft:
Seeing customer issues, need to revert for now


william-msft:
@pauld-msft could you share more context please? I wonder if this is regarding from Anthony's email yesterday to Oryx Team?

pauld-msft:
> @pauld-msft could you share more context please? I wonder if this is regarding from Anthony's email yesterday to Oryx Team?

Yes. This is in response to 2 ICMs: 
[Incident-402089880 Details - IcM (microsofticm.com)](https://portal.microsofticm.com/imp/v3/incidents/details/402089880/home)
[Incident-402107293 Details - IcM (microsofticm.com)](https://portal.microsofticm.com/imp/v3/incidents/details/402107293/home)

In both of these cases, the new versions of the php fpm image seemed to break customers, and reverting to 8.0 without the nginx upgrade fixed one of the customer's issue. The other was fixed through:
```
Issue was mitigated by the Client when he realized that folder to which the default.config file on the app was pointing to did not exist in the same way on the PHP 8.2.7 than on 8.2.5, so he implemented that changes to fix the 403
```

While we have an investigation going, we wanted to get the ball rolling on a revert for the nginx update, so we can mitigate further issues.

william-msft (approved):

daniv-msft (approved):

PR #2068, opened 2023-07-17
Wali/nuget package vulnerability fix

waliMSFT:
updated the version of the nuget package System.Runtime.Uri to 4.3.2

### DIFF
--- a/src/BuildServer/BuildServer.csproj
+++ b/src/BuildServer/BuildServer.csproj
@@ -18,7 +18,9 @@
     <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.17.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
     <PackageReference Include="RunProcessAsTask" Version="1.2.4" />
+    <PackageReference Include="runtime.unix.System.Private.Uri" Version="4.3.2" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.4.0" />
+    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
   </ItemGroup>
 
 </Project>
--- a/src/BuildServer/BuildServer.csproj
+++ b/src/BuildServer/BuildServer.csproj
@@ -18,9 +18,7 @@
     <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.17.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
     <PackageReference Include="RunProcessAsTask" Version="1.2.4" />
-    <PackageReference Include="runtime.unix.System.Private.Uri" Version="4.3.2" />
     <PackageReference Include="Swashbuckle.AspNetCore" Version="6.4.0" />
-    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
   </ItemGroup>
 
 </Project>
--- a/src/Detector/Detector.csproj
+++ b/src/Detector/Detector.csproj
@@ -58,6 +58,7 @@
     <PackageReference Include="Microsoft.Extensions.Http" Version="7.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
     <PackageReference Include="SemanticVersioning" Version="2.0.2" />
+    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
     <PackageReference Include="xunit" Version="2.4.2" />
   </ItemGroup>
 
--- a/src/Detector/Detector.csproj
+++ b/src/Detector/Detector.csproj
@@ -57,7 +57,9 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="7.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
+    <PackageReference Include="runtime.unix.System.Private.Uri" Version="4.3.2" />
     <PackageReference Include="SemanticVersioning" Version="2.0.2" />
+    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
     <PackageReference Include="xunit" Version="2.4.2" />
   </ItemGroup>
 
--- a/src/Detector/Detector.csproj
+++ b/src/Detector/Detector.csproj
@@ -57,9 +57,7 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="7.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
-    <PackageReference Include="runtime.unix.System.Private.Uri" Version="4.3.2" />
     <PackageReference Include="SemanticVersioning" Version="2.0.2" />
-    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
     <PackageReference Include="xunit" Version="2.4.2" />
   </ItemGroup>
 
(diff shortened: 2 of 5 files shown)


src/BuildServer/BuildServer.csproj
```diff
@@ -18,7 +18,9 @@
     <PackageReference Include="Microsoft.VisualStudio.Azure.Containers.Tools.Targets" Version="1.17.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
     <PackageReference Include="RunProcessAsTask" Version="1.2.4" />
+    <PackageReference Include="runtime.unix.System.Private.Uri" Version="4.3.2" />
```
cormacpayne, line 21: Would you mind providing some context as to why this `runtime.unix.System.Private.Uri` package was also added as a package reference throughout these `.csproj` files?

src/Detector/Detector.csproj
```diff
@@ -57,7 +57,9 @@
     <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="7.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
+    <PackageReference Include="runtime.unix.System.Private.Uri" Version="4.3.2" />
```
cormacpayne, line 60: nit: I'm not sure if we need to include these package references in `Detector.csproj` and `Detector.Test.csproj` since they already reference the `Common.csproj` project, so these package references should be propagated downstream to these projects, but it's worth double-checking this

src/BuildServer/BuildServer.csproj
william-msft: Adding a screenshot or a description of the vulnerability we are solving would be nice.

waliMSFT: In the `GenerateBuildScript.deps.json` file our partner shared with us, `system.private.uri` has a direct dependency on the `runtime.unix.System.Private.Uri`. `system.private.uri` is the general-purpose URI library provided by the .NET Core runtime, while the `runtime.unix.System.Private.Uri` is a Unix-specific URI library provided by the runtime.unix package. Both provides the implementation of System.Uri. As the package version `4.3.0` has vulnerability I felt it is better to upgrade the recommended version to `4.3.2`.

image

waliMSFT:
I have updated the dependency in the projects. Our partners showed concern on the `System.Private.Uri` version 4.3.0. `Xunit` has a inner dependency on that package, so updated the `System.Private.Uri` to 4.3.2 in the `Common` project as other projects have dependency on it.

src/Detector/Detector.csproj
waliMSFT: Please have a look on the comment: https://github.com/microsoft/Oryx/pull/2068#issuecomment-1640737635

src/BuildServer/BuildServer.csproj
waliMSFT: Please have a look on the comment: https://github.com/microsoft/Oryx/pull/2068#issuecomment-1640737635

src/Oryx.Common/Common.csproj
```diff
@@ -19,6 +19,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging" Version="7.0.0" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
+    <PackageReference Include="System.Private.Uri" Version="4.3.2" />
```
daniv-msft, line 22: In the end, were you able to validate that if we upgrade xunit to the newest version (2.5.0) in Detector.csproj, it would automatically upgrade System.Private.Uri in the process?
Hopefully Xunit also fixed the vulnerability on their side by upgrading to a newer version of their own dependencies, and as such we might not need to force this version ourselves.
That said, it's not 100% sure they did so it would be good to try to update the nuget and maybe regenerate the deps.json file to see if the wrong System.Private.Uri version is still there?

waliMSFT: Thanks Dan for the suggestion. I have checked the dependencies of `xUnit 2.5.0`. it is like that:
`xUnit 2.5.0 ---> xunit.assert (>= 2.5.0) ---> NETStandard.Library (>= 1.6.1)  ---> System.Runtime (>= 4.3.0) ---> System.Private.Uri (>= 4.3.0)`
This exactly the same as `xUnit 2.4.2`. That's why I have updated the `System.Private.Uri` version to 4.3.2.

daniv-msft:
> After discussing with Dan, I have updated the dependency in the projects. Our partners showed concern on the `System.Private.Uri` version 4.3.0. `Xunit` has a inner dependency on that package, so updated the `System.Private.Uri` to 4.3.2 in the `Common` project as other projects have dependency on it.

Could you please clarify why the other projects apart from Detector don't show the vulnerability? Given that here we need to take a direct dependency on a package we don't need directly, I'd rather make sure we minimize the impact on our project to where it is absolutely needed.

cormacpayne:
@daniv-msft @waliMSFT -- to follow-up on the above comment: it may be worth introducing `packages.lock.json` files for our projects to better understand the dependencies we're taking and where each one is coming from (_i.e.,_ direct vs. transitive). [This](https://devblogs.microsoft.com/nuget/enable-repeatable-package-restores-using-a-lock-file/) doc outlines the changes we'd make in each `.csproj` file if we wanted to generate it.

waliMSFT:
> > After discussing with Dan, I have updated the dependency in the projects. Our partners showed concern on the `System.Private.Uri` version 4.3.0. `Xunit` has a inner dependency on that package, so updated the `System.Private.Uri` to 4.3.2 in the `Common` project as other projects have dependency on it.
> 
> Could you please clarify why the other projects apart from Detector don't show the vulnerability? Given that here we need to take a direct dependency on a package we don't need directly, I'd rather make sure we minimize the impact on our project to where it is absolutely needed.

Among all the projects we have inside `Oryx/src` directory, only the `Detector` project uses the `Xunit` package. Though few other projects have dependency on `Detector` project, but the root dependency on the `Xunit` always comes from the `Detector` project. 
In addition, there are projects do not have dependency on the `Detector` project. for example: `BuildServer, BuildScriptGenerator.Common`, etc. So, they don't possess any vulnerability.

daniv-msft:
> > > After discussing with Dan, I have updated the dependency in the projects. Our partners showed concern on the `System.Private.Uri` version 4.3.0. `Xunit` has a inner dependency on that package, so updated the `System.Private.Uri` to 4.3.2 in the `Common` project as other projects have dependency on it.
> > 
> > 
> > Could you please clarify why the other projects apart from Detector don't show the vulnerability? Given that here we need to take a direct dependency on a package we don't need directly, I'd rather make sure we minimize the impact on our project to where it is absolutely needed.
> 
> @daniv-msft Among all the projects we have inside `Oryx/src` directory, only the `Detector` project uses the `Xunit` package. As few projects have dependency on `Detector` project, so the root dependency on the `Xunit` always comes from the `Detector` project. In addition, there are projects do not have dependency on the `Detector` project. for example: `BuildServer, BuildScriptGenerator.Common`, etc. So, they don't possess any vulnerability.

Thanks for clarifying! In that case, please consider moving the System.Private.Uri nuget reference to Detector.csproj only so that we keep the workaround we need to create here as minimal as possible.

waliMSFT:
> Thanks for clarifying! In that case, please consider moving the System.Private.Uri nuget reference to Detector.csproj only so that we keep the workaround we need to create here as minimal as possible.
@daniv-msft I have made the change as you said. I agree with you to make the change as minimum as possible. However, I was wondering the vulnerability on the `system.private.uri version 4.3.0` will be there if we import the `xUnit` package in any other project. It will create the same vulnerability. As all the projects use the `Common` project, previously I thought to add the updated `system.private.uri version 4.3.2` there.

daniv-msft (approved):

PR #2072, opened 2023-07-19
Williamhe/1853186 update automation to use staging account

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as: https://devdiv.visualstudio.com/DevDiv/_boards/board/t/Oryx/Stories/?workitem=1853186
Automation is switching from using Dev storage account to Stagging storage account, which requires a SAS token.
    - Successful test invocation of the Automation project for [dotnet](https://github.com/microsoft/Oryx/actions/runs/5604169421/jobs/10251782485#step:4:1) and [python](https://github.com/microsoft/Oryx/actions/runs/5604197609/jobs/10251845469#step:4:20).


snehapar9 (dismissed):

build/tools/Automation/Services/HttpService.cs
```diff
@@ -33,7 +33,6 @@ public async Task<string> GetDataAsync(string url)
         {
             try
             {
-                Console.WriteLine("Making GET request to: " + url);
                 HttpResponseMessage response = await this.httpClient.GetAsync(url);
                 Console.WriteLine($"Response received: {response.StatusCode}");
```
snehapar9, line 37: Should we remove this line also?

william-msft: I think we can leave this on as it just outputs: 
image

The reason I removed [line 36](https://github.com/microsoft/Oryx/pull/2072/files#diff-1f46b896008cc1f309d5d61a8c7268ce4d5abb3b683377f4ed0f536e9c278ecdL36) is because we're adding the SAS token to the URL and I don't want that to leak in our logs.

snehapar9: Makes sense!

william-msft: Yes we can change that. Would it be ok if we moved logging for another time? Since this update is needed to unblock Automation from releasing the new .NET SDKs.

build/tools/Automation/DotNet/DotNetAutomator.cs
```diff
@@ -58,8 +58,22 @@ public async Task RunAsync()
...
+                if (string.IsNullOrEmpty(sasToken))
+                {
+                    throw new ArgumentException($"The environment variable {Constants.OryxSdkStagingPrivateSasTokenEnvVar} " +
+                        $"must be provided in order to access {Constants.OryxSdkStagingStorageBaseUrl}");
+                }
+
+                sdkVersionsUrl += "&" + sasToken;
+            }
+
+            this.oryxDotNetSdkVersions = await this.httpService.GetOryxSdkVersionsAsync(sdkVersionsUrl);
```
cormacpayne, line 76: nit: if we have a `Common` project (or some equivalent set of common files shared amongst the different platforms), it would be nice to move this block of logic (from lines 55 to 76) out into its own method that's shared amongst the platforms, and all they need to do is provide the base URL to the method call.

cormacpayne (approved):
One minor refactoring comment, otherwise LGTM :shipit:

.github/workflows/automationTemplate.yaml
```diff
@@ -25,6 +25,7 @@ env:
   PYTHON_MIN_RELEASE_VERSION: ${{ vars.PYTHON_MIN_RELEASE_VERSION }}
   PYTHON_MAX_RELEASE_VERSION: ${{ vars.PYTHON_MAX_RELEASE_VERSION }}
   PYTHON_BLOCKED_VERSIONS_ARRAY: ${{ vars.PYTHON_BLOCKED_VERSIONS_ARRAY }}
+  ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN: ${{ secrets.ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN }}
```
waliMSFT, line 28: how the secrets are passed here?

build/tools/Automation/DotNet/DotNetAutomator.cs
waliMSFT: +1

build/tools/Automation/DotNet/DotNetAutomator.cs
```diff
@@ -58,8 +58,22 @@ public async Task RunAsync()
...
+            if (this.oryxSdkStorageBaseUrl == Constants.OryxSdkStagingStorageBaseUrl)
+            {
+                string sasToken = Environment.GetEnvironmentVariable(Constants.OryxSdkStagingPrivateSasTokenEnvVar);
+                if (string.IsNullOrEmpty(sasToken))
+                {
+                    throw new ArgumentException($"The environment variable {Constants.OryxSdkStagingPrivateSasTokenEnvVar} " +
+                        $"must be provided in order to access {Constants.OryxSdkStagingStorageBaseUrl}");
+                }
+
+                sdkVersionsUrl += "&" + sasToken;
```
waliMSFT, line 73: is it needed to add the parameter separator `&` in between the url and token?

build/tools/Automation/DotNet/DotNetAutomator.cs
```diff
@@ -58,8 +58,22 @@ public async Task RunAsync()
...
             }
 
-            this.oryxDotNetSdkVersions = await this.httpService.GetOryxSdkVersionsAsync(
-                this.oryxSdkStorageBaseUrl + DotNetConstants.DotNetSuffixUrl);
+            string sdkVersionsUrl = this.oryxSdkStorageBaseUrl + DotNetConstants.DotNetSuffixUrl;
+
+            // A SAS token is required for the staging account.
+            if (this.oryxSdkStorageBaseUrl == Constants.OryxSdkStagingStorageBaseUrl)
+            {
+                string sasToken = Environment.GetEnvironmentVariable(Constants.OryxSdkStagingPrivateSasTokenEnvVar);
```
waliMSFT, line 66: is there any chance we need to run the process locally? If we run the process locally it won't have the secrets from the Oryx-pipeline-library (I am assuming the secrets are retrieved from the oryx-pipeline-library).

build/tools/Automation/DotNet/DotNetAutomator.cs
william-msft: Yeah let me get this one in first and I can add that separately. Since the pipelines seem to have noticeable transient intermittent errors.

build/tools/Automation/Services/HttpService.cs
snehapar9: Sounds good to me!

.github/workflows/automationTemplate.yaml
william-msft: [This GitHub doc ](https://stackoverflow.com/questions/68008181/how-do-i-get-all-github-secrets-into-env-variables-for-actions-to-access-powers)can provide more details. Basically using `${{ secrets.ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN }}` will get the encrypted secret `ORYX_SDK_STORAGE_ACCOUNT_ACCESS_TOKEN`

build/tools/Automation/DotNet/DotNetAutomator.cs
william-msft: Yes, because the URL contains other multiple parameters, for accessing the list of SDKs. Example: https://oryxsdksprod.blob.core.windows.net/python?`restype=container&comp=list&include=metadata`

build/tools/Automation/DotNet/DotNetAutomator.cs
william-msft: I only run locally to test this. This primarily runs on GitHub Actions. If I need to run locally, I can store the token manually.

build/tools/Automation/DotNet/DotNetAutomator.cs
william-msft: https://github.com/microsoft/Oryx/pull/2092

PR #2105, opened 2023-08-21
New Python Sdks 2023-08-21

william-msft:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


waliMSFT:
@william-msft,
 
 We need to re-run automation from the beginning because of the below reasons. 
 
 There was a bug in the baseimage dockerfiles for python 3.10 and 3.11 where the `sasToken` was not passed in a `Run` command. For that, the `baseImage pipeline` was failing. However, as the `Platform-Binaries` pipeline ran successfully for python, it uploaded the latest binaries in the staging environment. Unfortunately, this is causing the mismatch and failing in the tests (which checks the python's latest versions) of the validation and nightly pipelines in other PRs. 

I have already created and merged a PR which fixed the `sasToken` issue in the baseImage dockerfile for python. Besides, to unblock the validation and nightly pipelines, I removed the uploaded latest versions of python from the staging environment.

Now, if you re-run this automation by taking a branch from the `main` which has the fix (related to the sasToken) then everything should be fine. (the `platformBinaries` pipelines needs to be run again as I deleted the latest python binaries that this automation uploaded earlier in the staging environment).

N.B. - the situation might be confusing. So, please don't hesitate to contact me.

PR #2108, opened 2023-08-23
hotfix baseimage scirpt

waliMSFT:
- buildRunImageBases.sh script was failing because the sastoken is not passed through it. 
- added condition to log the info when the sasToken is valid in a couple of platform-installing scripts.

### DIFF
--- a/build/buildRunTimeImageBases.sh
+++ b/build/buildRunTimeImageBases.sh
@@ -42,6 +42,9 @@ then
     fi
 fi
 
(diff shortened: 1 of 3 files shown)


william-msft (approved):

waliMSFT:
> @waliMSFT could you confirm we aren't leaking SAS token with these changes?

in these changes I hid the logs I dealt with `sasToken`. So, the token should not be leaked. Besides, I tested locally to confirm it. So, we should be good.

PR #2116, opened 2023-08-24
Add support for Node 16.20.2 and 18.17.1

cormacpayne:
This PR adds platform binary and runtime image support for Node 16.20.2 and 18.17.1.


doc/supportedPlatformVersions.md
```diff
@@ -623,6 +626,7 @@
 - 16.19.0
 - 16.20.0
 - 16.20.1
+- 16.20.2
```
snehapar9, line 629: Do we need to update minor versions for node runtimes [here](https://github.com/microsoft/Oryx/blob/main/doc/supportedRuntimeVersions.md#node)?

cormacpayne: We do not because our Node runtime images are published with _just_ the major version, but if we ever wanted to be more granular with the runtime image tags (specifically for the buildpack case where we'd want to do runtime swapping), then we'll want to add a tag that includes the minor version, and maybe patch as well.

snehapar9: Thanks for clarifying @cormacpayne :)

images/runtime/node/14/base.buster.Dockerfile
```diff
@@ -20,6 +20,7 @@ ARG DEBIAN_FLAVOR
 ENV NODE_VERSION ${NODE14_VERSION}
 ENV NPM_CONFIG_LOGLEVEL info
 
+ARG BUILD_DIR=/tmp/oryx/build
```
daniv-msft, line 23: Could you please share more info about this change? Why do we need to add this now?

daniv-msft (approved):
Looks good to me, but please don't merge it immediately as we're still releasing other changes that have a higher urgency.

images/runtime/node/14/base.buster.Dockerfile
cormacpayne: This line used to exist in this file (since it's used to reference the `__nodeVersions.sh` script a few lines down), but I think it was accidentally removed in a previous PR, so I'm just adding it back so this Dockerfile (and the others where this change was made) can successfully build.

cormacpayne: [Here](https://github.com/microsoft/Oryx/commit/cb7d9947017fe217a8664a401548f5080e472a36#diff-fe86717626fd70be5d55410ea401debfdb9ce42d571a8b4053d285bbf25ad2e4L24) is the commit where it was accidentally removed for context

william-msft (approved):

pauld-msft (approved):

PR #2117, opened 2023-08-24
Wali/pipeline token issue fix

waliMSFT:

### DIFF
--- a/images/build/installDotNetCore.sh
+++ b/images/build/installDotNetCore.sh
@@ -24,9 +24,9 @@ if [ "$sdkStorageAccountUrl" == "$PRIVATE_STAGING_SDK_STORAGE_BASE_URL" ]; then
     set -x
     
     if [ $isSasTokenEmpty -eq 0 ]; then
-      echo "sasToken cannot be empty for $sdkStorageAccountUrl."
+        echo "sasToken cannot be empty for $sdkStorageAccountUrl."
     else
-        echo "sasToken is valid for $sdkStorageAccountUrl."
+        echo "sasToken is empty for $sdkStorageAccountUrl."
     fi
 fi
 echo
--- a/images/installPlatform.sh
+++ b/images/installPlatform.sh
@@ -61,7 +61,7 @@ if [ "$sdkStorageAccountUrl" == "$PRIVATE_STAGING_SDK_STORAGE_BASE_URL" ]; then
     if [ $isSasTokenEmpty -eq 0 ]; then
       echo "sasToken cannot be empty for $sdkStorageAccountUrl."
     else
-        echo "sasToken is valid for $sdkStorageAccountUrl."
+      echo "sasToken is empty for $sdkStorageAccountUrl."
     fi
 fi
 if [ -z "$debianFlavor" ] || [ "$debianFlavor" == "stretch" ]; then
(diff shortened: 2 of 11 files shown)


images/build/installDotNetCore.sh
```diff
@@ -24,9 +24,9 @@ if [ "$sdkStorageAccountUrl" == "$PRIVATE_STAGING_SDK_STORAGE_BASE_URL" ]; then
     set -x
     
     if [ $isSasTokenEmpty -eq 0 ]; then
-      echo "sasToken cannot be empty for $sdkStorageAccountUrl."
+        echo "sasToken cannot be empty for $sdkStorageAccountUrl."
     else
-        echo "sasToken is valid for $sdkStorageAccountUrl."
+        echo "sasToken is empty for $sdkStorageAccountUrl."
```
cormacpayne, line 29: Is this correct? `isSasTokenEmpty` is set to 0 on line 22 if the value of `sasToken` provided is empty, so the above if-statement is correct and this `else` statement should only be hit if `sasToken` has some value.

images/installPlatform.sh
```diff
@@ -61,7 +61,7 @@ if [ "$sdkStorageAccountUrl" == "$PRIVATE_STAGING_SDK_STORAGE_BASE_URL" ]; then
     if [ $isSasTokenEmpty -eq 0 ]; then
       echo "sasToken cannot be empty for $sdkStorageAccountUrl."
     else
-        echo "sasToken is valid for $sdkStorageAccountUrl."
+      echo "sasToken is empty for $sdkStorageAccountUrl."
```
cormacpayne, line 64: Same comment here.

vsts/pipelines/baseImages/dotnetcore.yml
```diff
@@ -1,8 +1,13 @@
...
-    Packaging.EnableSBOMSigning: true
+- group: Oryx
+- name: ascName
+  value: oryx-automation-service-principal
+- name: acrName
+  value: oryxdevmcr.azurecr.io
+- name: skipComponentGovernanceDetection
+  value: true
+- name: Packaging.EnableSBOMSigning
+  value: true
```
cormacpayne, line 10: nit: do we need to change the format of all of these existing variables to add the Oryx group? Would we not be able to do something like the following:

```yaml
variables:
  group: Oryx
  ascName: oryx-automation-service-principal
  acrName: oryxdevmcr.azurecr.io
  skipComponentGovernanceDetection: true
  Packaging.EnableSBOMSigning: true
```

images/build/installDotNetCore.sh
```diff
@@ -24,9 +24,9 @@ if [ "$sdkStorageAccountUrl" == "$PRIVATE_STAGING_SDK_STORAGE_BASE_URL" ]; then
     set -x
     
     if [ $isSasTokenEmpty -eq 0 ]; then
-      echo "sasToken cannot be empty for $sdkStorageAccountUrl."
+        echo "sasToken cannot be empty for $sdkStorageAccountUrl."
```
snehapar9, line 27: Is the space here before `echo` intended?

william-msft:
> @waliMSFT Could you please confirm that we don't need to be checked in before the PR from @william-msft for .NET 8 preview? I want to make sure that we don't delay that last one.

@daniv-msft we'll need a fix for the SAS token issue in Oryx-CI pipeline: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8305442&view=logs&j=f45c9538-ea56-56fc-09d6-585ee0335e0c&t=4c7d1d52-4268-5964-8ee9-edcf23ad5c3c&l=2613

PR #2125, opened 2023-08-31
Retag runtime images with specific platform version used

cormacpayne:
Currently, our runtime images are tagged to include a version that encapsulates the "latest version" model rather than a "specific version". For example, our .NET runtime images are tagged as `dotnetcore:7.0-debian-buster`, which means that this runtime image will include the latest version of the .NET 7.0 runtime that we've adopted rather than a more granular version.

With this PR, I'm looking to retag this image to include the more granular version of the platform version used in the runtime. Using the same example as above, we'd continue to publish an image that's tagged as `dotnetcore:7.0-debian-buster`, but we'd also retag that image as `dotnetcore:7.0.10-debian-buster` and publish that as well.

This more granular versioning will allow for detecting patches in our source-to-cloud flow to be much easier as we can simply compare the specific versions in the tag rather than `7.0` vs. `7.0`.

This PR also some miscellaneous minor changes to support platform runtime images being published across multiple Debian flavors, but that will be completed in a separate PR.


build/buildRunTimeImageBases.sh
```diff
@@ -81,7 +81,7 @@ dockerFileName="base.$runtimeImageDebianFlavor.Dockerfile"
 stagingDockerFileName="base.$runtimeImageDebianFlavor.staging.Dockerfile"
 dockerFiles=$(find $runtimeImagesSourceDir -type f \( -name $dockerFileName -o -name $stagingDockerFileName \))
 
-busterNodeDockerFiles=()
+nodeDockerfiles=()
```
snehapar9, line 84: I'm guessing this was re-named because we support bookworm, buster and bullseye for node now but I'm curious why it was named `busterNodeDockerFiles` before? Didn't we support bullseye images from before?

images/runtime/node/18/bullseye.Dockerfile
```diff
@@ -13,11 +13,11 @@ ENV GIT_COMMIT=${GIT_COMMIT}
 ENV BUILD_NUMBER=${BUILD_NUMBER}
 RUN ./build.sh node /opt/startupcmdgen/startupcmdgen
 
-FROM mcr.microsoft.com/oryx/base:node-18-20221112.1
+FROM mcr.microsoft.com/oryx/base:node-18-debian-bullseye-20230707.2
```
snehapar9, line 16: For my own understanding - I'm not sure why we have this file checked in only for node 18 and not 16 and 14.
Shouldn't this file be generated while running the [generateDockerfiles](https://github.com/microsoft/Oryx/blob/main/images/runtime/node/generateDockerfiles.sh) script?

build/buildRunTimeImageBases.sh
cormacpayne: When we first introduced support for buster a few years ago, we added this section to build Node base images for _only_ buster, and there was an if-statement that would check the flavor for buster and would add the Dockerfiles to the `dockerfiles` variable. Over the years, this logic has changed a little and the aforementioned if-statement was removed so we could build Node base images for all supported Debian flavors (since we have Node 18 on bullseye and Node 14 and 16 on buster). When we removed the if-statement, we didn't change the name of the variable reflect the universal Debian flavor image building we were doing.

images/runtime/node/18/bullseye.Dockerfile
cormacpayne: That's true, this file _should_ be generated via `generateDockerfiles.sh` during CI -- this file pattern is actually covered by the `.gitignore`, so I'm assuming the file was uploaded before we updated the `.gitignore`, so I'll delete it since it shouldn't be maintained within the repository.

build/buildRunTimeImages.sh
```diff
@@ -15,6 +15,105 @@ source $REPO_DIR/build/__functions.sh
...
+        return 1
+    fi
+
+    if [ -z $2 ]
+    then
+        echo "Runtime platform version was not provided"
+        return 1
+    fi
+
+    PLATFORM_NAME=$1
```
harryli0108, line 41: In order to use these two args, do we need to modify the pipeline templates we have currently? Seems like the current `_buildTemplate.yml` doesn't have these two parameters yet.

cormacpayne: These arguments are scoped within the `getRuntimeTagVersion` function I added -- I call `getRuntimeTagVersion` later in this file and pass the `$platformName` and `$platformVersion` variables to it.

cormacpayne:
Closing this in favor of a PR I'll open at a later date that includes these changes along with the Debian multi-flavor change

PR #2128, opened 2023-09-07
Add support for latest patch versions of PHP composer 2.2 - 2.6

cormacpayne:
This PR aims to refresh the PHP composer versions available to customers to use from our storage account via the `PHP_COMPOSER_VERSION` environment variable, but **does not** change the current default version used.

The following versions were added as a part of this PR:
- 2.2.21
- 2.3.10
- 2.4.4
- 2.5.8
- 2.6.2

---


tests/Oryx.BuildImage.Tests/Php/PhpDynamicInstallationTest.cs
```diff
@@ -39,7 +40,7 @@ public PhpDynamicInstallationTest(ITestOutputHelper output) : base(output)
                 data.Add(PhpVersions.Php81Version, imageHelper.GetGitHubActionsBuildImage(ImageTestHelperConstants.GitHubActionsBuster), PhpVersions.ComposerVersion);
                 data.Add(PhpVersions.Php82Version, imageHelper.GetGitHubActionsBuildImage(ImageTestHelperConstants.GitHubActionsBuster), PhpVersions.ComposerVersion);
```
snehapar9, line 42: Curious why we have `PhpVersions.ComposerVersion`? Isn't that the same as `PhpVersions.Composer20Version`?

tests/Oryx.BuildImage.Tests/Php/PhpDynamicInstallationTest.cs
```diff
@@ -49,6 +50,39 @@ public PhpDynamicInstallationTest(ITestOutputHelper output) : base(output)
                 data.Add(PhpVersions.Php80Version, imageHelper.GetGitHubActionsBuildImage(ImageTestHelperConstants.GitHubActionsBuster), PhpVersions.Composer23Version);
                 data.Add(PhpVersions.Php81Version, imageHelper.GetGitHubActionsBuildImage(ImageTestHelperConstants.GitHubActionsBuster), PhpVersions.Composer23Version);
                 data.Add(PhpVersions.Php82Version, imageHelper.GetGitHubActionsBuildImage(ImageTestHelperConstants.GitHubActionsBuster), PhpVersions.Composer23Version);
+
+                // Test PHP composer version 2.4.x
```
snehapar9, line 54: Should we add a test for ` Composer22Version`?

tests/Oryx.BuildImage.Tests/Php/PhpDynamicInstallationTest.cs
```diff
@@ -59,12 +93,26 @@ public PhpDynamicInstallationTest(ITestOutputHelper output) : base(output)
...
+
+                // Test default PHP composer version
                 data.Add(PhpVersions.Php73Version, imageHelper.GetCliImage(),PhpVersions.ComposerVersion);
                 data.Add(PhpVersions.Php74Version, imageHelper.GetCliImage(), PhpVersions.ComposerVersion);
 
-                // test latest php-composer version
+                // Test PHP composer version 2.3.x
                 data.Add(PhpVersions.Php73Version, imageHelper.GetCliImage(), PhpVersions.Composer23Version);
                 data.Add(PhpVersions.Php74Version, imageHelper.GetCliImage(), PhpVersions.Composer23Version);
+
```
william-msft, line 104: I'm wondering if we should stop adding tests for stretch. Or maybe even start removing stretch tests when we come across them? Since Stretch has been EOL [for a while](https://endoflife.date/debian) now. If/when the stretch tests were to fail, I wonder if it'll be of value for us to mitigate. We'd also save some pipeline test time by opting out stretch.

tests/Oryx.BuildImage.Tests/Php/PhpDynamicInstallationTest.cs
```diff
@@ -115,16 +185,36 @@ public PhpDynamicInstallationTest(ITestOutputHelper output) : base(output)
...
+                data.Add(PhpVersions.Php81Version, imageHelper.GetCliBuilderImage(ImageTestHelperConstants.CliBuilderBullseyeTag), PhpVersions.Composer25Version);
+                data.Add(PhpVersions.Php82Version, imageHelper.GetCliBuilderImage(ImageTestHelperConstants.CliBuilderBullseyeTag), PhpVersions.Composer25Version);
+
+                // Test PHP composer version 2.6.x
+                data.Add(PhpVersions.Php74Version, imageHelper.GetCliBuilderImage(ImageTestHelperConstants.CliBuilderBullseyeTag), PhpVersions.Composer26Version);
+                data.Add(PhpVersions.Php80Version, imageHelper.GetCliBuilderImage(ImageTestHelperConstants.CliBuilderBullseyeTag), PhpVersions.Composer26Version);
+                data.Add(PhpVersions.Php81Version, imageHelper.GetCliBuilderImage(ImageTestHelperConstants.CliBuilderBullseyeTag), PhpVersions.Composer26Version);
+                data.Add(PhpVersions.Php82Version, imageHelper.GetCliBuilderImage(ImageTestHelperConstants.CliBuilderBullseyeTag), PhpVersions.Composer26Version);
```
pauld-msft, line 217: Not needed for this PR, but these seems like the perfect case for setting up combinatorial xunit test cases: https://github.com/AArnott/Xunit.Combinatorial

pauld-msft (dismissed):
LGTM

tests/Oryx.BuildImage.Tests/Php/PhpDynamicInstallationTest.cs
cormacpayne: It's the same version, but `PhpVersions.ComposerVersion` represents the default composer version that's used -- let me see if I can easily rename this property to make it more clear

tests/Oryx.BuildImage.Tests/Php/PhpDynamicInstallationTest.cs
cormacpayne: That's a good idea -- better to be consistent and have test coverage here

tests/Oryx.BuildImage.Tests/Php/PhpDynamicInstallationTest.cs
cormacpayne: Spoke offline about this -- I don't think there's any action required in this PR

tests/Oryx.BuildImage.Tests/Php/PhpDynamicInstallationTest.cs
cormacpayne: Thanks for providing this! I think we're in need of cleaning up our tests (removing references to stretch and unsupported platform versions, updating our testing matrix used by ADO, etc.), so this could be bundled in with that work for sure.

william-msft (approved):

PR #2129, opened 2023-09-07
Publish runtime images with specific version and multiple Debian flavors

cormacpayne:
This PR introduces the following changes to how our runtime images are published:
- To support patching in our source-to-cloud flow, runtime images are re-tagged with the specific version of the platform binaries that were baked with it; _e.g._, the runtime image `python:3.11-debian-bullseye` will also be tagged as `python:3.11.0-debian-bullseye`
  - This will enable our patching flow to compare the previous platform version an application image was built on top of to the other versions we have readily available on MCR
- Runtime images are currently only published for a single Debian flavor; depending on the platform binaries readily available in our storage account, we will create new runtime images for each Debian flavor supported for a given platform version
  - This change will allow us to support more platform versions per OS-specific builder that we produce for the source-to-cloud flow

---


images/runtime/php/8.2/8.2.buster.Dockerfile
```diff
@@ -0,0 +1,260 @@
...
+# https://bugs.php.net/bug.php?id=74125
+	if [ ! -d /usr/include/curl ]; then \
+		ln -sT "/usr/include/$debMultiarch/curl" /usr/local/include/curl; \
+	fi; \
+	./configure \
+		--build="$gnuArch" \
+		--with-config-file-path="$PHP_INI_DIR" \
+		--with-config-file-scan-dir="$PHP_INI_DIR/conf.d" \
+		\
+# make sure invalid --configure-flags are fatal errors intead of just warnings
```
william-msft, line 178: nit: comment spacing alignment with the code

build/buildRunTimeImages.sh
```diff
@@ -15,6 +15,140 @@ source $REPO_DIR/build/__functions.sh
...
 source $REPO_DIR/build/__stagingRuntimeConstants.sh
 
+# Load platform versions for runtimes
+source $REPO_DIR/build/__dotNetCoreRunTimeVersions.sh
+source $REPO_DIR/build/__nodeVersions.sh
+source $REPO_DIR/build/__phpVersions.sh
+source $REPO_DIR/build/__pythonVersions.sh
+source $REPO_DIR/build/__rubyVersions.sh
+
+# Get the specific platform version used for each runtime image to assist with future patching
```
william-msft, line 25: nit: we can shorten `getRuntimeTagVersion` a little:
```

#!/bin/bash

declare -A versionMappings

# Define version mappings
versionMappings["dotnetcore"]="8.0:NET_CORE_APP_80 7.0:NET_CORE_APP_70 6.0:NET_CORE_APP_60 5.0:NET_CORE_APP_50 3.1:NET_CORE_APP_31 3.0:NET_CORE_APP_30"
versionMappings["node"]="18:NODE18_VERSION 16:NODE16_VERSION 14:NODE14_VERSION"
versionMappings["php"]="8.2:PHP82_VERSION 8.1:PHP81_VERSION 8.0:PHP80_VERSION 7.4:PHP74_VERSION"
versionMappings["python"]="3.11:PYTHON311_VERSION 3.10:PYTHON310_VERSION 3.9:PYTHON39_VERSION 3.8:PYTHON38_VERSION 3.7:PYTHON37_VERSION"
versionMappings["ruby"]="2.7:RUBY27_VERSION 2.6:RUBY26_VERSION 2.5:RUBY25_VERSION"

# Function to get full runtime tag version
function getRuntimeTagVersion() {
    local PLATFORM_NAME="$1"
    local PLATFORM_VERSION="$2"
    
    # Check if the platform name is valid
    if [ -z "${versionMappings["$PLATFORM_NAME"]}" ]; then
        echo "Unable to retrieve version from the provided runtime platform name '$PLATFORM_NAME'"
        return 1
    fi

    # Use associative array for version mappings
    local versions="${versionMappings["$PLATFORM_NAME"]}"
    for versionMapping in $versions; do
        IFS=':' read -ra mapping <<< "$versionMapping"
        key="${mapping[0]}"
        value="${mapping[1]}"
        if [ "$PLATFORM_VERSION" == "$key" ]; then
            FULL_RUNTIME_TAG_VERSION="${!value}"
            return 0
        fi
    done

    # If version not found in mappings, use the provided version
    FULL_RUNTIME_TAG_VERSION="$PLATFORM_VERSION"
    return 0
}
```

william-msft:
```
# Example usage:
PLATFORM_NAME="dotnetcore"
PLATFORM_VERSION="7.0"
getRuntimeTagVersion "$PLATFORM_NAME" "$PLATFORM_VERSION"
echo "Full Runtime Tag Version: $FULL_RUNTIME_TAG_VERSION"
```

build/buildRunTimeImages.sh
```diff
@@ -15,6 +15,140 @@ source $REPO_DIR/build/__functions.sh
...
+
+    PLATFORM_NAME=$1
+    PLATFORM_VERSION=$2
+
+    if [ "$PLATFORM_NAME" == "dotnetcore" ]
+    then
+        case $PLATFORM_VERSION in
+
+            8.0)
+                FULL_RUNTIME_TAG_VERSION=$NET_CORE_APP_80
```
pauld-msft, line 49: This new method propagates the limitation that currently exists in our version selection today where we cannot release different patch versions of the same minor version to different OS types.

This means that if we stopped pushing patch versions to stretch (for example), `dotnetcore:7.0-debian-stretch` would still return the latest patch version that we have defined in the constants file.

Ideally, we can fix this in the future by writing scripts to read and sort the versionsToBuild.txt files by semver version, and then use the most recent version from in those.

Not particularly something that I think we should address here, but wanted to bring up that we likely want to rework this in the future to support different patch versions for different flavors / os types.

pauld-msft (commented):
Surprised there are no pipeline changes needed? I know that updates in the build images have [some hardcoded values](https://github.com/microsoft/Oryx/blob/main/vsts/scripts/pullAndTag.sh#L98-L116) that need to be changes when those tags change, but i guess we don't have to for runtime?

build/buildRunTimeImageBases.sh
```diff
@@ -91,20 +91,22 @@ if [ "$runtimeSubDir" == "node" ]; then
...
 
     NODE_VERSION_ARRAY=()
-    if  [ "$runtimeImageDebianFlavor" == "buster" ]; then
-        NODE_VERSION_ARRAY=(${NODE_BUSTER_VERSION_ARRAY[@]})
+    if [ "$runtimeImageDebianFlavor" == "bookworm" ];then
+        NODE_VERSION_ARRAY=("${NODE_BOOKWORM_VERSION_ARRAY[@]}")
     elif [ "$runtimeImageDebianFlavor" == "bullseye" ];then
         NODE_VERSION_ARRAY=("${NODE_BULLSEYE_VERSION_ARRAY[@]}")
+    elif [ "$runtimeImageDebianFlavor" == "buster" ]; then
+        NODE_VERSION_ARRAY=(${NODE_BUSTER_VERSION_ARRAY[@]})
```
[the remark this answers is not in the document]
cormacpayne: Nice catch 🙏

build/buildRunTimeImages.sh
cormacpayne: Just to make sure I understand the issue: it's not _guaranteed_ that the full version we're using in each scenario is the latest (or even supported) version for a given Debian flavor, correct?

If so, then I think you're right that we'll need to read from the `versionsToBuild.txt` file to guarantee that platform versions we have available for each Debian flavor in our storage account can be referenced here, but it would require just a bit more refactoring that we may or may not need to do at the moment 😃

images/runtime/php/8.2/8.2.buster.Dockerfile
cormacpayne: I think this is how the comments were aligned in the Dockerfiles as I copied them over from other PHP images -- I think I'm going to leave as-is for now if that's OK 😃

cormacpayne:
> Surprised there are no pipeline changes needed? I know that updates in the build images have [some hardcoded values](https://github.com/microsoft/Oryx/blob/main/vsts/scripts/pullAndTag.sh#L98-L116) that need to be changes when those tags change, but i guess we don't have to for runtime?

@pauld-msft -- let me get a successful Validation and Nightly run before making a final comment on this 😃 There may be something I overlooked when updating things across the board, but I'll know for sure from the logs what was and wasn't released, but so far it seems that everything was at least _tested_ successfully, which is a good start.

william-msft (dismissed):

cormacpayne:
@microsoft-github-policy-service agree

william-msft (approved):

build/buildRunTimeImages.sh
william-msft: Gentle reminder regarding using code above.

cormacpayne: This is a good refactoring comment -- I decided not to make the change for this PR since Paul left a comment below that may change how we're reading versions, so we may scrap this logic in the future, but I should be able to blend your feedback with a future refactor involving Paul's proposal.

PR #2132, opened 2023-09-08
Update automation pipeline check retries

william-msft:
  - I've noticed our pipelines are increasingly taking longer to run, which has started to mess with Automation. I've increased the number of retries to mitigate this obstacle. [Example](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8371936&view=results):
    -  image

### DIFF
--- a/.github/workflows/automationTemplate.yaml
+++ b/.github/workflows/automationTemplate.yaml
@@ -16,7 +16,7 @@ on:
         description: 'The platform name should be all lowercase'
 
 env:
-  MAX_RETRIES: 10
+  MAX_RETRIES: 15
   TIMEOUT_SECONDS: 900
   ORYX_SDK_STORAGE_BASE_URL: ${{ vars.ORYX_SDK_STORAGE_BASE_URL }}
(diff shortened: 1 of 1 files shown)


cormacpayne (approved):
:shipit:

.github/workflows/automationTemplate.yaml
```diff
@@ -16,7 +16,7 @@ on:
         description: 'The platform name should be all lowercase'
 
 env:
-  MAX_RETRIES: 10
+  MAX_RETRIES: 15
```
waliMSFT, line 19: I am wondering how increasing the number of retries will mitigate the issue?

william-msft: If the he pipeline does not finish within 1.5hrs then the pipeline will fail. Currently its taking more than that (shown in the picture of the description) takes over 2 hours.

william-msft: If the pipeline does not finish within 1.5hrs then the pipeline will fail. Currently its taking more than that (shown in the picture of the description) takes over 2 hours.

PR #2152, opened 2023-09-18
Sync php versions from constants.yaml

william-msft:
  - this [php versions file](https://github.com/microsoft/Oryx/pull/2148/files#diff-c8a0a40b2a20e54011fc6e980f570aabbcd662f0966f2e7219e606b1aad9334bR39) is out of sync with the constants.yaml file, which is causing automation to create PRs since there are changes in constants.yaml not yet distributed throughout Oryx.
  - This PR will sync the php versions with what exists in the constants.yaml file
  - This file gets populated after running `./build/generateConstants.sh`

### DIFF
--- a/tests/Oryx.Tests.Common/ImageTestHelper.cs
+++ b/tests/Oryx.Tests.Common/ImageTestHelper.cs
@@ -495,12 +495,6 @@ namespace Microsoft.Oryx.Tests.Common
                 NodeVersions.RuntimeVersions
             },
             {
-                PhpConstants.PlatformName,
-                PhpVersions.RuntimeVersions
-                .Concat(PhpVersions.FpmRuntimeVersions)
-                .ToList()
-            },
-            {
                 PythonConstants.PlatformName,
                 PythonVersions.RuntimeVersions
             },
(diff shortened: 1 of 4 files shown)


cormacpayne (dismissed):
:shipit:

tests/Oryx.Tests.Common/ImageTestHelper.cs
```diff
@@ -494,12 +494,6 @@ private string GetFullBuildImage(string buildImageTag = null)
                 NodeConstants.NodeToolName,
                 NodeVersions.RuntimeVersions
             },
-            {
```
cormacpayne, line 497: As mentioned offline: we should be OK to make the below change to the [`ShellOutput`](https://github.com/microsoft/Oryx/blob/main/build/tools/SharedCodeGenerator/Outputs/ShellOutput.cs#L56) file rather than making any changes in this PR to `.cs` files:

`!c.Key.Equals("runtime-versions")` --> `!c.Key.Contains("runtime-versions")`

cormacpayne (changes requested):

tests/Oryx.Tests.Common/ImageTestHelper.cs
william-msft: Created different PR for this: https://github.com/microsoft/Oryx/pull/2156/files

william-msft:
Closing in favor for: https://github.com/microsoft/Oryx/pull/2156/files

PR #2156, opened 2023-09-19
Update to ignore fpm-runtime-versions

william-msft:
- [x] The purpose of this PR is explained in this message or in an issue. If an issue please include a [reference ](url)as: 
  - We ignore `fpm-runtime-versions` because `/bin/sh` does not support arrays and will cause pipeline failures. 
  - Failure: 
image

### DIFF
--- a/build/tools/SharedCodeGenerator/Outputs/ShellOutput.cs
+++ b/build/tools/SharedCodeGenerator/Outputs/ShellOutput.cs
@@ -53,7 +53,7 @@ namespace Microsoft.Oryx.SharedCodeGenerator.Outputs
             if (this.collection.ListConstants?.Any() ?? false)
             {
                 // exclude runtime versions constants lists, as they will just overwrite each other in shell
-                foreach (var constant in this.collection.ListConstants.Where(c => !c.Key.Equals("runtime-versions")))
+                foreach (var constant in this.collection.ListConstants.Where(c => !c.Key.Equals("runtime-versions") && !c.Key.Equals("fpm-runtime-versions")))
                 {
                     string name = constant.Key.Replace(ConstantCollection.NameSeparator[0], '_').ToUpper();
                     var value = constant.Value.Count != 0


build/tools/SharedCodeGenerator/Outputs/ShellOutput.cs
```diff
@@ -53,7 +53,7 @@ public string GetContent()
             if (this.collection.ListConstants?.Any() ?? false)
             {
                 // exclude runtime versions constants lists, as they will just overwrite each other in shell
-                foreach (var constant in this.collection.ListConstants.Where(c => !c.Key.Equals("runtime-versions")))
+                foreach (var constant in this.collection.ListConstants.Where(c => !c.Key.Equals("runtime-versions") && !c.Key.Equals("fpm-runtime-versions")))
```
william-msft, line 56: We cannot use `.Contains` or `.EndsWith`, in the loop, since it will affect changes in this file:
image

cormacpayne: Ah, that makes sense -- if we decide to skip anything in the future, it may make sense to abstract the "strings to ignore" into a separate variable and iterate over those with LINQ, but this is good for now.

cormacpayne (approved):
Change looks good to me -- does `main` already include the change to remove the FPM array from the corresponding Bash script?

william-msft:
> Change looks good to me -- does `main` already include the change to remove the FPM array from the corresponding Bash script?

No, we closed that PR: https://github.com/microsoft/Oryx/pull/2152

build/tools/SharedCodeGenerator/Outputs/ShellOutput.cs
waliMSFT: @william-msft can you please add a little bit more hint on how it will gonna affect the changes in the mentioned file?
other than that looks good.

PR #2159, opened 2023-09-20
Wali/locking nginx version

waliMSFT:

The related work item: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1844164

To avoid customer issues, we decided to lock the NGINX version for PHP. As only php-fpm uses NGINX so the changes are reflected only on the php-fpm dockerfiles. 
For `buster` the stable NGINX version is 1.22.0.
For `bullseye` the stable NGINX version is 1.24.0.


images/runtime/php-fpm/template.base.Dockerfile
```diff
@@ -12,9 +16,22 @@ ENV NGINX_DOCUMENT_ROOT /home/site/wwwroot
...
+RUN echo "deb http://nginx.org/packages/debian `lsb_release -cs` nginx" | tee /etc/apt/sources.list.d/nginx.list
+RUN apt update
+RUN set -eux; \
+    if [ "${DEBIAN_FLAVOR}" = "buster" ]; then \
+        apt install -y nginx=${NGINX_BUSTER_VERSION}; \
+    elif [ "${DEBIAN_FLAVOR}" = "bullseye" ]; then \ 
+        apt install -y nginx=${NGINX_BULLSEYE_VERSION}; \
+    else \
+        apt install -y nginx; \
+    fi
```
cormacpayne, line 33: nit: I know the Dockerfile was structured like this previously, but it's generally best practice in Dockefiles to chain multiple commands into a single `RUN` command rather than having them executed one-by-one as individual `RUN` commands. This will minimize the number of layers that need to be cached and the size of the image as a result.

cormacpayne (dismissed):
One minor comment, otherwise LGTM :shipit:

images/runtime/php-fpm/template.base.Dockerfile
waliMSFT: thanks @cormacpayne for sharing the knowledge. I will update the code accordingly.

waliMSFT: I have addressed the changes. Please have a look whenever..

pauld-msft (changes requested):
Mentioned this offline, but how do we know that this is not going to have the same problems that we saw with [my old pull request](https://github.com/microsoft/Oryx/pull/2016) (which needed to be reverted) that broke a bunch of app service folks? I think we should add a test to cover the exact problem that we saw then. We need to ensure that this issue is not duplicated, since it went into Sev 2 territory for App Service and we discovered the root cause so we need to have testing for it before making further changes.

PR #2164, opened 2023-09-22
Adding prereleases for dotnet 8 in automator

harryli0108:
This PR allows Oryx DotNet Automator to ingest prereleases specifically for DotNet 8. 
It also updates some nit things I found in the automator as well.


build/tools/Automation/DotNet/DotNetAutomator.cs
```diff
@@ -69,6 +63,17 @@ public async Task RunAsync()
             }
         }
 
+        public async Task InitializeFieldsAsync()
```
cormacpayne, line 66: nit: let's make this a `private` method since it's only accessed within the scope of this class

build/tools/Automation/DotNet/Models/DotNetVersion.cs
```diff
@@ -35,8 +36,8 @@ public int CompareTo(object obj)
                 return 1;
             }
 
-            var thisVersion = new Version(this.Version);
-            var objVersion = new Version(objDotNetVersion.Version);
+            var thisVersion = SemanticVersion.Parse(this.Version);
+            var objVersion = SemanticVersion.Parse(objDotNetVersion.Version);
```
cormacpayne, line 40: Is there any validation with `SemanticVersion.Parse` that may bite us here that we wouldn't have seen previously with the `Version` constructor? For example, the `Version` constructor may just take the version string provided to it and attempt to piece together components of it to the properties of the `Version` class, whereas `SemanticVersion.Parse` will try to actually verify that each component aligns with a given property of the `SemanticVersion` class.

Along a similar path, below in the `VersionService` class, we're making a call to `SemanticVersion.TryParse`, which will swallow any exceptions possibly thrown when parsing a version string, so I'm wondering the order in which we'd be calling `TryParse` vs. `Parse` here.

build/tools/Automation/Services/VersionService.cs
```diff
@@ -21,9 +21,9 @@ public class VersionService : IVersionService
         /// <returns>True if the version is within the specified range and not in the list of blocked versions. False otherwise.</returns>
         public bool IsVersionWithinRange(string version, string minVersion = null, string maxVersion = null, List<string> blockedVersions = null)
         {
-            // Try to parse the version string into a SemanticVersion object and ignores pre-releases
+            // Try to parse the version string into a SemanticVersion object and ignores pre-releases that are not dotnet 8
             if (!SemanticVersion.TryParse(version, out var semanticVersion) ||
-                semanticVersion.IsPrerelease)
+                (semanticVersion.IsPrerelease && (semanticVersion.Major != 8)))
```
cormacpayne, line 26: nit: we should file a work item to remove this prerelease check for .NET 8 once it has become stable and link it in a comment here so we remember to update this automation logic just in case they were to release any additional prerelease versions in the future

cormacpayne (approved):
Some minor comments, otherwise LGTM

build/tools/Automation/DotNet/Models/DotNetVersion.cs
harryli0108: I think this should be ok here. `Version` will throw exceptions when it encounters releases, and `SemanticVersion.Parse` will do the same thing here. I think it should be safe since `SemVer.Parse` takes what `Version` could take and takes prereleases in addition. And if there's some issue, the behavior will be similar to what `Version` would be doing.

build/tools/Automation/DotNet/DotNetAutomator.cs
harryli0108: Just want to double check here. I made this a public method so the private fields could be initialized in test. Making this private will make the newly added unit test fail since mock can't access private method. Is it ok to remove the added unit test once I make this method private or should we keep the public method and the unit test here?

cormacpayne: OK in that case I think we should scrap this new method and move the instantiation of these variables to be in the constructor of this class -- that way they're still accessible in the tests since we're calling the constructor directly

build/tools/Automation/DotNet/Models/DotNetVersion.cs
cormacpayne: OK sounds good -- it may be worth adding some validation in the future around these versions, but I think we should be fine in any case since we're pulling the versions from .NET (and other public sources) and not some random strings.

build/tools/Automation/Services/VersionService.cs
cormacpayne: Would you mind adding an in-line comment for this conditional that links to the work item?

build/tools/Automation/DotNet/DotNetAutomator.cs
harryli0108: I think if we put this in the constructor, it'll bypass any mock http requests, which then causes the unit test to fail as well. Is there any other ways to get around this?

cormacpayne: Discussed offline -- we should be good with the current implementation based on how the class and its methods are being tested.

cormacpayne (approved):

william-msft:
I tested locally and looks good! 
image

william-msft (approved):

PR #2165, opened 2023-09-25
Snehapar/fix nginx version

snehapar9:

Using `mcr.microsoft.com/mirror/docker/library/nginx:1.21` as the build/final image breaks during docker build in [this step](https://github.com/Azure/container-apps-deploy-action/actions/runs/6302982762/job/17114097071#step:5:1113).
Updating the build/final image to nginx:1.24.0 fixes it.

Verification - https://github.com/Azure/container-apps-deploy-action/actions/runs/6304596653/job/17116265754

image

### DIFF
--- a/tests/SampleApps/DotNetCore/Blazor_Function_Sample/blazor-sample-app/Dockerfile
+++ b/tests/SampleApps/DotNetCore/Blazor_Function_Sample/blazor-sample-app/Dockerfile
@@ -5,7 +5,11 @@ COPY . ./
 WORKDIR /app/
 RUN dotnet publish -c Release -o published_app
 
+<<<<<<< Updated upstream
 FROM nginx:1.24.0 AS build
+=======
+FROM mcr.microsoft.com/mirror/docker/library/nginx:1.24 AS build
+>>>>>>> Stashed changes
 WORKDIR /src
 RUN apt-get update && apt-get install -y git wget build-essential libssl-dev libpcre3-dev zlib1g-dev libbrotli-dev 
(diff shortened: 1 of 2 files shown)


daniv-msft (dismissed):

tests/SampleApps/DotNetCore/Blazor_Function_Sample/blazor-sample-app/Dockerfile
```diff
@@ -5,23 +5,23 @@ COPY . ./
 WORKDIR /app/
 RUN dotnet publish -c Release -o published_app
 
-FROM mcr.microsoft.com/mirror/docker/library/nginx:1.21 AS build
+FROM nginx:1.24.0 AS build
```
cormacpayne, line 8: Any reason we can't bump the version in the tag of the MCR repository to `1.24`? It looks like it's available in the [list of MCR tags](https://[mcr.microsoft.com/v2/mirror/docker/library/nginx/tags/list).

If we move away from MCR, I think (depending on if Sample Apps are scanned) the security portion of the pipeline will fail because we're no longer allowed to use non-MCR/ACR bases within a Dockerfile without an explicit opt-out (which we have in some of our own Dockerfiles, but using the MCR version of this library would be optimal)

snehapar9: Works with `mcr.microsoft.com/mirror/docker/library/nginx:1.24`. Thanks

cormacpayne (approved):
:shipit:

william-msft (approved):

waliMSFT (approved):

PR #2166, opened 2023-09-25
Snehapar/remove shell tracing output

snehapar9:

https://github.com/Azure/static-web-apps/issues/1183

`set -ex` enables shell tracing and the shell tracing output is written to stderr by design. So, there is no straightforward way to just change the color of the shell tracing output.

We can remove it if it's not valuable information for customers and just exit on error instead be running `set -e` or let the customer know why they are seeing it and evaluate if we should remove it or not based on their response.

Here is how it looks after the change (without the statements beginning with ++)

image

Before change : 

image

### DIFF
--- a/images/build/installHugo.sh
+++ b/images/build/installHugo.sh
@@ -4,7 +4,7 @@
 # Licensed under the MIT license.
 # --------------------------------------------------------------------------------------------
 
-set -ex
+set -e
 
 __CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 source "$__CURRENT_DIR/../../build/__hugoConstants.sh"


william-msft:
The tracing might be helpful to keep in our pipelines and also if the user does a `--debug` in the during `oryx build`
We could adjust like so:
```
# Check if the debug flag is set to true
if [ "$DEBUG" = true ]; then
    set -ex  # Enable exit on error and print each command before executing
else
    set -e   # Enable exit on error only
fi
```
We'd also have to add the flag in our `/vsts` pipeline folder.

snehapar9:
> The tracing might be helpful to keep in our pipelines and also if the user does a `--debug` in the during `oryx build` We could adjust like so:
> 
> ```
> # Check if the debug flag is set to true
> if [ "$DEBUG" = true ]; then
>     set -ex  # Enable exit on error and print each command before executing
> else
>     set -e   # Enable exit on error only
> fi
> ```
> 
> We'd also have to add the flag in our `/vsts` pipeline folder.

Okay so enable shell tracing in debug mode or the pipeline? Is there a way to determine the build is executing in the pipeline? This is just for Hugo? So do we want to see shell tracing for running Hugo build tests?

william-msft:
> > The tracing might be helpful to keep in our pipelines and also if the user does a `--debug` in the during `oryx build` We could adjust like so:
> > ```
> > # Check if the debug flag is set to true
> > if [ "$DEBUG" = true ]; then
> >     set -ex  # Enable exit on error and print each command before executing
> > else
> >     set -e   # Enable exit on error only
> > fi
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We'd also have to add the flag in our `/vsts` pipeline folder.
> 
> Okay so enable shell tracing in debug mode or the pipeline? Is there a way to determine the build is executing in the pipeline? This is just for Hugo. So do we want to see shell tracing while running Hugo build tests in the pipeline?

To determine if the build is running in the pipeline, we can set a flag (either an arg or env var) in our pipeline `/vsts`, enabling `set -ex`. I've found the trace useful when I debug, so I think keeping it around in the pipelines would be beneficial. This is only if we need to switch over to `set -e`.

cormacpayne:
> > > The tracing might be helpful to keep in our pipelines and also if the user does a `--debug` in the during `oryx build` We could adjust like so:
> > > ```
> > > # Check if the debug flag is set to true
> > > if [ "$DEBUG" = true ]; then
> > >     set -ex  # Enable exit on error and print each command before executing
> > > else
> > >     set -e   # Enable exit on error only
> > > fi
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > We'd also have to add the flag in our `/vsts` pipeline folder.
> > 
> > 
> > Okay so enable shell tracing in debug mode or the pipeline? Is there a way to determine the build is executing in the pipeline? This is just for Hugo. So do we want to see shell tracing while running Hugo build tests in the pipeline?
> 
> To determine if the build is running in the pipeline, we can set a flag (either an arg or env var) in our pipeline `/vsts`, enabling `set -ex`. I've found the trace useful when I debug, so I think keeping it around in the pipelines would be beneficial. This is only if we need to switch over to `set -e`.

@william-msft -- I don't know if simply having this `$DEBUG` conditional will work since this `installHugo.sh` script will be run within the context of a Dockerfile, _not_ our VSTS pipelines (or anywhere that would have access to VSTS environment variables). It seems like we would need to expose this `$DEBUG` environment variable to every Dockerfile calling the script, which doesn't seem feasible to me. Let me know if I'm looking at this incorrectly.

william-msft:
> > > > The tracing might be helpful to keep in our pipelines and also if the user does a `--debug` in the during `oryx build` We could adjust like so:
> > > > ```
> > > > # Check if the debug flag is set to true
> > > > if [ "$DEBUG" = true ]; then
> > > >     set -ex  # Enable exit on error and print each command before executing
> > > > else
> > > >     set -e   # Enable exit on error only
> > > > fi
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > We'd also have to add the flag in our `/vsts` pipeline folder.
> > > 
> > > 
> > > Okay so enable shell tracing in debug mode or the pipeline? Is there a way to determine the build is executing in the pipeline? This is just for Hugo. So do we want to see shell tracing while running Hugo build tests in the pipeline?
> > 
> > 
> > To determine if the build is running in the pipeline, we can set a flag (either an arg or env var) in our pipeline `/vsts`, enabling `set -ex`. I've found the trace useful when I debug, so I think keeping it around in the pipelines would be beneficial. This is only if we need to switch over to `set -e`.
> 
> @william-msft -- I don't know if simply having this `$DEBUG` conditional will work since this `installHugo.sh` script will be run within the context of a Dockerfile, _not_ our VSTS pipelines (or anywhere that would have access to VSTS environment variables). It seems like we would need to expose this `$DEBUG` environment variable to every Dockerfile calling the script, which doesn't seem feasible to me. Let me know if I'm looking at this incorrectly.

Yeah I also confirmed we'd have to add it to every Dockerfile. I'm fine with removing the `-x` parameter. We can always add it back if we need to.

william-msft (approved):

PR #2174, opened 2023-09-30
Harrli/dotnet8automator

harryli0108:
This PR contains changes in DotNet 8 automator's github action changes where it uses PAT instead of Github Token; added bookworm's versionToBuild to auto populate; includes new DotNet 8 RC package.


build/tools/Automation/Services/VersionService.cs
```diff
@@ -21,9 +21,10 @@ public class VersionService : IVersionService
         /// <returns>True if the version is within the specified range and not in the list of blocked versions. False otherwise.</returns>
         public bool IsVersionWithinRange(string version, string minVersion = null, string maxVersion = null, List<string> blockedVersions = null)
         {
-            // Try to parse the version string into a SemanticVersion object and ignores pre-releases
+            // Try to parse the version string into a SemanticVersion object and ignores pre-releases that are not dotnet 8
+            // Block pre release once stabled. Workitem: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1891562
```
daniv-msft, line 25: Could you please share more info about what prevents us from having a logic that would work for both the pre-releases and the GA'd versions?

.github/workflows/automationTemplate.yaml
```diff
@@ -75,7 +79,7 @@ jobs:
         with:
           branch: automation/add-new-${{ inputs.platformName }}-versions-${{ env.RELEASE_DATE }}
           title: New ${{ env.PLATFORM_NAME_CAMEL_CASE }} Sdks ${{ env.RELEASE_DATE }}
-          token: ${{ secrets.GH_TOKEN }}
+          token: ${{ secrets.PAT_TEMP }}
```
daniv-msft, line 82: Could you please confirm we have a work item to not forget to change that soon? Also, my understanding is that we need this today so that we unblock the pipelines. Is this correct?

build/tools/Automation/Services/VersionService.cs
harryli0108: For this one, this is more to just include prereleases for dotnet 8. But this check doesn't prevent us from getting GA'd versions. And because we need to make sure things are compatible when new major version releases, it would be better to limit the prereleases to just the versions we want to ingest rather than having it taking all prereleases imo.

.github/workflows/automationTemplate.yaml
harryli0108: Yes, this unblocks us from the issue. I have two workitems created and assigned to me for now: `1894797` (Allowing GH token to create PR) and `1894896` (Refresh PAT to make sure the token is still valid). The token refresh should be just performing logging in from my side every 30 days according to Microsoft security policy in their org account.

build/tools/Automation/Services/VersionService.cs
daniv-msft: OK sorry I want to be sure I understand: what prereleases do we expect outside of the .NET 8 ones? Do they provided other packages labeled as prereleases that we don't care about?

.github/workflows/automationTemplate.yaml
daniv-msft: OK thanks!

build/__dotNetCoreRunTimeVersions.sh
```diff
@@ -30,7 +30,7 @@ NET_CORE_APP_70='7.0.9'
 NET_CORE_APP_70_SHA='09552e5ae6ac014dadf17545ff0a30ab32921075a31fb33e7be148c13078e30d097f592ffa1b8d306563aaa3f6302e40c5c0ba815c1473bbd5d72e3bef55d91e'
 ASPNET_CORE_APP_70='7.0.9'
 ASPNET_CORE_APP_70_SHA='aabf4fa5ca726dc52774e5d644800ef7477815b22a982b7a2752dec6569186aabca93d5386e195e7ead377144601a786ae6a5d76ff28435bdabfad495cfe554b'
-NET_CORE_APP_80='8.0.0-preview.7.23375.6'
-NET_CORE_APP_80_SHA='bfd8491550178b86a7a72fe06bdc82f0dd66771d5b60d7e4e1133cdde29f84bd57857d846722e027bd209db087123b2d12b2e23590d77991052269fa265814e1'
-ASPNET_CORE_APP_80='8.0.0-preview.7.23375.9'
-ASPNET_CORE_APP_80_SHA='b8c8a5cd579a8ee1e082363a73a05a745499365445f784e6ff87547f9acdbe8b7ba525140ef10555bc2802c13af131c2d568ec6af020bd0dd2fdf82d4c258442'
+NET_CORE_APP_80='8.0.0-rc.1.23419.4'
```
daniv-msft, line 33: Nit: looks like we need to rebase the PR to get the latest changes you pushed in parallel.

build/tools/Automation/Services/VersionService.cs
harryli0108: Oh i see. There's no other prereleases outside of .NET 8 for now. But say if .NET 9 is out, this would prevent it accidentally takes in the prereleases of the hypothetical .NET 9. Just a preventive measure here.

build/__dotNetCoreRunTimeVersions.sh
harryli0108: i think this one should be ok. The .NET changes were merged into this branch, and I merge all the changes (automator change and .NET 8 RC change) into main.

.github/workflows/automationTemplate.yaml
cormacpayne, line 82: nit: would you mind renaming this secret so that it's a bit more descriptive with regards to what the personal access token is for? `PAT_TEMP` let's me know it's a temporary PAT, but outside of that, I wouldn't be sure what the PAT has access to or is used for without a bit more context, so renaming this secret would be great

.github/workflows/automationTemplate.yaml
```diff
@@ -29,6 +29,10 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
+    permissions: 
+      actions: write
+      contents: write
+      pull-requests: write
```
cormacpayne, line 35: nit: just curious as to the impact defining this `permissions` property has on the workflow

cormacpayne (approved):
Two minor comments, otherwise LGTM

.github/workflows/automationTemplate.yaml
harryli0108: By explicitly calling the permissions, it'll override the organizational setting on the github action permissions just for this specific workflow. By default, microsoft prevents the usage of actions outside of the org and prevents actions open PR. So just to be cautious, i put the permissions needed for this workflow.

actions permission allows this workflow to cancel or trigger another workflow
contents allow this workflow to commit things
pull-requests allow it to create/modify PRs

build/__dotNetCoreRunTimeVersions.sh
daniv-msft: Harry and I synced this morning. I indeed missed that the previous PR wasn't targeting main.

build/tools/Automation/Services/VersionService.cs
daniv-msft: Thanks! At the same time having .NET 9 pre-releases would sound good pretty good to me as well. :D Let's sync separately

daniv-msft (approved):

doc/supportedPlatformVersions.md
```diff
@@ -65,6 +66,7 @@
 - 7.0.304
 - 7.0.305
 - 7.0.306
+- 8.0.100-rc.1.23463.5
```
waliMSFT, line 69: do we release dotnet 8 other than `bookworm` flavor? if it is then why we are including the dotnet 8 version in all the flavors (in this file only)?

PR #2183, opened 2023-10-05
Wali/dotnet test release with public binaires to private mcr

waliMSFT:

-- This PR is created for the below testing purpose.

We need to test if we can release the `baseImages and runTimeImages` to the staging repo (staging/oryx/base) of MCR. For the testing purpose I took the public binaries so that if anything went wrong in the release won't create any mess.


images/runtime/dotnetcore/8.0/base.bookworm.staging.Dockerfile
```diff
@@ -0,0 +1,67 @@
...
+ARG BUILD_DIR=/tmp/oryx/build
+ADD build ${BUILD_DIR}
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        # .NET Core dependencies
+        libc6 \
+        libgcc1 \
+        libgssapi-krb5-2 \
+        libicu63 \
```
daniv-msft, line 24: Could you please confirm that you copied the Dockerfiles for each version from their corresponding prod version, and not just duplicating the .NET 6 one?
For example in .NET 8 libicu should be targeting `libicu72` (cf. https://github.com/microsoft/Oryx/blob/cde2209830994c33600eb9e6658d2eba3c480710/images/runtime/dotnetcore/8.0/base.bookworm.Dockerfile)

There might be other differences so we need to rely on the existing .NET 7/8/6 Dockerfiles for these staging ones.

waliMSFT: Thanks Dan for pointing this out.  I have copied dockerfiles from dotnet 7 and reused it. As, I am using these dockerfiles just for testing purposes, I did not pay attention to the details here. Was thinking to create the actual Dockerfiles for release purposes from their corresponding version's base.Dockerfile in the next PR (PR-2 to release the valid images created from private binaires).
Should I update these dockerfiles in these PR?

daniv-msft: OK thanks for clarifying. I believe it should be fine as we're mostly checking the presence of the images generated and not their content? Let's not lose time with re-triggering a new iteration in that case.

daniv-msft (dismissed):

daniv-msft (dismissed):

images/runtime/dotnetcore/generateDockerfiles.sh
```diff
@@ -54,9 +54,7 @@ do
 	sed -i "s|$RUNTIME_BASE_IMAGE_TAG_PLACEHOLDER|$RUNTIME_BASE_IMAGE_TAG|g" "$TARGET_DOCKERFILE"
 	sed -i "s|$DOTNET_VERSION_PLACEHOLDER|$VERSION_DIRECTORY|g" "$TARGET_DOCKERFILE"
 
-	# Added the [ "$ImageDebianFlavor" != "bullseye" ] condition to avoid creating bullseye images from `staging/oryx` repo.
-	# This is a temporary change only to release the dotnet 6,7,8 privately.
-	if [ shouldStageRuntimeVersion "dotnetcore" $VERSION_DIRECTORY ] && [ "$ImageDebianFlavor" != "bullseye" ] ; then
+	if shouldStageRuntimeVersion "dotnetcore" $VERSION_DIRECTORY ; then
```
waliMSFT, line 57: please ignore this change. In previous commit I tried to fix one issue. In this commit I just reverted it back..

daniv-msft (approved):

images/runtime/dotnetcore/8.0/base.bookworm.staging.Dockerfile
cormacpayne: We need to double-check the dependencies outlined here because some of them are specific to certain versions of Debian.

For example, `libicu63` is for buster, `libicu67` is for bullseye and `libicu72` is for bookworm.

Similarly, both buster and bullseye used `libssl1.1`, but bookworm uses `libssl3`.

I _think_ these are the only cases, but we'll need to check out Dockerfiles to be certain.

PR #2186, opened 2023-10-09
Split runtime jobs based on Debian flavor

cormacpayne:

### DIFF
--- a/tests/Oryx.RuntimeImage.Tests/DotNetCore/DotnetCoreImageVersionsTest.cs
+++ b/tests/Oryx.RuntimeImage.Tests/DotNetCore/DotnetCoreImageVersionsTest.cs
@@ -20,6 +20,7 @@ namespace Microsoft.Oryx.RuntimeImage.Tests
         }
 
         [SkippableTheory]
+        [Trait("category", "runtime-bullseye")]
         [InlineData("3.1")]
         public void DotNetCoreBullseyeRuntimeImage_Contains_VersionAndCommit_Information(string version)
         {
@@ -57,6 +58,7 @@ namespace Microsoft.Oryx.RuntimeImage.Tests
         }
 
         [SkippableTheory]
+        [Trait("category", "runtime-buster")]
         [InlineData("3.0")]
         [InlineData("5.0")]
         [InlineData("6.0")]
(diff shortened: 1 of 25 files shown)


tests/Oryx.RuntimeImage.Tests/DotNetCore/DotnetCoreImageVersionsTest.cs
```diff
@@ -97,21 +99,46 @@ public void DotNetCoreBusterRuntimeImage_Contains_VersionAndCommit_Information(s
...
             var result = _dockerCli.Run(new DockerRunArguments
             {
-                ImageId = _imageHelper.GetRuntimeImage("dotnetcore", version, osType),
+                ImageId = _imageHelper.GetRuntimeImage("dotnetcore", version, ImageTestHelperConstants.OsTypeDebianBuster),
+                CommandToExecuteOnRun = "dotnet",
+                CommandArguments = new[] { "--info" }
+            });
+
+            // Assert
+            var actualOutput = string.Join("", result.StdOut.ReplaceNewLine().Where(c => !char.IsWhiteSpace(c)));
```
daniv-msft, line 120: I know we want this available quickly so I'm not suggesting that we fix all this tonight, but would we see any alternatives in the future to us duplicating a lot of this code? Would helper functions be a path to reduce code duplication?

daniv-msft (dismissed):

tests/Oryx.RuntimeImage.Tests/DotNetCore/DotnetCoreImageVersionsTest.cs
cormacpayne: Yep we can definitely refactor this once things have settled -- we have examples of doing this across other places from previous refactors, so it's just a matter of finding some time to do it.

cormacpayne: In these cases, the only thing that's different per test is the Debian flavor that we're using, so we could just turn the guts of these tests into a common private method that is then called from the tests themselves, and the tests will provide the Debian flavor they want to us (along with the other arguments they're already providing, like platform version).

daniv-msft (approved):

waliMSFT:
need to merge this PR asap. `No disk space` is blocking the pipelines now.

PR #2194, opened 2023-10-12
Ensure all runtime images are pushed to production during release phase

cormacpayne:
Currently, during the release phase of our CI pipeline that pushes the runtime images to production, only one set of Debian flavor images will be pushed (whichever job finished last) -- this is due to the fact that all three Debian flavor runtime jobs are writing the list of runtime images built and pushed within the scope of the job to a `runtime-images-acr.txt` file that is dropped as an artifact of the job. This artifact is seemingly overridden when it shares its name with the same artifact dropped by a separate job later, so if the bookworm job finishes first and drops an artifact named `runtime-images-acr.txt`, it will be overridden when the buster job finishes next and drops an artifact with the same name.

To resolve this, each job is already dropping an artifact that includes the Debian flavor in the name (_e.g._, `runtime-images-acr.bookworm.txt`), so before we read the list of runtime images built and pushed in the `runtime-images-acr.txt` file, we will consolidate all of the images in each Debian flavor `.txt` file to this `runtime-images-acr.txt` file.

### DIFF
--- a/vsts/pipelines/templates/_buildTemplate.yml
+++ b/vsts/pipelines/templates/_buildTemplate.yml
@@ -198,12 +198,6 @@ steps:
       BuildDropPath: '$(Build.ArtifactStagingDirectory)'
       AdditionalComponentDetectorArgs: '--DirectoryExclusionList **/SampleApps/**'
 
-- script: |
-    echo "Writing Debian-specific runtime image tags to known file"
-    cat $(Build.ArtifactStagingDirectory)/images/runtime-images-acr.${{ parameters.imageType }}.txt >> '$(Build.ArtifactStagingDirectory)/images/runtime-images-acr.txt'
-  displayName: 'Writing Debian-specific runtime image tags to known file'
-  condition: and(succeeded(), eq(variables['BuildRuntimeImages'], 'true'))
-
 - task: PublishTestResults@2
   inputs:
     testRunner: 'xUnit'
--- a/vsts/pipelines/templates/_buildTemplate.yml
+++ b/vsts/pipelines/templates/_buildTemplate.yml
@@ -148,7 +148,7 @@ steps:
   inputs:
     type: FilePath
     scriptPath: ./vsts/scripts/pullAndTag.sh
-    args: $(System.ArtifactsDirectory)/drop/images/build-images-acr.txt $(System.ArtifactsDirectory)/drop/images/runtime-images-acr.txt
+    args: $(System.ArtifactsDirectory)/drop/images/build-images-acr.txt
   condition: and(succeeded(), eq(variables['TestIntegration'], 'true'))
 
 - task: ms-devlabs.utilitytasks.task-Shellpp.Shell++@0
@@ -224,7 +224,7 @@ steps:
     azureContainerRegistry: ${{ parameters.acrName }}
     command: 'Push an image'
     pushMultipleImages: true
-    imageNamesPath: '$(Build.ArtifactStagingDirectory)/images/runtime-images-acr.txt'
+    imageNamesPath: '$(Build.ArtifactStagingDirectory)/images/runtime-images-acr.${{ parameters.imageType }}.txt'
     includeLatestTag: false
     enforceDockerNamingConvention: false
   condition: and(succeeded(), eq(variables['PushRuntimeImages'], 'true'), eq(variables['BuildRuntimeImages'], 'true'))
--- a/vsts/scripts/pullAndTag.sh
+++ b/vsts/scripts/pullAndTag.sh
@@ -127,10 +127,10 @@ if [ -n "$TESTINTEGRATIONCASEFILTER" ];then
 		# Note: we don't write this file to the drop folder as we don't want this file written to for every integration test job
 		sourceFile="$BUILD_SOURCESDIRECTORY/temp/images/runtime-images-acr.txt"
 
-		# Consolidate the different Debian runtime image files into one to be read from
-		(cat "$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-acr.buster.txt"; echo) >> '$sourceFile'
-		(cat "$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-acr.bullseye.txt"; echo) >> '$sourceFile'
-		(cat "$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-acr.bookworm.txt"; echo) >> '$sourceFile'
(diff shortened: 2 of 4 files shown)


vsts/pipelines/templates/_buildTemplate.yml
```diff
@@ -198,12 +198,6 @@ steps:
       BuildDropPath: '$(Build.ArtifactStagingDirectory)'
       AdditionalComponentDetectorArgs: '--DirectoryExclusionList **/SampleApps/**'
 
-- script: |
-    echo "Writing Debian-specific runtime image tags to known file"
```
daniv-msft, line 202: Just mentioning this for other people reviewing and being confused on why this part is being removed: this logic existed in the build case but @cormacpayne determined it wasn't used there. Removing it makes sense so that we don't keep things around that are not used.

vsts/scripts/tagRunTimeImagesForRelease.sh
```diff
@@ -11,6 +11,14 @@ acrPmeProdRepo="oryxprodmcr"
 sourceBranchName=$BUILD_SOURCEBRANCHNAME
 outFilePmeMCR="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/$acrPmeProdRepo-runtime-images-mcr.txt"
 sourceFile="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-acr.txt"
+debianBusterSourceFile="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-acr.buster.txt"
+debianBullseyeSourceFile="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-acr.bullseye.txt"
+debianBookwormSourceFile="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-acr.bookworm.txt"
+
```
daniv-msft, line 17: At this point is the logic that creates `runtime-images-acr.txt` in the runtime case still useful, if it also produces the Debian-specific versions as well? Could we remove it so that we don't keep a file around that we know is incomplete?

cormacpayne: Resolved offline -- the `runtime-images-acr.txt` file is no longer created in each job but rather only when we're consolidating all of the Debian-specific `.txt` files, so no lingering files will exist here.

daniv-msft (dismissed):

daniv-msft (dismissed):

vsts/scripts/pullAndTag.sh
```diff
@@ -124,8 +123,22 @@ if [ -n "$TESTINTEGRATIONCASEFILTER" ];then
 		# Always convert filter for runtime images to lower case
 		echo "Runtime image filter is set for $platformFilter with version $platformVersionFilter"
 
+		# Create a local file that consolidates the different runtime image files into one to be read from
+		# Note: we don't write this file to the drop folder as we don't want this file written to for every integration test job
+		sourceFile="$BUILD_SOURCESDIRECTORY/temp/images/runtime-images-acr.txt"
```
waliMSFT, line 128: is there any other way to elaborate why writing all the files in a single file?

vsts/scripts/pullAndTag.sh
```diff
@@ -124,8 +123,25 @@ if [ -n "$TESTINTEGRATIONCASEFILTER" ];then
...
+
+		if [[ ! -f "$sourceFile" ]]; then
+			echo "Creating consolidated runtime image file '$sourceFile'..."
+			touch "$sourceFile"
+		fi
+
```
william-msft, line 134: We'll have to do something similar to:
```
		if [[ ! -f "$sourceFile" ]]; then
			echo "Creating consolidated runtime image file '$sourceFile'..."
                        mkdir -p $BUILD_SOURCESDIRECTORY/temp/images/
			touch "$sourceFile"
		fi
```

To fix:
image

william-msft: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=8550982&view=logs&j=913ce52b-8ad3-5a93-c0fe-123235d07c67&t=674b07f0-e3ef-53f4-9cc2-286a791ee0ed

vsts/scripts/tagRunTimeImagesForRelease.sh
```diff
@@ -12,10 +12,23 @@ sourceBranchName=$BUILD_SOURCEBRANCHNAME
 outFilePmeMCR="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/$acrPmeProdRepo-runtime-images-mcr.txt"
 sourceFile="$BUILD_ARTIFACTSTAGINGDIRECTORY/drop/images/runtime-images-acr.txt"
 
+if [[ ! -f "$sourceFile" ]]; then
+			echo "Creating consolidated runtime image file '$sourceFile'..."
+			touch "$sourceFile"
```
william-msft, line 17: Might have to do `mkdir -p` the directory repo here as well.
https://github.com/microsoft/Oryx/pull/2194/files#r1362827563

william-msft: Might have to do `mkdir -p` the directory repo here as well, before the `touch`.
https://github.com/microsoft/Oryx/pull/2194/files#r1362827563

vsts/scripts/pullAndTag.sh
cormacpayne: Thanks for the catch -- for some reason, I just assumed that directory existed, but it makes sense why it doesn't since it's within the scope of the repository code that we clone down.

vsts/scripts/pullAndTag.sh
cormacpayne: Each job of a pipeline is running in its own independent container, and we moved from having all of the runtime images being built within the scope of one job to having a job for each Debian flavor that we build a set of runtime images for. Within these jobs, we're building the runtime images associated with the given Debian flavor, as well as pushing those images to our dev ACR instance. When those jobs are done, and we move on to future steps of the pipeline, such as integration tests or pushing the images to our production ACR instance, we need to have a record of what images we built within the scope of this pipeline. These individual Debian flavor jobs will each write their own `.txt` file that is an inventory of all runtime images built and pushed to our dev ACR instance so that when we reach future jobs, we can consolidate all of those `.txt` files into one to have one complete list of all runtime images created during this pipeline that we can either test or bulk push to our production ACR instance.

william-msft (dismissed):

william-msft (dismissed):

daniv-msft (approved):

PR #2196, opened 2023-10-17
expand jar file to workspace

yilims:

### DIFF
--- a/builders/container-apps-internal-registry-demo/startup-script.sh
+++ b/builders/container-apps-internal-registry-demo/startup-script.sh
@@ -32,7 +32,7 @@ tar -xzf "$temp_app_source_path"
 fileCount=$(ls | wc -l)
 if [ "$fileCount" = "1" ]; then
   # Find .jar file in the directory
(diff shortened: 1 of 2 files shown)


builders/container-apps-internal-registry-demo/startup-script.sh
```diff
@@ -29,6 +29,20 @@ mkdir -p $CNB_APP_DIR
 cd $CNB_APP_DIR
 tar -xzf "$temp_app_source_path"
 
+fileCount=$(ls | wc -l)
+if [ "$fileCount" = "1" ]; then
+  # Find .jar file in the directory
+  jarfile=$(find "$CNB_APP_DIR" -maxdepth 1 -name "*.jar" | head -n 1)
```
pauld-msft, line 35: do we also need to support `war` and `zip` file extension as well?

pauld-msft (approved):
LGTM if we don't need to support any file extensions besides jar

builders/container-apps-internal-registry-demo/startup-script.sh
yilims: war will be supported soon, zip is not support.

PR #2197, opened 2023-10-17
Container App Build Image

pauld-msft:
Adds the image that is based off of the buildpack builder image and will download source code from blob storage and build an image from it


builders/container-apps-internal-registry-demo/app.tar.gz
william-msft, line 1: Do we need this binary?

builders/container-apps-internal-registry-demo/startup-script.sh
```diff
@@ -0,0 +1,100 @@
...
+  then
+    echo "Unzip file $jarfile"
+    unzip -qq $jarfile -d $CNB_APP_DIR
+    rm $jarfile
+  fi
+fi
+
+# public cert should be in this env var
+ca_pem_decoded=$(printf "%s" "$REGISTRY_HTTP_TLS_CERTIFICATE" | base64 -d)
+echo "$ca_pem_decoded" >> /usr/local/share/ca-certificates/internalregistry.crt
```
william-msft, line 45: Do we plan on having multiple certs inside `internalregistry.crt`? If not, we probably replace the cert instead of append. Something like:
`echo "$ca_pem_decoded" > /usr/local/share/ca-certificates/internalregistry.crt`

builders/container-apps-internal-registry-demo/app.tar.gz
pauld-msft: removed, thanks!

william-msft (approved):

PR #2201, opened 2023-10-19
Add support for node.js version 18.18.2

nunoalima:
This PR adds platform binary and runtime image support for Node 18.18.2, which includes CVE fixes as a follow up to Node.js [latest Security Releases](https://nodejs.org/en/blog/vulnerability/october-2023-security-releases).

This is a requirement for developers that are using Oryx to build both static applications and Azure Functions using Microsoft's `Azure/static-web-apps-deploy` GitHub actions step.

A few questions:
- Do I need to update the `NODE_RUNTIME_BASE_TAG`?
- Shall I include older patch versions for the same minor (`18.18.0` and `18.18.1`)?
- Are there any tests I should be running other than `testStartupScriptGenerators.sh`?


kumaraksh1:
Hey [nunoalima](https://github.com/nunoalima),
We are going to generate node images with latest patch version soon.

PR #2230, opened 2023-10-30
Update script for better telemetry

pauld-msft:
https://msazure.visualstudio.com/Antares/_sprints/taskboard/ContainerApps/Antares/Germanium/CY23Q4?workitem=25625685

Update the builder's lifecycle's headers format to be easier to parse by the CLI

### DIFF
--- a/builders/container-apps-internal-registry-demo/startup-script.sh
+++ b/builders/container-apps-internal-registry-demo/startup-script.sh
@@ -52,11 +52,11 @@ token=$(printf "%s" "$REGISTRY_AUTH_USERNAME:$REGISTRY_AUTH_PASSWORD" | base64)
 acr_access_string="Basic $token"
 export CNB_REGISTRY_AUTH='{"'$ACR_RESOURCE_NAME'":"'$acr_access_string'"}'
 
+echo "Initiating buildpack build with correlation id: '$CORRELATION_ID'"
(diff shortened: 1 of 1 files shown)


builders/container-apps-internal-registry-demo/startup-script.sh
```diff
@@ -52,48 +52,41 @@ token=$(printf "%s" "$REGISTRY_AUTH_USERNAME:$REGISTRY_AUTH_PASSWORD" | base64)
 acr_access_string="Basic $token"
 export CNB_REGISTRY_AUTH='{"'$ACR_RESOURCE_NAME'":"'$acr_access_string'"}'
 
+echo "Initiating buildpack build..."
+echo "Correlation id: '$CORRELATION_ID'"
+
```
cormacpayne, line 57: nit: it may be worth either putting an extra `echo` here or putting a `\n` at the end of the correlation ID `echo` so that there's separation between this message and the "Executing the analyze phase" message below

cormacpayne (dismissed):
One very minor comment, otherwise :shipit:

daniv-msft (dismissed):

pauld-msft:
closing in favor of https://github.com/microsoft/Oryx/pull/2240 which also contains these updates

PR #2236, opened 2023-11-01
Allow .NET SDK versions to be provided with `--platform-version` argument

cormacpayne:
Currently, only runtime versions can be provided to the `--platform-version` argument for `oryx build`, preventing users from specifying a specific version of the .NET SDK they'd like to build with (instead they'd need to know the corresponding runtime version shipped with the SDK version they want to use).

This PR cleans up the logic around the version that can be provided and detected to account for SDK versions, while still respecting the runtime version logic we currently have in place.


src/BuildScriptGenerator/DotNetCore/DotnetCorePlatform.cs
```diff
@@ -435,24 +459,5 @@ private string GetRuntimeVersionUsingHierarchicalRules(string detectedVersion)
             var defaultVersion = this.versionProvider.GetDefaultRuntimeVersion();
             return defaultVersion;
         }
-
-        private bool TryGetExplicitVersion(out string explicitVersion)
```
daniv-msft, line 439: Was this one just never used?

daniv-msft:
@Cormacpayne Overall looks good to me, but I would prefer not to push it before .NET 8 GA. Do you see any urgency in getting this one in?

william-msft (dismissed):

PR #2240, opened 2023-11-03
Pauld/retry buildpack phases

pauld-msft:
https://msazure.visualstudio.com/Antares/_sprints/taskboard/ContainerApps/Antares/Germanium/CY23Q4?workitem=25625680

retries the various buildpack phases if they fail


builders/container-apps-internal-registry-demo/startup-script.sh
```diff
@@ -52,49 +52,124 @@ token=$(printf "%s" "$REGISTRY_AUTH_USERNAME:$REGISTRY_AUTH_PASSWORD" | base64)
...
+echo "Correlation id: '$CORRELATION_ID'"
+echo 
+
+RETRY_DELAY=2
+RETRY_ATTEMPTS=5
+
+function fail_if_retry_exceeded() {
+  retries=$1
+  if [ "$retries" -ge $RETRY_ATTEMPTS ]; then
+    echo "===== Retry attempts exceeded ====="
```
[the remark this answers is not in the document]
pauld-msft: Do we not want this to show up for customers in the CLI? This seems to me like fairly valuable information?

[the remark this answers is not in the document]
pauld-msft: I added this, because previously the retry attempt logs were very hidden in the log stream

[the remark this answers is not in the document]
pauld-msft: I can maybe change the format to `----- * -----`?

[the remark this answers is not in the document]
pauld-msft: previous:
image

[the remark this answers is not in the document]
pauld-msft: > I can maybe change the format to ----- * -----?

implemented this

builders/container-apps-internal-registry-demo/startup-script.sh
```diff
@@ -52,49 +52,124 @@ token=$(printf "%s" "$REGISTRY_AUTH_USERNAME:$REGISTRY_AUTH_PASSWORD" | base64)
...
+    echo "----- Retrying analyze phase (attempt $retryCount) -----"
+  fi
+
+  /lifecycle/analyzer \
+    -log-level debug \
+    -run-image mcr.microsoft.com/oryx/builder:stack-run-debian-bullseye-20230926.1 \
+    $APP_IMAGE \
+    && break
+
+  retryCount=$((retryCount+1))
```
daniv-msft, line 86: Nit: not a blocker, just curious: do we need the double parentheses?

builders/container-apps-internal-registry-demo/startup-script.sh
```diff
@@ -52,49 +52,124 @@ token=$(printf "%s" "$REGISTRY_AUTH_USERNAME:$REGISTRY_AUTH_PASSWORD" | base64)
...
-  -run-image mcr.microsoft.com/oryx/builder:stack-run-debian-bullseye-20230926.1 \
-  $APP_IMAGE
+retryCount=0
+until [ "$retryCount" -ge $RETRY_ATTEMPTS ]
+do
+  if [ "$retryCount" -ge 1 ]; then
+    echo "----- Retrying analyze phase (attempt $retryCount) -----"
+  fi
+
+  /lifecycle/analyzer \
```
daniv-msft, line 80: As of today, the CLI stops as soon as "ERROR:" is detected in the logs. However with the retry this wouldn't work anymore. I believe we might be able to have something where the CLI checks for "ERROR:" only if the logs are done streaming? If not, we might need to have some error interception here maybe. Nothing to change for this PR.

builders/container-apps-internal-registry-demo/startup-script.sh
```diff
@@ -52,49 +52,124 @@ token=$(printf "%s" "$REGISTRY_AUTH_USERNAME:$REGISTRY_AUTH_PASSWORD" | base64)
...
+  retryCount=$((retryCount+1))
+  sleep $RETRY_DELAY
+done
+
+fail_if_retry_exceeded $retryCount
 
 # Execute the extend phase
 echo
-echo "======================================"
 echo "===== Executing the extend phase ====="
```
daniv-msft, line 136: The extend phase is when the code is being built. We cannot retry it, or else in case of build failure we would slow down the experience too much. From my experience I've not seen it failing without a good reason.

builders/container-apps-internal-registry-demo/startup-script.sh
pauld-msft: so once all the retries complete, we exit with code 1. This actually will cause another build to be created (up to 4 times i think, provided time limit of 30 min has not been reached), so we get tons of retrying altogether for each build

builders/container-apps-internal-registry-demo/startup-script.sh
pauld-msft: I'll remove this retry then, good point

builders/container-apps-internal-registry-demo/startup-script.sh
pauld-msft: https://tldp.org/LDP/abs/html/arithexp.html

builders/container-apps-internal-registry-demo/startup-script.sh
pauld-msft: done

william-msft (approved):

daniv-msft (approved):

PR #2243, opened 2023-11-06
Pauld/improve error handling builder

pauld-msft:
Updates the builder error handling so that we parse the exit codes from each of the lifecycle commands rather than using the `e` flag to immediately exit. This allows us to have a custom error message that we can parse using the CLI:
- `Build process failed with exit code` (this can be whatever we want, I did not spend much time thinking about it so let me know if anyone has suggestions)

### DIFF
--- a/builders/container-apps-internal-registry-demo/buildCappsBuilder.sh
+++ b/builders/container-apps-internal-registry-demo/buildCappsBuilder.sh
@@ -12,7 +12,6 @@ source $REPO_DIR/build/__variables.sh
 destinationFqdn="oryxprodmcr.azurecr.io"
 destinationRepo="public/oryx/builder"
 destinationTag="capps-20230208.1"
-baseBuilderImage="mcr.microsoft.com/oryx/builder:debian-bullseye-20231016.1"
 
 PARAMS=""
 while (( "$#" )); do
@@ -27,10 +26,6 @@ while (( "$#" )); do
       ;;
     -t|--destination-tag)
       destinationTag=$2
-      shift 2
-      ;;
-    -b|--base-builder-tag)
-      baseBuilderImage=$2
       shift 2
       ;;
     --) # end argument parsing
@@ -60,7 +55,6 @@ echo "Building '$BUILD_IMAGE'..."
 echo
 cd $SCRIPT_DIR
 docker build \
-  --build-arg BASE_BUILDER_IMAGE=$baseBuilderImage \
   -t $BUILD_IMAGE \
   -f Dockerfile \
   .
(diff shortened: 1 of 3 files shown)


cormacpayne (dismissed):

builders/container-apps-internal-registry-demo/buildCappsBuilder.sh
```diff
@@ -12,7 +12,6 @@ source $REPO_DIR/build/__variables.sh
 destinationFqdn="oryxprodmcr.azurecr.io"
 destinationRepo="public/oryx/builder"
 destinationTag="capps-20230208.1"
-baseBuilderImage="mcr.microsoft.com/oryx/builder:debian-bullseye-20231016.1"
```
daniv-msft, line 15: Could you please share more details about this part of the change?

builders/container-apps-internal-registry-demo/startup-script.sh
```diff
@@ -140,10 +160,18 @@ echo "===== Executing the extend phase ====="
   -log-level debug \
   -app $CNB_APP_DIR
 
+lifecycleExitCode=$?
+if [ $lifecycleExitCode -ne 0 ]; then
+    echo "----- Build failed -----"
```
daniv-msft, line 165: Nit: I'm not sure that this "----- Build failed -----" header adds much. Please consider removing it as the next error line should be sufficient from my perspective.

builders/container-apps-internal-registry-demo/startup-script.sh
```diff
@@ -140,10 +160,18 @@ echo "===== Executing the extend phase ====="
...
+    echo "Build process failed with exit code '$lifecycleExitCode'. Exiting..."
```
daniv-msft, line 166: Please consider:
* Replacing "Build process failed" by "Cloud Build failed". In the CLI we're pushing the "Cloud Build" wording, so it would make sense to be consistent.
* Please consider removing the "Exiting..." part. I don't feel strongly so feel free to keep it if you do, but I don't think it adds much.

builders/container-apps-internal-registry-demo/buildCappsBuilder.sh
pauld-msft: This just makes it so that I don't need to update the builder image in 2 places. So running this script will use whatever builder is defined in the Dockerfile, rather than overriding it.

builders/container-apps-internal-registry-demo/startup-script.sh
pauld-msft: updated this to be a single line that still has the `-----` for better visibility and distinction from the rest of the logs

daniv-msft: OK sounds good, thanks!

daniv-msft (dismissed):

daniv-msft (dismissed):

daniv-msft (approved):

PR #2254, opened 2023-11-13
Supporting running FastAPI apps via Uvicorn in Python runtimes

cormacpayne:
This PR adds support for detecting and running applications targeting the FastAPI Python framework. Rather than using the traditional Gunicorn server for running the application, we are now supporting Uvicorn as the server of choice for FastAPI applications.


tests/Oryx.Integration.Tests/Python/PythonEndToEndTests.cs
```diff
@@ -370,5 +370,101 @@ public async Task CanBuildAndRunPythonApp_UsingIntermediateDir_AndNestedOutputDi
                     Assert.Contains("Hello World!", data);
                 });
         }
+
+        [Fact]
+        [Trait("category", "python-3.11")]
```
harryli0108, line 375: Should we add the same test for 3.12 as well?

harryli0108 (approved):

tests/Oryx.Integration.Tests/Python/PythonEndToEndTests.cs
cormacpayne: There's a test for 3.12 down below

william-msft (approved):

images/runtime/python/3.10/bullseye.Dockerfile
```diff
@@ -81,6 +81,7 @@ LABEL io.buildpacks.stack.id="oryx.stacks.skeleton"
 RUN ${IMAGES_DIR}/runtime/python/install-dependencies.sh
 RUN pip install --upgrade pip \
     && pip install gunicorn \
+    && pip install uvicorn \
```
daniv-msft, line 84: Unfortunately, I don't think we can add  these for the existing images as it can have some implications in terms of size and side-effects for existing customers.
The easiest ways I would see is:
* Waiting for a new Python version and only enable this here.
* Create a new image just for this case, which is a bit annoying as well.
We can sync offline if you want.

PR #2257, opened 2023-11-14
support build environment variable

yilims:
**Problem:**
The build environment variable is not supported in cloud build service and user can't configure buildpacks.
**Solution:**
Add prefix "USER_ENV_" for  user defined environment variable to avoid collision with build-in environment variables. The prefix will be removed in the build pod to restore origianl name. User defined environment variables will be passed to Buildpack when building image for the app.


builders/container-apps-internal-registry-demo/startup-script.sh
```diff
@@ -13,6 +13,22 @@ temp_app_source_dir="/tmp/appsource"
 temp_app_source_path="$temp_app_source_dir/$FILE_UPLOAD_BLOB_NAME"
 mkdir $temp_app_source_dir
 
+
+# List all the environment variables and filter the environment variable with prefix "USER_ENV_", then write them to folder "/platform/env", 
+# file name is the environment variable name without prefix USER_ENV_, file content is the environment variable value.
+build_env_dir="/platform/env"
```
cormacpayne, line 19: nit: do we need to create this `/platform/env` folder before we start writing files to it?

cormacpayne, line 19: I'm assuming these environment variables are only used by the Java buildpacks until we move to a model where our Oryx-based buildpacks are reading the environment variables from this `/env` folder, correct?

cormacpayne (approved):

builders/container-apps-internal-registry-demo/startup-script.sh
yilims: When environment varaibles are stored in /platform/env, lifecycle will make it available to all the subsequent buildpack, buildpack doesn't need to read environment vairable from /platform/env manually. I'd propse Oryx-based buidpacks adopt the same implentation to avoid read/update the environment variable to a file manually.

yilims: /platform/env folder already exists in the builder image, no need to create it.

pauld-msft (approved):

PR #2262, opened 2023-11-15
Update the builder with new lifecycle binaries and base builder image

cormacpayne:


pauld-msft (dismissed):

builders/container-apps/Dockerfile
```diff
@@ -8,15 +8,20 @@ FROM ${BASE_BUILDER_IMAGE}
 # ACR_SCOPE: scope that the image will need to push to the ACR, e.g. repository:app/capps-runnable-app:pull,push
 # MI_PRINCIPAL_ID: principal id of the managed identity that will allow the image to push to the specified ACR
 ENV APP_IMAGE="" \
-	ACR_RESOURCE_NAME="" \
+	ACR_RESOURCE_NAME="private-repository-k8s.k8se-system" \
```
daniv-msft, line 11: Could you please confirm why we need to specify this specifically? If we could avoid having a hard dependency on the Kubernetes resource name/namespace, it would be a lot more resilient to breaking changes.

daniv-msft (approved):

PR #2273, opened 2023-12-04
Adding Blazor test app with Dockerfile back

harryli0108:
Adding Blazor test app back to resolve the test failure from github actions validation workflow.


pauld-msft (approved):
:shipit:

PR #2276, opened 2023-12-11
Ensure that the .NET 8.0 runtime images aren't pushed to staging

cormacpayne:
When initially publishing the GA .NET 8.0 runtime images before they were available to the public, we were publishing them to our private staging MCR instance for internal testing -- since .NET 8.0 has GA'd, we never removed the flag that told our pipelines to publish these runtime images to the private staging account, so our automation hasn't been publishing new versions of the .NET 8.0 runtimes to our public MCR instance.

As a result, our builder isn't able to pick up any stable .NET 8.0 runtime images (since the _only_ stable version we have was _manually tagged_ from the private staging MCR instance and pushed to our public MCR instance), and will default back to RC.2, which is causing some issues for customers.

Once this PR is merged, our CI pipelines should publish stable .NET 8.0 runtime images to our public MCR instance that include both the regular and patching tags required for Oryx and builder flows.


daniv-msft (dismissed):

build/__stagingRuntimeConstants.sh
```diff
@@ -1,6 +1,6 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
-DOTNETCORE_STAGING_RUNTIME_VERSIONS=("8.0")
+DOTNETCORE_STAGING_RUNTIME_VERSIONS=()
```
pauld-msft, line 3: can we update this https://github.com/microsoft/Oryx/blob/main/build/constants.yaml#L482 constant and then re-generate the constants? I think that should actually update it throughout the codebase so it won't get staged anymore

cormacpayne: Ah good catch -- I'm running the pipeline to create new base runtime images for .NET that are on the public MCR instance, so I'll re-generate the constants as a part of this and get your comment resolved with it. Thanks!

william-msft (approved):

pauld-msft (approved):
:shipit:

PR #2279, opened 2023-12-15
Remove pulled .NET 7.0 version from our list of buster versions to build

cormacpayne:
There's a lingering .NET 7.0.404 version in our buster `versionsToBuild.txt` file that doesn't exist in the production storage account since it was manually pulled, so this change removes that version from the file so the CI pipeline can succeed.


pauld-msft (approved):
:shipit:

PR #2294, opened 2023-12-20
Pauld/fix pipelines

pauld-msft:
image

extension was updated for the first time in years:
image


daniv-msft (approved):
Thank you for fixing this!

PR #2359, opened 2024-02-22
Add Support for PYTHON|3.12.2 and PYTHON|3.11.8

kumaraksh1:

Adding support for PYTHON|3.12.2, PYTHON|3.11.8 for following debian flavors.
Bulllseye
Buster
Bookworm


images/runtime/python/3.11/base.bullseye.Dockerfile
```diff
@@ -29,7 +29,7 @@ ADD build ${BUILD_DIR}
 RUN find ${IMAGES_DIR} -type f -iname "*.sh" -exec chmod +x {} \;
 RUN find ${BUILD_DIR} -type f -iname "*.sh" -exec chmod +x {} \;
 
-ENV PYTHON_VERSION 3.11.0
+ENV PYTHON_VERSION 3.11.8
```
sarsharma, line 32: Are we sure we need to update this for base.bullseye Dockerfile also?

kumaraksh1: Let me come back on this to you.

william-msft: We shouldn't need to update this, unless there has been an issue associated with this.

kumaraksh1: I also realized from the code that template.dockerfile is copied to base.bullseye.Dockerfile

kumaraksh1: But this should not create an issue as it would be overridden and it's fine to keep the parity like bookworm.

daniv-msft (approved):

sarsharma (approved):

PR #2352, opened 2024-02-05
Convert Bases Image pipelines to 1ES Template

daniv-msft:
- [ ] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as \#\<issue\_number\>.
- [ ] Tests are included and/or updated for code changes.
- [ ] Proper license headers are included in each file.

PR #2397, opened 2024-03-26
Moved 1ES compliant pipeline templates to a separate folder temporarily

harryli0108:
Based on the changes from #2352 but moved the new templates to a new folder and reverted other templates back to the older version for fallbacks.


snehapar9 (approved):
:shipit:

PR #2407, opened 2024-04-18
Update Packages 

kumaraksh1:
Updating some packages like npm to fix cves.

Pardon my branch name #ds.


tests/Oryx.RuntimeImage.Tests/Node/NodeRuntimeImageContainsRequiredPrograms.cs
```diff
@@ -265,7 +265,14 @@ public void NodeBullseyeImages_Contains_Correct_NPM_Version(string version, stri
                 () =>
                 {
                     Assert.True(result.IsSuccess);
-                    Assert.Contains(NodeVersions.NpmVersion, result.StdOut.ReplaceNewLine());
+                    if (version == "18" || version == "20")
```
sarsharma, line 268: should we just do version >=18 here?

sarsharma (approved):

tests/Oryx.RuntimeImage.Tests/Node/NodeRuntimeImageContainsRequiredPrograms.cs
kumaraksh1: It's fine i think

PR #2485, opened 2024-07-29
fix: deprecated poetry flag

### LINKED ISSUE #2019: Support poetry install --only
### Feature Request
Currently the script
[PythonBashBuildSnippet.sh.tpl](https://github.com/microsoft/Oryx/blob/9dbab9750ced045e9b4bd30104c3f9b680edf717/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl#L79-L110) assumes that pyproject.toml implies that the user always intends to use 'poetry run --no-env'. First of all, 'poetry install --no-env' is deprecated in favor of poetry install --only no-dev. Second of all, we have a use case where we deploy a monorepo containing many dependencies using github actions to an Azure App, which uses Oryx as the build system. We have a smaller dependency group, which contains only the dependencies that are needed to the deployed application. So, in this case, what we would like to run is poetry install --only on this subset of dependencies.

It would be nice if we can set an environment variable that allows to specify additional poetry install arguments. The documentation makes no mention of this.

raffifu:
This issue close #2019

Currently flag --no-dev in poetry is deprecated and the new flag option is --only. this PR change the poetry command to use --only

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -100,9 +100,9 @@ fi
         printf %s " , $InstallPipCommand" >> "$COMMAND_MANIFEST_FILE"
         pip install poetry
         echo "Running poetry install..."
-        InstallPoetryCommand="poetry install --no-dev"
+        InstallPoetryCommand="poetry install --only"
         printf %s " , $InstallPoetryCommand" >> "$COMMAND_MANIFEST_FILE"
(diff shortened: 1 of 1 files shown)


dimbleby:
The "--only" option requires a value.

I suppose you intend `--only main`

tsukhu:
shouldn't this be `--without dev` as a replacement to `--no-dev`

- [Docs, FAQ: Replace deprecated --no-dev with --without dev](https://github.com/python-poetry/poetry/pull/8659)

- [Change Log 2.0.0](https://github.com/python-poetry/poetry/blob/main/CHANGELOG.md#200---2025-01-05)

PR #2499, opened 2024-08-29
User/vchintalapat/monitor versions

Saipriya-1144:
This pull request introduces a new pipeline for monitoring stack versions and automating updates.

- Scrapes the latest available versions of each stack from the internet.
- Updates constants.yaml based on values in Oryx_variables variable group:
   - If the value is latest, updates constants.yaml with the latest scraped version.
   - If the value is dont_change, does not update the version.
   - If the value is pinned to a fixed value, updates constants.yaml with that fixed value.
- Updates versionstoBuild files for SDKs after updating constants.yaml.
- Creates a new branch and pushes all changes to it.
- Creates a Pull Request (PR). (Sample PR: https://github.com/microsoft/Oryx/pull/2500)

After PR approval and merging, on running other pipelines (AAPT-Antares-Publish Binaries and AAPT-Antares-Official) the Oryx Images gets generated.


CodingIsBliss (dismissed):

monitor_version_scripts/web_scrap_files/dotnet_versions.py
```diff
@@ -0,0 +1,120 @@
+from bs4 import BeautifulSoup
```
sarsharma, line 1: Can we use this API to get latest versions instead of crawling the webpage?
https://endoflife.date/docs/api

platforms/nodejs/versions/bullseye/versionsToBuild.txt
```diff
@@ -7,7 +7,6 @@
 12.22.9
 12.22.11
 12.22.12
-
```
sarsharma, line 10: nit: any reason we removed newlines from these versionsToBuild text files?

platforms/python/prereqs/build.sh
```diff
@@ -59,6 +59,7 @@ if [ "$debianFlavor" == "stretch" ]; then
 
     PIP_VERSION="20.2.3"
 else
+    PIP_VERSION="21.2.4"
```
sarsharma, line 62: we can remove the debian-flavor = stretch condition entirely

monitor_version_scripts/monitor_variables.sh
```diff
@@ -0,0 +1,62 @@
+# #!/bin/bash
+
+apt-get update
+apt-get install -y python3 python3-pip
```
sarsharma, line 4: nit: python & pip should be available preinstalled on the agents 
also we can install multiple pip packages in one go - `pip install bs4 lxml requests`

monitor_version_scripts/update_constants.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
+
+Old_constants_FILE=$(cd .. && pwd)/images/constants.yml
```
sarsharma, line 3: any reason we are using absolute paths here?

platforms/python/prereqs/build.sh
Saipriya-1144: Will take this one in next pr

platforms/nodejs/versions/bullseye/versionsToBuild.txt
Saipriya-1144: Removing these so we can sort the versionsToBuild file whenever a new item gets added

CodingIsBliss (approved):

monitor_version_scripts/web_scrap_files/dotnet_versions.py
Saipriya-1144: Yes, updated node_versions, php_versions , python_versions to take latest versions from the api

monitor_version_scripts/update_constants.sh
Saipriya-1144: No, it works with relative path as well

Saipriya-1144: Updated

CodingIsBliss (approved):

sarsharma (approved):

Issue #2557, opened 2025-01-05
Deployment of Python app fails because of usage of --no-dev parameter


MaartenUreel:
### Bug Report

Since a few hours, a deployment of a Python 3.12 app on Azure Web App fails.

On a working deployment, we saw this warning:
```
Running poetry install...
The `--no-dev` option is deprecated, use the `--only main` notation instead.
Installing dependencies from lock file
```

Now since the last deploy 24 hours ago, we see this message:
```
Running poetry install...
The option "--no-dev" does not exist
```

And the deployment fails.

It seems on working deployment, this version was pulled:
```
Running pip install poetry...
Collecting poetry
  Using cached poetry-1.8.5-py3-none-any.whl.metadata (6.9 kB)
```

While now it is this version:
```
Collecting poetry
  Using cached poetry-2.0.0-py3-none-any.whl.metadata (7.0 kB)
```

But the version of Poetry to use is not something I chose, it is part of the build process if I am not mistaken.
--no-dev is remove in version 2, so either the parameter needs to be replaced with the correct replacement or the version needs to be pinned to <2.

MaartenUreel:
I have worked around this on my own project by adding this to my pre-build script so that we do not rely on the built in Poetry handling:

```
# Temporary fix for Poetry, export requirements.txt
curl -sSL https://install.python-poetry.org | python3 -
/home/.local/bin/poetry self add poetry-plugin-export
/home/.local/bin/poetry export -f requirements.txt --output requirements.txt
```

lloyd-lighthouse:
im having the same issue.
where do you add this 'pre-build' script to?

lloyd-lighthouse:
**_Update:_**

I spent hours doing all sorts ....
I believe that this is impossible to fix from our side, until Microsoft resolve the issue themselves and force Oryx to either

a) Use Poetry 1.8.5
b) Remove anything in their script that refers to `--no-dev` 

I spent a few hours waiting for someone from Azure Support to contact me, and when they finally did, I spent a bunch of time troubleshooting and doing pretty much the same things I'd tried already.

Only when we had exhausted all options, did the Support Admin take a breath and listen to what I thought was wrong.
As of now, there is a promise to '_coordinate with the internal team to ensure everything is aligned_'.

**_My Interim Solution:_**

To remove anything to do with poetry from my project and revert to requirements.txt instead. 
[...]

PR #2559, opened 2025-01-07
Fix for issue #2557

tsukhu:
python 2.0 has removed --no-dev and we need to use `--without dev` instead


luismelo1999 (approved):
Thank you, I am also blocked by this issue

PR #2561, opened 2025-01-07
Revert poetry

### LINKED ISSUE #2557: Deployment of Python app fails because of usage of --no-dev parameter

sarsharma:
pinning poetry version to 1.8.5

This should fix #2557

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -96,9 +96,9 @@ fi
     then
         set +e
         echo "Running pip install poetry..."
-        InstallPipCommand="pip install poetry"
+        InstallPipCommand="pip install poetry==1.8.5"
         printf %s " , $InstallPipCommand" >> "$COMMAND_MANIFEST_FILE"
-        pip install poetry
+        pip install poetry==1.8.5
         echo "Running poetry install..."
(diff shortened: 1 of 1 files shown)


kumaraksh1 (approved):
LGTM!!

surenderssm:
Let's link the reported issue in this PR.

surenderssm (approved):

MaartenUreel:
And why not just switch the flag `--no-dev` which has been deprecated for a long time to `--only main`?

It is a minimal change that would allow to use a recent up-to-date Poetry version instead if pinning an old version...

PR #2540, opened 2024-11-13
.NET9 post GA cleanup

sarsharma:
reverting changes that are no longer needed post .NET9 public availability

PR #2569, opened 2025-02-13
Only install .NET Monitor 9 on .NET 9 images

jander-msft:
The .NET 9 Oryx images are unconditionally installing the latest .NET Monitor version onto the image. This is opposed to the prior .NET Oryx images where they are range bound to the major version that corresponds with the .NET version. If this constraint is not added, when .NET Monitor releases a newer version that only works on a future version of .NET, image build failure will occur. This change is intended to mitigate that risk by applying the same range bound as the prior images have e.g. https://github.com/microsoft/Oryx/blob/main/images/runtime/dotnetcore/8.0/bookworm.Dockerfile#L9


wiktork:
It looks like the removal was intended to cleanup the rc version with #2540.
@sarsharma as FYI

surenderssm (approved):

PR #2578, opened 2025-03-31
Update yaml.v2 to 2.4.0 and remove old crypto package reference in go

keshava-kandikonda:
Updated go.mod for the following 2 packages and updated few other modules to latest versions - 

> Updated yaml.v2 to 2.4.0
> Removed old crypto package references


CodingIsBliss (dismissed):

src/startupscriptgenerator/src/php/php.exe
sarsharma, line 1: let's remove?

keshava-kandikonda: Is this not in use? If we are sure, I can remove or else I will have one more PR soon to update Node packages as well I can remove it in that PR.

sarsharma (approved):

PR #2598, opened 2025-06-11
Search for checksum tag irrespective of case.

CodingIsBliss:
This fixes the issue where the checksum tags for some of the older stacks starts with a 'c'. Made the search for checksum case insensitive.

image

### DIFF
--- a/src/BuildScriptGenerator/ExternalSdkProvider.cs
+++ b/src/BuildScriptGenerator/ExternalSdkProvider.cs
@@ -184,7 +184,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator
         }
         catch (Exception ex)
         {
-          this.outputWriter.WriteLine($"Failed to get blob for platform {platformName}, blobName {blobName} from external provider. Exception : {ex}");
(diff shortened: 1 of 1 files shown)


src/BuildScriptGenerator/ExternalSdkProvider.cs
```diff
@@ -161,16 +172,19 @@ public async Task<bool> RequestBlobAsync(string platformName, string blobName)
...
           else
           {
+            this.outputWriter.WriteLine($"Failed to get blob for platform {platformName}, blobName {blobName} from external provider.");
             this.logger.LogError("Failed to get blob for platform {platformName}, blobName {blobName}", platformName, blobName);
             return false;
           }
         }
         catch (Exception ex)
         {
+          this.outputWriter.WriteLine($"Failed to get blob for platform {platformName}, blobName {blobName} from external provider. Exception : {ex}");
```
siriande: Is it required to print complete exception details here? Can we print just the message?

SarathBussa: done

siriande (approved):

kumaraksh1 (approved):
LGTM

prathambande (approved):

PR #2617, opened 2025-07-22
[Oryx] Stack update for dotnet 10 preview-6

Shi1810:
Stack update: 

Dotnet 10 preview-6: https://dotnet.microsoft.com/en-us/download/dotnet/10.0


Saipriya-1144:
Can you update sdk version in DotNetSdkVersions file as well otherwise the sdk tests will fail

Issue #2614, opened 2025-07-10
Support uv


GPetrites:
### Feature Request

Please add support for uv as an alternative to pip and poetry for installing package dependencies. uv is becoming the preferred tool for many organizations.

PR #2652, opened 2025-09-17
Add uv support

hallvictoria:
Fixes https://github.com/microsoft/Oryx/issues/2614. Adds support for `uv`.

If `pyproject.toml` and `uv.lock` are present and `requirements.txt` is not, run `uv pip install -r uv.lock`.

- [X] The purpose of this PR is explained in this message or in an issue. If an issue please include a reference as as #<issue_number>.


src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -94,23 +94,40 @@ fi
...
-        if [[ $pythonBuildExitCode != 0 ]]
+        if [ -e "uv.lock" ] && [ ! -e "$REQUIREMENTS_TXT_FILE" ];
         then
-            LogWarning "${output} | Exit code: {pythonBuildExitCode} | Please review message | ${moreInformation}"
-            exit $pythonBuildExitCode
-        fi
+            # Install using uv
+            set +e
+            echo "Detected uv.lock. Installing dependencies with uv..."
+            InstallUvCommand="uv pip install -r uv.lock"
```
larohra: Given that `uv` is faster for installing packages, should we also consider installing `requirements.txt` from `uv` too? (relevant for line-67)

larohra: These scripts are assuming `uv` is already installed in the build image which might not be the case. We might have to install `uv` first and then install the dependencies (similar to how they're doing for poetry below - 

```bash
pip intall uv
```

larohra: I was talking with Gavin about this, and we can potentially provide an override for opt-in (or opt-out) based on the Environment variables (provided via AppSettings) to always prioritize `uv` installs if possible.

gavin-aguiar: `uv sync`

vrdmr: Ideally - yes.

But let this be not an AppSetting but more a host.json setting - where we create a new config

vrdmr: Whats the technical difference between the two?

vrdmr: Yes - needs to be installed similar to poetry.

larohra: Orynx currently has a list of settings that it supports for Python APps - https://github.com/microsoft/Oryx/blob/4a8bbc9093a2f58375f945f0db3ae04cad24c989/doc/configuration.md?plain=1#L49C1-L50C1

```
Setting name for Python apps | Description                                                    | Default | Example
-----------------------------|----------------------------------------------------------------|---------|----------------
PYTHON\_VERSION              | Specify which Python version the app is using                  | ""      | "2.7.1"
PYTHON\_DEFAULT\_VERSION     | Specify which Python version the app defaults to if none detected | ""      | "2.7.1"
DISABLE\_PYTHON\_BUILD       | Do not apply Python build even if repo indicates it            | `false` | `true`, `false`
DISABLE\_COLLECTSTATIC       | Disable running `collectstatic` when building Django apps.     | `false` | `true`, `false`
CUSTOM\_REQUIREMENTSTXT\_PATH| Specify where a requirements.txt is locating. If not set, default is at root of the repo.| ""    | "subdir/requirements.txt"
PYTHON\_ENABLE\_GUNICORN\_MULTIWORKERS| Enable Gunicorn multi worker multi thread config.     | `false` | `true`, `false`
PYTHON\_GUNICORN\_CUSTOM\_WORKER\_NUM| Only works when `PYTHON\_ENABLE\_GUNICORN\_MULTIWORKERS` is set to `true`. Specify Gunicorn multi worker number. If not set, default is (2 * CPU core num) + 1| `(2 * CPU core num) + 1`, `1`    | "2"
PYTHON\_GUNICORN\_CUSTOM\_THREAD\_NUM| Only works when `PYTHON\_ENABLE\_GUNICORN\_MULTIWORKERS` is set to `true`. Specify Gunicorn multi thread number. If not set, default is single thread. | ""    | "4"
ORYX\_DISABLE\_PIP\_UPGRADE  | Remove the --upgrade flag from the pip install command when targeting a specific package installation directory. | `false` | `true`, `false`
NGINX\_CONF\_FILE            | Specify a customized configuration file to modify nginx.conf file        | ""      | "newconfigfile.conf"
```

All these are propagated via AppSettings for FunctionApps. Might just be cleaner and simpler to add it on here to reduce cx confusion. 

What benefits do you think adding another config provides here over these app settings?

anthonychu: Suggest investigating using uv for requirements.txt in a later PR.

anthonychu: `uv sync` automatically activates a virtual env, need to check if it conflicts with the virtual env activation logic above.

anthonychu: I don't think `uv pip install` takes a uv.lock file. See if `uv sync --locked` works with the virtual env enabled above.

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -164,26 +181,47 @@ fi
...
         then
-            LogWarning "${output} | Exit code: {pythonBuildExitCode} | Please review message | ${moreInformation}"
-            exit $pythonBuildExitCode
-        fi
+            # Install using uv
+            set +e
+            echo "Detected uv.lock. Installing dependencies with uv..."
+            START_TIME=$SECONDS
+            SITE_PACKAGES_PATH="{{ PackagesDirectory }}"
+            InstallUvCommand="uv pip install --target $SITE_PACKAGES_PATH -r uv.lock"
```
anthonychu: I don't think `uv pip install` takes a uv.lock file. I think it might require 2 steps:

- `uv export --locked --format requirements-txt > $REQUIREMENTS_TXT_FILE`
- `$python -m pip install --cache-dir $PIP_CACHE_DIR --prefer-binary -r $REQUIREMENTS_TXT_FILE --target="{{ PackagesDirectory }}" {{ PipUpgradeFlag }} | ts $TS_FMT` (taken from requirements.txt section above)

`uv pip install` might also work in the second step but not sure how it differs from regular pip.

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: It doesnt conflict with the existing venv (if we use `--active`) logic as long as the original venv is created for the same python version that the cx code was written for (basically the build image python version should match the python version specified in `pyproject.toml`). 

If this does not match, then `uv sync` recreates the virtual env with the version specified in the `pyproject.toml` - 

```
(yoloenv) root@2366eaf1e071:/repo# uv sync --active
Using CPython 3.11.13
**Removed virtual environment at: yoloenv**
**Creating virtual environment at: yoloenv**
Resolved 4 packages in 4ms
```

larohra: Fixed, just using `uv sync`

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: Thanks for sharing the tips. Given how poetry also ignores `PackagesDirectory` in this case, I was thinking of just keeping it similar (i.e. using `uv sync` and ignoring `PackagesDirectory`). Will let the Oryx team confirm how to handle this case though.

anthonychu: `PackagesDirectory` is needed for function app builds to work. This is an example command from Kudu: `[Kudu-OryxBuildStep] Running oryx build command: oryx build /tmp/zipdeploy/extracted -o /home/site/wwwroot --platform python --platform-version 3.12.2 -p packagedir=.python_packages/lib/site-packages`

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: Targeting this for later, for now just scoping to supporting only the cases where `uv` is the only possibility

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: Fixed it (screenshot below)

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -44,17 +44,28 @@ fi
...
     echo "Python Virtual Environment: $VIRTUALENVIRONMENTNAME"
 
-    if [ -e "$REQUIREMENTS_TXT_FILE" ]; then
-        VIRTUALENVIRONMENTOPTIONS="$VIRTUALENVIRONMENTOPTIONS --system-site-packages"
+    if [ -e "pyproject.toml" ] && [ -e "uv.lock" ] && [ ! -e "$REQUIREMENTS_TXT_FILE" ]; then
+        echo "Detected uv.lock (and no $REQUIREMENTS_TXT_FILE); creating virtual environment with uv..."
+        echo "Install uv first..."
+        InstallUv="pip install uv"
+        printf %s " , $InstallUv" >> "$COMMAND_MANIFEST_FILE"
+        pip install uv
```
gavin-aguiar: python -m pip install uv

gavin-aguiar: If `pip install uv` fails, the script will continue and attempt to run `uv venv`. Should we exit if pip fails?

`pip install uv || { echo "Failed to install uv"; exit 1; }`

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -94,22 +105,40 @@ fi
...
+            echo "Running poetry install..."
+            InstallPoetryCommand="poetry install --no-dev"
+            printf %s " , $InstallPoetryCommand" >> "$COMMAND_MANIFEST_FILE"
+            output=$( ( poetry install --no-dev; exit ${PIPESTATUS[0]} ) 2>&1)
+            pythonBuildExitCode=${PIPESTATUS[0]}
+            set -e
+            echo "${output}"
+            if [[ $pythonBuildExitCode != 0 ]]
+            then
+                LogWarning "${output} | Exit code: {pythonBuildExitCode} | Please review message | ${moreInformation}"
```
gavin-aguiar: This was old code but it's missing a $.
${pythonBuildExitCode}.

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -94,22 +105,40 @@ fi
...
+                LogError "${output} | Exit code: ${uvExitCode} | Please review your uv.lock | ${moreInformation}"
+                exit $uvExitCode
+            fi
+        else
+            # Fallback to poetry
+            set +e
+            echo "Running pip install poetry..."
+            InstallPipCommand="pip install poetry==1.8.5"
+            printf %s " , $InstallPipCommand" >> "$COMMAND_MANIFEST_FILE"
+            pip install poetry==1.8.5
```
gavin-aguiar, line 129: Why are we pinning poetry? It wasn't pinned before

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -164,25 +193,54 @@ fi
...
-        echo "${output}"
-        if [[ $pythonBuildExitCode != 0 ]]
+        if [ -e "uv.lock" ];
         then
-            LogWarning "${output} | Exit code: {pythonBuildExitCode} | Please review message | ${moreInformation}"
-            exit $pythonBuildExitCode
+            # Install using uv
+            set +e
+            echo "Detected uv.lock. Installing dependencies with uv..."
+            START_TIME=$SECONDS
```
gavin-aguiar, line 200: What is $SECONDS set to?

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
hallvictoria: poetry was pinned to 1.8.5 originally. Should we consider pinning the uv version? It's not at a 1.x major version yet - they could still introduce breaking changes

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -164,25 +193,54 @@ fi
...
+            printf %s " , $InstallPoetryCommand" >> "$COMMAND_MANIFEST_FILE"
+            output=$( ( poetry install --no-dev; exit ${PIPESTATUS[0]} ) 2>&1 )
+            pythonBuildExitCode=${PIPESTATUS[0]}
+            ELAPSED_TIME=$(($SECONDS - $START_TIME))
+            echo "Done in $ELAPSED_TIME sec(s)."
+            set -e
+            echo "${output}"
+            if [[ $pythonBuildExitCode != 0 ]]
+            then
+                LogWarning "${output} | Exit code: {pythonBuildExitCode} | Please review message | ${moreInformation}"
```
hallvictoria: ${pythonBuildExitCode}

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -94,22 +105,40 @@ fi
...
-        echo "${output}"
-        if [[ $pythonBuildExitCode != 0 ]]
+        if [ -e "uv.lock" ];
         then
-            LogWarning "${output} | Exit code: {pythonBuildExitCode} | Please review message | ${moreInformation}"
-            exit $pythonBuildExitCode
+            # Install using uv
+            set +e
+            echo "Detected uv.lock. Installing dependencies with uv..."
+            InstallUvCommand="uv sync --active --link-mode copy"
```
hallvictoria, line 113: Do we need to install uv first in this case before running uv sync?

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -164,25 +193,54 @@ fi
...
+                LogError "${output} | Exit code: ${uvExitCode} | Please review your uv.lock | ${moreInformation}"
+                exit $uvExitCode
+            fi
+        else
+            # Fallback to poetry
+            set +e
+            echo "Running pip install poetry..."
+            InstallPipCommand="pip install poetry"
+            printf %s " , $InstallPipCommand" >> "$COMMAND_MANIFEST_FILE"
+            pip install poetry
```
hallvictoria, line 228: old code but should poetry be pinned here as well?

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: We're already installing it in line #52. Basically in this case, I'm installing uv first, using uv to create a venv and then using uv to install depdendnecies here

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: Ohh good catch!

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: On deeper investigation, the script would exit on failure because this is set at the top - 
`set -e`

I think that's good enough for here, don't see any major benefits of adding a specific exit here for the same thing.

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: Its a preset variable for bash - 

```
High-Level Concept
$SECONDS acts like a stopwatch that starts automatically when your shell session or script begins. It continuously counts the number of seconds that have passed.

Granular Details
Origin: It's provided directly by the bash shell itself, not from an external command or environment variable. Shells like ksh and zsh also have this feature.
Behavior:
It starts at 0 when the script is launched.
It increments by one every second.
Reading $SECONDS gives you the total number of seconds the script has been running.
Usage in Your Code: The line START_TIME=$SECONDS captures the current value of the stopwatch at that specific moment and saves it into a new variable called START_TIME. This is a common and efficient way to time how long your script, or a portion of it, takes to execute.
```

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: I'm not too familiar with their scenarios to change this here so will leave it as is for now.

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: I'm not a fan of pinning versions since that can lead to using outdated builds. Will let the Oryx team guide us here on what they prefer :)

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -44,17 +44,28 @@ fi
 
     echo "Python Virtual Environment: $VIRTUALENVIRONMENTNAME"
 
-    if [ -e "$REQUIREMENTS_TXT_FILE" ]; then
-        VIRTUALENVIRONMENTOPTIONS="$VIRTUALENVIRONMENTOPTIONS --system-site-packages"
+    if [ -e "pyproject.toml" ] && [ -e "uv.lock" ] && [ ! -e "$REQUIREMENTS_TXT_FILE" ]; then
+        echo "Detected uv.lock (and no $REQUIREMENTS_TXT_FILE); creating virtual environment with uv..."
+        echo "Install uv first..."
```
sarsharma: nit: Installing uv...

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -44,17 +44,28 @@ fi
...
+        InstallUv="python -m pip install uv"
+        printf %s " , $InstallUv" >> "$COMMAND_MANIFEST_FILE"
+        python -m pip install uv
+        CreateVenvCommand="uv venv --link-mode=copy --system-site-packages $VIRTUALENVIRONMENTNAME"
```
sarsharma, line 53: do we also want to pass VIRTUALENVIRONMENTOPTIONS here while creating venv?

larohra: This corresponds to `--copies` that's used for ` python -m venv` - 

https://vscode.dev/github/hallvictoria/Oryx/blob/hallvictoria/support-uv/src/BuildScriptGenerator/Python/PythonPlatform.cs#L643-L646

```py
case "3":
                    virtualEnvModule = "venv";
                    virtualEnvParams = "--copies";
                    break;
```

I have the corresponding parameters for `uv` - 

`uv venv **--link-mode=copy --system-site-packages** $VIRTUALENVIRONMENTNAME`

PR #2677, opened 2025-10-15
Update golang for runtime script generator

kumaraksh1:
Updating golang image for runtime script generator from 1.23 to 1.25


Saipriya-1144:
Can you update version in node 24 as well

prathambande:
Are we not changing for this file? 
images/build/Dockerfiles/full.Dockerfile

kumaraksh1:
@prathambande changing for runtime images only.

PR #2684, opened 2025-10-22
Fixing Python Package Installation

prathambande:
This PR fixes the python package installation, which got changed in recent PRs.

With $python, packages are installed in local instead of venv.

Without fix:
image

With fix:
image

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -169,7 +169,7 @@ fi
     fi
 
     # For virtual environment, we use the actual 'python' alias that as setup by the venv,
-    python_bin=$python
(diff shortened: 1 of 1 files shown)


kumaramit-msft (approved):

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra, line 52: Should this be updated too?

kumaramit-msft (approved):

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
larohra: Flagging the comment just incase you missed it @kumaramit-msft @prathambande.
Feel free to resolve the comment if this indeed is working as expected :)

PR #2728, opened 2025-12-09
Disable Java Detection and Build

Saipriya-1144:
This PR disables Java Detection and Build

Testing:
Created a node 22 zip which has java files as well and triggered oryx build with current image and with new image which has JAVA disabled

With Current Image:
for oryx build:
image

for detect:
image
(Java is detected)

After Disabling JAVA:
for oryx build:
image

for detect:
image
(Java is not detected)


src/BuildScriptGenerator/BuildScriptGeneratorServiceCollectionExtensions.cs
```diff
@@ -27,8 +27,7 @@ public static IServiceCollection AddBuildScriptGeneratorServices(this IServiceCo
                 .AddDotNetCoreScriptGeneratorServices()
                 .AddPhpScriptGeneratorServices()
                 .AddScriptGeneratorServicesRuby()
```
surenderssm, line 29: How do we ensure that change is only picked in noble?

surenderssm, line 29: How about relooking at Ruby and Go as well?

Saipriya-1144: Sure, I will disable Ruby, Go and Hugo as well

Saipriya-1144: This error is coming for bookworm and noble. But disabling it for all debian versions because Java is not supported in Oryx Build and SDKs are also not getting updated

PR #2737, opened 2026-01-08
[Python] Add zstd compression support for build output

Saipriya-1144:
This PR introduces zstd compression support for Oryx build outputs to improve performance. Previously, build artifacts were compressed using gzip before copying to /home. Now, when enabled via the ORYX_COMPRESS_WITH_ZSTD environment variable from Kudu, the build process uses zstd compression instead, offering significantly faster compression and decompression.

Zstd compression is currently enabled only for Python 3.10+ web apps (excluding Azure Functions). When enabled, the build script generates output.tar.zst instead of output.tar.gz.

The startup script generator now checks for zstd-compressed tarballs first (output.tar.zst) and falls back to gzip-compressed tarballs (output.tar.gz) if not found. Zstd extraction uses the system tar command, while gzip continues using the existing Go-based extraction.

If **ORYX_COMPRESS_WITH_ZSTD** env is **false**, it uses gzip to compress

image

image

-------------------------------------------------------------------------------------------------------------------------------------

If **ORYX_COMPRESS_WITH_ZSTD** env is **true**, it uses zst to compress

image
image

-------------------------------------------------------------------------------------------------------------------------------------

Extraction errors can be monitored via panic messages:
image

### DIFF
--- a/images/runtime/python/install-dependencies.sh
+++ b/images/runtime/python/install-dependencies.sh
@@ -15,7 +15,6 @@ else
     odbcPackage="libodbc1"
 fi
 
-# GIS libraries for GeoDjango (https://docs.djangoproject.com/en/3.2/ref/contrib/gis/install/geolibs/)
 # libpq-dev is for PostgreSQL
 apt-get update \
     && apt-get upgrade -y \
@@ -30,6 +29,7 @@ apt-get update \
         $odbcPackage \
         apt-transport-https \
         swig \
+        # GIS libraries for GeoDjango (https://docs.djangoproject.com/en/3.2/ref/contrib/gis/install/geolibs/)
         binutils \
         libproj-dev \
         gdal-bin \


src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
```diff
@@ -190,7 +190,24 @@ then
 	DESTINATION_DIR="$OLD_DESTINATION_DIR"
 	echo "Compressing content of directory '$preCompressedDestinationDir'..."
 	cd "$preCompressedDestinationDir"
-	tar -zcf "$DESTINATION_DIR/output.tar.gz" .
+
+    echo "ORYX_COMPRESS_WITH_ZSTD: $ORYX_COMPRESS_WITH_ZSTD"
```
Saipriya-1144: Added this just for testing, I will remove this before merging the PR

src/startupscriptgenerator/src/common/compressionHelper.go
```diff
@@ -49,10 +51,35 @@ func ExtractTarball(tarballFile string, destinationDir string) {
 	}
 }
 
+func untar(dst string, tarballFile string) error {
+    // Detect compression format based on file extension
+	if strings.HasSuffix(tarballFile, ".tar.zst") {
+        // Use zstd for decompression
+        return untarWithZstd(dst, tarballFile)
```
kumaramit-msft, line 59: Lets add a log here - Extracting filename using zstd/gzip

Saipriya-1144: Added a log message

kumaramit-msft (approved):

images/runtime/python/install-dependencies.sh
```diff
@@ -15,20 +15,21 @@ else
     odbcPackage="libodbc1"
 fi
 
+# GIS libraries for GeoDjango (https://docs.djangoproject.com/en/3.2/ref/contrib/gis/install/geolibs/)
```
kumaraksh1: Whats this comment?

src/startupscriptgenerator/src/python/scriptgenerator.go
```diff
@@ -64,7 +64,20 @@ func (gen *PythonStartupScriptGenerator) GenerateEntrypointScript() string {
...
+
+		// Check for zstd-compressed tarball first, then fall back to gzip
+		zstdTarballFile := filepath.Join(gen.AppPath, "output.tar.zst")
+		gzipTarballFile := filepath.Join(gen.AppPath, "output.tar.gz")
+
+		var tarballFile string
+		if common.PathExists(zstdTarballFile) {
+			tarballFile = zstdTarballFile
+			println("Found zstd-compressed tarball.")
+		} else{
```
vijaysaayi: nit: space after else

images/runtime/python/install-dependencies.sh
Saipriya-1144: Earlier this comment was in between packages download command. It interrupted the download of packages beneath this, so moved it to the start of the command

src/startupscriptgenerator/src/python/scriptgenerator.go
```diff
@@ -64,7 +64,20 @@ func (gen *PythonStartupScriptGenerator) GenerateEntrypointScript() string {
 
 	if gen.Manifest.CompressDestinationDir == "true" {
 		println("Output is compressed. Extracting it...")
-		tarballFile := filepath.Join(gen.AppPath, "output.tar.gz")
+
+		// Check for zstd-compressed tarball first, then fall back to gzip
```
kumaraksh1, line 68: why not this in if else condition?
wont create much difference but would be better :)

images/runtime/python/install-dependencies.sh
```diff
@@ -15,20 +15,21 @@ else
...
         curl \
+        zstd \
         gnupg \
         libpq-dev \
         default-libmysqlclient-dev \
         unzip \
         $odbcPackage \
         apt-transport-https \
         swig \
-        # GIS libraries for GeoDjango (https://docs.djangoproject.com/en/3.2/ref/contrib/gis/install/geolibs/)
```
vijaysaayi, line 31: This comment seems to indicate that the following libs are for GIS. Should we keep it here only ?
binutils \
        libproj-dev \
        gdal-bin \

vijaysaayi (approved):
Approving with suggestions

src/startupscriptgenerator/src/python/scriptgenerator.go
kumaraksh1: or something like this:
tarballFile = filepath.Join(gen.AppPath, "output.tar.zst")

if tarballFile does not exist:
  tarballFile = filepath.Join(gen.AppPath, "output.tar.gz")

images/runtime/python/install-dependencies.sh
Saipriya-1144: image

All the packages after this comment were not getting installed till now. They were installed only after removing this comment

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
kumaraksh1, line 1: If you want you can add one more test case where extraction will fail.

prathambande (approved):

src/startupscriptgenerator/src/python/scriptgenerator.go
Saipriya-1144: // Check for zstd-compressed tarball first, then fall back to gzip
		zstdTarballFile := filepath.Join(gen.AppPath, "output.tar.zst")
		gzipTarballFile := filepath.Join(gen.AppPath, "output.tar.gz")

		var tarballFile string
		if common.PathExists(zstdTarballFile) {
			tarballFile = zstdTarballFile
			println("Found zstd-compressed tarball.")
		} else if common.PathExists(gzipTarballFile) {
			tarballFile = gzipTarballFile
			println("Found gzip-compressed tarball.")
		} else {
			panic(fmt.Sprintf("Could not find compressed output. Expected '%s' or '%s'", zstdTarballFile, gzipTarballFile))
		}
		
		Updated the code to extract only for .gz or .zst

images/runtime/python/install-dependencies.sh
Saipriya-1144: Reverting this chance, since we are fine without these packages in our images

tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
Saipriya-1144: I will add tests for all scenarios in next PR

PR #2750, opened 2026-01-12
Install uv and poetry in Python SDKs

Saipriya-1144:
This PR adds support for installing both `uv` and `poetry` in the SDK

Tested this by building SDK locally 
Old SDK:
image
Poetry and uv are not present

After this change:
image
poetry and uv are present

### DIFF
--- a/platforms/python/prereqs/build.sh
+++ b/platforms/python/prereqs/build.sh
@@ -110,6 +110,15 @@ then
         --disable-pip-version-check \
         --no-cache-dir \
         --no-warn-script-location
+
+    # Install uv and poetry
+    echo "Installing uv and poetry..."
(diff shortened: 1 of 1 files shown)


vijaysaayi (approved):

platforms/python/prereqs/build.sh
```diff
@@ -128,6 +137,16 @@ else
         --no-cache-dir \
         --no-warn-script-location \
         pip==$PIP_VERSION
+    
+    # Install uv and poetry
+    echo "Installing uv and poetry..."
+    LD_LIBRARY_PATH=/usr/src/python \
+    /usr/src/python/python -m pip install uv poetry \
```
vijaysaayi, line 144: Should we use pinned version of uv and poetry here ?

vijaysaayi (changes requested):

PR #2751, opened 2026-01-13
Validate configuration file size before parsing

Saipriya-1144:
This change adds file size validation before loading configuration files (TOML, YAML, JSON) into memory to improve build reliability in resource-constrained environments. 

A default limit of 10 MB is enforced for configuration files, which can be adjusted using the ORYX_MAX_CONFIG_FILE_SIZE_MB environment variable. 

This helps prevent out-of-memory errors in containerized build environments when processing unexpectedly large configuration files.

Testing:
Created a kudu image with this change and tried deploying an app which has hugo.yaml

-> If the size is less than 10MB, oryx-build was successful 
-> ORYX_MAX_CONFIG_FILE_SIZE_MB is set to 1MB and deployed a 10KB hugo.yaml file, oryx-build was successful
-> ORYX_MAX_CONFIG_FILE_SIZE_MB is set to 1MB and deployed a 2MB hugo.yaml file, oryx build failed
image

### DIFF
--- a/src/Detector/ParserHelper.cs
+++ b/src/Detector/ParserHelper.cs
@@ -22,7 +22,7 @@ namespace Microsoft.Oryx.Detector
         /// <summary>
         /// Default maximum allowed file size for configuration files (10 MB).
         /// </summary>
-        private const long DefaultMaxConfigurationFileSizeInMB = 10;
+        private const int DefaultMaxConfigurationFileSizeInMB = 10;
 
         /// <summary>
         /// Environment variable name to override the maximum configuration file size.
@@ -33,14 +33,14 @@ namespace Microsoft.Oryx.Detector
         /// Gets the maximum allowed file size for configuration files in MB.
         /// Can be overridden via ORYX_MAX_CONFIG_FILE_SIZE_MB environment variable.
         /// </summary>
-        private static long MaxConfigurationFileSizeInMB
+        private static int MaxConfigurationFileSizeInMB
         {
             get
             {
                 var envValue = Environment.GetEnvironmentVariable(MaxFileSizeEnvironmentVariable);
                 if (!string.IsNullOrEmpty(envValue) && double.TryParse(envValue, out var customSize) && customSize > 0)
                 {
-                    return (long)Math.Ceiling(customSize);
+                    return (int)Math.Ceiling(customSize);
                 }
 
                 return DefaultMaxConfigurationFileSizeInMB;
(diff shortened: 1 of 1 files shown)


src/Detector/ParserHelper.cs
```diff
@@ -19,6 +19,11 @@ namespace Microsoft.Oryx.Detector
     /// </summary>
     internal static class ParserHelper
     {
+        /// <summary>
+        /// Maximum allowed file size for configuration files (10 MB).
+        /// </summary>
+        private const long MaxConfigurationFileSizeInBytes = 10 * 1024 * 1024; // 10 MB
```
vijaysaayi: Could we keep this in MB ? anyway we divide by (1024 * 1024)

src/Detector/ParserHelper.cs
```diff
@@ -82,8 +112,32 @@ public static YamlNode ParseYamlFile(ISourceRepo sourceRepo, string filePath)
...
+
+            if (fileSize.HasValue)
+            {
+                var fileSizeMB = fileSize.Value / (1024.0 * 1024.0);
+
+                if (fileSizeMB > MaxConfigurationFileSizeInMB)
+                {
+                    throw new InvalidOperationException(
+                        $"Configuration file '{filePath}' is too large ({fileSizeMB:F2} MB). " +
+                        $"The maximum allowed size is {MaxConfigurationFileSizeInMB} MB.");
```
prathambande: Should we add the suggestion in error "if it is expected then set the env var "ORYX_MAX_CONFIG_FILE_SIZE_MB" to preferred val" or along those line?

src/Detector/ParserHelper.cs
```diff
@@ -82,8 +112,32 @@ public static YamlNode ParseYamlFile(ISourceRepo sourceRepo, string filePath)
...
             return JObject.Parse(jsonContent);
         }
+
+        /// <summary>
+        /// Validates that a file size does not exceed the maximum allowed size before reading it into memory.
+        /// </summary>
+        /// <param name="sourceRepo">Source repo for the application.</param>
+        /// <param name="filePath">The path to the file to validate.</param>
+        /// <exception cref="InvalidOperationException">Thrown when file size exceeds the maximum allowed size.</exception>
+        private static void ValidateFileSizeBeforeReading(ISourceRepo sourceRepo, string filePath)
```
vijaysaayi: nit: We can call the method as ValidateFileSizeOrThrow()

src/Detector/ParserHelper.cs
```diff
@@ -19,6 +19,34 @@ namespace Microsoft.Oryx.Detector
...
+        /// <summary>
+        /// Environment variable name to override the maximum configuration file size.
+        /// </summary>
+        private const string MaxFileSizeEnvironmentVariable = "ORYX_MAX_CONFIG_FILE_SIZE_MB";
+
+        /// <summary>
+        /// Gets the maximum allowed file size for configuration files in MB.
+        /// Can be overridden via ORYX_MAX_CONFIG_FILE_SIZE_MB environment variable.
+        /// </summary>
+        private static long MaxConfigurationFileSizeInMB
```
vijaysaayi: instead of long could we use int? Do we forsee this value getting very large?

PR #2755, opened 2026-01-14
Implement automatic fallback from zstd to gzip

Saipriya-1144:
This PR adds error handling and automatic fallback mechanism for tar compression in oryx build. When zstd compression fails, the build will now automatically fall back to gzip compression instead of failing completely.

testing:
Created a Kudu Image which doesn't have zstd installed to trigger compression failure
image
 compression with zstd failed, it switched to gzip and compression is successful with gzip

### DIFF
--- a/src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
+++ b/src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
@@ -176,7 +176,6 @@ then
 			else
 				echo "WARNING: Direct compression with zstd failed: $output, exit code: $compressionExitCode"
 				echo "Falling back to gzip compression."
-				BASE_START_TIME=$SECONDS
 			fi
 		fi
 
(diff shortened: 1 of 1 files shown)


src/BuildScriptGenerator/BaseBashBuildScript.sh.tpl
```diff
@@ -160,15 +160,30 @@ then
...
+				COMPRESSION_DONE=true
+			else
+				echo "WARNING: Direct compression with zstd failed: $output, exit code: $compressionExitCode"
+				echo "Falling back to gzip compression."
+				BASE_START_TIME=$SECONDS
+			fi
+		fi
+
+		if [ "$COMPRESSION_DONE" = "false" ]; then
+			if [ -f "$DESTINATION_DIR/output.tar.zst" ]; then
```
kumaramit-msft, line 183: Shall me move it here - BASE_START_TIME=$SECONDS

PR #2777, opened 2026-01-21
Update hugo version

kumaraksh1:

Update hugo version that is compatible with bullseye, bookworm and noble.

Testing
image

### DIFF
--- a/tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
+++ b/tests/Oryx.BuildImage.Tests/Python/PythonSampleAppsTest.cs
@@ -1632,7 +1632,7 @@ namespace Microsoft.Oryx.BuildImage.Tests
                 () =>
                 {
                     Assert.True(result.IsSuccess);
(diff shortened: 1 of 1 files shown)


build/__hugoConstants.sh
```diff
@@ -1,6 +1,6 @@
 # This file was auto-generated from 'constants.yaml'. Changes may be overridden.
 
-VERSION='0.152.2'
+VERSION='0.121.2'
```
Saipriya-1144, line 3: How will it only be applicable to bullseye

kumaraksh1: It will be applicable to all and we will generate all images but only bullseye prod image will be retagged.

kumaramit-msft: Its ok for now, but we will have to put a working solution. Lets get the images and revert this code

kumaraksh1: this is also working @kumaramit-msft 
Or can you elaborate more on this ?

PR #2808, opened 2026-01-29
changes to copy requirements.txt file

CodingIsBliss:
Copy the requirements.txt file to the dest file. Will be used later on  to know what all libraries are being used by the site.

Tested the same by running the oryx build command. -

Screenshot 2026-01-28 193515

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -565,6 +565,18 @@ else
     rm "$COMMAND_MANIFEST_FILE"
 fi
 
+# Copy requirements.txt to the destination directory if it exists
+if [ "$SOURCE_DIR" != "$DESTINATION_DIR" ]
+then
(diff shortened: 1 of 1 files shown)


src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -565,6 +565,18 @@ else
     rm "$COMMAND_MANIFEST_FILE"
 fi
 
+# Copy requirements.txt to the destination directory if it exists
+if [ "$SOURCE_DIR" != "$DESTINATION_DIR" ]
+then
```
prathambande, line 570: should we also add a delete req.txt first? so that we do not have stale req.txt if cx shifts to other mode of package installation?

surenderssm (approved):

src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
SarathBussa: Makes sense. Lets take it in next iteration.

PR #2817, opened 2026-02-04
Copy requirements.txt file to the destination folder.

CodingIsBliss:
Copy the requirements.txt file to the dest file. Will be used later on to know what all libraries are being used by the site.

image

### DIFF
--- a/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
@@ -572,7 +572,7 @@ then
     then
         echo
         echo "Copying '$REQUIREMENTS_TXT_FILE' to destination directory..."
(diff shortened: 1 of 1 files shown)


src/BuildScriptGenerator/Python/PythonBashBuildSnippet.sh.tpl
```diff
@@ -565,6 +565,18 @@ else
...
 fi
 
+# Copy requirements.txt to the destination directory if it exists
+if [ "$SOURCE_DIR" != "$DESTINATION_DIR" ]
+then
+    if [ -e "$SOURCE_DIR/$REQUIREMENTS_TXT_FILE" ]
+    then
+        echo
+        echo "Copying '$REQUIREMENTS_TXT_FILE' to destination directory..."
+        cp "$SOURCE_DIR/$REQUIREMENTS_TXT_FILE" "$DESTINATION_DIR/requirements.txt"
```
Saipriya-1144: Can we write || true so that deployment won't fail even if copy step fails

PR #2837, opened 2026-04-01
ACR-based SDK distribution

CodingIsBliss:
## Summary

**Multiple ways to pull SDKs from different resources based on environment variables:**

1. ORYX_ENABLE_EXTERNAL_ACR_SDK_PROVIDER
     Oryx -> Socket -> ACR

2. ORYX_ENABLE_EXTERNAL_SDK_PROVIDER (Existing flow for App Services)
   Oryx -> Socket -> Storage Account

3. ORYX_ENABLE_ACR_SDK_PROVIDER
   Oryx -> ACR
4. Oryx -> Storage Account

### **Priority and Fallback order**
1. Oryx -> Socket -> ACR
2. Oryx -> Socket -> Storage Account 
3. Oryx -> ACR
4. Oryx -> Storage Account

### **Review Guide**

Start at DefaultPlatformsInformationProvider.cs → GetPlatformsInfo(). This is the entry point that detects platforms and logs SDK provider status (new). It calls platform.Detect() for each platform.

Follow the flow for each platform:

Python — PythonPlatform.Detect() → ResolveVersions() → GetInstallerScriptSnippet()
PHP — PhpPlatform → InstallPhp() / InstallPhpComposer() (refactored with ACR fallback chain)
Node — NodePlatform → same pattern as Python
DotNet — DotnetCorePlatform → has extra SDK↔Runtime version mapping via compound ACR tags
Key new shared components:

Key new shared components:

AcrSdkProvider.cs — Direct OCI API client (downloads SDK tarballs from ACR)
OciRegistryClient.cs — HTTP client for OCI Distribution API
ExternalAcrSdkProvider.cs / ExternalAcrVersionProviderBase.cs — Socket-based ACR path
Modified existing files:

PlatformInstallerBase.cs → skipSdkBinaryDownload block added for ACR tarball lookup + extraction

### **Some other points:**
1. PHP uses a different pattern than Python/Node/DotNet
Python, Node, DotNet: GetInstallerScriptSnippet returns a string — each TryInstall* returns the script or null
PHP: Uses StringBuilder — each TryInstall* appends directly to the builder and returns bool. This is because PHP installs two things (PHP + Composer) in a single GetInstallerScriptSnippet call
2. DotNet ACR tags use compound format
All other platforms: {osFlavor}-{version} (e.g., noble-3.14.3)
DotNet: {osFlavor}-{sdkVersion}_{runtimeVersion} (e.g., noble-10.0.201_10.0.5) because .NET requires runtime→SDK version mapping. DotNetCoreAcrVersionProvider parses this; AcrSdkProvider passes runtimeVersion as an optional parameter to construct the correct tag
3. OCI layer unwrapping done in C# (AcrSdkProvider), not bash (PlatformInstallerBase)
FROM scratch; COPY sdk.tar.gz / produces a layer that wraps the tarball inside a tar
AcrSdkProvider.ExtractFileFromTar extracts the inner SDK tarball using System.Formats.Tar.TarReader — so by the time PlatformInstallerBase runs, the file at dynamicPath is already the raw SDK tarball, same as what CDN/blob providers produce
4. Four SDK sources with independent fallback
Each provider is independent — if one fails, the next is tried without any state leaking
Order: External-ACR (socket→host→ACR) → External-SDK (socket→host→blob) → Direct-ACR (OCI API) → CDN
The first two are controlled by the App Service platform; the last two are Oryx-native
5. Token auth for anonymous ACR
Even "anonymous pull" ACR registries require a bearer token exchange — OciRegistryClient handles this transparently with per-scope token caching

**Testing**

With ACR Provider enabled (direct ACR pull from Oryx)

* Dotnet
image

* Python
image

* Php
image

* Nodejs
image

With External ACR Provider enabled (pulled via socket communication)

* Php
image

* Python
image

* Nodejs
image

* Dotnet
image

### DIFF
--- a/src/BuildScriptGenerator/DefaultPlatformsInformationProvider.cs
+++ b/src/BuildScriptGenerator/DefaultPlatformsInformationProvider.cs
@@ -112,7 +112,7 @@ namespace Microsoft.Oryx.BuildScriptGenerator
                 return false;
             }
 
-            if (this.commonOptions.EnableExternalAcrSdkProvider && platform.Name != this.commonOptions.PlatformName)
+            if (this.commonOptions.EnableExternalAcrSdkProvider && !string.IsNullOrEmpty(this.commonOptions.PlatformName) && platform.Name != this.commonOptions.PlatformName)
             {
                 this.logger.LogDebug(
                     "Skipping detection for platform '{PlatformName}' because External ACR SDK provider is enabled and only user provided platform '{UserPlatform}' is considered.",


src/BuildScriptGenerator.Common/SdkStorageConstants.cs
```diff
@@ -21,5 +21,12 @@ public static class SdkStorageConstants
         public const string DotnetRuntimeVersionMetadataName = "Dotnet_runtime_version";
         public const string LegacyDotnetRuntimeVersionMetadataName = "Runtime_version";
         public const string OsTypeMetadataName = "Os_type";
+
+        // ACR-based SDK distribution constants
+        public const string EnableExternalAcrSdkProviderKey = "ORYX_ENABLE_EXTERNAL_ACR_SDK_PROVIDER";
+        public const string EnableAcrSdkProviderKey = "ORYX_ENABLE_ACR_SDK_PROVIDER";
+        public const string AcrSdkRegistryUrlKeyName = "ORYX_ACR_SDK_REGISTRY_URL";
+        public const string DefaultAcrSdkRegistryUrl = "https://oryxacr.azurecr.io";
```
sarsharma: before merging, lets update this to mcr if required

src/BuildScriptGenerator.Common/SdkStorageConstants.cs
```diff
@@ -21,5 +21,12 @@ public static class SdkStorageConstants
...
+        public const string AcrSdkRepositoryPrefix = "sdks";
```
sarsharma: same here, for prod case + mcr, this might become `<mcr>/oryx/sdks`

src/BuildScriptGenerator/Contracts/IExternalSdkProvider.cs
```diff
@@ -17,7 +17,7 @@ public interface IExternalSdkProvider
     /// <summary>
     /// The directory where SDKs are cached by the external provider.
     /// </summary>
-    public const string ExternalSdksStorageDir = "/var/OryxSdksCache";
+    public const string ExternalSdksStorageDir = "/var/OryxSdks";
```
kumaraksh1, line 20: **Crew:**
Why do we even need to change this?

src/BuildScriptGenerator/Helpers/OciContainerConfig.cs
```diff
@@ -0,0 +1,16 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    public class OciContainerConfig
```
kumaraksh1: **Crew:**
Are we using it anywhere ?

src/BuildScriptGenerator/Helpers/OciImageConfig.cs
```diff
@@ -0,0 +1,15 @@
+// --------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+// --------------------------------------------------------------------------------------------
+
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    public class OciImageConfig
```
kumaraksh1: **Crew:**
seems like this is also not used.

src/BuildScriptGenerator/Php/VersionProviders/PhpComposerExternalAcrVersionProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            var availableVersions = this.GetAvailableSdkVersions(platformName: "php-composer", debianFlavor: this.DebianFlavor);
+            if (availableVersions == null || availableVersions.Count == 0)
+            {
+                return null;
+            }
+
+            return PlatformVersionInfo.CreateOnDiskVersionInfo(
+                supportedVersions: availableVersions.ToArray(),
+                defaultVersion: availableVersions[0]);
```
kumaraksh1: **Crew:**
Look at this @sarsharma

src/BuildScriptGenerator/ExternalAcrVersionProviderBase.cs
```diff
@@ -0,0 +1,198 @@
...
+                using (var cts = new CancellationTokenSource(
+                    TimeSpan.FromSeconds(MaxTimeoutForSocketOperationInSeconds)))
+                {
+                    await socket.ConnectAsync(new UnixDomainSocketEndPoint(SocketPath), cts.Token);
+                    var requestJson = JsonSerializer.Serialize(
+                        new { Action = action, PlatformName = platformName, DebianFlavor = debianFlavor }) + "$";
+                    var requestBytes = Encoding.UTF8.GetBytes(requestJson);
+
+                    await socket.SendAsync(new ArraySegment<byte>(requestBytes), SocketFlags.None, cts.Token);
+                    var buffer = new byte[4096];
```
kumaraksh1: **Crew:**

Lets re-think on this size

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -146,13 +146,30 @@ protected string GetInstallerScriptSnippet(
 
             if (skipSdkBinaryDownload)
                 {
```
kumaraksh1, line 148: **Crew:**

Need to relook at this.

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreVersionProvider.cs
```diff
@@ -105,5 +71,113 @@ public Dictionary<string, string> GetSupportedVersions()
...
+                }
+            }
+
+            this.outputWriter.WriteLine("DotNet version resolved using blob SDK storage provider(CDN).");
+            return this.sdkStorageVersionProvider.GetDefaultRuntimeVersion();
+        }
+
+        private Dictionary<string, string> ResolveDynamicSupportedVersions()
+        {
+            if (this.cliOptions.EnableExternalSdkProvider)
```
kumaraksh1, line 107: **Crew:**
Are we not trying ExternalAcr option? @sarsharma

src/BuildScriptGenerator/Helpers/OciContainerConfig.cs
kumaraksh1: we can delete it.

src/BuildScriptGenerator/Helpers/OciImageConfig.cs
kumaraksh1: we can delete it

src/BuildScriptGenerator/Php/VersionProviders/PhpComposerExternalAcrVersionProvider.cs
kumaraksh1: use ComposerDefaultVersion instead of availableVersions[0]

src/BuildScriptGenerator/PlatformInstallerBase.cs
```diff
@@ -146,13 +146,26 @@ protected string GetInstallerScriptSnippet(
...
+                    // dynamic install dir (used by AcrSdkProvider).
                     var tarFileName = BlobNameHelper.GetBlobNameForVersion(platformName, version, this.CommonOptions.DebianFlavor);
-                    var tarFilePath = Path.Combine(ExternalSdkProvider.ExternalSdksStorageDir, platformName, tarFileName);
-                    snippet.AppendLine($"echo \"Skipping download of {platformName} version {version} as it is available in external sdk provider cache...\"")
-                        .AppendLine($"echo \"Extracting contents...\"")
-                        .AppendLine($"tar -xzf {tarFilePath} -C .")
-                        .AppendLine($"rm -f {tarFileName}")
-                        .AppendLine($"echo \"Successfully extracted {platformName} version {version} from external sdk provider cache...\"");
+                    var externalPath = Path.Combine(ExternalSdkProvider.ExternalSdksStorageDir, platformName, tarFileName);
+                    var dynamicPath = Path.Combine(this.CommonOptions.DynamicInstallRootDir, platformName, tarFileName);
```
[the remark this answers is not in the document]
kumaraksh1: Old comment

src/BuildScriptGenerator/Contracts/IExternalSdkProvider.cs
prathambande: ++, Wouldn't this make rolling out tightly coupled with changes to the external provider

src/BuildScriptGeneratorCli/SettingsKeys.cs
```diff
@@ -75,5 +75,9 @@ public static class SettingsKeys
         public const string DebianFlavor = "DEBIAN_FLAVOR";
         public const string CallerId = "CALLER_ID";
         public const string OryxDisablePipUpgrade = "ORYX_DISABLE_PIP_UPGRADE";
+        public const string EnableExternalAcrSdkProvider = "ORYX_ENABLE_EXTERNAL_ACR_SDK_PROVIDER";
```
surenderssm, line 80: How about changing following for consistency and easy to manage/recall.
`ORYX_ENABLE_EXTERNAL_ACR_SDK_PROVIDER` to `ORYX_ENABLE_ACR_SDK_PROVIDER_EXTERNAL`

src/BuildScriptGenerator.Common/SdkStorageConstants.cs
```diff
@@ -21,5 +21,24 @@ public static class SdkStorageConstants
...
         public const string OsTypeMetadataName = "Os_type";
+
+        // OCI image based SDK distribution constants
+        public const string EnableExternalAcrSdkProviderKey = "ORYX_ENABLE_EXTERNAL_ACR_SDK_PROVIDER";
+        public const string EnableAcrSdkProviderKey = "ORYX_ENABLE_ACR_SDK_PROVIDER";
+        public const string AcrSdkRegistryUrlKeyName = "ORYX_ACR_SDK_REGISTRY_URL";
+        public const string DefaultAcrSdkRegistryUrl = "https://mcr.microsoft.com";
+        public const string AcrSdkRepositoryPrefixKeyName = "ORYX_ACR_SDK_REPOSITORY_PREFIX";
+        public const string DefaultAcrSdkRepositoryPrefix = "oryx";
+
```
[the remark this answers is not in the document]
surenderssm: We should refer to MCR everywhere, let's remove the reference to anything internal

src/BuildScriptGenerator/Helpers/OciRegistryClient.cs
```diff
@@ -0,0 +1,319 @@
...
+                    var json = await response.Content.ReadAsStringAsync();
+                    var tagList = JsonSerializer.Deserialize<OciTagList>(json);
+                    if (tagList?.Tags != null)
+                    {
+                        allTags.AddRange(tagList.Tags);
+                    }
+
+                    // Handle OCI pagination via Link header (RFC 5988)
+                    url = null;
+                    if (response.Headers.TryGetValues("Link", out var linkValues))
```
surenderssm, line 95: Link: rel="next" header URL is followed without host validation. A compromised registry could redirect requests (with auth tokens!) to an attacker-controlled domain. Validate that absolute URLs match this.registryHost.

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreExternalAcrVersionProvider.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        /// <summary>
+        /// Gets the single SDK version dictated by the external host, or null if unavailable.
+        /// </summary>
+        public string GetSdkVersion()
+        {
+            if (!this.resolved)
+            {
+                this.resolvedVersion = this.GetCompanionSdkVersion(DotNetCoreConstants.PlatformName, debianFlavor: this.DebianFlavor);
```
prathambande: null handling ?

src/BuildScriptGenerator/ExternalAcrVersionProviderBase.cs
```diff
@@ -0,0 +1,198 @@
...
+                    var received = await socket.ReceiveAsync(
```
prathambande: @copilot -- only explain, if a single ReceiveAsync guarantees entire message at once with $ end?

images/constants.yml
sarsharma, line 1: let's revert

sarsharma: done

src/BuildScriptGenerator.Common/SdkStorageConstants.cs
```diff
@@ -21,5 +21,20 @@ public static class SdkStorageConstants
         public const string DotnetRuntimeVersionMetadataName = "Dotnet_runtime_version";
         public const string LegacyDotnetRuntimeVersionMetadataName = "Runtime_version";
         public const string OsTypeMetadataName = "Os_type";
+
+        // OCI image based SDK distribution constants
+        public const string DefaultAcrSdkRegistryUrl = "https://mcr.microsoft.com";
+        public const string DefaultAcrSdkRepositoryPrefix = "oryx";
+
```
SarathBussa: can we have a different file for images?

src/BuildScriptGenerator/Contracts/IExternalSdkProvider.cs
```diff
@@ -17,7 +17,7 @@ public interface IExternalSdkProvider
     /// <summary>
     /// The directory where SDKs are cached by the external provider.
     /// </summary>
-    public const string ExternalSdksStorageDir = "/var/OryxSdksCache";
```
SarathBussa, line 20: this is still the folder where lwasv2 puts the sdks right?

SarathBussa: in the old way

sarsharma: no, the actual folder in use is in ExternalSdkProvider `/var/OryxSdks`, this is just interface default which was set to wrong value, so just updating it for consistency

src/BuildScriptGenerator/AcrSdkProvider.cs
```diff
@@ -0,0 +1,212 @@
...
+                return false;
+            }
+        }
+
+        /// <summary>
+        /// Extracts the expected SDK .tar.gz file from an OCI layer tar archive.
+        /// OCI layers use media type "application/vnd.docker.image.rootfs.diff.tar.gzip",
+        /// so the blob must be decompressed before reading tar entries.
+        /// </summary>
+        private void ExtractFileFromTar(string layerPath, string outputPath, string expectedFileName)
```
siriande, line 191: **non-blocking:** This has no locking on the output path, so concurrent calls targeting the same file could race. Not an issue today (all callers run sequentially) but worth noting if SDK fetches are ever parallelized.

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreAcrVersionProvider.cs
```diff
@@ -0,0 +1,120 @@
...
+            : base(commonOptions, ociClient, loggerFactory)
+        {
+            this.commonOptions = commonOptions.Value;
+            this.logger = loggerFactory.CreateLogger<DotNetCoreAcrVersionProvider>();
+        }
+
+        public string GetDefaultRuntimeVersion()
+        {
+            this.EnsureVersionInfo();
+            if (this.versionMap == null || this.versionMap.Count == 0)
```
siriande, line 47: this check needed?

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreSdkStorageVersionProvider.cs
```diff
@@ -70,7 +70,17 @@ public void GetVersionInfo()
                 {
                     this.logger.LogWarning(ex, "Failed to get list of blobs from primary storage. Trying backup storage.");
                     var sdkStorageBackupBaseUrl = this.GetPlatformBinariesBackupStorageBaseUrl();
-                    xdoc = ListBlobsHelper.GetAllBlobs(sdkStorageBackupBaseUrl, DotNetCoreConstants.PlatformName, httpClient);
+                    if (sdkStorageBackupBaseUrl != null)
```
Shi1810, line 73: shall we check with string.IsNullOrEmpty() ?

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreSdkStorageVersionProvider.cs
```diff
@@ -70,7 +70,17 @@ public void GetVersionInfo()
...
+                    {
+                        xdoc = ListBlobsHelper.GetAllBlobs(sdkStorageBackupBaseUrl, DotNetCoreConstants.PlatformName, httpClient);
```
Shi1810, line 75: what happens if GetAllBlobs fails and xdoc is still null? I think we should add a try catch here

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreSdkStorageVersionProvider.cs
Shi1810, line 98: xdoc.XPathSelectElements will fail when xdoc is null

src/BuildScriptGenerator/Helpers/OciDescriptor.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Microsoft.Oryx.BuildScriptGenerator
+{
+    public class OciDescriptor
+    {
+        [JsonPropertyName("mediaType")]
+        public string MediaType { get; set; }
+
+        [JsonPropertyName("digest")]
+        public string Digest { get; set; }
```
siriande, line 16: Shall we add OCI digest format validator here ([ref](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests))?

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreAcrVersionProvider.cs
```diff
@@ -0,0 +1,120 @@
...
+                {
+                    continue;
+                }
+
+                // Strip the osFlavor prefix to get "{sdkVersion}_{runtimeVersion}"
+                var versionPart = tag.Substring(prefix.Length);
+                var parts = versionPart.Split('_', 2);
+                var sdkVersion = parts[0];
+                var runtimeVersion = parts.Length > 1 ? parts[1] : null;
+                if (parts.Length != 2 || string.IsNullOrEmpty(sdkVersion) || string.IsNullOrEmpty(runtimeVersion))
```
siriande, line 104: `NullOrWhiteSpace`?

src/BuildScriptGenerator/Helpers/OciRegistryClient.cs
```diff
@@ -0,0 +1,340 @@
...
+
+        /// <summary>
+        /// Gets the manifest digest for a tag via a HEAD request.
+        /// Returns the <c>Docker-Content-Digest</c> header value (e.g. <c>sha256:abc...</c>),
+        /// or null if unavailable.
+        /// </summary>
+        public async Task<string> GetManifestDigestAsync(string repository, string tag)
+        {
+            var url = $"{this.registryUrl}/v2/{repository}/manifests/{tag}";
+            using (var request = await this.CreateAuthenticatedRequestAsync(HttpMethod.Head, url, repository))
```
Shi1810, line 168: I feel we should have some type of exception handling here as well

src/BuildScriptGenerator.Common/SdkStorageConstants.cs
kumaraksh1: I think its file, this file contains constants related SDKs so images or storage don't matter.

src/BuildScriptGenerator.Common/SdkStorageConstants.cs
```diff
@@ -21,5 +21,20 @@ public static class SdkStorageConstants
...
+        /// <summary>
+        /// Maps a platform name to its OCI SDK image repository path.
+        /// e.g. "nodejs" → "oryx/nodejs-sdk", "php" → "oryx/php-sdk".
+        /// Final image ref: mcr.microsoft.com/oryx/nodejs-sdk:bookworm-20.20.2
+        /// </summary>
+        public static string GetSdkImageRepository(string platformName, string prefix = null)
```
SarathBussa: also we could find some other place for the function.. helper class will be better

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreSdkStorageVersionProvider.cs
kumaraksh1: GetPlatformBinariesBackupStorageBaseUrl()

in this method itself, its handeled.. if the url is null or empty, it will return null.

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreSdkStorageVersionProvider.cs
kumaraksh1: If GetAllBlobs fails then an exception will be thrown

src/BuildScriptGenerator.Common/SdkStorageConstants.cs
```diff
@@ -21,5 +21,20 @@ public static class SdkStorageConstants
...
+        {
+            prefix = string.IsNullOrEmpty(prefix) ? DefaultAcrSdkRepositoryPrefix : prefix;
+            return $"{prefix}/{platformName}-sdk";
```
SarathBussa: this will not work for dotnet and node... for node the platform is nodejs and image is node [runtime too]

for dotnet platform is dotnet and image is dotnetcore

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreSdkStorageVersionProvider.cs
kumaraksh1: all these things should be handled in GetPlatformMetaDataAsync() method who is providing xdoc

src/BuildScriptGenerator/AcrSdkProvider.cs
kumaraksh1: Sure @siriande !!
Resolving it for now!

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreAcrVersionProvider.cs
kumaraksh1: This will never happen @siriande 
Tag can not have empty spaces value.

src/BuildScriptGenerator/Helpers/OciRegistryClient.cs
```diff
@@ -0,0 +1,340 @@
...
+        /// </summary>
+        public async Task<List<string>> GetAllTagsAsync(string repository)
+        {
+            var allTags = new List<string>();
+            var url = $"{this.registryUrl}/v2/{repository}/tags/list";
+
+            while (!string.IsNullOrEmpty(url))
+            {
+                this.logger.LogDebug("Fetching tags from {url}", url);
+                using (var request = await this.CreateAuthenticatedRequestAsync(HttpMethod.Get, url, repository))
```
SarathBussa, line 77: Generally you should create an authenticated request if the request fails with 401, as the specs tells the same - 

**Commenting to test with mcr.microsoft.io before merging it that works -**

Image

src/BuildScriptGenerator.Common/SdkStorageConstants.cs
sarsharma: discussed offline, resolving

src/BuildScriptGenerator/Helpers/OciRegistryClient.cs
sarsharma: we are excluding auth header for mcr.microsoft* prefix registries

src/BuildScriptGenerator/Python/PythonConstants.cs
```diff
@@ -20,5 +22,18 @@ public static class PythonConstants
...
         public const string CondaExecutablePath = "/opt/conda/condabin/conda";
+
+        /// <summary>
+        /// Default Python major.minor version per OS flavor, matching platforms/python/versions/*/defaultVersion.txt.
+        /// </summary>
+        public static readonly Dictionary<string, string> DefaultVersionPerFlavor = new Dictionary<string, string>
+        {
+            { "bookworm", "3.8" },
+            { "bullseye", "3.8" },
+            { "buster", "3.8" },
```
prathambande, line 33: is 3.8 for all intentional? 
Or should we be using latest per flavor

src/BuildScriptGenerator/Helpers/OciRegistryClient.cs
kumaraksh1: It should not fail silently and caller should decide what need to be done.
But thank yuo :)

src/BuildScriptGenerator/Python/PythonConstants.cs
kumaraksh1: Kept is same as there in defaultVersion.txt !!

sarsharma: this is just following the values we have in prod today from defaultversion.txt
we can choose to update in a follow up

src/BuildScriptGenerator/ExternalAcrSdkProvider.cs
```diff
@@ -0,0 +1,182 @@
...
+        public ExternalAcrSdkProvider(
+            IStandardOutputWriter outputWriter,
+            ILogger<ExternalAcrSdkProvider> logger)
+        {
+            this.logger = logger;
+            this.outputWriter = outputWriter;
+        }
+
+        /// <inheritdoc/>
+        /// Update this Method to pull SDK tarball from ACR via the external provider and place it in the local cache.
```
Saipriya-1144: nit: can we remove this comment as the method is updated already

src/BuildScriptGenerator.Common/SdkStorageConstants.cs
kumaraksh1: Done!!

src/BuildScriptGenerator/ExternalAcrSdkProvider.cs
kumaraksh1: Done, thanks !!

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreAcrVersionProvider.cs
sarsharma: just a defensive check so ensure map is populated

src/BuildScriptGenerator/Helpers/OciDescriptor.cs
sarsharma: I don't think it's necessary here. OciDescriptor is a plain DTO for JSON deserialization — the digest comes directly from the registry's own response. Downstream, DownloadLayerBlobAsync (line 208-210) already parses the sha256: prefix and does a full content hash verification

src/BuildScriptGeneratorCli/SettingsKeys.cs
kumaraksh1: discussed offline

src/BuildScriptGenerator/DotNetCore/VersionProviders/DotNetCoreAcrVersionProvider.cs
sarsharma: added comment explaining the rationale

PR #2887, opened 2026-04-06
Decouple runtime image generation from pre-built SDKs

Saipriya-1144:
Decouple runtime image builds from SDK publishing pipeline so runtime images can be generated independently.

Python: Introduced a multi-stage pythonSdkBuilder intermediate stage that builds Python from source using build.sh (the same script used by the SDK pipeline). The final image receives only the compiled /opt/python/<version> directory, eliminating storage account dependency and keeping the final image clean. Applied to both template.Dockerfile and noble.Dockerfile

Node: Introduced a multi-stage nodeDownloader intermediate stage that installs Node via n version manager. The final image receives only the Node binaries

Testing:
image
20260427.2 image is created with n, 20260427.3 is created with nvm


images/runtime/python/template.Dockerfile
```diff
@@ -27,15 +27,30 @@ ENV BUILD_NUMBER=${BUILD_NUMBER}
...
+
 FROM ${BASE_IMAGE} as main
 
 ARG IMAGES_DIR=/tmp/oryx/images
-ARG BUILD_DIR=/tmp/oryx/build
-ARG SDK_STORAGE_BASE_URL_VALUE
 
 ARG DEBIAN_FLAVOR
 ENV DEBIAN_FLAVOR=${DEBIAN_FLAVOR}
-ENV ORYX_SDK_STORAGE_BASE_URL=${SDK_STORAGE_BASE_URL_VALUE}
```
sarsharma, line 27: any reason for removing this env var?

images/runtime/node/18/bullseye.Dockerfile
```diff
@@ -14,6 +14,14 @@ ENV GIT_COMMIT=${GIT_COMMIT}
 ENV BUILD_NUMBER=${BUILD_NUMBER}
 RUN chmod +x build.sh && ./build.sh node /opt/startupcmdgen/startupcmdgen
 
+# Download Node.js via n in a disposable stage
+FROM ${BASE_IMAGE} AS nodeDownloader
+ARG NODE18_VERSION
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends curl git make ca-certificates \
+    && curl -sL https://git.io/n-install | bash -s -- -ny - \
```
sarsharma: in the long term we should ideally move to using nvm
if we are using n-install, we can maybe use the raw url pointing to a particular commit for the script
git.io is already deprecated
we can use this instead
`https://raw.githubusercontent.com/mklement0/n-install/0cd24ac9d350f30d5ae1049f020e61818ece03b6/bin/n-install`

sarsharma (approved):

sarsharma (approved):

Shi1810 (approved):

PR #2907, opened 2026-04-28
Use official nginx.org instead of third-party sources

Saipriya-1144:
Migrate PHP-FPM nginx from sury.org/ondrej PPA to official nginx.org repository (https://nginx.org/en/linux_packages.html)

**Changes**

1. Nginx source migration

 - Replaced sury.org APT source with the official nginx.org signed repository
 - Pinned nginx to version 1.30.0 for reproducible builds and also added it to apt-hold list, so it doesn't get auto upgraded
 - Removed nginx-core, nginx-common, nginx-full — the nginx.org package is self-contained

image

The nginx package from the official nginx.org repo is a unified package that already includes everything nginx-common, nginx-core. These split packages only exist in Debian/Ubuntu's default apt repos and conflict with the nginx.org version, so they cannot be installed separately.

2. Fixed silent nginx config failures (critical)

 - Site config path: nginx.org includes conf.d/*.conf, not sites-available/ or sites-enabled/. Previous COPY to sites-enabled/ was silently ignored — PHP-FPM proxy config was never loaded. Now copies to /etc/nginx/conf.d/default.conf
 - worker_connections: nginx.org defaults to 1024 (not 768). Previous sed matched 768, silently no-op'd. Fixed to match 1024 → 10068
 - multi_accept: nginx.org config has no commented # multi_accept on line (Debian-only). Changed to append multi_accept on; after the worker_connections line
 - User directive: Explicitly set user www-data; and fixed temp directory ownership (/var/cache/nginx)
 - Additional tuning: Uncommented tcp_nopush on, gzip on, added types_hash_max_size 2048
 - Verification guards: Added grep -q assertions after each sed to fail the build if any replacement silently misses

3. Pinned msodbcsql17 to prevent interactive EULA prompt

 - Pinned msodbcsql17=17.10.6.1-1 during install
 - Added msodbcsql17 to apt-mark hold list so apt-get upgrade doesn't trigger an interactive EULA acceptance that breaks non-interactive Docker builds


prathambande (approved):

images/runtime/php-fpm/8.1/bullseye.Dockerfile
```diff
@@ -277,13 +277,13 @@ ENV PHP_ORIGIN php-fpm
...
-RUN curl -sSL https://packages.sury.org/nginx/README.txt | bash -x
-RUN apt-get update
-RUN yes '' | apt-get install nginx-core nginx-common nginx nginx-full -y
+# Install NGINX stable from official nginx.org repository
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends curl gnupg2 nano \
+    && curl -fsSL https://nginx.org/keys/nginx_signing.key | gpg --dearmor -o /usr/share/keyrings/nginx-archive-keyring.gpg \
+    && echo "deb [signed-by=/usr/share/keyrings/nginx-archive-keyring.gpg] http://nginx.org/packages/debian bullseye nginx" > /etc/apt/sources.list.d/nginx.list \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends nginx
```
sarsharma: `--no-install-recommends` will this omit any packages which we bake in today?

images/runtime/php-fpm/8.1/bullseye.Dockerfile
```diff
@@ -277,13 +277,13 @@ ENV PHP_ORIGIN php-fpm
...
-RUN apt-get update
-RUN apt install curl nano -y
-RUN curl -sSL https://packages.sury.org/nginx/README.txt | bash -x
-RUN apt-get update
-RUN yes '' | apt-get install nginx-core nginx-common nginx nginx-full -y
+# Install NGINX stable from official nginx.org repository
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends curl gnupg2 nano \
+    && curl -fsSL https://nginx.org/keys/nginx_signing.key | gpg --dearmor -o /usr/share/keyrings/nginx-archive-keyring.gpg \
+    && echo "deb [signed-by=/usr/share/keyrings/nginx-archive-keyring.gpg] http://nginx.org/packages/debian bullseye nginx" > /etc/apt/sources.list.d/nginx.list \
```
sarsharma: will this pull the stable version? also do we require pinning?
https://nginx.org/en/linux_packages.html

Image

Saipriya-1144: Do you mean pinning repo or version? I have pinned the nginx version.

images/runtime/php-fpm/8.1/bullseye.Dockerfile
```diff
@@ -277,13 +277,13 @@ ENV PHP_ORIGIN php-fpm
 ENV NGINX_RUN_USER www-data
 # Edit the default DocumentRoot setting
 ENV NGINX_DOCUMENT_ROOT /home/site/wwwroot
-# Install NGINX latest stable version using APT Method with Nginx Repository instead of distribution-provided one:
-# - https://www.linuxcapable.com/how-to-install-latest-nginx-mainline-or-stable-on-debian-11/
+# Install NGINX from official nginx.org repository
```
vijaysaayi, line 280: **Code Review (AI)**
⚠️ **Critical: nginx.org packages do NOT use `sites-available`/`sites-enabled`**
 
 The official nginx.org package ships an `nginx.conf` that includes `conf.d/*.conf` — it does **not** include `sites-enabled/*`. These COPY commands place the PHP-FPM proxy config in a directory that nginx will never read.
 
 `nginx -t` won't catch this because it only validates syntax of loaded configs — files in `sites-enabled/` are simply never loaded (no error, just silent omission). At runtime, nginx won't forward requests to PHP-FPM.
 
 **Suggested fix:**
 
 ```dockerfile
 RUN rm -f /etc/nginx/conf.d/default.conf
 COPY images/runtime/php-fpm/nginx_conf/default.conf /etc/nginx/conf.d/default.conf

Also update the sed commands — nginx.org defaults use worker_connections 1024 (not 768), so the current sed will silently no-op.

This applies to all 7 Dockerfiles in this PR. ```

images/runtime/php-fpm/8.1/bullseye.Dockerfile
vijaysaayi: **Code Review (AI)** 

⚠️ **`sed` patterns won't match nginx.org defaults**
 
 The nginx.org package ships different defaults than the Debian/PPA packages:
 
 | Setting | Debian/PPA default | nginx.org default |
 |---|---|---|
 | `worker_connections` | `768` | `1024` |
 | `# multi_accept on` | present | not present |
 
 `sed -i` doesn't error when a pattern doesn't match — it silently writes the file unchanged. So after this PR, `worker_connections` stays at 1024 (never bumped to 10068) and `multi_accept` remains off.
 
 **Suggested fix:** Supply a complete `nginx.conf` via COPY instead of relying on sed against upstream defaults, or update the patterns:
 
 ```dockerfile
 RUN sed -ri -e 's!worker_connections\s+[0-9]+!worker_connections 10068!g' /etc/nginx/nginx.conf

This applies to all 7 Dockerfiles in this PR. ```

sarsharma (approved):

images/runtime/php-fpm/8.1/bullseye.Dockerfile
Saipriya-1144: Fixed this bug, Thanks

sarsharma (approved):

images/runtime/php-fpm/8.1/bullseye.Dockerfile
```diff
@@ -31,7 +31,7 @@ RUN set -eux \
 	&& curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add - \
 	&& curl https://packages.microsoft.com/config/debian/11/prod.list > /etc/apt/sources.list.d/mssql-release.list \
 	&& apt-get update \
-	&& ACCEPT_EULA=Y apt-get install -y msodbcsql17 msodbcsql18=18.1.2.1-1 odbcinst1debian2=2.3.7 odbcinst=2.3.7 unixodbc=2.3.7 unixodbc-dev=2.3.7
+	&& ACCEPT_EULA=Y apt-get install -y msodbcsql17=17.10.6.1-1 msodbcsql18=18.1.2.1-1 odbcinst1debian2=2.3.7 odbcinst=2.3.7 unixodbc=2.3.7 unixodbc-dev=2.3.7
```
surenderssm, line 34: Is this pinning intentional `17.10.6.1-1`?

Saipriya-1144: Yes, it's intentional. The version is pinned just like the other ODBC packages and added to the apt-mark hold list so that apt-get upgrade doesn't trigger an interactive EULA acceptance prompt that would break non-interactive Docker builds.

images/runtime/php-fpm/8.1/bullseye.Dockerfile
```diff
@@ -277,26 +277,37 @@ ENV PHP_ORIGIN php-fpm
...
 RUN nginx -t
 # Edit the default port setting
 ENV NGINX_PORT 8080
 
 # Install common PHP extensions
 # TEMPORARY: Holding odbc related packages from upgrading.
-RUN apt-mark hold msodbcsql18 odbcinst1debian2 odbcinst unixodbc unixodbc-dev \
+RUN apt-mark hold msodbcsql17 msodbcsql18 odbcinst1debian2 odbcinst unixodbc unixodbc-dev \
     && apt-get update \
     && apt-get upgrade -y \
```
vijaysaayi, line 316: I see we are doing `apt-get upgrade -y` here.
Would this upgrade the pinned version of nginx here `RUN yes '' | apt-get install nginx=1.30.0-1~bullseye -y`

I feel we should add nginx to apt-mark hold list along with msodbc

images/runtime/php-fpm/8.2/bullseye.Dockerfile
```diff
@@ -277,26 +277,37 @@ ENV PHP_ORIGIN php-fpm
...
+    && echo "deb [signed-by=/usr/share/keyrings/nginx-archive-keyring.gpg] https://nginx.org/packages/debian bullseye nginx" > /etc/apt/sources.list.d/nginx.list
 RUN apt-get update
-RUN yes '' | apt-get install nginx-core nginx-common nginx nginx-full -y
-RUN ls -l /etc/nginx
-COPY images/runtime/php-fpm/nginx_conf/default.conf /etc/nginx/sites-available/default
-COPY images/runtime/php-fpm/nginx_conf/default.conf /etc/nginx/sites-enabled/default
-RUN sed -ri -e 's!worker_connections 768!worker_connections 10068!g' /etc/nginx/nginx.conf
-RUN sed -ri -e 's!# multi_accept on!multi_accept on!g' /etc/nginx/nginx.conf
-RUN ls -l /etc/nginx
+RUN yes '' | apt-get install nginx=1.30.0-1~bullseye -y
```
vijaysaayi, line 290: ## Upgrade Risk Analysis 

### 🔴 High — Module parity: `nginx-full` → bare `nginx`
 
 Debian's `nginx-full` included these extra modules not present in the nginx.org base package:
 
 | Module | Impact |
 |--------|--------|
 | `ngx_http_dav_ext` | WebDAV extended methods (PROPFIND, MKCOL, etc.) |
 | `ngx_http_upstream_fair` | Fair load-balancing algorithm |
 | `ngx_http_auth_pam` | PAM authentication |
 | `ngx_http_echo` | Debug/test directive |
 | `ngx_http_subs_filter` | Response body substitution |
 | GeoIP2 (static) | Available as dynamic `nginx-module-geoip` from nginx.org |
 
 **Risk:** Customers with custom nginx configs using these directives will get `unknown directive` errors at container startup.

images/runtime/php-fpm/8.2/bullseye.Dockerfile
```diff
@@ -277,26 +277,37 @@ ENV PHP_ORIGIN php-fpm
...
 # Edit the default DocumentRoot setting
 ENV NGINX_DOCUMENT_ROOT /home/site/wwwroot
-# Install NGINX latest stable version using APT Method with Nginx Repository instead of distribution-provided one:
-# - https://www.linuxcapable.com/how-to-install-latest-nginx-mainline-or-stable-on-debian-11/
+# Install NGINX version from official nginx.org repository
 RUN apt-get update
-RUN apt install curl nano -y
-RUN curl -sSL https://packages.sury.org/nginx/README.txt | bash -x
+RUN apt install curl gnupg2 nano -y
+RUN curl -fsSL https://nginx.org/keys/nginx_signing.key | gpg --dearmor -o /usr/share/keyrings/nginx-archive-keyring.gpg \
```
vijaysaayi, line 283: Should we verify the fingerprint ?

eg:
```bash
 RUN curl -fsSL https://nginx.org/keys/nginx_signing.key \
       | gpg --dearmor -o /usr/share/keyrings/nginx-archive-keyring.gpg \
     && gpg --no-default-keyring \
            --keyring /usr/share/keyrings/nginx-archive-keyring.gpg \
            --list-keys 573BFD6B3D8FBC641079A6ABABF5BD827BD9BF62 \
     || { echo "ERROR: nginx signing key fingerprint mismatch"; exit 1; }
```

images/runtime/php-fpm/8.1/bullseye.Dockerfile
Saipriya-1144: This currently don't auto upgrade because the nginx version in apt repos is very old.  I will add it to the hold list

images/runtime/php-fpm/8.2/bullseye.Dockerfile
Saipriya-1144: These third-party modules are not available in the official nginx.org repository.
 Customers using custom nginx configurations that reference these modules will be impacted -
 will monitor for any such failures during rollout.

images/runtime/php-fpm/8.1/bullseye.Dockerfile
Saipriya-1144: Done

images/runtime/php-fpm/8.2/bullseye.Dockerfile
Saipriya-1144: done

surenderssm:
NIT: last two lines in PR description can be updated

surenderssm (approved):

Issue #2917, opened 2026-05-05
Allow disablement of platform detection when the platform and version details are provided explictly


sarsharma:
Today oryx runs the detector for each supported platform regardless of whether we have passed `--platform` and `--platform-version` values during build. 
For most scenarios, we might want to disable running of all detectors and just rely on user provided values to do the env setup and build. Running of all detectors results in fetching/installing of all detected platforms even if they are not required for the build.
Oryx already has config flags which allow disablement of select platforms(eg. DISABLE_NODEJS_BUILD). It might be nice to have another config like `DISABLE_PLATFORM_DETECTION` to skip the detection cycle completely.
The `oryx prep` command also has a `--skip-detection` flag but there is no such equivalent for `oryx build` command.

PR #2918, opened 2026-05-06
Add skip detection option for oryx build command

sarsharma:
Implements #2917 
This change introduces a new `--skip-detection` CLI option for build commands, which allows users to skip automatic platform detection and only use a user-specified platform. The change is implemented across the core build logic and all relevant command-line interfaces. This will help provide users with more control over platform selection during builds and improve the build robustness and preventing false detections.

## Testing
Deployed an app which consists both nodejs and python files
* Before this change -  when passing nodejs as specified platform 
image
* after this change but without --skip-detection flag
image

* after this change with --skip-detection flag
image


kumaraksh1 (approved):
Looks Good !!

But if possible add a test in description.

Saipriya-1144 (approved):

PR #2922, opened 2026-05-11
Add support for custom build commands

sarsharma:
This pull request introduces support for specifying a custom build command(via CUSTOM_BUILD_COMMAND env var) in the generated build scripts for .NET Core, Python, and PHP platforms. This support already exists for nodejs platform. When a custom build command is provided, it overrides the default build steps (like `dotnet publish`, `composer install`, or `pip install`) in the generated scripts. The implementation includes new configuration options, updates to script templates, and ensures the build process and file copying logic adapt accordingly.
These changes make the build process more flexible and customizable for advanced scenarios.

Also introduced a new configuration key `SKIP_PLATFORM_DETECTION` and ensured the CLI correctly merges this setting from multiple sources. Follow up from #2918 to add parity to allow skip detection via env variable/cli flag.

## Testing
Deployed test apps for all platforms - nodejs, python, php, dotnet and set a CUSTOM_BUILD_COMMAND for each scenario and checked if the custom build command was used instead of oryx generated build commands. To verify this, the custom build command also writes a custom txt file which is served from `/` for each app

* Nodejs
image
image

* Python
image

* Dotnet
image

### DIFF
--- a/src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
+++ b/src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
@@ -9,6 +9,12 @@ echo "Using .NET Core SDK Version: $dotnetCoreVersion"
     {{ InstallBlazorWebAssemblyAOTWorkloadCommand }}
 {{ end }}
 
+{{ if CustomBuildCommand | IsNotBlank }}
+echo
+echo "Running custom build command '{{ CustomBuildCommand }}'..."
+echo
+{{ CustomBuildCommand }}
+{{ else }}
 doc="https://docs.microsoft.com/en-us/azure/app-service/configure-language-dotnetcore?pivots=platform-linux"
 suggestion="Please build your app locally before publishing." 
 msg="${suggestion} | ${doc}"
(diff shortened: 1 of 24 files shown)


kumaraksh1 (approved):
LGTM

Shi1810 (approved):

src/BuildScriptGenerator/DotNetCore/DotNetCoreBashBuildSnippet.sh.tpl
```diff
@@ -9,6 +9,12 @@ echo "Using .NET Core SDK Version: $dotnetCoreVersion"
     {{ InstallBlazorWebAssemblyAOTWorkloadCommand }}
 {{ end }}
 
+{{ if CustomBuildCommand | IsNotBlank }}
+echo
+echo "Running custom build command '{{ CustomBuildCommand }}'..."
+echo
```
prathambande, line 16: Not a blocker: Should we surface this message for anyone trying custom build command in dotnet ? 
`The custom command must output to `$DESTINATION_DIR` (e.g., `dotnet publish -o $DESTINATION_DIR`).`

prathambande (approved):
LGTM

prathambande (approved):

PR #2931, opened 2026-06-04
[Security] Unpin Go to 1.26 to fix CVEs

Shi1810:
[Security] Unpin Go to 1.26.3 to fix CVEs

### DIFF
--- a/src/startupscriptgenerator/src/python/go.mod
+++ b/src/startupscriptgenerator/src/python/go.mod
@@ -1,6 +1,6 @@
 module python
 
-go 1.26.3
+go 1.26
 
 require github.com/stretchr/testify v1.11.1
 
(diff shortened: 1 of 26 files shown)


src/startupscriptgenerator/src/python/go.mod
```diff
@@ -1,6 +1,6 @@
 module python
 
-go 1.26.3
+go 1.26
```
Saipriya-1144, line 3: I think this value is the minimum go version required to compile the binary. I think the better way would be using latest version here. Can you check what happens if this version is diff from the one used in image

Saipriya-1144 (approved):

src/startupscriptgenerator/src/python/go.mod
Shi1810: Discussed offline

vijaysaayi (approved):

PR #2934, opened 2026-06-15
Remove nvm dependency for nodejs docker builds

Saipriya-1144:
This PR eliminates the nvm (Node Version Manager) supply chain dependency from all Node.js Dockerfiles and the platform SDK build script. 

Node binaries are now downloaded directly from nodejs.org with GPG signature and SHA256 integrity verification. The clearsigned SHASUMS256.txt.asc file is verified against the official nodejs/release-keys keyring before validating the tarball SHA256. This ensures the checksums file itself hasn't been tampered with, protecting against a nodejs.org server compromise scenario.

An empty grep guard ([ -s node.sha256 ]) prevents silent bypass if the filename is missing from SHASUMS256.txt.

Additional cleanup:
 - Removed Node.js 18 support as it has reached End of Life.
 - Removed dead upgradeNpm() function from build.sh — it only applied to npm 6.x. All actively built versions (Node 20, 22, 24) ship with npm 10+.


Shi1810 (approved):

images/runtime/node/18/bullseye.Dockerfile
```diff
@@ -1,94 +0,0 @@
-ARG BASE_IMAGE
-
-# Startup script generator
-FROM mcr.microsoft.com/oss/go/microsoft/golang:1.26-bullseye as startupCmdGen
```
vijaysaayi, line 4: Are we removing node-18 dockerfile as it EOL ?

Saipriya-1144: Yes Vijay, it's EOL in 2025

platforms/nodejs/build.sh
```diff
@@ -10,32 +10,25 @@ version="$1"
...
+
+INSTALL_DIR="/opt/nodejs/${version}"
+mkdir -p "$INSTALL_DIR"
+
+echo "Downloading Node.js v${version} from official source..."
+curl -fsSLO "$NODE_DOWNLOAD_URL"
+curl -fsSL "$NODE_SHASUM_URL" -o SHASUMS256.txt
+
+# Verify SHA256 integrity
+grep "$nodeFileName" SHASUMS256.txt | sha256sum -c -
```
vijaysaayi: `[Security][Blocking]` Same SHA bypass as in the Dockerfiles. Empty grep output makes sha256sum -c succeed on empty stdin.
 
```bash
  set -o pipefail
  grep "$nodeFileName" SHASUMS256.txt > node.sha256
  [ -s node.sha256 ] || { echo "ERROR: $nodeFileName not found in SHASUMS256.txt"; exit 1; }
  sha256sum -c node.sha256
  rm -f node.sha256
```
 
 (Or add `set -euo pipefail` at the top of the script and switch shebang to `#!/usr/bin/env bash`.)

images/runtime/node/20/bookworm.Dockerfile
```diff
@@ -14,16 +14,17 @@ ENV GIT_COMMIT=${GIT_COMMIT}
...
 ARG NODE20_VERSION
-ENV NVM_DIR /usr/local/nvm
 RUN apt-get update \
-    && apt-get install -y --no-install-recommends curl ca-certificates \
-    && mkdir -p $NVM_DIR \
-    && curl -sL https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.4/install.sh | bash \
-    && . $NVM_DIR/nvm.sh \
-    && nvm install ${NODE20_VERSION}
+    && apt-get install -y --no-install-recommends curl ca-certificates xz-utils \
+    && curl -fsSLO "https://nodejs.org/dist/v${NODE20_VERSION}/node-v${NODE20_VERSION}-linux-x64.tar.xz" \
```
vijaysaayi, line 23: I see we are now hardcoding `linux-x64` for all arch , was nvm install previously handling this ?

platforms/nodejs/build.sh
vijaysaayi: Also should we Pin nodeXXVersion_SHA256 in images/constants.yml and pass it as a build arg, then verify against that pinned value.

images/runtime/node/18/bullseye.Dockerfile
vijaysaayi: If so seems like there is more code we need to remove. Can we take this over a separate PR ?

1. **`images/runtime/build_runtime_images.sh:78-81`** — still has a `"18")` case branch that invokes the deleted `./images/runtime/node/18/$os_flavor.Dockerfile` and uses the now-removed `$node18Version`. Please delete this branch.
 2. **`src/BuildScriptGenerator/Node/NodeVersions.cs:31`** — `Node18Version => ConstantsYamlReader.Get("node18Version")` will now return null/empty since the YAML key is gone. Remove the property (and its consumers).
 3. **`build/__nodeVersions.sh:13`** — `NODE18_VERSION='18.20.4'` should be removed.
 4. **`monitor_version_scripts/override_constants.yaml:11`** — `node18DebianFlavors: bullseye` should be removed.
 5. **Tests still reference Node 18 and will break:**
    - `tests/Oryx.Integration.Tests/Node/NodeDynamicInstallationTest.cs` (lines ~30, 92, 145)
    - `tests/Oryx.BuildImage.Tests/Node/NodeDynamicInstallationTest.cs` (lines 44, 335)
    - `tests/Oryx.BuildImage.Tests/CommandTests/PlatformsCommandTest.cs:67` — asserts `Versions.Contains("18.20.7")`
    - `tests/Oryx.BuildImage.Tests/Node/NodeJsSampleAppsOtherTests.cs:311`
    - `tests/Oryx.Integration.Tests/Node/NodeOtherEndtoEndTests.cs:565`

platforms/nodejs/build.sh
Saipriya-1144: - Agreed for the 1st point. Added logic to check if file is actually present in SHASUM256.txt
- We already verify the SHASUMS256.txt file's authenticity via GPG signature using the official nodejs/release-keys keyring. This makes pinning redundant

images/runtime/node/20/bookworm.Dockerfile
Saipriya-1144: nvm auto detects the architecture and installs corresponding sdk. It was installing sdk for linux-x64 arch

prathambande (approved):

vijaysaayi (approved):

vijaysaayi (approved):

PR #2939, opened 2026-06-29
Add support for node 26

vijaysaayi:


images/runtime/node/26/install-dependencies-slim.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
prathambande, line 1: this install-dependencies.sh functions differently from the existing install-depencies.sh we have for existing node images
Is this change to alter that ?

PR #2951, opened 2026-07-10
Onboarding Python 3.15

prathambande:
This PR brings Dockerfile and supporting shell scripts addition/changes to get about getting Python 3.15 on resolute raccoon based OS flavor. 
 
Why new common base ? 
- To keep ongoing stack versions untouched and cut down on unnecessary packages/ script files to prevent bloating of intermediate layers 
- This will also help with shipping upcoming newer frameworks with resolute raccoon

Comparing with 3.14 Noble:
image


images/runtime/python/resolute-based/resolute.Dockerfile
```diff
@@ -0,0 +1,113 @@
...
+COPY images/runtime/python/resolute-based/install-dependencies.sh /tmp/install-dependencies.sh
+RUN chmod +x /tmp/install-dependencies.sh \
+    && bash /tmp/install-dependencies.sh \
+    && rm -f /tmp/install-dependencies.sh
+
+ARG PYTHON_FULL_VERSION
+ARG PYTHON_VERSION
+ARG PYTHON_MAJOR_VERSION
+
+ENV PYTHON_VERSION ${PYTHON_FULL_VERSION}
```
vijaysaayi, line 69: Line 66 declares `ARG PYTHON_VERSION` (expected value: `3.15`), but line 75 sets `ENV PYTHON_VERSION ${PYTHON_FULL_VERSION}` (value: `3.15.0b3`). Per Docker semantics, the ENV permanently shadows the ARG of the same name in all subsequent RUN commands.

The symlink command on line 77 then expands to:
```bash
ln -s 3.15.0b3 3.15.0b3   # self-referencing — fails with 'File exists'
```

Since `set -ex` is active, the entire build aborts. No Python 3.15 image can be produced.

images/runtime/python/resolute-based/install-dependencies.sh
```diff
@@ -0,0 +1,43 @@
...
+        gnupg \
+        libpq-dev \
+        default-libmysqlclient-dev \
+        unzip \
+        libodbc2
+
+# Microsoft SQL Server 2018
+# https://docs.microsoft.com/en-us/sql/connect/odbc/linux-mac/installing-the-microsoft-odbc-driver-for-sql-server
+export ACCEPT_EULA=Y \
+    && curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add - \
```
vijaysaayi: `apt-key` was deprecated in Ubuntu 22.04 and removed from newer releases. On Ubuntu 26.04 (Resolute Raccoon), this line exits non-zero. Since `set -ex` is active (line 7), the entire script aborts before the correct modern approach on line 28 (`gpg --dearmor`) ever executes.

Additionally, line 26 uses `curl` without `-f`, so HTTP errors would be silently piped to `apt-key` as garbage data.

**Suggested fix:** Remove line 26 entirely and update line 27's source list to use scoped signing:
```bash
export ACCEPT_EULA=Y \
    && curl -fsSL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor -o /usr/share/keyrings/microsoft-prod.gpg \
    && echo "deb [signed-by=/usr/share/keyrings/microsoft-prod.gpg] https://packages.microsoft.com/ubuntu/24.04/prod noble main" > /etc/apt/sources.list.d/mssql-release.list \
    && apt-get update \
    && ACCEPT_EULA=Y apt-get install -y msodbcsql18=18.6.1.1-1
```

images/runtime/python/resolute-based/install-dependencies.sh
```diff
@@ -0,0 +1,43 @@
...
+    && curl https://packages.microsoft.com/config/ubuntu/24.04/prod.list > /etc/apt/sources.list.d/mssql-release.list \
```
vijaysaayi: This fetches the Ubuntu 24.04 (Noble) Microsoft package repository, but the base OS is Ubuntu 26.04 (Resolute Raccoon). Packages from the 24.04 repo are compiled against Noble's glibc and libssl — on Resolute, this can cause:
- Hard install failure (unresolvable dependencies)
- Silent ABI mismatch → runtime `SIGSEGV` when `pyodbc` loads the driver

Also uses `curl` without `-f`, so a 404 would silently write an HTML error page as the source list.

**Suggested fix:** At minimum, add `curl -fsSL` and document as a temporary workaround:
```bash
# TODO: Update to ubuntu/26.04/prod.list when Microsoft publishes it.
# Using 24.04 as a temporary workaround — pinned to msodbcsql18=18.6.1.1-1 to limit drift.
&& curl -fsSL https://packages.microsoft.com/config/ubuntu/24.04/prod.list > /etc/apt/sources.list.d/mssql-release.list \
```

platforms/python/prereqs/build.sh
```diff
@@ -52,7 +52,6 @@ fi
         liblzma-dev \
         libsqlite3-dev \
         lzma \
-        lzma-dev \
```
vijaysaayi, line 55: This script is shared across all OS flavors. `lzma-dev` is used for what ? is it safe to remove for other python versions ?

Or can we just remove it for python 3.15 ?

prathambande: Yes, it is safe to remove for other python versions too.
1) This .sh file is executed in intermediate image stage where we build python binary, and only /opt/python is copied from it to the main. 
2) lzma-dev seems to have been added for building this python binary -- this package is old/obsolete and `liblzma-dev` package should suffice

prathambande: If lzma-dev is something which is needed, then the build will fail and we would know before shipping. As far, build did not fail in our testing. 

lzma-dev package is not present in ubuntu 26

images/runtime/python/resolute-based/resolute.Dockerfile
prathambande: Done, removed the legacy carry forward of not needed code. Thanks !

images/runtime/python/resolute-based/install-dependencies.sh
prathambande: Good Catch ! 
Since msodbc driver is not currently present for resolute, added a todo placeholder

images/runtime/python/resolute-based/install-dependencies.sh
prathambande: Good Catch ! 
Since msodbc driver is not currently present for resolute, added a todo placeholder

vijaysaayi (approved):

kumaraksh1:
Can you add test cases for this stack?

kumaraksh1 (approved):
LGTM

vijaysaayi (approved):

PR #2956, opened 2026-07-12
Onboarding Node 26

prathambande:
This PR brings Dockerfile and supporting shell scripts addition/changes to get about getting Node 26 on resolute raccoon based OS flavor.

Why new common base ?

- To keep ongoing stack versions untouched and cut down on unnecessary packages/ script files to prevent bloating of intermediate layers
- Similar to https://github.com/microsoft/Oryx/pull/2951

Comparing with Node 24 Noble
image

Corepack is not being shipped by default in Node 25+: https://github.com/nodejs/TSC/pull/1697


images/constants.yml
```diff
@@ -12,6 +12,7 @@ variables:
   node20osFlavors: bullseye,bookworm
   node22osFlavors: bullseye,bookworm
   node24osFlavors: noble
+  node26osFlavors: resolute
```
vijaysaayi, line 15: **Node 26 / `resolute` is declared but not wired into the build orchestrator**

`node26osFlavors: resolute` (and the expanded `nodejsosFlavors`) make the flavor visible to the build matrix, but `images/runtime/build_runtime_images.sh` only has node cases for 18/20/22/24 — there is no `26` case, and it never passes the `NODE_FULL_VERSION` / `NODE_SHA256` build-args that `node/26/resolute.Dockerfile` requires. The new `commonbase/resolute/Dockerfile` is also never selected (the orchestrator always builds the shared `commonbase/Dockerfile`). Net effect: the Node 26 leg cannot be built, and a half-wired flavor in the global list risks failing the release matrix.

**Recommendation:** Add an explicit Node 26 case that selects `node/26/$os_flavor.Dockerfile`, builds/uses the `commonbase/resolute` base, and passes `NODE_FULL_VERSION` + `NODE_SHA256`. Confirm which script CI actually runs (the pipeline invokes `./build/buildRunTimeImages.sh`) and add an end-to-end smoke build for `node:26-resolute`.

images/runtime/commonbase/resolute/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+		ca-certificates \
+		curl \
+		gnupg \
+		netbase \
+		wget \
+        unzip \
+		# Not shipped in Resolute, but will be required by Node runtime. It is tiny, would save from unexpected runtime failures in other frameworks too
+		libatomic1
+
+RUN  rm -rf /var/lib/apt/lists/*
```
vijaysaayi: Can we combine this with the layer which runs apt-get update ?

images/runtime/node/26/resolute.Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+ARG RELEASE_TAG_NAME=unspecified
+ENV RELEASE_TAG_NAME=${RELEASE_TAG_NAME}
+ENV GIT_COMMIT=${GIT_COMMIT}
+ENV BUILD_NUMBER=${BUILD_NUMBER}
+RUN chmod +x build.sh && ./build.sh node /opt/startupcmdgen/startupcmdgen
+
+
+# Download Node.js directly from official source and verify SHA256
+FROM ${BASE_IMAGE} AS nodeDownloader
+ARG NODE24_VERSION
```
vijaysaayi: Do we need this arg ?

images/runtime/node/26/resolute.Dockerfile
```diff
@@ -0,0 +1,95 @@
...
+ARG NODE_FULL_VERSION
+ARG NODE_SHA256
+WORKDIR /tmp/node-download
+RUN set -ex \
+    && curl -fsSLO "https://nodejs.org/dist/v${NODE_FULL_VERSION}/node-v${NODE_FULL_VERSION}-linux-x64.tar.xz" \
+    && echo "${NODE_SHA256} node-v${NODE_FULL_VERSION}-linux-x64.tar.xz" | sha256sum -c - \
```
vijaysaayi: This verifies the tarball with a bare `sha256sum` against a build-arg hash. Node 24 (`node/24/noble.Dockerfile`) downloads the signed `SHASUMS256.txt.asc` and verifies it against Node's release GPG keys *before* the SHA check. Trusting only a build-arg SHA means a compromised or mis-set arg would accept a malicious tarball that matches the supplied hash, with no upstream-signature validation. Severity: Medium.

**Recommendation:** Restore the Node 24 pattern — fetch `SHASUMS256.txt.asc`, import the nodejs release-keys, verify the signature, then `sha256sum -c`. If central SHA-pinning is the intended model instead, pin the SHA in `constants.yml` and document why signature verification is dropped.

images/runtime/commonbase/resolute/Dockerfile
prathambande: Done

images/runtime/node/26/resolute.Dockerfile
prathambande: Done

vijaysaayi:
In `build_runtime_images.sh` , we need to add a switch case for node 26

images/runtime/node/26/resolute.Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+ENV BUILD_NUMBER=${BUILD_NUMBER}
+RUN chmod +x build.sh && ./build.sh node /opt/startupcmdgen/startupcmdgen
+
+
+# Download Node.js directly from official source and verify SHA256
+FROM ${BASE_IMAGE} AS nodeDownloader
+ARG NODE_FULL_VERSION
+ARG NODE_SHA256
+WORKDIR /tmp/node-download
+RUN curl -fsSLO "https://nodejs.org/dist/v${NODE_FULL_VERSION}/node-v${NODE_FULL_VERSION}-linux-x64.tar.xz" \
```
vijaysaayi, line 21: node24's `noble.Dockerfile` has explicit `dpkg --print-architecture` mapping.
Considering this, Will an arm64 build fails silently with `exec format error` at container start.

vijaysaayi:
We have some constants in `NodeVersions.cs` , Do we use this ? If so do we need a constant for Node 26 ?

images/runtime/node/26/resolute.Dockerfile
```diff
@@ -0,0 +1,93 @@
+ARG BASE_IMAGE
+ARG NODE_FULL_VERSION
+ARG NODE_SHA256
+
+# build the Oryx startup-script generator (the `oryx` CLI).
+FROM mcr.microsoft.com/oss/go/microsoft/golang:1.26.4-bookworm as startupCmdGen
```
kumaraksh1: golang:1.26.5 is also available in community, check if you can use that?
https://mcr.microsoft.com/en-us/artifact/mar/oss/go/microsoft/golang/tag/1.26.5-2-bookworm

images/runtime/commonbase/resolute/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+	&& apt-get install -y --no-install-recommends \
+		xz-utils \
+		ca-certificates \
+		curl \
+		gnupg \
+		netbase \
+		wget \
+        unzip \
+		# Not shipped in Resolute, but will be required by Node runtime. It is tiny, would save from unexpected runtime failures in other frameworks too
+		libatomic1 \
```
kumaraksh1, line 32: libatomic1 for what?

images/runtime/node/26/resolute.Dockerfile
```diff
@@ -0,0 +1,93 @@
+ARG BASE_IMAGE
+ARG NODE_FULL_VERSION
```
kumaraksh1: This line can be rmeoved.

images/runtime/node/26/resolute.Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+ENV NODE_VERSION=${NODE_FULL_VERSION}
+ENV NPM_CONFIG_LOGLEVEL=info
+
+COPY --from=nodeDownloader /opt/nodejs/ /usr/local/
+RUN ln -s /usr/local/bin/node /usr/local/bin/nodejs
+
+COPY images/runtime/node/installDependencies.sh /tmp/installDependencies.sh
+
+ARG NPM_VERSION
```
kumaraksh1: Unused

images/runtime/node/26/resolute.Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+ARG PM2_VERSION
+ARG NODE_APP_INSIGHTS_SDK_VERSION
```
kumaraksh1: unused

images/runtime/commonbase/resolute/Dockerfile
prathambande: It is needed for running nodejs

prathambande: Without it, node errors and does not start
`node: error while loading shared libraries: libatomic.so.1: cannot open shared object file: No such file or directory`
https://github.com/nodejs/node/issues/60790

images/runtime/node/26/resolute.Dockerfile
prathambande: The line of code was 
```
RUN ARCH= && dpkgArch="$(dpkg --print-architecture)" \
  && case "${dpkgArch##*-}" in \
    amd64) ARCH='x64';; \
    ppc64el) ARCH='ppc64le';; \
    s390x) ARCH='s390x';; \
    arm64) ARCH='arm64';; \
    armhf) ARCH='armv7l';; \
    i386) ARCH='x86';; \
    *) echo "unsupported architecture"; exit 1 ;; \
  esac
```

looking at it, its purpose is failing docker build if arch is not in the list. It does nothing else, the value (ARCH) is not preserved or propagated, and we curl download node `-linux-x64`. 

`arm64 build fails silently with exec format error at container start` - it would have failed even with the presence of that RUN command

images/runtime/node/26/resolute.Dockerfile
prathambande: Done, passes smoke tests, thanks !

images/runtime/node/26/resolute.Dockerfile
prathambande: Addressed

images/runtime/node/26/resolute.Dockerfile
prathambande: Addressed

images/runtime/node/26/resolute.Dockerfile
prathambande: Addressed

vijaysaayi (approved):

kumaraksh1 (approved):
LGTM

Shi1810 (approved):

PR #3036, opened 2026-08-13
Fix PHP source archive download URLs

vijaysaayi:
Update PHP source archive URLs from the legacy `/get/.../from/this/mirror` route to PHP's direct `/distributions/` route.

## Changes

- Update PHP FPM runtime Dockerfiles for the maintained PHP 8.1–8.5 variants.
- Update the shared PHP prerequisite build script.
- Update the corresponding detached-signature URLs.
- Preserve the existing `secure.php.net` hostname in the shared prerequisite script.

## Reason

The existing `/get/php-<version>.tar.xz/from/this/mirror` route currently returns HTTP 404. The equivalent `/distributions/php-<version>.tar.xz` archives and `.asc` signatures are available.

## Validation

- Confirmed the archive and signature URLs return HTTP 200 for the active PHP versions referenced by the repository.
- Confirmed `platforms/php/prereqs/build.sh` passes `bash -n`.
- Confirmed no legacy `php.net/get` source archive URLs remain.

### DIFF
--- a/images/runtime/php-fpm/8.1/bullseye.Dockerfile
+++ b/images/runtime/php-fpm/8.1/bullseye.Dockerfile
@@ -62,7 +62,7 @@ ENV GPG_KEYS 528995BFEDFBA7191D46839EF9BA0ADA31CBD89E 39B641343D8C104B2B146DC3F9
 ARG PHP_VERSION
 ARG PHP_SHA256
 ENV PHP_VERSION ${PHP_VERSION}
-ENV PHP_URL="https://www.php.net/get/php-${PHP_VERSION}.tar.xz/from/this/mirror" PHP_ASC_URL="https://www.php.net/get/php-${PHP_VERSION}.tar.xz.asc/from/this/mirror" PHP_MD5=""
+ENV PHP_URL="https://www.php.net/distributions/php-${PHP_VERSION}.tar.xz" PHP_ASC_URL="https://www.php.net/distributions/php-${PHP_VERSION}.tar.xz.asc" PHP_MD5=""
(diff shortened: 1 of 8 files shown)


Saipriya-1144 (approved):

images/runtime/php-fpm/8.1/bullseye.Dockerfile
```diff
@@ -62,7 +62,7 @@ ENV GPG_KEYS 528995BFEDFBA7191D46839EF9BA0ADA31CBD89E 39B641343D8C104B2B146DC3F9
 ARG PHP_VERSION
 ARG PHP_SHA256
 ENV PHP_VERSION ${PHP_VERSION}
-ENV PHP_URL="https://www.php.net/get/php-${PHP_VERSION}.tar.xz/from/this/mirror" PHP_ASC_URL="https://www.php.net/get/php-${PHP_VERSION}.tar.xz.asc/from/this/mirror" PHP_MD5=""
+ENV PHP_URL="https://www.php.net/distributions/php-${PHP_VERSION}.tar.xz" PHP_ASC_URL="https://www.php.net/distributions/php-${PHP_VERSION}.tar.xz.asc" PHP_MD5=""
```
kumaraksh1, line 65: Do we need to update this as we wont be generating 8.1 images anymore?

kumaraksh1 (approved):
LGTM

images/runtime/php-fpm/8.1/bullseye.Dockerfile
vijaysaayi: The current pipeline is still building php 8.1 images. We can remove this on a separate PR.

PR #3087, opened 2026-09-18
Pin PHP 8.2 Bullseye package sources to Debian snapshot

Saipriya-1144:
This PR pins the Debian Bullseye package sources used by the PHP 8.2 runtime, common base, and PHP SDK images to snapshot  20260831T211305Z , 

Bullseye reached end of life and packages referenced by the live repository are no longer consistently available. With this change, customer startup commands using apt-get update, apt-get install, or apt-get upgrade will use the frozen snapshot and avoid package download failures.

Old apt-sources file:
image

New apt-sources file:
image


platforms/php/prereqs/Dockerfile
```diff
@@ -2,6 +2,14 @@ ARG OS_FLAVOR
 FROM mcr.microsoft.com/mirror/docker/library/buildpack-deps:${OS_FLAVOR} AS php-buildpack-prereqs
 ARG OS_FLAVOR
 ENV OS_FLAVOR=$OS_FLAVOR
+
+COPY images/runtime/scripts/configure-debian-bullseye-snapshot.sh /usr/local/bin/
```
vijaysaayi, line 6: Thinking out loud, is there a way to scope these changes only for bullseye ?

Saipriya-1144: We can make the  COPY  conditional using a multi-stage build, but that would add unnecessary complexity. These images are used only for SDK generation, and the pipeline publishes only the contents of  /tmp/compressedSdk . The helper is executed only for Bullseye and is removed immediately afterward, so it does not affect the generated SDK artifacts.

vijaysaayi (approved):

images/runtime/php-fpm/8.2/bullseye.Dockerfile
```diff
@@ -18,6 +18,11 @@ RUN chmod +x build.sh && ./build.sh php /opt/startupcmdgen/startupcmdgen
 FROM ${BASE_IMAGE}
 ARG IMAGES_DIR=/tmp/oryx/images
 
+COPY images/runtime/scripts/configure-debian-bullseye-snapshot.sh /usr/local/bin/
+RUN chmod +x /usr/local/bin/configure-debian-bullseye-snapshot.sh \
+	&& /usr/local/bin/configure-debian-bullseye-snapshot.sh \
+	&& rm /usr/local/bin/configure-debian-bullseye-snapshot.sh
```
prathambande, line 24: this will run twice right then ? once in commonbase and again here

prathambande (approved):

